Issue #3, opened 2016-05-30
Review Comments to SC1-4-1-a


YetAnotherEye:
[https://github.com/auto-wcag/auto-wcag/blob/master/drafts/SC1-4-1-a.md](https://github.com/auto-wcag/auto-wcag/blob/master/drafts/SC1-4-1-a.md)

Please add your comments regarding the test-case to this issue.

WilcoFiers:
Updated, please also review #6

PR #6, opened 2016-06-09
Improved on formatting

WilcoFiers:
Changed formatting for #3.


YetAnotherEye:
* Assumtions and Step 1: What happens, if the border has the same color as the background-color?
* Selector: ancestor has [...] A different text color then the text color of the link. Please explain

Issue #13, opened 2016-08-26
Restructure the format


WilcoFiers:
We've been using the same template for two years now, and given what we've learned I think we can do better. 

https://github.com/auto-wcag/auto-wcag/blob/master/pages/rule-template.md

A few topics to address:
- Move the test procedure above background and assumptions. It seems to me this is a more important part of any rule.
- We should consider including validation tests. I would suggest that we include at least one code snippet for each unique outcome a rule creates. The discussions about ACT have shown there is a lot of interest in this.
- The way we present results currently isn't semantically correct. In the wiki we had a template for this, but markup does not have these. We'll need a better way to show these in markdown, without them getting too complicated.

WilcoFiers:
This has been done :)

PR #15, opened 2016-09-28
updated rule-template

WilcoFiers:
as in #13 

List of changes:
- Renamed 'criteria' to 'success_criterion'
- Moved the background below the test procedure
- Removed test properties table, as all of the information was moved to other places
- Moved outcomes outside the test steps, to a separate section **outcome**
- Added a user question
- Removed test mode from each step. Since all manual steps are always in the shape of a user question, it is straight forward enough to understand which parts require user input
- Move the identifier instruction to rule-design.md
- Added an **Implementation tests** section, which should reference to a place where implementation tests for the rule can be found.
- Added a **change log** section


josehilera:
In the rule template, there are propeties in TEST PROPERTIES section with different name that the same in the heading of the rule template.

- "name" (in heading) and "Test name" (in section). 
- "test_mode" (in heading) and "Test mode" (in section)
- "criteria" (in heading) and "success criterion" (in section).

Maybe they should be deleted from the section, because they already are in heading.

In the case of "criterion", this term should be "criteria", because the rule can be associated to more than one criterion. For example the rule "SC2-2-1+SC3-2-5-meta-refresh".

pages/design/rule-template.md
```diff
@@ -30,62 +22,87 @@ authors:
...
+| Property     | Value
+|--------------|---------
+| Highlight    | Element A1 and B2
+| Question     | Does T2 provide an extended description of the image additionally to T1?
+| Help         | If the image contributes meaning to the page or provide any functionality or conveys information additional to the pages text, this must be described.
+| User profile | Requires <sight / hearing / fine motor control / HTML Knowledge / Accessibility knowledge / ...>
+| context      | yes | Optional
+| Interaction  | yes | Optional
+
+## Background
```
EmmaJP: The position of this section is inconsistent with rule-design.md

PR #22, opened 2016-11-15
P as heading

WilcoFiers:
Added a rule to identify when paragraphs are used as headings. Looking forward to what people think!

### DIFF
--- a/_drafts/SC1-3-1-p-as-heading.md
+++ b/_drafts/SC1-3-1-p-as-heading.md
@@ -1,6 +1,6 @@
 ---
 rule_id: SC1-3-1-p-as-heading
-name: 
+name:
 test_mode: semi-automatic
 
 criteria:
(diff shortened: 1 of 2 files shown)


dd8:
Wouldn't the text size rule pick up tag clouds that use <p></p> for the words in the cloud?

https://en.wikipedia.org/wiki/Tag_cloud#/media/File:Word_population_tagcloud_2011.png

WilcoFiers:
@dd8 I think you are right. We can maybe solve this by checking that both paragraphs are display:block and that they don't have absolute position?

dd8:
I think it's worth assembling a bunch of different texts to test possible algorithms

For example, you can find the full text of Lord of the Rings online. The only headings are the Chapter headings plus sections for Foreword and notes. Any algorithm should only detect these and no others. Might not want to use LotR due to copyright issues, but should be able to find something similar in the public domain.

Should also have other test texts like US Constitution, song lyrics, creative commons license, etc. where a human can check the number of headings that should be present, then the algorithm can be tested against this.

_drafts/SC1-3-1-p-as-heading.md
```diff
@@ -0,0 +1,245 @@
...
+## Test procedure
+
+### Selector
+
+Test mode: [automatic][]
+
+Select each `p` element that meets the following:
+
+- It contains non-empty text
+- It has one or more sibling `p` elements after it
```
EmmaJP: This could be true of paragraph and heading elements.

dd8:
Definition of textParagraph looks like it might be prone to false positives/negatives: 
"Any textNode containing punctuation marks: `.`, `:`, `!`, `?`

The probably won't work for Wikipedia pages with punctation in the title:
https://en.wikipedia.org/wiki/Mars_Attacks!
https://en.wikipedia.org/wiki/O_Brother,_Where_Art_Thou%3F

PR #29, opened 2017-01-12
Outline the breakup for SC1.1.1

WilcoFiers:


dd8:
I added some more comments to the original issue, which are relevant to this:
https://github.com/auto-wcag/auto-wcag/issues/25

It's difficult to see how the rules interact, and what gaps there (e.g. IMG without ALT isn't covered in the PR, but that's hard to see at the moment).

It would be easier to see if there was a high level selector, which then called the test procedure in the split rules, so something like this

```
If image contains text then
  call SC1-1-1-images-of-text.md
else if image parent is link, button, input then
   call SC1-1-1-interactive-images.md
else ...
```

There are also no checks for other types of images like the SVG and CANVAS elements.

The complex-images rule assumes the user has correctly used aria-describedby, but people often get confused about aria-describedby and aria-labelledby, so this may not be a reliable assumption (https://www.aaron-gustafson.com/notebook/aria-quicktip-labelledby-vs-describedby/).

PR #30, opened 2017-01-12
Sc4 1 1 updates

WilcoFiers:


EmmaJP:
Agreed to create a new pull request with only template changes, and then suggested adjustments as a second pull request.

EmmaJP:
I was asked to split this into two separate pull requests - one moving to revised rule format with no other changes, and a second with other suggested changes. I think this pull request can be rejected.

EmmaJP:
Impact from https://github.com/auto-wcag/auto-wcag/pull/31 and https://github.com/auto-wcag/auto-wcag/pull/32/files

WilcoFiers:
Has been superseded by #53, #54, #55

PR #31, opened 2017-04-09
Delete rules based on F17

WilcoFiers:
I recently discovered that one of the failure techniques that three of our rules were based on had been dropped as of Febuary 2015. It's a unfortunate that this went unnoticed for 2 years, so in addition to dropping these rules now, we should have a discussion on how we can avoid this in the future.

Publication that dropped F17: https://www.w3.org/TR/2015/NOTE-WCAG20-TECHS-20150226/


EmmaJP:
Shadi shared https://github.com/w3c/wcag/pull/45

EmmaJP:
Seems OK ... has an impact on https://github.com/auto-wcag/auto-wcag/pull/30, which I still haven't had time to rework.

PR #33, opened 2017-05-03
Adding "Cannot tell" as outcome to SC3-1-1-text.md

josehilera:
Example of using "Cannot tell" value for outcome when an user is not availabe for manual evaluation.

### DIFF
--- a/_rules/SC3-1-1-text.md
+++ b/_rules/SC3-1-1-text.md
@@ -2,8 +2,9 @@
 rule_id: SC3-1-1-text
 name: HTML lang matches text
 test_mode: semi-automatic
+environment: DOM Structure
 
-criteria:
+success_criterion:
 - 3.1.1 # Language of Page (Level A)
 
 authors:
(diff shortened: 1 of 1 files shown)


josehilera:
All rules in the repository only offer two possible output values: failed or passed. However, [ACT Rules Format 1.0](https://www.w3.org/TR/act-rules-format/) and [EARL 1.0](https://www.w3.org/TR/EARL10-Schema/#OutcomeValue) also consider "Cannot tell" as a possible value.

I believe that in all semi-automatic rules, a "Cannot tell" exit point should be included just before each manual step. In this way, an automatic evaluation tool that does not allow interaction with the user to answer questions, will offer as output the "cannot tell" value together with a description of the reason for which it could not be evaluated.

_rules/SC3-1-1-text.md
```diff
@@ -34,44 +35,28 @@ This test does not prescribe which of these definitions to use as long as one of
...
 
-If yes, return
-
-| Outcome  | Passed
-|----------|-----
-| Testcase | {{ page.rule_id }}
-| ID       | {{ page.rule_id }}-text-pass1
-
+If yes, return [step1-pass](#step1-pass)
+Else If the user in not available, return [step1-cannottell](#step1-cannottell)
```
WilcoFiers: Move this statement into step 2 before the user question.

josehilera:
Pull request closed because I have created a similar but updated pull request (#41) considering all rules in _rules folder.

PR #38, opened 2017-06-06
Table headers-id rules refactor

kensgists:
Original draft rule was refactored into two rules, one fully automatic and the other semi-automatic. Rules were made to conform with the rule template.

### DIFF
--- a/_drafts/SC1-3-1-tables-headers-id-correct.md
+++ b/_drafts/SC1-3-1-tables-headers-id-correct.md
@@ -1,7 +1,7 @@
 ---
-rule_id: SC1-3-1-tables-headers-id
-name: Tables headers-id associations
-test_mode: automatic
+rule_id: SC1-3-1-tables-headers-id-correct
+name: Tables headers-id associations correct
+test_mode: semi-automatic
 environment:  DOM Structure
 
 success_criterion:
@@ -13,7 +13,7 @@ authors:
 
 ## Description
 
-This rule checks that data table headers-id associations are used correctly.
+This rule checks that data table headers-id associations reference correct targets in the correct order.
 
 ## Background
 
@@ -29,7 +29,9 @@ This rule checks that data table headers-id associations are used correctly.
 
 ### Selector
 
-Select all elements that match the following CSS selector:
+Loop outward from most deeply nested table.
+
+For each un-marked table, select all elements that match the following CSS selector:
 
     table:not([role=presentation]) th[headers],
     table:not([role=presentation]) td[headers],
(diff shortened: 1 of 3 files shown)


_drafts/SC1-3-1-tables-headers-id-correct.md
```diff
@@ -0,0 +1,131 @@
+---
+rule_id: SC1-3-1-tables-headers-id-correct
```
WilcoFiers: I suggest removing "-correct" from the title to keep it terse.

_drafts/SC1-3-1-tables-headers-id-correct.md
```diff
@@ -0,0 +1,131 @@
...
+success_criterion:
+- 1.3.1 # Info and relationships (level A)
+
+authors:
+- Ken Petri
+---
+
+## Description
+
+This rule checks that data table headers-id associations reference correct targets in the correct order.
```
WilcoFiers: How would this rule work with other rules for testing tables?

_drafts/SC1-3-1-tables-headers-id-correct.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Test procedure
+
+### Selector
+
+Loop outward from most deeply nested table.
+
+For each un-marked table, select all elements that match the following CSS selector:
+
+    table:not([role=presentation]) th[headers],
```
WilcoFiers: The following table would match this, even though I don't think that it should:

```html
<table><tr><td>
   <table role="presentation"><tr><td headers="">Matching cell</td></tr></table>
</td></tr></table>
```

The same is true for the selector in the second rule

_drafts/SC1-3-1-tables-headers-id-correct.md
```diff
@@ -0,0 +1,131 @@
...
+find the the 'td' or 'th' referenced,
+
+give the user the following question:
+
+| Property             | Value
+|----------------------|---------
+| Presented item       | Highlight a table cell and the related "header" cell.
+| Question             | Is the cell appropriately categorized by or semantically related with the associated "header" cell?
+| Requires context     | Yes. An understanding of the data relationships in the table.
+| Requires interaction | Yes. Answering the question allows test to proceed to the next step or iteration.
```
WilcoFiers: These should only be 'yes' or 'no'.

_drafts/SC1-3-1-tables-headers-id-correct.md
kensgists: Need a way to differentiate between the two rules -- one "valid," one "correct"

_drafts/SC1-3-1-tables-headers-id-correct.md
kensgists: In latest pull, I created a dependency on tables-headers-id-valid rule. Seems like there needs to be a first pass that validates the associations in terms of pure conformance to spec (or, maybe this would be caught by the parser?? and there is no need for the tables-headers-id-value rule???)

_drafts/SC1-3-1-tables-headers-id-correct.md
kensgists: Tables are treated in isolation -- loop from innermost out, marking that a table has been checked as you move from table to table. Contents of marked tables is ignored on subsequent loops. 

I'm open to a change to the selector if there is a way to exclude nested tables, but I couldn't think of a way to do it in CSS.

kensgists:
Changes to files address some of @WilcoFiers comments and updates both rules so that headers-id usage conforms to HTML specification.

_drafts/SC1-3-1-tables-headers-id-correct.md
EmmaJP: Is it worth filtering out layout tables first? Or would a more specific selector do this?

cpandhi: Checking for role of presentation or none is one options, but not all authors implement that. HTML 5.1 spec has some guidance on what one can check - https://www.w3.org/TR/2014/WD-html51-20140617/tabular-data.html

kensgists: @EmmaJP and @cpandhi I changed the rule so that it does a very basic check for the existence of the th[headers] or td[headers] and then just assumes that the author is using these within a data table. Thought this was the simplest approach.

WilcoFiers: @kensgists Where can I find these changes? I want to make sure we've got a PR for this.

kensgists:
Changes are on the issue and I have pushed them to my fork. File is here: https://github.com/kensgists/auto-wcag/blob/master/_drafts/SC1-3-1-tables-headers-id.md. 

Do you want me to do another PR?

Note, that the previous files (tables-headers-id-correct and tables-headers-id-valid) can be removed. They are superseded by this new rule, which merges the other two.

WilcoFiers:
Yes, please open a PR. I'll add it to the agenda for Thursday's call.

PR #41, opened 2017-10-02
Added Undetermined outcome before user questions

josehilera:
To satisfy [ACT Rules Format 1.0](https://w3c.github.io/wcag-act/act-rules-format.html#output-outcome), I have added a "Undetermined" outcome value before user questions if the user is not available. In rules included in _rules folder.

### DIFF
--- a/_rules/SC1-1-1-aria-describedby.md
+++ b/_rules/SC1-1-1-aria-describedby.md
@@ -60,9 +60,9 @@ else, return [step2-fail](#step2-fail)
 
 Concatenate the results of [Text Alternative Computation][TXTALT] Algorithm run on the element itself and assign it to variable T1 and on all elements referenced by the `aria-describedby` attribute and assign it to variable T2.
 
-**User Input Question:**
-
 If the user in not available, return [step3-cannottell](#step3-cannottell)
+
+**User Input Question:**
 
 | Property     | Value
 |--------------|---------
--- a/_rules/SC1-1-1-aria-describedby.md
+++ b/_rules/SC1-1-1-aria-describedby.md
@@ -60,8 +60,6 @@ else, return [step2-fail](#step2-fail)
 
 Concatenate the results of [Text Alternative Computation][TXTALT] Algorithm run on the element itself and assign it to variable T1 and on all elements referenced by the `aria-describedby` attribute and assign it to variable T2.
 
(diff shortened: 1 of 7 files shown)


_rules/SC1-1-1-aria-describedby.md
```diff
@@ -104,6 +106,14 @@ The resulting assertion is as follows,
 | outcome     | Failed
 | description | None of the elements referenced by aria-describedby exists.
 
+### step3-undetermined
+
+| Property    | Value
+|-------------|----------
+| type        | TestResult
+| outcome     | Undetermined
```
WilcoFiers: I think these should be `CannotTell` instead of `Undetermined`.

_rules/SC1-1-1-aria-describedby.md
```diff
@@ -60,6 +60,8 @@ else, return [step2-fail](#step2-fail)
 
 Concatenate the results of [Text Alternative Computation][TXTALT] Algorithm run on the element itself and assign it to variable T1 and on all elements referenced by the `aria-describedby` attribute and assign it to variable T2.
 
+If the user in not available, return [step3-undetermined](#step3-undetermined)
+
```
kensgists: Syntax issue. Line should read: "If the user is...."

_rules/SC1-1-1-aria-describedby.md
```diff
@@ -104,6 +106,14 @@ The resulting assertion is as follows,
 | outcome     | Failed
```
kensgists: Line 72: "If no, could you suggest a...."

_rules/SC1-1-1-aria-describedby.md
josehilera: We have here a problem because the inconsistence between two WC3 documents:

- ACT 1.0 establishes "Undetermined": https://www.w3.org/TR/act-rules-format/#output-outcome
- EARL 1.0 establishes "Cannot tell" coded as "earl:cantTell": https://www.w3.org/TR/EARL10-Schema/#OutcomeValue

It should be convenient that both documents editors unify this notation: ACT editor: Wilco and Maureen , EARL editor: Shadi.

_rules/SC1-1-1-aria-describedby.md
josehilera: Thanks

Issue #43, opened 2017-10-12
SC4-1-1-accesskey test procedure may only check first accesskey 


EmmaJP:
**Selector**

Select any element with an accesskey attribute. The following CSS selector could be used: *[accesskey]

**Step 1**

- IF there is no list called 'knownAccesskeys':
  - Create an empty list 'knownAccesskeys;
- Take the first character of the ACCESSKEY attribute as FirstChar
- IF the FirstChar exists in 'knownAccesskeys':
  - Return step1-fail
- ELSE:
  - Add the FirstChar to 'knownAccesskeys'
  - Return step1-pass

...

Fail "The accesskey of this element has occurred before on this page"

EmmaJP:
I had suggested the following as steps:

**Step 1**

If there is no list called 'knownAccesskeys', create an empty list 'knownAccesskeys'.

**Step 2**

For each element, check if the first character of the accesskey attribute value exists in 'knownAccesskeys'.

if yes, return [step2-fail](#step2-fail)

else, add the first character of the accesskey attribute value for the selected element to 'knownAccesskeys'

**Step 3**

Confirm there were no fails.

if yes, return [step3-pass](#step3-pass)

...

Fail "Accesskey <attribute-value> is duplicated on the page."

Pass "All accesskey attributes are unique on the page."

EmmaJP:
Based on the comment on https://github.com/auto-wcag/auto-wcag/issues/46 it may be possible to use the node list from the selector as the list of known access keys.

Something like below? With additional assumption and fail/pass as previously suggested.

...

**Selector**

Select all elements that match the following CSS selector: *[accesskey]

**Step 1**

For each element, check if the first character of the accesskey attribute matches the first character of any other selected element's accesskey attribute.

if yes, return step1-fail

**Step 2**

Confirm there were no fails.

if yes, return step2-pass

...

WilcoFiers:
I suggest:

**Step 1**

Check if the first character of the accesskey attribute matches the first character of any other selected element's accesskey attribute.

if yes, return step1-fail

Else return step2-pass

EmmaJP:
As with unique ids, it should check the whole page before returning a pass. The success criteria requires all accesskeys in the web-page to be unique, so it cannot pass just because one accesskey is unique. Hence the additional assumption as well as the second step before returning a pass.

2biazdk:
If one fails, it never reaches step 2, does it?

EmmaJP:
[...]

PR #53, opened 2017-11-13
New draft accesskey rule

WilcoFiers:
Based on the work in issue #43.

### DIFF
--- a/_drafts/SC4-1-1-accesskey
+++ b/_drafts/SC4-1-1-accesskey
@@ -1,0 +1,69 @@
+---
+rule_id: SC4-1-1-accesskey
+name: Provide unique accesskeys
+test_mode: automatic
+environment: DOM Structure
+
+success_criterion:
+- 4.1.1 # Parsing (Level A)
+
+author:
+- Kamyar Rasta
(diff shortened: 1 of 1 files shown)


EmmaJP:
The rule is about the uniqueness of the accesskey values in a page, rather than whether or not they are sensible values.

WilcoFiers:
We need to add tests to this as well.

WilcoFiers:
@gianwild Merging this into draft, we'll open an issue to discuss your comment.

EmmaJP:
https://github.com/auto-wcag/auto-wcag/issues/58

gianwild:
I think we should make sure that people use this only for web applications, not standard web sites. It's a bit like role=menu - only makes sense on a complex web app not a standard web site.
And I always thought accesskeys were ALT + a character - not CTRL. Has that changed?

EmmaJP:
@gianwild I think it's OK for either, as this rule about the uniqueness of the attribute value rather than how suitably it has been applied. That would be a different test, if there is need for one.

gianwild:
Thanks @EmmaJP

PR #54, opened 2017-11-13
SC4-1-1-id revised

### LINKED ISSUE #46: SC4-1-1-id test procedure may only check first id and requires full page load
Current rule https://github.com/auto-wcag/auto-wcag/blob/master/_rules/SC4-1-1-id.md includes:

**Assumptions**
_no known assumptions_

...

**Selector**

Select all elements that match the following CSS selector: *[id]

**Step 1**

- IF there is no list called 'knownIDs':
  - Create an empty list 'knownIDs;
- IF the value of the ID attribute exists in 'knownIDs':
  - Return step1-fail
- ELSE:
  - Add the value of the ID attribute of the selected element to 'knownIDs'
  - Return step1-pass

...

Fail: "The ID of this element has occurred before on this page"

WilcoFiers:
Closes #46.

### DIFF
--- a/_drafts/SC4-1-1-id.md
+++ b/_drafts/SC4-1-1-id.md
@@ -1,0 +1,69 @@
+---
+rule_id: SC4-1-1-id
+name: Define ids for elements
+test_mode: automatic
+environment: DOM Structure
+
+success_criterion:
+- 4.1.1 # Parsing (Level A)
+
+author:
+- Kamyar Rasta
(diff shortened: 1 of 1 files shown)


WilcoFiers:
This rule needs test cases before we publish.

EmmaJP:
https://github.com/auto-wcag/auto-wcag/issues/58

PR #55, opened 2017-11-13
Created SC4-1-1-unique-attrs.md

### LINKED ISSUE #47: Is SC4-1-1-unique-attrs test 'page' wide or element specific?
While I understand that the guideline is element specific, the rule refers to HTML document as a whole as well as the individual elements.

Current rule https://github.com/auto-wcag/auto-wcag/blob/master/_rules/SC4-1-1-unique-attrs.md includes:

...

**Selector**

Select all opening tags in the HTML document

**Step 1**

Make a list of all the attribute names on the current tag

Check if any of the attributes occurs more the once:

Return [step1-fail][#step1-fail]

Else, return [step1-pass][#step1-fail]

...

Fail: "Attribute was used more then once on the element"

WilcoFiers:
Closes #47

### DIFF
--- a/_drafts/SC4-1-1-unique-attrs.md
+++ b/_drafts/SC4-1-1-unique-attrs.md
@@ -1,0 +1,84 @@
+---
+rule_id: SC4-1-1-unique-attrs
+name: Attributes are unique
+test_mode: automatic
+environment: Markup Document
+
+success_criterion:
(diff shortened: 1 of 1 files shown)


EmmaJP:
Issue raised for adding tests ... https://github.com/auto-wcag/auto-wcag/issues/58

PR #56, opened 2017-11-14
Added CannotTell outcome before user questions

josehilera:
This PR replaces [PR#41](https://github.com/auto-wcag/auto-wcag/pull/41), because I have changed the name of the branch, now is "CannotTell" instead "Undetermined". 

To satisfy ACT Rules Format 1.0, I have added a "CannotTell" out come value before user questions if the user is not available. In rules included in _rules folder.

I have include the suggestions from Wilco and Ken in [PR#41](https://github.com/auto-wcag/auto-wcag/pull/41).


WilcoFiers:
Thanks for the update Jose!

EmmaJP:
Looks good ... will definitely be a useful addition for a few of the draft rules.

PR #62, opened 2017-12-06
Add "for each selected item" line to every rule

WilcoFiers:


EmmaJP:
Thank you - good call Carlos

WilcoFiers:
Updated. @EmmaJP @carlosapaduarte Please update your review.

PR #80, opened 2018-03-29
Draft: SC3-1-1-html-lang-valid

jkodu:
- Added Drafts
- Added Tests
- Closes Issue - https://github.com/auto-wcag/auto-wcag/issues/76


Brynanders:
[There is documentation here](https://www.w3.org/International/questions/qa-html-language-declarations) from the W3C which says to "Ensure that the values for both attributes are identical" when serving XHTML. Wouldn't this sufficiently validate the 3rd Failed Test Case?

Brynanders:
For what its worth I also couldn't find any documentation on the ordering of lang attributes but running some basic tests in Chrome and NVDA revealed that only changes to the `lang` attribute had any affect on the way text was read out by NVDA, regardless of in which order `xml;lang` and `lang` appear in the document.

_draft-tests/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,48 @@
...
+```html
+<html xml:lang="fr">
+```
+
+```html
+<html xml:lang="fr" lang="fr">
+```
+
+```html
+<html lang="fr" xml:lang="xyz">
```
WilcoFiers: This is identical to the last fail. The test case should be:

```html
<html lang="xyz" xml:lang="fr">
```

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,43 @@
+---
+name: SC3-1-1-html-lang-valid
+
+description:
+This rule checks the `xml:lang` and/or the `lang` attribute, and that it conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
```
WilcoFiers: This is invalid yaml. Needs to be:
```yaml
description: |
  This rule checks...
```

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,43 @@
...
+
+success_criterion:
+- 3.1.1
+
+test_aspects:
+- DOM Tree: The tree that HTML is parsed into.
```
WilcoFiers: Invalid YAML. Should be:

```yaml
test_aspects:
- DOM Tree # The tree that 
```

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,43 @@
...
+authors:
+- Annika Nietzio
+- Jey Nandakumar
+---
+
+## Test Procedure
+
+### Applicability
+
+The root node of the page, if it is an `<html>` element with a `lang` and/or `xml:lang` attribute.
```
WilcoFiers: It should be `html` element and not `<html>` element.

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Expectation
+
+- The value of the `lang` or `xml:lang` attribute conforms to one of the values from the list - [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
+- NOTE: If `lang` and `xml:lang` exists, then only `lang` is tested for conformance.
```
WilcoFiers: These should be two paragraphs, not two lists. There needs to be an empty line between them. Also the note should start with `**Note**: `, not `NOTE:`.

WilcoFiers:
@Brynanders @kensgists If you completed your review, please approve merging the draft.

_drafts/SC3-1-1-html-lang-valid.md
jkodu: Kept it simple, to just DOM Tree

_draft-tests/SC3-1-1-html-lang-valid.md
jkodu: Don't agree on this one. We are trying to depict a a pass case, where `lang` has the right value and `xml:lang` has an invalid value.

## Passed

```html
<html lang="fr">
```

```html
<html xml:lang="fr">
```

```html
<html xml:lang="fr" lang="fr">
```

```html
<html lang="fr" xml:lang="xyz">
```

## Failed

```html
<html lang="xyz">
```

```html
<html xml:lang="xyz">
```

```html
<html xml:lang="xyz" lang="fr">
```

jkodu:
@kensgists @2biazdk @jribbens @rdeltour @WilcoFiers @annethyme @nitedog @skotkjerra @Brynanders

Pushed changes.
Could do with some approvals/ reviews for this one.

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,44 @@
...
+authors:
+- Annika Nietzio
+- Jey Nandakumar
+---
+
+## Test Procedure
+
+### Applicability
+
+The root node of the page, if it is an `html` element with a `lang` and/or `xml:lang` attribute.
```
kasperisager: The root node of the page will always be a `Document`; you likely mean root element instead.

jkodu: Cool, changed to root element.

jkodu: @kasperisager 
done - review/ approve please.

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Test Procedure
+
+### Applicability
+
+The root element of the page, if it is an `html` element with a `lang` and/or `xml:lang` attribute.
+
+### Expectation
+
+The value of the `lang` or `xml:lang` attribute conforms to one of the values from the list - [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
```
kasperisager: This sentence could be clearer. BCP47 is not a list, but a document describing a "best current practice" for constructing language identifiers. IANA, on the other hand, maintains an authoritative list of tags conforming to BCP 47. Maybe something along the lines of:

> The value of the `lang` or `xml:lang` attribute conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).

One might also add a short description of what BCP 47 actually is.

jribbens: This still has the problem that the text is ambiguous as to whether it is a pass if both attributes are present and only one passes.

_draft-tests/SC3-1-1-html-lang-valid.md
WilcoFiers: > If `lang` and `xml:lang` exists, then only `lang` is tested for conformance.

This means that the following would pass:

```html
<html xml:lang="xyz" lang="fr">
```

_drafts/SC3-1-1-html-lang-valid.md
jkodu: Thanks @kasperisager

_draft-tests/SC3-1-1-html-lang-valid.md
jkodu: You are right. I have edited the test case to:

```html
<html xml:lang="xyz" lang="xyz">
```

jkodu:
@WilcoFiers 
@kasperisager 
Changes done. Kindly review/ approve.

_draft-tests/SC3-1-1-html-lang-valid.md
```diff
@@ -34,7 +34,7 @@ success_criterion:
 ```
 
 ```html
-<html xml:lang="xyz" lang="fr">
+<html xml:lang="xyz" lang="xyz">
```
WilcoFiers: There should be a test case that checks that xml:lang is ignored when lang is set. This doesn't do that.

jkodu: That is the last case in the passed scenario. (The xml:lang is ignored with lang is set).

Although you are right. I have amended this fail scenario like so:
```html
<html xml:lang="en" lang="xyz">
```

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,48 @@
+---
+name: SC3-1-1-html-lang-valid
+
+description: |
+ This rule checks the `xml:lang` and/or the `lang` attribute, and that it conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
```
WilcoFiers: Should be:
> This rule checks the `lang` or `xml:lang` attribute conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Note**: If `lang` and `xml:lang` exists, then only `lang` is tested for conformance.
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+This rule is only applicable in a scenario where assistive technologies are inconsistent about how they determine the language of the page.
```
WilcoFiers: > There are known combinations of a popular operating system with browsers and assistive technologies that do not support the `lang` and `xml:lang` attributes.

annethyme: Did we want Wilco's text to completely replace the paragraph that is now in the Accessibility Support section? I can't see it from the comment, but I think that was what we talked about.

_drafts/SC3-1-1-html-lang-valid.md
jkodu: Done!

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,48 @@
+---
+name: SC3-1-1-html-lang-valid
+
+description: |
+This rule checks the lang or xml:lang attribute conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
```
WilcoFiers: Double space indent needed here

_drafts/SC3-1-1-html-lang-valid.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Applicability
+
+The root element of the page, if it is an `html` element with a `lang` and/or `xml:lang` attribute.
+
+### Expectation
+
+The value of the `lang` or `xml:lang` attribute conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
+
+**Note**: If `lang` and `xml:lang` exists, then only `lang` is tested for conformance.
```
WilcoFiers: Editorial: I would suggest saying "if **both** `lang` and `xml:lang` exist,..."

PR #81, opened 2018-04-05
SC2-1-2-no-keyboard-trap-r1: Added rule and test case

geirsf:


_drafts/SC2-1-2-no-keyboard-trap-r1.md
```diff
@@ -0,0 +1,38 @@
...
+- Carlos Duarte
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+### Applicability
+The rule applies to all elements on a web page where focus can be moved to through keyboard navigation.
+
+### Expectation
+For each target element, focus can be moved to the next element on the web page by using arrow keys, ESC key, TAB key, Shift+TAB or Enter key.
```
WilcoFiers: I don't think the logic of this works. Firstly because the "next element on the page" can be any element, I was going to suggest the "next element in the focus order", but I don't think that's true either. In things like ARIA tabs the next tab shouldn't be in the focus order. You should make sure this rule doesn't become a test that checks that the focus order makes sense. That's entirely different SC, and not the problem you are trying to describe.

This one is difficult to review without knowing what is going to be in the R2 rule.

geirsf:
We close this pull-request because we now see it needs more work.

PR #96, opened 2018-04-16
FINAL: SC3-1-1-html-xml-lang-match

jkodu:
- Create Rule, based on issue -https://github.com/auto-wcag/auto-wcag/issues/77
- Editorial amendments - https://github.com/auto-wcag/auto-wcag/pull/86#pullrequestreview-111679506
- Fixes issue #165


_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -23,9 +23,9 @@ The root node of the page, if it is an `html` element with both `lang` and `xml:
 
 ### Expectation
 
-- The value of the primary language sub-tag for the `lang` and `xml:lang` attributes are the same.
+The value of the primary language sub-tag for the `lang` and `xml:lang` attributes are the same.
```
jribbens: I don't know why this is saying only the "primary language sub-tags" need to match. Why allow mismatches in the other sub-tags? There is no benefit to this and it is a violation of the HTML 5 specification. Just say "The values must be identical when compared case-insensitively".

WilcoFiers: @jribbens This is testing compliance to WCAG 2.0, not to HTML 5. To the best of my knowledge, a mismatch between secondary languages does not lead to accessibility problems, unlike what can happen if the primary language sub-tag is incorrect.

jribbens: @WilcoFiers Clearly a secondary sub-tag mismatch is *less likely* to lead to accessibility problems than a primary mismatch, but stating outright that it *does not* seems brave. I can see no benefit to allowing mismatches, and it seems like erring on the side of incaution.

WilcoFiers: Can you give an example of when getting the secondary sub-tag wrong can be a violation of WCAG? Even for example if you do "en-US" on a page with British spelling, that shouldn't be a violation of WCAG. Even something non-existent such as "en-XYZ" doesn't lead to a violation, so a mismatch between a secondary sub-tag shouldn't make a difference either.

jribbens: Well, `<span lang="en-us">This is a boot sale</span>` vs `<span lang="en-gb">This is a boot sale</span>` have quite different meanings, if that's the sort of thing you're looking for, or similarly with "This is the first floor" :-)

Honestly I don't know if there are languages where words might be mispronounced in a way as to make them ambiguous or incomprehensible if the wrong secondary sub-tag is used, it just seems to bizarre to make the test *more* complicated on the off-chance that *maybe* a situation that will almost never come up won't matter.

I do think there needs to be a discussion as to what the possible outcomes of a auto-wcag test can be... we have pass/fail/inapplicable but I think at least "possible fail" and "warning" are needed. Then if we do want this test to be complicated then it could have "fail" for primary mismatch and "warning" if not identical.

jkodu:
@jribbens 
This is a PR to migrate a draft to a rule.

Would appreciate early comments either in draft PR or in the issue.
- https://github.com/auto-wcag/auto-wcag/pull/86#pullrequestreview-111679506
- https://github.com/auto-wcag/auto-wcag/issues/77

_rules/SC3-1-1-html-xml-lang-match.md
WilcoFiers: Would you agree to covering this with an assumption, something like:

> Although there is no known way that an inappropriate secondary sub-tag in the `lang` or `xml:lang` attribute (such as "en-XYZ") can lead to violations of 3.1.2, it is conceivable that with specific assistive technologies on some languages, there can be exceptions. In that case `lang` and `xml:lang` should have matching secondary sub-tags.

WilcoFiers: It's not. Assumptions are just that, things that weren't addressed by the rule that may lead to incorrect results. It's almost impossible to write a rule that's entirely air tight, so where it isn't, we include that information.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -23,13 +23,13 @@ The root node of the page, if it is an `html` element with both `lang` and `xml:
...
 
-**Note:** language tags are not case sensitive
+**Note:** Language tags are not case sensitive.
```
WilcoFiers: Add note:

> **Note**: HTML 5 specification requires the `lang` and `xml:lang` attributes to match exactly. This is not known to impact accessibility, which is why it is permitted in this rule.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -19,18 +19,18 @@ authors:
 
 ### Applicability
 
-The root node of the page, if it is an `html` element with both `lang` and `xml:lang` attributes.
+The root node of the page, if it is an `html` element with both [non-empty][] `lang` and `xml:lang` attributes.
```
kasperisager: Should be "root element"; the root node will always be a `Document`.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -19,18 +19,18 @@ authors:
...
 
 ### Expectation
 
-- The value of the primary language sub-tag for the `lang` and `xml:lang` attributes are the same.
+The value of the primary language sub-tag for the `lang` and `xml:lang` attributes are the same.
 
-**Note:** language tags are not case sensitive
+**Note:** Language tags are not case sensitive.
```
kasperisager: Maybe add a reference to BCP 47?

jkodu: Available int the background.

jkodu:
Changes made. Please review again.

jkodu:
@WilcoFiers @kasperisager @skotkjerra @Brynanders @annethyme - review appreciated.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,81 @@
+---
+name: SC3-1-1-html-xml-lang-match
+
+description: |
+ The rule checks that for the `html` element, there is no mismatch between the primary language in [non-empty](#non-empty) `lang` and `xml:lang` attributes, if both are used, and that it conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
```
annethyme: Remove ", and that it conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt)", since this is for another rule.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,81 @@
+---
+name: SC3-1-1-html-xml-lang-match
+
+description: |
```
annethyme: In the "view" mode it seems that link insertion is not possible in the description.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Background
+
+- https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/H57
+- https://www.ietf.org/rfc/bcp/bcp47.txt
+- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang
+- https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xml:lang
+- https://www.w3.org/TR/WCAG20-TECHS/H57.html
+
+## Test Cases
```
annethyme: For some of the other rules we have added html comments to explain why each test case has been included. Should we do that here?

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,81 @@
...
+<html lang="en" xml:lang="en">
+```
+
+```html
+<html lang="en" xml:lang="En">
+```
+
+```html
+<html lang="en" xml:lang="en-GB">
+```
```
annethyme: Should we also include the opposite test case of the last one?

```html
<html lang="en-GB" xml:lang="en">
```
I suggest including a passing test case that does not conform to BCP 47: 

```html
<html lang="en-XYZ" xml:lang="en">
```

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,81 @@
...
+- Although there is no known way that an inappropriate secondary sub-tag in the `lang` or `xml:lang` attribute (such as "en-XYZ") can lead to accessibility issues, it is conceivable that with specific assistive technologies on some languages, there can be exceptions. In that case `lang` and `xml:lang` should have matching secondary sub-tags.
+
+## Accessibility Support
+
+Since most assistive technologies will consistently use `lang` over `xml:lang` when both are used, violation of this rule may not necessarily be a violation of WCAG 2. Only when there are inconsistencies between assistive technologies, as to which attribute is used to determine the language, does this lead to a violation of SC 3.1.1.
+
+## Background
+
+- https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/H57
+- https://www.ietf.org/rfc/bcp/bcp47.txt
```
annethyme: Is BCP 47 a relevant resource for this rule when it doesn't test conformance to BCP 47?

_rules/SC3-1-1-html-xml-lang-match.md
WilcoFiers: Yeah, I don't think GH view supports it. I had to do some magic to make markup in the description work in the website. It should work there.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,81 @@
...
+### Applicability
+
+The root element of the page, if it is an `html` element with both [non-empty](#non-empty) `lang` and `xml:lang` attributes.
+
+### Expectation
+
+The value of the primary language sub-tag for the `lang` and `xml:lang` attributes are the same.
+
+**Note**: HTML 5 specification requires the `lang` and `xml:lang` attributes to match exactly. This is not known to impact accessibility, which is why it is permitted in this rule.
+
```
WilcoFiers: Remove double empty line.

_rules/SC3-1-1-html-xml-lang-match.md
kasperisager: https://github.com/auto-wcag/auto-wcag/pull/96#pullrequestreview-121399025

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,94 @@
...
+### Failed
+
+```html
+<html lang="fr" xml:lang="en">
+```
+
+```html
+<!-- lang/xml:lang value does not confirm to BCP 47 -->
+<html lang="en-XYZ" xml:lang="en">
```
Brynanders: This shouldn't be a fail if we agree that BCP 47 isn't relevant for this rule.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -0,0 +1,94 @@
...
+```html
+<svg lang="en" xml:lang="en">
+```
+
+```html
+<html lang="fr" xml:lang="">
+```
+
+```html
+<html lang="" xml:lang="">
```
Brynanders: I think this test case is redundant (and is covered in the previous test case) considering the applicability requires both `lang` and `xml:lang` attributes to be non-empty.

PR #97, opened 2018-04-17
Refactor: Layout & Plugin for - Demo Frame Embedding

jkodu:
This refactor tackles the below.

- Enhanced the layout of various pages. A quick summary via screenshots is as below.
image
image
image

- A plugin for iframe embedding of a code snippet.
Usage:
image

Open for discussion on what the decorating tags <<EMBED_START>> & <<EMBED_END>> should be named.

This creates an output like so:
image


_layouts/default.html
```diff
@@ -1,173 +1,193 @@
...
+      </p>
+    </main>
+    
+    <script src="{{ site.baseurl }}/assets/js/jquery.min.js"></script>
+    <script src="{{ site.baseurl }}/assets/js/bootstrap.min.js"></script>
+    <script>
+      $(document).ready(function () {
+        // append class names to table
+        $('table').each(function() {
+          $(this).addClass('table table-striped table-hover');
```
WilcoFiers: This seems like a problem that could be solved without JS, no?

_layouts/default.html
```diff
@@ -1,173 +1,193 @@
...
-        About Us
-      </a></li>
-      <li><a href="{{ site.baseurl }}/pages/contribute.html">
-        How to contribute
-      </a></li>
-      <li><a href="https://www.w3.org/community/auto-wcag/about-us/">GitHub repository</a></li>
-    </ul>
-  </div>
+        // find every language-html and iframe next to eachother and flex them in 2 column layout
+        $('div.language-html')
```
WilcoFiers: This too seems like it should work without JS.

assets/css/main.scss
```diff
@@ -2,51 +2,171 @@
...
-    color: #FFF;
-    text-decoration:underline;
-  }
-  ul {
-    padding-left:1em;
-  }
-}
\ No newline at end of file
+
+// table {
```
WilcoFiers: Either this shouldn't be commented out, or this shouldn't be here at all :)

development.md
```diff
@@ -0,0 +1,20 @@
...
+
+- bundle install (once, if you do not have ruby, install via home brew).
+- bundle exec jekyll serve
+
+This will serve the site at - http://127.0.0.1:4000/auto-wcag/
+
+All changes are watched and served.
+
+
+<!-- To document
```
WilcoFiers: ??

spec/frame_embed_generator_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+require 'rspec'  
+require 'rack/test'
+
+set :environment, :test
+
+describe 'FrameEmbedGenerator Plugin Testing' do
+
+    include Rack::Test::Methods
+  
+    # TODO: Test
```
WilcoFiers: Should probably change that :-)

Guardfile
```diff
@@ -0,0 +1,73 @@
+# A sample Guardfile
```
WilcoFiers: So this is a watcher for frame_embed_generator.rb? Please document the function of this file.

jkodu: I have removed Guard, and watching using foreman for spec files. As I still have not learnt how to write tests especially for Jekyll Plugins.

On the contrary to writing rspec for a simple functionality like so - https://github.com/auto-wcag/auto-wcag-deployer/blob/master/spec/server_spec.rb

_layouts/default.html
jkodu: Yes, can do whilst listening to the Liquid Template Generation, but it is an unnecessary plugin to author. 

I will see what I can do about this.
Will create an issue to solve in the coming days.

assets/css/main.scss
jkodu: Removed commented styles.

development.md
jkodu: removed...

spec/frame_embed_generator_spec.rb
jkodu: removed spec file, for reason explained in earlier comment.

_layouts/default.html
jkodu: Have to master overriding Liquid Template Generation, will give it a stab once I learn more on that.

_layouts/default.html
jkodu: Same comment as above. Need to master Liquid Template generation.

kasperisager:
Going over this again, I do have some concerns regarding the complexity introduced. We're effectively going from four well-documented and established languages used for authoring pages (YAML, Liquid, Markdown, and HTML) to additionally bolting on our own custom language when we want to do things that we should be able to accomplish in the other languages. On top of that maintainers now also have to know Ruby if they want to touch the custom language that we invent. I see a couple good ways to solve this without introducing too much complexity:

1. Use a templating language that supports components/macros defined in that same templating language.
2. Use web components on the client instead of custom syntax as part of the build process.

All that being said, I still think the end result is great and hat tip to @jkodu for your work on this!

jkodu:
@WilcoFiers 
@kasperisager 

Magic! Got there.
Solved for all the above arguments in commit - https://github.com/auto-wcag/auto-wcag/commit/f95b006b3943115cc292c37c41df4ef486d8beda

This solves the enhancement issue - https://github.com/auto-wcag/auto-wcag/issues/98

No overhead in terms of writing rules, the snippets look like below:
image

The output is as so:
image

jkodu:
@WilcoFiers @skotkjerra @Brynanders 
Need a few more approvals on this to merge.

PR #107, opened 2018-04-27
FINAL: SC3-1-1-html-has-lang

jkodu:
- Rule based on draft - https://github.com/auto-wcag/auto-wcag/pull/85


_rules/SC3-1-1-html-has-lang.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Test Procedure
+
+### Applicability
+
+The root element of the page, if it is an `html` element.
+
+### Expectation
+
+The test target has a `lang` or `xml:lang` attribute.
```
annethyme: Add non-empty here like in the description: 
"The test target has a [non-empty][] `lang` or `xml:lang` attribute."

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+```html
+<html xml:lang="" lang="en">
+```
+
+```html
+<html xml:lang="en" lang="">
+```
+
```
annethyme: I suggest adding a passed test case with a non-valid value too, to make clear that we don't test the validity of the values in this rule, e.g.:

```html
<html lang="xyz">
```

```html
<html xml:lang="xyz">
```

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -0,0 +1,94 @@
+---
+name: Existence of HTML Lang attribute
```
WilcoFiers: Why not "HTML has lang attribute"?

_rules/SC3-1-1-html-has-lang.md
WilcoFiers: Agreed.

_rules/SC3-1-1-html-has-lang.md
jkodu: Smart 🥇

jkodu:
Changes done. Review again.

PR #115, opened 2018-05-01
Draft: SC1-1-1-non-decorative-image-accessible-name

annethyme:


DagfinnRomen:
Images that are non-decorative, but have an empty alt attribute should have an accessible name and should fail the SC. How is this situation handled by the rule?

skotkjerra:
If an image has an empty alt attribute, it is considered marked as decorative by this rule. wether this is correct is not assessed in this rule.

PR #119, opened 2018-05-02
Draft: SC3-1-2-lang

jkodu:
Rule based on issue - https://github.com/auto-wcag/auto-wcag/issues/78

### DIFF
--- a/_drafts/SC3-1-2-lang.md
+++ b/_drafts/SC3-1-2-lang.md
@@ -1,0 +1,75 @@
+---
+name: Validity of HTML Attribute with in elements of page body
+
+description:
+- This rule checks that `lang` and/or `xml:lang` attribute on elements with in the `body` of a web page are correct.
+
+success_criterion:
+- 3.1.2
+
+test_aspects:
+- DOM Tree
+
+authors:
+- Jey Nandakumar
+---
+
+## Test Procedure
+
+### Applicability
+
+- Any DOM element, other than the root DOM node with a `lang` and/or `xml:lang` attribute.
+
+### Expectation
+
+- The value of the `lang` and/or `xml:lang` conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility support
+
+This rule is only applicable in a scenario where assistive technologies are inconsistent about how they determine the language of the page.
+
+## Background
+
+- https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/H58
+- https://www.ietf.org/rfc/bcp/bcp47.txt
+- http://wiki.egovmon.no/wiki/SC3.1.2#Element_descendent-or-self::body.5B.40lang.5D_or_descendent-or-self::body.5B.40xml:lang.5D
+- https://www.w3.org/WAI/WCAG20/quickref/?showtechniques=312#qr-meaning-other-lang-id
+
+## Test Cases
+
+### Selector
+
+Retrieve all DOM nodes with a `lang` and or `xml:lang` attribute, other than the root node
+`descendent-or-self::body[@lang]` -or-  `descendent-or-self::body[@xml:lang]`
+
+### Passed
+
+```html
+<article lang="en">
+```
+
+```html
+<p xml:lang="DE">
+```
+
+```html
+<blockquote lang="fr-CH">
+```
+
+### Failed
+
+```html
+<article lang="dutch">
+```
+
+```html
+<p xml:lang="english">
+```
+
+### Inapplicable
+
+*none*
--- a/_drafts/SC3-1-2-lang.md
+++ b/_drafts/SC3-1-2-lang.md
@@ -11,6 +11,7 @@ test_aspects:
 - DOM Tree
 
 authors:
+- Bryn Anderson
 - Jey Nandakumar
 ---
 


jkodu:
@ShadowBB 
It is not just the body element, but the contents of the body element. Your suggestion does not cover that. Thoughts?

Brynanders:
@jkodu I think we could just add an "s" to @shadowBB suggestion to cover all elements within the body.

ShadowBB:
@Brynanders 

Yes! That was my intention all along! XD

That is a good suggestion Brynanders, and thanks for the feedback @jkodu .

jkodu:
@annethyme - could you add an approval. I will make the editorial change of adding an 's' in the final pr.

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
...
+
+- The value of the `lang` and/or `xml:lang` conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility support
+
+This rule is only applicable in a scenario where assistive technologies are inconsistent about how they determine the language of the page.
```
annethyme: I don't think we should use the term "is only applicable" here, to not confuse this with the Applicability section of the rule.

I suggest something more along these lines: "Most assistive technologies will rely solely on the `lang` attribute for determining the language of the page. The requirement tested in this rule will only give rise to accessibility issues in cases where assistive technologies are inconsistent about how they determine the language of the page."

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
...
+## Background
+
+- https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/H58
+- https://www.ietf.org/rfc/bcp/bcp47.txt
+- http://wiki.egovmon.no/wiki/SC3.1.2#Element_descendent-or-self::body.5B.40lang.5D_or_descendent-or-self::body.5B.40xml:lang.5D
+- https://www.w3.org/WAI/WCAG20/quickref/?showtechniques=312#qr-meaning-other-lang-id
+
+## Test Cases
+
+### Selector
```
annethyme: The new ACT Rules format doesn't have test cases for "Selector"...

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
...
+
+Retrieve all DOM nodes with a `lang` and or `xml:lang` attribute, other than the root node
+`descendent-or-self::body[@lang]` -or-  `descendent-or-self::body[@xml:lang]`
+
+### Passed
+
+```html
+<article lang="en">
```
annethyme: Should the tags in these test cases have closing tags too, e.g. <article lang="en"></article> ?

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
+---
+name: Validity of HTML Attribute with in elements of page body
+
+description:
+- This rule checks that `lang` and/or `xml:lang` attribute on elements with in the `body` of a web page are correct.
```
annethyme: Maybe "correct" is too broad a term? This could imply that the rule should also check of it has been correctly applied to the content. 
Consider aligning it with the description in https://github.com/auto-wcag/auto-wcag/pull/109/files

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
...
+<article lang="dutch">
+```
+
+```html
+<p xml:lang="english">
+```
+
+### Inapplicable
+
+*none*
```
annethyme: If we follow the logic from the other language rules, inapplicable test cases are empty lang attributes (see https://github.com/auto-wcag/auto-wcag/pull/109/files). 

However, the applicability of the other one narrows it down to "non-empty attributes", this one's applicability doesn't. I think we create some inconsistency if they are not scoped similarly.

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
...
+authors:
+- Bryn Anderson
+- Jey Nandakumar
+---
+
+## Test Procedure
+
+### Applicability
+
+- Any DOM element, other than the root DOM node with a `lang` and/or `xml:lang` attribute.
```
annethyme: Is it in fact any DOM element or just the ones that are e.g. exposed to assistive technologies that should be applicable to this rule?

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
+---
+name: Validity of HTML Attribute with in elements of page body
```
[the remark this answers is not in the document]
jkodu: @annethyme 
- The new title is - Validity of lang attribute value in html body elements
- Took care of within.

_drafts/SC3-1-2-lang.md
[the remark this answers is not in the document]
jkodu: @annethyme - ok.

_drafts/SC3-1-2-lang.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Test Procedure
+
+### Applicability
+
+- Any DOM element, other than the root DOM node with a `lang` and/or `xml:lang` attribute.
+
+### Expectation
+
+- The value of the `lang` and/or `xml:lang` conforms to [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt).
```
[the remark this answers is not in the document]
jkodu: @annethyme - leaving this as is.

PR #122, opened 2018-05-07
Rule: SC1-1-1-image-has-name

jkodu:
- Create FINAL Rule, based on PR https://github.com/auto-wcag/auto-wcag/pull/115


kasperisager:
@WilcoFiers Making this rule apply to decorative images as well would result in a bit of a tautology as they would always pass since they don't need an accessible name.

WilcoFiers:
@kasperisager The problem I have with making "decorative" part of applicability, is that you're hiding the fact that marking an image as decorative is a technique to have this pass. I think either this needs to be part of the expectation, or we need to create a small group of it for `img` elements that pass it in rule 1) When it is decorative or 2) When there is an accessible name. I know the logic of what has been proposed here works, but it obfuscates one of the remediation options in the applicability.

kasperisager:
I tend to disagree. Yes, marking as image as "decorative" just so happens to make AT treat it as something other than an image, but this shouldn't be something the rule explicitly accounts for. It should, in my opinion, keep things simple by just stating that it applies to "semantic images"; whatever techniques might be used for taking something that would otherwise be treated as an image and making it not be treated as such shouldn't matter.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,21 +19,21 @@ authors:
 
 ### Applicability
 
-The rule applies to any HTML element with the [semantic role][] of `image` that is [exposed to assistive technologies][].
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `image` that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
```
Brynanders: this links don't work correctly, they load this rule and not the algorithm.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,21 +19,21 @@ authors:
...
 
-**NOTE**: Any image marked as decorative, by using either `role="presentation"`, `role="none"` or an empty `alt` attribute in combination with either no `title` attribute or an empty `title` attribute, does not have the role of `image` according to [Accessible Rich Internet Applications (WAI-ARIA) 1.1](https://www.w3.org/TR/wai-aria-1.1/)
+**NOTE**: Any image marked as decorative, by using either `role="presentation"`, `role="none"` or an empty `alt` attribute, does not have the role of `image` according to [ARIA in HTML](https://www.w3.org/TR/html-aria/#img-alt).
 
 ### Expectation
 
-Each target element has an [accessible name][] that is [non-empty][].
+Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty).
```
Brynanders: Algorithm links don't work.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,21 +19,21 @@ authors:
...
 
 ## Assumptions
 
 _There are currently no assumptions_
 
 ## Accessibility Support
 
-Certain assistive technologies can be set up to ignore the `title` attribute, which means that to some users the `title` attribute will not act as an accessible name.
+Certain assistive technologies can be set up to ignore the `title` attribute, which means that to some users the `title` attribute will not act as an [accessible name](#accessible-name).
```
Brynanders: Either `title` should be considered as not supported as removed as a test case, or be included in the 'accessible name' algorithm.

kasperisager:
@ShadowBB Your point about `title` is valid, though I would suggest we include a note in the accessibility support section instead.

annethyme:
No browser support for title as accessible name in Chrome + JAWS. Accessibility Support section needs to be updated accordingly.

ShadowBB:
@kasperisager I agree.

_rules/SC1-1-1-image-has-name.md
Brynanders: @jkodu I can't remember what we said about this, does this need to get resolved now or will it resolve itself once this is published?

Brynanders:
@ShadowBB I haven't removed the Pass test case for `title` because it can be technically used (according to spec) for providing an accessible name. Do you agree with that logic?

ShadowBB:
@Brynanders After having read this once more I have to agree with you and have now approved this PR: https://github.com/auto-wcag/auto-wcag/blob/master/pages/structure/accessibility-support.md

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,20 +19,18 @@ authors:
...
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `image` that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
 
-**NOTE**: Any image marked as decorative, by using either `role="presentation"`, `role="none"` or an empty `alt` attribute in combination with either no `title` attribute or an empty `title` attribute, does not have the role of `image` according to [Accessible Rich Internet Applications (WAI-ARIA) 1.1](https://www.w3.org/TR/wai-aria-1.1/)
+**NOTE**: Any image marked as decorative, by using either `role="presentation"`, `role="none"` or an empty `alt` attribute, does not have the role of `image` according to [ARIA in HTML](https://www.w3.org/TR/html-aria/#img-alt).
 
 ### Expectation
 
-Each target element has an [accessible name][] that is [non-empty][].
-
-## Assumptions
```
WilcoFiers: Why did the assumptions section get removed? Clearly there is an assumption here that role=img is only given to actual images.

kasperisager:
@WilcoFiers I agree on your point regarding `alt=""`. I'm guessing the issue is partly caused by the fact that ARIA in HTML is still a working draft. Should we maybe remove the associated note and add an additional paragraph to the accessibility support section?

_rules/SC1-1-1-image-has-name.md
Brynanders: What if we changed the applicability and expectation to the following...
Applicability
The rule applies to HTML image elements or any HTML element with the semantic role of image that is exposed to assistive technologies.

Expectation
Each target element has an accessible name that is non-empty or is marked as decorative.

NOTE: An image can be marked as decorative, by using either role="presentation", role="none" or an empty alt attribute alt="".

_rules/SC1-1-1-image-has-name.md
Brynanders: @jkodu can you confirm the above please?

Brynanders:
@WilcoFiers @kasperisager what if we changed the applicability and expectation to the following...
Applicability
The rule applies to HTML image elements or any HTML element with the semantic role of image that is exposed to assistive technologies.

Expectation
Each target element has an accessible name that is non-empty or is marked as decorative.

NOTE: An image can be marked as decorative, by using either role="presentation", role="none" or an empty alt attribute alt="".

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,20 +19,18 @@ authors:
 
 ### Applicability
 
-The rule applies to any HTML element with the [semantic role][] of `image` that is [exposed to assistive technologies][].
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `img` that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
```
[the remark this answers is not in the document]
WilcoFiers: It's in a PR: https://github.com/auto-wcag/auto-wcag/pull/153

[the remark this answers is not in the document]
WilcoFiers: (Could use your review as well Carlos)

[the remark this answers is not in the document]
carlosapaduarte: OK. I'll do it.
Strange that I used Find File to look for it and it didn't. But now it does...

WilcoFiers:
@carlosapaduarte Please review.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,21 +19,17 @@ authors:
...
-
-**NOTE**: Any image marked as decorative, by using either `role="presentation"`, `role="none"` or an empty `alt` attribute in combination with either no `title` attribute or an empty `title` attribute, does not have the role of `image` according to [Accessible Rich Internet Applications (WAI-ARIA) 1.1](https://www.w3.org/TR/wai-aria-1.1/)
+The rule applies to HTML `img` elements or any HTML element with the [semantic role](#semantic-role) of `img` that is exposed to assistive technologies.
 
 ### Expectation
 
-Each target element has an [accessible name][] that is [non-empty][].
-
-## Assumptions
+Each target element has an accessible name that is non-empty or is marked as decorative.
```
carlosapaduarte: Link to definitions of accessible name, non-empty and decorative

jkodu: Updated, please review again.

jkodu:
@carlosapaduarte - Updated, plz review again.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -19,21 +19,17 @@ authors:
 
 ### Applicability
 
-The rule applies to any HTML element with the [semantic role][] of `image` that is [exposed to assistive technologies][].
-
-**NOTE**: Any image marked as decorative, by using either `role="presentation"`, `role="none"` or an empty `alt` attribute in combination with either no `title` attribute or an empty `title` attribute, does not have the role of `image` according to [Accessible Rich Internet Applications (WAI-ARIA) 1.1](https://www.w3.org/TR/wai-aria-1.1/)
+The rule applies to HTML `img` elements or any HTML element with the [semantic role](#semantic-role) of `img` that is exposed to assistive technologies.
```
Brynanders: This needs punctuation to clearly indicate that "exposed to AT" is applicable to both HTML `img` elements and HTML elements with the semantic role of `img`.
For example:
The rule applies to HTML `img` elements or any HTML element with the [semantic role](#semantic-role) of `img`, that are exposed to assistive technologies.

Brynanders: Could this also be simplified to simply say...
The rule applies to any HTML element with the [semantic role](#semantic-role) of `img` that is exposed to assistive technologies.

PR #129, opened 2018-05-08
Rule: SC4-1-2-button-has-name

jkodu:
Final PR based on draft PR - https://github.com/auto-wcag/auto-wcag/pull/118


kasperisager:
What's the deal with leaving out "input elements of `type="image"`"?

kasperisager:
`<input type='image' vale='download'>` => `<input type='image' value='download'>`

kasperisager:
I'd love to see some more test cases as well.

WilcoFiers:
This test case is incorrect:

```html
<button value='Save Page'></button>
```

The `value` attribute does not give accessible names to button elements. It does do this for input element, so I understand the confusion.

WilcoFiers:
```html
<button aria-hidden='true'>
</button>
```

This looks like a failure to me, not a pass, since it is visible on the page, and because it is focusable it I would think it is also exposed to assistive technologies. That definition will need to be finished and part of this PR before we can complete this rule.

I also agree with Kasper that there are test cases missing, especially inapplicable cases.

WilcoFiers:
@kasperisager The reason to isolate `type=image` inputs is because those are also images, so 1.1.1 is applicable to them. I think a separate rule that combines `type=image` and `area` elements makes more sense.

kasperisager:
Per ARIA, an `input` of `type=image` can _never_ be a semantic image though: https://www.w3.org/TR/html-aria/#input-image. I therefore don't quite see how it also falls under 1.1.1?

jkodu:
@WilcoFiers - all algorithm pages are in place on the base already. Like visible on page etc. No need to cherry pick across.

WilcoFiers:
@kasperisager An element's role factor in when determining if that element is non-text content.

> **non-text content**
> any content that is not a sequence of characters that can be programmatically determined or where the sequence is not expressing something in human language
> https://www.w3.org/TR/WCAG21/#dfn-non-text-content

I think `input[type=image]` can reasonably considered content, and it isn't a sequence of characters, so I think it's a safe bet to say it's applicable. There's also this technique suggesting that it is applicable: https://www.w3.org/TR/WCAG20-TECHS/H36.html

kasperisager:
Alright, a separate rule it is then 👍

_rules/SC4-1-2-button-has-name.md
```diff
@@ -19,7 +19,7 @@ authors:
 
 ### Applicability
 
-The rule applies to any element with the role of `button`, except for `input` elements of `type="image"` that are [visible on the page][] or [exposed to assistive technologies][].
+The rule applies to any element with the [semantic role](#semantic-role) of `button`, except for `input` elements of `type="image"` that are [visible on the page][] or [exposed to assistive technologies][].
```
WilcoFiers: This is ambiguous. Does the exception only apply to element visible and exposed to assistive technologies, or does it refer to all buttons?

jkodu:
@ShadowBB @kasperisager @EmmaJP @cpandhi @DagfinnRomen @annethyme @Brynanders  - please review again.

WilcoFiers:
@ShadowBB @DagfinnRomen @carlosapaduarte Could you please review this rule?

ShadowBB:
I would review this again but by comment above has not been addressed. I believe that the first failure test case is suppose to pass.

ShadowBB:
@WilcoFiers The new test case follows H91 and should pass right? https://www.w3.org/TR/WCAG20-TECHS/H91.html

why do buttons not get their accessible name from the value attribute?

_rules/SC4-1-2-button-has-name.md
```diff
@@ -0,0 +1,137 @@
...
+<button aria-label="My button"></button>
+```
+
+```html
+<!-- span tag with role button and has name defined by aria-label -->
+<span role="button" aria-label="My button"></button>
+```
+
+```html
+<!-- summary element has a default sematic role of button -->
```
carlosapaduarte: r/sematic/semantic

_rules/SC4-1-2-button-has-name.md
```diff
@@ -0,0 +1,137 @@
...
+  </style>
+  <body>
+    <!-- Off screen elements should be tested -->
+    <button class='notInPage'>Save</button>
+  </body>
+</html>
+```
+
+## Failed
+
```
carlosapaduarte: Add one failed test case of an off-screen element without an accessible name

jkodu: Sorted.

jkodu:
@carlosapaduarte - please review again.

PR #152, opened 2018-06-01
Rule: SC3-3-2-form-field-has-name

Brynanders:
# Checklist for Pull Request
🚨Please review the [guidelines for contributing](https://auto-wcag.github.io/auto-wcag/pages/contribute.html) to this repository.

- [x] Kindly prefix the pull request name with `DRAFT` or `FINAL` accordingly. For more details, read [Rule Design Guidelines](https://auto-wcag.github.io/auto-wcag/pages/design/review.html)
- [x] Provide a reference to the issue being addressed.

#Reviewing Guidelines
- For `DRAFT` PR's, leave any comments you have and approve if you’ve reviewed the entire pull request. 
- Once there have been three such reviews the pull request can be merged. 
- Any comments left will be taken up in an updated pull request, which will be necessary before a FINAL release”.
- For `FINAL` PR's, review and approve changes to be merged.

# How to Review And Approve

### Description
Please describe your pull request.

💔Thank you!


_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,78 @@
...
+authors:
+- Anne Thyme Nørregaard
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rules applies to:
```
WilcoFiers: I don't think this list passes the "plain language" requirement.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,78 @@
...
+- any element that is or [exposed to assistive technologies](#exposed-to-assistive-technologies),
+- that has a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and
```
WilcoFiers: This has two meanings, and so it's not objective (as required by ACT-RF). This can mean either roles that inherit from select or input in ARIA 1.1, or roles defined in any future standard (regardless of who published it!) and use ARIA 1.1 select to inherit from.

It seems to me this should only be the first one... although a better solution IMO is to just use the list from the note. That solves "plain language" as well. I realise this is trying to be future proof, but in my view it's doing quite the opposite, tying this to the inheritance in ARIA 1.1. If in ARIA 1.2 `menuitemcheckbox` or some other role is changed to inherit from something else, this rule has to be updated to fit.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,78 @@
...
+## Accessibility Support
+
+Certain assistive technologies can be set up to ignore the title attribute, which means that to some users the title attribute will not act as an accessible name.
+
+## Background
+
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
+
+## Test cases
```
WilcoFiers: More test cases please. Especially around use of the role attribute and disabled elements.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,78 @@
...
+- Anne Thyme Nørregaard
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rules applies to:
+- any element that is or [exposed to assistive technologies](#exposed-to-assistive-technologies),
```
WilcoFiers: Depending on how that definition shakes out - should it include focusable elements as well? What about visible on the page?

carlosapaduarte: Shouldn't the applicability be restricted to form elements? Or is it to be applied to elements with the semantic role described below even if they are not used inside a form?

_drafts/SC3-3-2-form-field-has-name.md
carlosapaduarte: When reading the rule I was only thinking about ARIA 1.1. But Wilco's comment makes all the sense.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,78 @@
...
+- that has a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
+- does not have a required context role that itself inherits from one of those roles.
```
carlosapaduarte: You could link required context role also to its ARIA definition: https://www.w3.org/TR/wai-aria/#scope

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,78 @@
...
+_There are currently no assumptions_
+
+## Accessibility Support
+
+Certain assistive technologies can be set up to ignore the title attribute, which means that to some users the title attribute will not act as an accessible name.
+
+## Background
+
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
```
carlosapaduarte: The two entries are the same

_drafts/SC3-3-2-form-field-has-name.md
kasperisager: I do not agree that hard coding a list of roles that only the rule authors know the reasoning behind is a good idea. That's a maintenance nightmare waiting to happen. As for limiting it to ARIA 1.1, I don't really see the point. If ARIA 1.2 indeed changes a role that was previously considered a form control to inherit from something else, then the rule should no longer apply to that role. All form controls inherit from either `input` or `select` and I don't see that changing in the future. In the contrary, if a role is changed in ARIA 1.2 to inherit from `input` or `select` then the rule needs no further changes to accommodate this.

Brynanders:
I made some commits without resolving all outstanding issues. This is not ready for another review yet.

_drafts/SC3-3-2-form-field-has-name.md
Brynanders: @WilcoFiers @carlosapaduarte can you please comment on kasper's reply.

_drafts/SC3-3-2-form-field-has-name.md
Brynanders: @WilcoFiers are you referring to the use of a ":" and a list? I was under the impression that plain language meant readable and understandable and I think that creating a single paragraph here goes against that from a readability perspective.

_drafts/SC3-3-2-form-field-has-name.md
Brynanders: @WilcoFiers we propose that focusable elements be covered in the definition for 'exposed to assistive technologies'
We also don't think that the applicability should include elements that are simply 'visible on the page' on the premise that this rule is only concerned with ensuring AT form controls are labeled.

Brynanders: @carlosapaduarte we don't want to restrict the rule to from elements as a form control could exist outside of a form element but would still need an accessible name.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,77 @@
...
+---
+
+## Test procedure
+
+### Applicability
+
+This rule applies to:
+- any element that is [exposed to assistive technologies](#exposed-to-assistive-technologies), and
+- that has a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
+- does not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
```
WilcoFiers: I don't even know what that means. This clearly doesn't pass the "plain language" requirement - just give people a list of roles to look for.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Note**: All elements that are semantic form controls, either implicitly through native HTML elements or explicitly through use of ARIA roles, inherit from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` roles. The required context role defines the owning container in which a given role is allowed. Form controls with a required context role that is itself a form control, e.g. `<option>` within a `<select>`, are not applicable to this rule as only the top-most form control requires an accessible name.
+
+**Note**: At the time of writing this rule the semantic roles that inherit from the abstract `input` or `select` role and do not have a required context role that itself inherits from one of those roles are: `checkbox`, `combobox`, `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider, spinbutton`, `switch`, `textbox`.
```
WilcoFiers: There is no "at the time of writing" - this is what the ARIA 1.1 spec says, that won't change. Maybe ARIA 1.2 comes along and this changes, or some module is added. If I write my own ARIA module tomorrow, inherits from the ARIA 1.1 select role, that doesn't immediately mean it gets included in this list. This needs to be unambiguous - we can't be arguing over who's spec we consider inheritance from.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,77 @@
...
+## Background
+
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
+
+## Test cases
+
+### Passed
+
+```html
+<input type="text" name="firstname" id="firstname" />
```
WilcoFiers: `name` doesn't give an accessible name to elements. This is a failure.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Accessibility Support
+
+Certain assistive technologies can be set up to ignore the title attribute, which means that to some users the title attribute will not act as an accessible name.
+
+## Background
+
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
+
+## Test cases
```
WilcoFiers: This rule needs more test cases.

_drafts/SC3-3-2-form-field-has-name.md
kasperisager: Last point could be changed to:

> - does not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from the `input` or `select` role.

Does that help?

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,77 @@
+---
+name: Form field has accessible name
+description: |
```
jkodu: nit: break new line between name and description

_drafts/SC3-3-2-form-field-has-name.md
jkodu: I think point 2 and 3 can be joined to one.

- that has a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and does not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role.

_drafts/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Background
+
+- https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html
+
+## Test cases
+
+### Passed
+
+```html
```
jkodu: Please add comments/ description for each test case

WilcoFiers:
@kasperisager @carlosapaduarte do you mind taking another look. We've updated the applicability and added a bunch of test cases.

_rules/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,140 @@
...
+<input />
+```
+
+```html
+<!-- Non-focusable still need an accessible name -->
+<input tabindex="-1" />
+```
+
+```html
+<!-- aria-Label with empty text string -->
```
carlosapaduarte: editorial: "aria-Label" to "aria-label"

_rules/SC3-3-2-form-field-has-name.md
```diff
@@ -0,0 +1,140 @@
...
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+This rule applies to any element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and that has one of the following [semantic roles](#semantic-role): `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider, spinbutton`, `switch`, `textbox`.
+
+**Note**: The list of roles is derived by taking all the ARIA 1.1 roles that:
```
carlosapaduarte: Shouldn't we link to the ARIA 1.1 recommendation?

Brynanders:
@carlosapaduarte I have added your change requests, please review and approve.

PR #167, opened 2018-06-24
Rule: SC4-1-2-role-has-required-states-and-properties

jkodu:
# Checklist for Pull Request
🚨Please review the [guidelines for contributing](https://auto-wcag.github.io/auto-wcag/pages/contribute.html) to this repository.

- [x] Kindly prefix the pull request name with `DRAFT` or `FINAL` accordingly. For more details, read [Rule Design Guidelines](https://auto-wcag.github.io/auto-wcag/pages/design/review.html)
- [x] Provide a reference to the issue being addressed.

#Reviewing Guidelines
- For `DRAFT` PR's, leave any comments you have and approve if you’ve reviewed the entire pull request. 
- Once there have been three such reviews the pull request can be merged. 
- Any comments left will be taken up in an updated pull request, which will be necessary before a FINAL release”.
- For `FINAL` PR's, review and approve changes to be merged.

# How to Review And Approve

### Description
Final PR based on Draft - https://github.com/auto-wcag/auto-wcag/pull/163

💔Thank you!


kasperisager:
I sure hope I'm wrong for the same reason you mention; it would make implicit values pretty useless.

annethyme:
@ShadowBB and @WilcoFiers, I think @Brynanders and I have handled all of your comments. 
Could you guys review again?

ShadowBB:
@WilcoFiers If it would not be mandatory, this entire rule (how it is currently written) would be moot correct?

jkodu:
@annethyme - could you please look into the above Q.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+### Failed
+
+```html
+<!-- Element does not list required states and properties (´aria-controls´ is required property for ´combobox´) -->
+<select role="combobox"></select>
+```
+
+```html
```
EmmaJP: Wondering if the pass an fail examples should be in the same order both times, or if that matters.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,75 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has a [semantic role](#semantic-role).
```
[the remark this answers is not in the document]
WilcoFiers: Some more thoughts. For the aria-valid-attribute-value rule I'm going to allow elements not exposed to assistive technologies, otherwise elements with aria-hidden="true" would be inapplicable instead of passed. I think we should be consistent between these rules about the exposed to assistive technologies part.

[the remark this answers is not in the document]
kasperisager: Per the HTML AAM, the selectedness of `<option>` is mapped to `aria-selected` (https://www.w3.org/TR/html-aam/#details-id-103) so `aria-selected` is implicitly set.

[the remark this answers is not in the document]
kasperisager: Unless there's a case of an element that is implicitly mapped to an ARIA role, but doesn't define native attributes that are mapped to all of the required ARIA states and properties, I'd also agree that we can leave out elements with implicit roles.

_rules/SC4-1-2-role-has-required-states-and-properties.md
annethyme: Agree. Updated now.

annethyme:
All comments should be handled now. Please review again, @WilcoFiers, @EmmaJP and @kasperisager, and see if you can approve. 

I hope that we don't have to wait until the ARIA Working Group clarifies the implicit values question for us to push this rule through our workflow, but can make do info about in the Assumptions section.

However, now that we rely on user agent support for default values, it seems that the ONLY attribute that can ever be required is ´aria-controls´. Which has narrowed down the scope of the rule quite extensively.

_rules/SC4-1-2-role-has-required-states-and-properties.md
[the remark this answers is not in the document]
annethyme: @WilcoFiers: I agree that we should have consistency between rules, but I don't see it as a good solution to create consistency between these two rules and sacrificing the consistency between them and the rest that handle things that are (primarily) important for screen reader users, e.g. all of rules that checks for accessible names.

jkodu:
@annethyme - I have resolved the conflicts now.

Brynanders:
Hi @jkodu could you give this a review?

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role. 
+
+### Expectation
+
+For each test target all attributes listed under [required states and properties](https://www.w3.org/TR/wai-aria/#requiredState) in [WAI-ARIA](https://www.w3.org/TR/wai-aria) for that role are present and [non-empty](#non-empty), unless they have a default value listed under [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole).
```
nitedog: "For each test target all ..." -> "For each test target, all ..."

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+The conflicting information around user agent support for implicit values for roles has been raised with the ARIA Working Group as https://github.com/w3c/aria/issues/787. Until the Auto-WCAG Community Group gets clarification from the ARIA Working Group, we assume for this rule, that relying on the default values is enough to satisfy this accessibility requirement.
+
+See more about the possible implications of any lacking user agent support for implicit values for roles in the ”Accessibility Support” section for this rule.
+
+## Accessibility Support
+
+If browsers and/or assistive technologies do not support leaving out [required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) if an [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole) is specified in [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole), passing this rule may not be enough to avoid accessibility issues due to missing required states and properties. Testing with assistive technologies have however showed that there seems to be general support for implicit values for roles.
```
nitedog: This is a confusing sentence. Is it correct to say the following instead?
[[
This rule relies on browsers and assistive technologies that leave out [required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) when an [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole) is specified in [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole).
]]

nitedog: Note that the last link refers to a section within the WAI-ARIA specification but the link name does not indicate that. It should reflect the name of the section (the target).

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+Any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role. 
+
+### Expectation
+
+For each test target all attributes listed under [required states and properties](https://www.w3.org/TR/wai-aria/#requiredState) in [WAI-ARIA](https://www.w3.org/TR/wai-aria) for that role are present and [non-empty](#non-empty), unless they have a default value listed under [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole).
+
+**Note:** This rule does not test whether the required states and properties has a correct value, only that the attributes are present and [non-empty](#non-empty).
+
+## Assumptions
```
nitedog: This section seems really long. Can it be cut down to the essential bits? For example, do we need the first two paragraphs? They seem to re-explain what is already defined by the applicability and expectation, or what is stated later in the accessibility support section. Putting this information in to different places will likely lead to confusion.

_rules/SC4-1-2-role-has-required-states-and-properties.md
WilcoFiers: This can be simplified quite a bit:
- Make the "non-empty" thing its own expectation
- Make the expectation something like: "For each test target, the required attributes for the role is present, unless the attribute has a default value. The list of required attributes for a role can be found in ARIA 1.1" That breaks it up, makes it easier to read.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role.
```
WilcoFiers: Again, ARIA can impact elements that are not exposed to assistive technologies. The rule shouldn't be limited in that way.

_rules/SC4-1-2-role-has-required-states-and-properties.md
WilcoFiers: I second that, lets see if we can simplify this.

_rules/SC4-1-2-role-has-required-states-and-properties.md
WilcoFiers: This needs some editorial work. Shorter, clearer sentences.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+If browsers and/or assistive technologies do not support leaving out [required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) if an [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole) is specified in [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole), passing this rule may not be enough to avoid accessibility issues due to missing required states and properties. Testing with assistive technologies have however showed that there seems to be general support for implicit values for roles.
+
+**Note:** The required states and properties that could be impacted by this can be found in the Core Accessibility API Mappings 1.1 [Overview of default values for missing required attributes](https://www.w3.org/TR/core-aam-1.1/#authorErrorDefaultValuesTable).
+
+## Background
+
+- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
+- [WAI-ARIA required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState)
+-  [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)
```
WilcoFiers: Remove the double spacing in "`-  [RFC 3986]`"

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+Element has required states (no properties required for this role)
+
+```html
+<div role="option" aria-selected="false"></div>
+```
+
+#### Passed example 2
+
+Implicit value for roles: ´option´ role has an implicit value (´false´) specified for the required state ´aria-selected´
```
WilcoFiers: Lets have a case where the attribute is defined without a value, and one with only whitespace.

_rules/SC4-1-2-role-has-required-states-and-properties.md
WilcoFiers: Should elements without required states/props be applicable?

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -18,45 +18,29 @@ authors:
...
+Any HTML or SVG element that has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role. 
 
 ### Expectation
 
-For each test target all attributes listed under [required states and properties](https://www.w3.org/TR/wai-aria/#requiredState) in [WAI-ARIA](https://www.w3.org/TR/wai-aria) for that role are present and [non-empty](#non-empty), unless they have a default value listed under [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole).
+For each test target, the [required states and properties](https://www.w3.org/TR/wai-aria/#requiredState) for the role is present, unless the attribute has a default value listed under [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole). The list of required attributes for a role can be found in [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria)"
```
kasperisager: "for the role _are_".

_rules/SC4-1-2-role-has-required-states-and-properties.md
annethyme: I removed almost all of it. Let me know, if we still needed any of the bits.

_rules/SC4-1-2-role-has-required-states-and-properties.md
annethyme: Test cases added

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any HTML or SVG element that has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role. 
+
+### Expectation
+
+For each test target, the [required states and properties](https://www.w3.org/TR/wai-aria/#requiredState) for the role are present, unless the state or property has a default value listed under [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole). The list of required states and properties for a role can be found in [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria)"
```
carlosapaduarte: The second sentence can be moved to a note, since it is not part of the expectation, but just something to help assess it.

_rules/SC4-1-2-role-has-required-states-and-properties.md
carlosapaduarte: This assumption is confusing to me for the following reasons:
1 - The applicability is for elements with an explicit semantic role. This assumption applies to elements with implicit semantic roles. 
2 - The assumption's reason to exist seems to be to justify the applicability. If that is case, then why isn't it just a note on the applicability?
3 - It is hard to read. There are too many negatives (no element that has an implicit semantic role + does not define native attributes). Can't this be rephrased?

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+**Note:** This rule does not test whether the required states and properties has a correct value, only that the attributes are present and [non-empty](#non-empty).
+
+## Assumptions
+
+- This rule assumes that there is no element that has an [implicit semantic role](#semantic-role), and at the same time does not define native attributes that are mapped to all of the [required states and properties](https://www.w3.org/TR/wai-aria/#requiredState) for this role. This assumption is the basis for limiting the Applicability to [explicit semantic roles](#semantic-role) only.
+
+## Accessibility Support
+
+This rule relies on browsers and assistive technologies to support leaving out [required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) when an [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole) is specified in [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1).
```
carlosapaduarte: Once more, the applicability is for elements with explicit semantic roles, and this section mentions issues for elements with implicit semantic roles. I'm sorry, but I must be having problems with understanding this rule.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+Implicit value for roles: ´option´ role has an implicit value (´false´) specified for the required state ´aria-selected´, here defined with only whitespace instead of a value
+
+```html
+<div role="option" aria-selected=" "></div>
+```
+
+#### Passed example 5
+
+Element has required properties (no states required for this role)
```
carlosapaduarte: There is a required state (`aria-expanded`) but it has a default value, so this is a pass anyway. But the comment needs to be updated.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Background
+
+- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
+- [WAI-ARIA required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState)
+- [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)
+
+## Test Cases
+
+### Passed
```
carlosapaduarte: Add a test case with a value for a state or a property that is non-empty but not correct. The rule should still pass since it does not test the correctness of this value.

_rules/SC4-1-2-role-has-required-states-and-properties.md
annethyme: @carlosapaduarte,
1 - Assumption changed to make relation between rule and assumption clearer
2  - My impression is that all assumptions that could change the outcome of the rule if they were not true should be listed under Assumptions, whereas notes are more informative.
3 - Tried rephrasing. Hope the meaning is conserved.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Applicability
+
+Any HTML or SVG element that has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role. 
+
+### Expectation
+
+For each test target, the [required states and properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) for the role are present, unless the state or property has a default value listed under [implicit value for role](https://www.w3.org/TR/wai-aria-1.1/#implictValueForRole). 
+
+**Note:** The list of required states and properties for each role can be found in [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1).
```
nitedog: Minor suggestion: change the link text in the above paragraph from "required states and properties" to "WAI-ARIA required states and properties" or to "required states and properties in WAI-ARIA" or to "required states and properties (WAI-ARIA)", so that you clarify the link target in the link text and can get rid of this note.

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -0,0 +1,141 @@
...
+
+**Note:** This rule does not test whether the required states and properties has a correct value, only that the attributes are present and [non-empty](#non-empty).
```
nitedog: "whether the required states and properties has a correct value" -> "whether the required states and properties have correct values"

_rules/SC4-1-2-role-has-required-states-and-properties.md
nitedog: Awesome work cutting this down!

_rules/SC4-1-2-role-has-required-states-and-properties.md
nitedog: Maybe prefixing or suffixing these terms with "WAI-ARIA" will help clarify the source of these terms (as well as the target of the link)?

PR #185, opened 2018-07-12
chore: site refactor (layout, styling and plugin updates)

jkodu:
# Description
This PR tackles website updates that cover the below:
- Fix 404 for test iframe generation from test code snippets
- Delete drafts from navigation header
- Syntax highlight(js) for all code snippets.
- Remove references to testcase.html in config.yml and other layout templates
- Layout change in rules table, add description column
- Change directory and usages of algorithm to definitions (./root/_definitions) (update ruby plugin to watch correct directory for auto generation)
- Styling changes, other layout updates & links to have underline
- Preliminary work for test case generation

Closes Issue: 
- https://github.com/auto-wcag/auto-wcag/issues/169

Screenshots of updates:
image

image

image

## Pull Request Checks

## How to Review And Approve


WilcoFiers:
Couple of thoughts:
1. How are we going to present composed / atomic rules in this table? I'm not sure it lends itself for that. Probably worth picking that up.
2. While it looks like the color pass the contrast requirement - I think a slightly darker blue would be better - just so you don't get every third accessibility tester getting out there color picker
3. Margins above headings are quite big
4. Can we put an icon with each test case for Pass / Fail / Inapplicable - with so many test cases on the page it's easy to lose track.

jkodu:
@WilcoFiers 
My notes on your thoughts below:

1) Going to tackle that in a separate issue (https://github.com/auto-wcag/auto-wcag/issues/174). Will mock something up & attach to the issue.
2) Noted. Took it up a notch.
3) Fixed. 
4) Will tackle this in a separate issue, as test cases is something to re-work based on the new format and it may be wise to do this then. Have created an issue (https://github.com/auto-wcag/auto-wcag/issues/186)

PR #203, opened 2018-08-03
Add composite rule template

WilcoFiers:
<< Describe the changes >>

Closes issue:

## Reviewer checks

# How to Review And Approve

### DIFF
--- a/pages/design/atomic-template-empty.md
+++ b/pages/design/atomic-template-empty.md
@@ -1,5 +1,7 @@
 ---
 name:
+test_type: atomic
+
 description: |
   This rule checks ...
 
@@ -37,7 +39,9 @@ There are no major accessibility support issues known for this rule.
 
 ## Test Cases
 
-### Pass example 1
+### Passed
+
+#### Pass example 1
 
 Description...
 
@@ -45,11 +49,13 @@ Description...
 <!-- code -->
 ```
 
-### Pass example 2
+#### Pass example 2
 
 ...
 
-### Failure example 1
+### Failed
+
+#### Failure example 1
 
 Description...
 
@@ -57,11 +63,13 @@ Description...
 <!-- code -->
 ```
 
-### Failure example 2
+#### Failure example 2
 
 ...
 
-### Inapplicable example 1
+### Inapplicable
+
+#### Inapplicable example 1
 
 Description...
 
@@ -69,6 +77,6 @@ Description...
 <!-- code -->
 ```
 
-### Inapplicable example 2
+#### Inapplicable example 2
 
 ...
--- a/pages/design/atomic-template-empty.md
+++ b/pages/design/atomic-template-empty.md
@@ -6,13 +6,13 @@ description: |
   This rule checks ...
 
 success_criterion:
-- x.x.x
+- x.x.x # (Name of Success Criterion)
 
 test_aspects:
-- 
+- # (e.g. HTTP Messages, DOM Tree, CSS Styling, Accessibility Tree, Language, etc.,)
 
 authors:
--
+- # (full names as found in /_data/contributors.yml@master - if not yet listed, please have authors added to the list)
 ---
 
 ## Test Procedure
@@ -35,13 +35,13 @@ Each target element ...
 
 ## Background
 
-- 
+- (e.g. WCAG Techniques or links with background information mentioned in Applicability, Expectations or Assumptions)
 
 ## Test Cases
 
 ### Passed
 
-#### Pass example 1
+#### Passed example 1
 
 Description...
 
@@ -49,13 +49,13 @@ Description...
 <!-- code -->
 ```
 
-#### Pass example 2
+#### Passed example 2
 
 ...
 
 ### Failed
 
-#### Failure example 1
+#### Failed example 1
 
 Description...
 
@@ -63,7 +63,7 @@ Description...
 <!-- code -->
 ```
 
-#### Failure example 2
+#### Failed example 2
 
 ...
 
--- a/pages/design/atomic-template-empty.md
+++ b/pages/design/atomic-template-empty.md
@@ -1,6 +1,5 @@
 ---
 name:
-group: # Optional
 description: |
   This rule checks ...
 
@@ -35,3 +34,23 @@ There are no major accessibility support issues known for this rule.
 ## Background
 
 - 
+
+## Test Cases
+
+### Passed
+
+```html
+<!--  -->
+```
+
+### Failed
+
+```html
+<!-- -->
+```
+
+### Inapplicable
+
+```html
+<!--  -->
+```
--- a/pages/design/atomic-template-empty.md
+++ b/pages/design/atomic-template-empty.md
@@ -37,20 +37,38 @@ There are no major accessibility support issues known for this rule.
 
 ## Test Cases
 
-### Passed
+### Pass example 1
+
+Description...
 
 ```html
-<!--  -->
+<!-- code -->
 ```
 
-### Failed
+### Pass example 2
+
+...
+
+### Failure example 1
+
+Description...
 
 ```html
-<!-- -->
+<!-- code -->
 ```
 
-### Inapplicable
+### Failure example 2
+
+...
+
+### Inapplicable example 1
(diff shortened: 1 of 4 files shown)


pages/design/rule-template.md
```diff
@@ -4,16 +4,15 @@ title: Auto-WCAG Rule Template
 
 The rule template contains a plain language description of the rule, some background information, and a list of all assumptions. The test procedure is defined by the selector, a number of steps and a description of the possible outcomes.
 
-Use the [empty test template](rule-template-empty.html) to create new auto-wcag rule. When creating a new rule, first read [rule design](rule-design.html).
+Use the [empty atomic rule template](atomic-template-empty.html) to create new auto-wcag rule. When creating a new rule, first read [rule design](rule-design.html).
 
-```markdown
+````markdown
```
jkodu: why ```` 4 backticks?

WilcoFiers: Markdown magic. I needed to put triple backticks inside this block for the test cases. By giving the outer one an extra backtick, the markdown parser knows not to stop at my triple backticks for the test cases, but to continue until the expression with four backticks.

jkodu: Right, truly magical.

pages/design/composed-template-empty.md
```diff
@@ -0,0 +1,78 @@
+---
+name:
+description: |
+  This rule checks ...
+
```
WilcoFiers: Should add `type: atomic|composed`

pages/design/composed-template-empty.md
```diff
@@ -0,0 +1,78 @@
...
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- 
+
+
+## Test Cases
+
+### Pass example 1
```
annethyme: Use "Passed", "Failed", "Inapplicable" as defined in https://www.w3.org/TR/act-rules-format/#output-outcome

pages/design/composed-template-empty.md
jkodu: Believe the agreement on the last call, was we will be able to decipher if a rule is atomic or composed, based on listing of atomic rules in the front-matter, so this might be not necessary.

pages/design/atomic-template-empty.md
```diff
@@ -34,4 +33,42 @@ There are no major accessibility support issues known for this rule.
 
 ## Background
 
-- 
\ No newline at end of file
+- 
+
+## Test Cases
+
```
jkodu: I think, there needs to be a clear heading for all each of the test type groups, namely `Passed`, `Failed` and `Inapplicable`.

@annethyme seems to agree on this as well.

I understand the need to have a heading for each of the test cases like `pass example 1`, but this to me is redundant information, would rather use the short description that we aim to address via https://github.com/auto-wcag/auto-wcag/issues/177 as a heading?

Also for automatic exporting of test cases as per this PR - https://github.com/auto-wcag/auto-wcag/pull/213 there needs to be a way to decipher which group or test type a code snippet belongs to, I use the group heading to discern if a test case is passed, failed of inapplicable, see https://github.com/auto-wcag/auto-wcag/pull/213/files#diff-b0a4bd50b949f5f4482b61265a629b5aR99

I suggest we keep the group headings for test cases.

pages/design/composed-template-empty.md
```diff
@@ -0,0 +1,85 @@
...
+- 
+- 
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility support
+
+There are no major accessibility support issues known for this rule.
```
annethyme: italics as well?

WilcoFiers:
@annethyme @Brynanders please add your review.

pages/design/rule-template.md
```diff
@@ -58,4 +58,54 @@ Support for XXX is known to be limited in some assistive technologies. If any of
...
+<!-- inapplicable code snippet -->
 ```
+
+#### Inapplicable example X
+
+... Add one test case for each "reason" an element can be inapplicable
+
+````
+
+## Composed rules
```
annethyme: Is this section meant to go into all Composed rules? 

I would suggest that we create one page and link to that one instead, so that we don't risk having to update the wording in all composed rules if something is being changed.

pages/design/composed-template-empty.md
```diff
@@ -0,0 +1,85 @@
+---
+name:
+rule_type: composed
+description: |
+  This rule checks ...
+
+success_criterion:
+- x.y.z
```
annethyme: align with the Atomic one, either "x.x.x" or "x.y.z"

pages/design/atomic-template-empty.md
```diff
@@ -0,0 +1,82 @@
+---
+name:
+test_type: atomic
+
+description: |
+  This rule checks ...
+
+success_criterion:
+- x.x.x
```
annethyme: Add "# (Name of Success Criterion)" here? We have that in most rules...
Applies to Composed rules as well

pages/design/atomic-template-empty.md
```diff
@@ -0,0 +1,82 @@
...
+
+test_aspects:
+-
```
annethyme: add "(e.g. HTTP Messages, DOM Tree, CSS Styling, Accessibility Tree, Language, etc.)"?

pages/design/atomic-template-empty.md
```diff
@@ -0,0 +1,82 @@
...
+  This rule checks ...
+
+success_criterion:
+- x.x.x
+
+test_aspects:
+- 
+
+authors:
+-
```
annethyme: Add "(full names as found in https://github.com/auto-wcag/auto-wcag/blob/master/_data/contributors.yml - if not yet listed, please have authors added to the list)" ?

pages/design/atomic-template-empty.md
```diff
@@ -0,0 +1,82 @@
...
+
+*There are currently no assumptions*
+
+## Accessibility support
+
+*There are no major accessibility support issues known for this rule.*
+
+## Background
+
+-
```
annethyme: Add "(e.g. WCAG Techniques or links with background information mentioned in Applicability, Expectations or Assumptions)" ?

pages/design/composed-template-empty.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Applicability
+
+The rule applies to any (??) element ...
+
+### Expectation
+
+For each test target, the outcome of << Logic here >>:
+
+-
```
annethyme: Add "(list atomic rules that are used in this composed rule)" ?

pages/design/composed-template-empty.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Aggregation Definition
+
+### Applicability
+
+The rule applies to any (??) element ...
+
+### Expectation
+
+For each test target, the outcome of << Logic here >>:
```
annethyme: "For each test target, the outcome of << Logic here >>:" 

--> 

"For each test target, the outcome of (at least one of / all of / any of etc.) the following rules is (passed / failed / etc.):"

pages/design/rule-template.md
```diff
@@ -4,16 +4,16 @@ title: Auto-WCAG Rule Template
...
 
-Use the [empty test template](rule-template-empty.html) to create new auto-wcag rule. When creating a new rule, first read [rule design](rule-design.html).
+Use the [empty atomic rule template](atomic-template-empty.html) to create new auto-wcag rule. When creating a new rule, first read [rule design](rule-design.html).
 
-```markdown
+````markdown
 ---
 name: Short descriptive name
-group: group-id # Optional
+rule_type: atomic
```
annethyme: Can we align the two templates, so that people get the same level of guidance no matter where they find the template? And like me, don't get confused about which is which - I think I wrote some comments for the wrong ones now... 
Aligning the two would also make it easier to keep them aligned.

jkodu:
@annethyme - updated, please review.
@WilcoFiers 

Can you both confirm, that we have decided to rename `composed` to `composite`?

Issue #206, opened 2018-08-06
Chore: layout updates


jkodu:
- [ ] Number of rules/ counter/ index in rules table.
- [ ] Autocomplete valid rule does not link to SC (as it is WCAG2.1).
- [ ] `\n` is shown after every definition in rule page.

PR #207, opened 2018-08-07
chore: site updates - issue #206

jkodu:
The below fixes have been done.

- [x] Number of rules/ counter/ index in rules table - (Solution: Just added an index column).

- [x] Autocomplete valid rule does not link to SC (as it is WCAG2.1) - (Solution: re-generated site.data for both 2.0 and 2.1 wcag sc's see - https://github.com/auto-wcag/auto-wcag/compare/chore-site-updates?expand=1#diff-24211d7b22f4dfc78076ff0c494e1213, the generated file is here: https://github.com/auto-wcag/auto-wcag/blob/9513f707fdd6cdab059e65171bb5d3b74581d81d/_data/sc-urls.json)

- [x] `\n` is shown after every definition in rule page. (Fixed)

- [x] Can we put an icon with each test case for Pass / Fail / Inapplicable - with so many test cases on the page it's easy to lose track. (Solution: Instead of an icon, have created floating headers of the section that has just been scrolled past)
Screenshot: (here passed have been scrolled, but the heading is sticky...)
image

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/206

## Reviewer checks

# How to Review And Approve


wcag-url-generator.js
```diff
@@ -0,0 +1,40 @@
+// https://github.com/w3c/wai-wcag-quickref/blob/gh-pages/_data/wcag21.json
```
WilcoFiers: We don't need to commit this file, do we? Are we ever going to need it again?

jkodu: Yup. Keep this file for now.

As the commits in [wcag21](https://github.com/w3c/wai-wcag-quickref/blob/gh-pages/_data/wcag21.json) changes, I would like to update the success_criteria mapping. I have an idea to automate it with webhooks, so this file can be generated on the fly. But that is an overhead at the moment.

PR #220, opened 2018-08-20
New Rule: Links with identical accessible names serve equivalent purpose

annethyme:
<< Describe the changes >>

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/210

## Reviewer checks

- [X] Make sure the pull request is prefixed with `Rule:` or `Algorithm:` or `Chore:` based on work involved.

# How to Review And Approve


ShadowBB:
This looks very good. I will approve as soon as I can verify the test assets.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,233 @@
...
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of link, are [exposed to assistive technologies](#exposed-to-assistive-technologies) and that have the same [accessible name](#accessible-name).
+
+**Note:** Leading and trailing whitespace and difference in case sensitivity should be ignored when deciding whether the [accessible names](#accessible-name) are the same.
```
jkodu: This seems to be implementation detail and may be it is best if it lived within the definition of accessible name?

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Expectation
+
+For each set of target elements, activating the links resolves, after any redirects, to resources that fulfil the same purpose in relation to the purpose indicated by the [accessible names](#accessible-name) of the link.
```
jkodu: `activating the links resolves` is not simple to understand, also there are multiple breaks in this sentence.

How about:
`Each target element resolves to resources .....`

I believe resolve inherently tackles `redirect`

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,233 @@
...
+
+**Note:** Web pages and documents (e.g. PDFs, office formats etc.) may fulfil the same purpose in relation to the link, even if the resources:
+* are located on different URLs, including different domains
+* present different navigation options, e.g. through bread crumbs or local sub menus
+* contain different amounts of information and/or differently worded information
+* are using different layouts.
+
+**Note:** If the same content is presented in different formats, the format itself is often part of the link purpose, e.g. an article as both HTML and PDF.
```
jkodu: if the format of the resulting page is different, the links would be ideally different, and can be ignored for the rule itself?

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,233 @@
...
+
+**Note:** If the [normalized](#url-normalization) value of the ´href´ is identical, the resources are identical, thus fulfilling the same purpose. [Relative URLs](https://www.w3.org/TR/WD-html40-970917/htmlweb.html#relative-urls) first need to be resolved to full URLs before doing the comparison.
+
+## Assumptions
+
+* This rule assumes that the purpose of the links for links with identical link texts would not be ambiguous to users in general, which is the exception mentioned in the success criterion 2.4.9 Link Purpose (Link Only).
```
jkodu: A link to SC2.4.9 in the background can be beneficial.

test-assets/my-university/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
jkodu: Ideally these file names are prefixed with the relevant SC, and not with in nested folders. 

In this case, it could be

`test-assets/SC2-4-9-rule-name-my-test-asset-file.html`

annethyme: @jkodu, can we create some guidance on this? And maybe a clean-up... I just had a look in the folder and tried to copy what I saw.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
annethyme: With the direction this rule is taking, it should be evaluated if the purpose is actually the same...

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
annethyme: Put into seperate PR since other rules are also relying on the definition: https://github.com/auto-wcag/auto-wcag/pull/254

test-assets/my-university/index.html
annethyme: Thinking further about it, wouldn't it make sense that each rule had it's own folder for test assets, where the folder name is the same as the rule name?

annethyme:
Test assets still missing, while waiting for decision on test asset folder structure and naming conventions.

Please do review the rest in the meantime.

jkodu:
@nitedog please review

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Note:** Web pages and documents (e.g. PDFs, office formats etc.) may fulfil the same purpose in relation to the link, even if the resources:
+* are located on different URLs, including different domains
+* present different navigation options, e.g. through bread crumbs or local sub menus
+* contain different amounts of information and/or differently worded information
+* are using different layouts.
+
+**Note:** If the same content is presented in different formats, the format itself is often part of the link purpose, e.g. an article as both HTML and PDF.
+
+**Note:** If the [normalized](#url-normalisation) value of the ´href´ is identical, the resources are identical, thus fulfilling the same purpose. [Relative URLs](https://www.w3.org/TR/WD-html40-970917/htmlweb.html#relative-urls) first need to be resolved to full URLs before doing the comparison.
```
nitedog: This is referring to a Working Draft of HTML4. Is there a comparable reference in HTML5?

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Assumptions
+
+* This rule assumes that the purpose of the links for links with identical link texts would not be ambiguous to users in general, which is the exception mentioned in [Understanding Success Criterion 2.4.9: Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html).
```
nitedog: I recommend removing the second half of the sentence "which is the exception ...". The rule already refers to SC 2.4.9 which spells out this exception. Adding another link to the Understanding document is superfluous and also adds potential maintenance overhead.

nitedog: Separately from the above, is it usual to have such aspects of the requirement in the assumptions section? Seems a little odd to me. I would expect this to be part of this or another rule rather than as an assumption.

pages/algorithms/url-normalisation.md
```diff
@@ -0,0 +1,15 @@
...
+A process to transform URLs in a standardised and consistent way, to make it possible to determine if two syntactically different URLs may be equivalent. (Source: [Wikipedia](https://en.wikipedia.org/wiki/URL_normalisation)).
+
+Several types of URL normalisation exists. Some types of normalisation preserve the semantics of the URL, some usually preserve the semantics, and some types change the semantics. 
+
+For the purpose of ACT rules, we only consider the types of URL normalisation that preserve the semantics of the URL. These normalisation types are: 
+- [Case normalisation](https://tools.ietf.org/html/rfc3986#section-6.2.2.1)
+- [Percent-Encoding normalisation](https://tools.ietf.org/html/rfc3986#section-6.2.2.2)
+- [Scheme-Based normalisation](https://tools.ietf.org/html/rfc3986#section-6.2.3)
+
+[Relative URLs](https://www.w3.org/TR/WD-html40-970917/htmlweb.html#relative-urls) first need to be resolved to full URLs before doing the comparison.
```
nitedog: Referring to Working Drafts is not good practice.

pages/algorithms/url-normalisation.md
```diff
@@ -0,0 +1,15 @@
+---
+title: URL normalisation 
+key: url-normalisation 
+---
+
+A process to transform URLs in a standardised and consistent way, to make it possible to determine if two syntactically different URLs may be equivalent. (Source: [Wikipedia](https://en.wikipedia.org/wiki/URL_normalisation)).
```
nitedog: Is there something more stable than Wikipedia to refer to? For example, IETF and W3C specs will not change.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of link, are [exposed to assistive technologies](#exposed-to-assistive-technologies) and that have the same [accessible name](#accessible-name).
+
+### Expectation
+
+When activated, the links in each set of target elements resolve to resources that fulfil the same purpose in relation to the purpose indicated by the [accessible names](#accessible-name) of the link.
```
WilcoFiers: Can we simplify the language here a bit?

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
WilcoFiers: There's also an assumption that the accessible name is the same as the link text. I'm not sure if that should be an assumption, or if the rule should be changed to compare the visible text, instead of the accessible name.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+<a href="/test-assets/link-purpose/index.html">Link text</a>
+```
+
+#### Passed example 2
+
+Links resolves to same page due to case-insensitivity on server:
+
+```html
+<a href="/Test-Assets/Link-Purpose/index.html">Link text</a>
+<a href="/test-assets/link-purpose/index.html>Link text</a>
```
WilcoFiers: A quote (") is missing after the URL.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+<a href="/test-assets/link-purpose/humanities/contact.html">Contact us</a>
+```
+
+#### Failed example 3
+
+Case-sensitivity in file name:
+
+```html
+<a href="/test-assets/link-purpose/page1.html">Link text</a> 
+<a href="/test-assets/link-purpose/Page1.html">Link text</a>
```
WilcoFiers: Why does this fail? Please explain in the description.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of link, are [exposed to assistive technologies](#exposed-to-assistive-technologies) and that have the same [accessible name](#accessible-name).
```
nitedog: "The rule" -> "This rule"

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
nitedog: - "fulfil" -> "fulfill" (US spelling)
- I think you can safely remove "in relation to the purpose" to read just "fulfill the same purpose indicated..."

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+When activated, the links in each set of target elements resolve to resources that fulfil the same purpose in relation to the purpose indicated by the [accessible names](#accessible-name) of the link.
+
+**Note:** Resolving the links includes potential redirects.
+
+**Note:** Web pages and documents (e.g. PDFs, office formats etc.) may fulfil the same purpose in relation to the link, even if the resources:
+* are located on different URLs, including different domains
+* present different navigation options, e.g. through bread crumbs or local sub menus
+* contain different amounts of information and/or differently worded information
+* are using different layouts.
```
nitedog: This sounds like it could be a definition for "equivalent purpose" or such a phrase?

test-assets/my-university/index.html
jkodu: We reached a consensus on this in a previous meeting.

jkodu:
@annethyme - could you address the review comments.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
carlosapaduarte: I don't really understand what is the purpose of this note, or how it would help interpret the rule

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of link, are [exposed to assistive technologies](#exposed-to-assistive-technologies) and that have the same [accessible name](#accessible-name).
```
carlosapaduarte: Not really an issue, but in most other rules applicability specifies a single element and the other elements are identified in the expectation. Here, applicability specifies a set of elements. I think the way is done here improves legibility.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Expectation
+
+When activated, the links in each set of target elements resolve to resources that fulfill the same purpose indicated by the [accessible names](#accessible-name) of the link.
```
carlosapaduarte: These are 2 expectations:
1 - resolve to resources that fulfil the same purpose
2 - the purpose is indicated by the accessible name
I don't think the second should be addressed in this rule.

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+Identical link text leads to identical URLs:
+
+```html
+<a href="/test-assets/link-purpose/index.html">Link text</a>
+<a href="/test-assets/link-purpose/index.html">Link text</a>
+```
+
+#### Passed example 2
+
+Links resolves to same page due to case-insensitivity on server:
```
carlosapaduarte: Do we want test cases that are server dependent? What if I run the test in a case-sensitive server?

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+Pages contain different amounts of information and/or differently worded information, but fulfils same purpose in relation to the link:
+
+```html
+<a href="/test-assets/link-purpose/page1.html">Link text</a>
+<a href="/test-assets/link-purpose/page2.html">Link text</a>
+```
+
+#### Passed example 8
+
+Has the same content but are using different layouts:
```
carlosapaduarte: Rephrase to "Pages have the same content but using different layouts:"

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [Understanding Success Criterion 2.4.9: Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html)
+- [URL normalization](#url-normalisation)
+
+## Test Cases
+
+### Passed
```
carlosapaduarte: Several test cases could be used exactly as they are written but as examples of failed test cases because they are dependent on the content of the files. Don't we need to provide the linked files also for the test cases to be meaningful?

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+Link is not exposed to assistive technologies;
+
+```html
+<a href="/test-assets/link-purpose" style="display:none">Link text 1</a>
+<a href="/test-assets/link-purpose">Link text 2</a>
+```
+
+#### Inapplicable example 4
+
+Links created via scripting, but doesn't have semantic role of link:
```
carlosapaduarte: replace "but doesn't have semantic role of link" with "but without the semantic role of link"

pages/algorithms/url-normalisation.md
carlosapaduarte: Link is pointing to an article that does not exist on Wikipedia (at least not today). This is what @nitedog is referring to

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
```diff
@@ -0,0 +1,234 @@
...
+
+description: |
+  This rule checks that identical accessible names are only used for links that have the same purpose
+
+success_criterion: 
+- 2.4.9 # Link Purpose (Link Only)
+
+test_aspects:
+- DOM Tree
+- CSS Styling
```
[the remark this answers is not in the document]
annethyme: We need CSS Styling to know whether something is "included in the accessibility tree", since it could be done with style"display:none" :)

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
annethyme: @carlosapaduarte, I agree that the second one should not be part of this rule. However, the point here is that "the same purpose" needs to be interpreted in the context of the accessible name.
---
Example:  

Page A: only has a phone number on it
Page B: has a phone number, email, fax, address, etc.

Example 1: Link text is "Call us" --> Rule is Passed, since Page A and Page B fulfil the same purpose
Example 2: Link text is "See all the different ways that you can contact us" --> Rule is Failed, since Page A and Page B doesn't fulfil the same purpose. 
---
Would one of these be better/clearer: 

"fulfil the same purpose in relation to the [accessible names](#accessible-name) of the link."
"fulfil the same purpose, indicated by the [accessible names](#accessible-name) of the link."
"fulfil the same purpose as indicated by the [accessible names](#accessible-name) of the link."

_rules/SC2-4-9-links-with-identical-accessible-names-have-same-purpose.md
annethyme: @carlosapaduarte, yes, this is part of the work load left for me on this one.

pages/algorithms/url-normalisation.md
annethyme: @nitedog, the normative references further down are good enough as stable references, I think. But the Wikipedia definition has such a nice way of explaining it. The link for Wikipedia was merely to give credit to the source. 
Any suggestions for how to handle this?

dd8: The parsing URLs definition in the W3 HTML 5.2 spec defines this, by reference to the WHATWG URL standard:
https://www.w3.org/TR/html52/infrastructure.html#parsing-urls

This WHATWG parser deals with unicode normalisation, IDNs, resolving relative URLs, percent encoding and many backwards compatibility issues. The resulting serialized URLs can be reliably compared for identity. So this algorithm produces the same serialized URLs for all these cases

`http://www.example.com/about` base `http://www.example.com/`
`//www.example.com/about` base `http://www.example.com/`
`/about` base `http://www.example.com/`

**Note:** fully resolved URLs may be different, but still lead to the same page after resolving the HTTP request, due to redirects and DNS aliasing. For example:

```
http://example.com/
http://www.example.com/
https://www.example.com/
```

annethyme: Linking to the HTML URL parsing instead of defining anything myself.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of link, are [included in the accessibility tree](#included-in-the-accessibility-tree) and that have the same [non-empty](#non-empty) [accessible name](#accessible-name).
```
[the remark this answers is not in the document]
Brynanders: Also, I am not sure that an applicability should be limited to non-emty accessible names. In other words, should we not also flag up sets of 'empty' links?

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+Identical link text leads to identical URLs:
+
+```html
+<a href="/test-assets/6cbcbe/index.html">Contact us</a>
+<a href="/test-assets/6cbcbe/index.html">Contact us</a>
+```
+
+#### Passed example 2
+
+Links resolves to same page after redirect:
```
Brynanders: Should address the terminology on the expectation.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+```html
+<a href="/test-assets/6cbcbe/index.html">Contact us</a>
+<a href="/test-assets/6cbcbe/redirect.html">Contact us</a>
+```
+
+#### Passed example 3
+
+Identical pages are located on different URLs:
```
Brynanders: Isn't this covered in the previous test case in that both the same page located somewhere else and a redirected page serve up content of an equivalent purpose?

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+Same link text used for links going to pages where the content section is the same, but where the navigation options (bread crumbs and local sub menus) differ due to different placement in navigation hierarchy:
+
+```html
+<a href="/test-assets/6cbcbe/about/contact.html">Contact us</a>
+<a href="/test-assets/6cbcbe/careers/contact.html">Contact us</a>
+```
+
+#### Passed example 5
+
+URLs differ due to trailing slashes:
```
Brynanders: This is a good point but its not included in the applicability. Maybe a note could cover trailing slashes and absolute and relative URLs.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+```html
+<a href="/test-assets/6cbcbe/">Contact us</a> 
+<a href="/test-assets/6cbcbe">Contact us</a>
+```
+
+#### Passed example 6
+
+Pages contain different amounts of information and/or differently worded information, but fulfils same purpose in relation to the link:
```
Brynanders: I don't get this one, isn't that like saying that two links to completely different contact pages for different department is okay?

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+```html
+<a href="/test-assets/6cbcbe/page1.html">Contact us</a>
+<a href="/test-assets/6cbcbe/page2.html">Contact us</a>
+```
+
+#### Passed example 7
+
+Pages have the same content but use different layouts:
```
Brynanders: same content is ambiguous, you could change it to equivalent purpose but I think that would make this a duplicate test case.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+Same link text used for links going to different resources:
+
+```html
+<a href="http://facebook.com">Follow us</a> 
+<a href="http://twitter.com">Follow us</a>
+```
+
+#### Failed example 2
+
+Same link text used for links going to web pages with same name, but with different information:
```
Brynanders: Nothing in the applicability the name of resources.

Also be good to stay consistent with either "information" or "content" or "that do not serve an equivalent purpose"

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+```html
+<a href="/test-assets/6cbcbe/about/contact.html">Contact us</a> 
+<a href="/test-assets/6cbcbe/admissions/contact.html">Contact us</a>
+```
+
+#### Failed example 3
+
+Links created via scripting with explicit role of link, but leads to resources offering different information:
```
Brynanders: terminology consistency

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+No identical link texts:
+
+```html
+<a href="/test-assets/6cbcbe/about/contact.html">Contact main office</a>
+<a href="/test-assets/6cbcbe/admissions/contact.html">Contact admissions office</a>
+```
+
+#### Inapplicable example 3
+
+Link is not exposed to assistive technologies;
```
Brynanders: exposed to assistive technologies > included in the accessibility tree

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
+---
+name: Links with identical accessible names serve equivalent purpose
+test_type: atomic
+
+description: |
+  This rule checks that identical accessible names are only used for links that serve an equivalent purpose
```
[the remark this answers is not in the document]
annethyme: How about: 

> This  rule checks that for links with identical accessible names, the links serve the same purpose.

_rules/6cbcbe.md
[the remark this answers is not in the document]
annethyme: @Brynanders, empty links will be flagged by the "Link has name" rule that asks to give the links an accessible name. 
I would find it confusing to have two different rules giving different advice on what to do with an item...

_rules/6cbcbe.md
annethyme: I still think they are different conceptually.

_rules/6cbcbe.md
annethyme: This is included in URL parsing in the definition used in the Expectation.

_rules/6cbcbe.md
annethyme: What if the link text is "call us", and the pages are the same as they are now? 
The point is, that as long as you still get the same phone number, they can fulfil the same purpose, whereas if you are actually calling different departments, it's not the same purpose anymore.

_rules/6cbcbe.md
annethyme: I *think* it is included in URL parsing, but actually not completely sure. 
@jkodu, can you see whether example.com and example.com/ would translate into the same URL after URL parsing?

_rules/6cbcbe.md
annethyme: Updated it to say "call us". @Brynanders, do you think it works now?

_rules/6cbcbe.md
carlosapaduarte: They resolve to the same resource (I'm not sure if that means that they translate to the same URL or if the user agent handles the difference)

annethyme: Researched with @kasperisager. Seems that it's protocols in the user agent that defines these redirects.

annethyme:
@ShadowBB, I did a few changes that has to do with user agent redirects. (only changed test case description for Passed example 5 and the "Same resource" definition) 
Can I have you review again?

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Expectation
+
+When activated, the links in each set of target elements resolve to the [same resource](#same-resource) or to different resources that fulfill an [equivalent purpose](#equivalent-purpose).
+
+**Note:** Resolving the links includes potential redirects.
+
+## Assumptions
+
+This rule assumes that the purpose of the links for links with identical names would not be ambiguous to users in general.
```
carlosapaduarte: You could simplify "the purpose of the links for links with identical names" to "the purpose of the links with identical names"

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+URLs differ due to trailing slashes, but resolves to the same resource after redirects caused by user agent:
+
+```html
+<a href="/test-assets/6cbcbe/">Contact us</a> 
+<a href="/test-assets/6cbcbe">Contact us</a>
+```
+
+#### Passed example 6
+
+Pages contain different amounts of information and/or differently worded information, but fulfils same purpose in relation to the link:
```
carlosapaduarte: "fulfils" > "fulfill the"

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+Same link text used for links going to different resources:
+
+```html
+<a href="http://facebook.com">Follow us</a> 
+<a href="http://twitter.com">Follow us</a>
+```
+
+#### Failed example 2
+
+Same link text used for links going to web pages that are similar, but has different information in the content:
```
carlosapaduarte: "has" > "have"
"the content" > "their content"

pages/algorithms/equivalent-purpose.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Equivalent purpose 
+key: equivalent-purpose 
+---
+
+Resources that are not identical can still serve an equivalent purpose by both/all living up to the expectation that was created when sending the user to the resources. This would usually involve that the advertised key content is the same.  
+
+Web pages and documents (e.g. PDFs, office formats etc.) may however fulfil an equivalent purpose, even if the resources:
```
carlosapaduarte: Remove "however"

pages/algorithms/equivalent-purpose.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Equivalent purpose 
+key: equivalent-purpose 
+---
+
+Resources that are not identical can still serve an equivalent purpose by both/all living up to the expectation that was created when sending the user to the resources. This would usually involve that the advertised key content is the same.
```
carlosapaduarte: I don't think this definition is as clear as it can be. Here is my suggestion (see if you think it improves):
Not identical resources can still serve an equivalent purpose by equally complying to the expectation formed by the user when navigating to them.

pages/algorithms/equivalent-purpose.md
```diff
@@ -0,0 +1,17 @@
...
+
+Resources that are not identical can still serve an equivalent purpose by both/all living up to the expectation that was created when sending the user to the resources. This would usually involve that the advertised key content is the same.  
+
+Web pages and documents (e.g. PDFs, office formats etc.) may however fulfil an equivalent purpose, even if the resources:
+* are located on different URLs, including different domains
+* present different navigation options, e.g. through bread crumbs or local sub menus
+* contain different amounts of information and/or differently worded information
+* use different layouts.
+
+If both/all resources cover the user expectations created by e.g. the link name equally well, the resources are considered to serve an equivalent purpose.
```
carlosapaduarte: Remove "both/" to simplify
"user" > "user's"

pages/algorithms/equivalent-purpose.md
carlosapaduarte: It should be "Non-identical resources"

_rules/6cbcbe.md
WilcoFiers: I'm not sure I understand what this means. Can you add to it to make this clearer?

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+## Accessibility support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [Understanding Success Criterion 2.4.9: Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html)
+- [URL parsing](https://www.w3.org/TR/html52/infrastructure.html#parsing-urls)
+
+## Test Cases
```
WilcoFiers: You are missing tests for SVG.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of link, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have the same [non-empty](#non-empty) [accessible name](#accessible-name).
```
WilcoFiers: I was wondering if this shouldn't also include links that have a different semantic role. For example, there are several `doc-*` roles that inherit from link. Those are clearly links. Another example of a link with a different role I found is where a an `a[href]` element has `role=menuitem`.

I would say elements with a role that inherit from link should be applicable. I'll leave the menuitem question up to the author.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+<a href="/test-assets/6cbcbe/about/contact.html">Contact main office</a>
+<a href="/test-assets/6cbcbe/admissions/contact.html">Contact admissions office</a>
+```
+
+#### Inapplicable example 3
+
+Link is not included in the accesssibility tree:
+
+```html
+<a href="/test-assets/6cbcbe/page1.html" aria-hidden="true">Contact Us</a>
```
WilcoFiers: I'd put `tabindex="-1"` on this link, make sure the example is accessible.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,206 @@
...
+Link text
+</span>
+
+<span role="link"
+   onclick="location='/test-assets/6cbcbe/admissions/contact.html'">
+Link text
+</span>
+```
+
+### Inapplicable
```
WilcoFiers: There is no example for links with an empty accessible name.

_rules/6cbcbe.md
annethyme: SVG test cases added.

_rules/6cbcbe.md
annethyme: Tabindex"-1" added - even though test cases don't have to be accessible...

_rules/6cbcbe.md
annethyme: @WilcoFiers, Roles that inherit from the `link` role has now been included. 
For `role=menuitem`, I'd rather not include it in this rule, as the menuitems could be many other things than just links. We might consider writing another rule for it though, like the "iframe has unique name" rule.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,269 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-text-content) [accessible names](#accessible-name) that contain [text content](#text-content).
```
kasperisager: What does it mean for a rule to apply to "any two or more HTML or SVG elements"?

_rules/6cbcbe.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Expectation
+
+When activated, the links in each set of target elements resolve to the [same resource](#same-resource) or to different resources that fulfill an [equivalent purpose](#equivalent-purpose).
```
kasperisager: What is meant by "activated"?

_rules/6cbcbe.md
annethyme: What if it is "a set of two or more..."?

_rules/6cbcbe.md
annethyme: Suggestions for a better wording for activating/clicking/following?

_rules/6cbcbe.md
kasperisager: If I'm not mistaken, an ACT rule can only apply to individual test targets and not sets of test targets. The same goes for expectations.

_rules/6cbcbe.md
kasperisager: Let's tackle https://github.com/auto-wcag/auto-wcag/pull/220#discussion_r267256695 first and then look at this.

_rules/6cbcbe.md
annethyme: I can't see why it couldn't apply to a set... For reference: https://w3c.github.io/wcag-act/act-rules-format.html#applicability

kasperisager: I do think I'm mistaken then 😄

_rules/6cbcbe.md
annethyme: https://github.com/auto-wcag/auto-wcag/pull/220#discussion_r267256695 tackled. Any suggestions for this one then?

annethyme: Changed to: "Following the link..."

_rules/6cbcbe.md
```diff
@@ -0,0 +1,271 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to a set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characaters) [accessible names](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
kasperisager: Maybe:

> This rule applies to any set of two or more HTML or SVG elements [...]

_rules/6cbcbe.md
```diff
@@ -0,0 +1,271 @@
...
+
+This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characaters) [accessible names](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name](#accessible-name).
+
+### Expectation
+
+When followed, the links in each set of target elements resolve to the [same resource](#same-resource) or to different resources that fulfill an [equivalent purpose](#equivalent-purpose).
+
+**Note:** Resolving the links includes potential redirects.
```
kasperisager: What's should happen if redirects are delayed? Consider:

```
<a href="/foo/a.html">Foo</a>
<a href="/foo/b.html">Foo</a>
```

If `/foo/b.html` redirects to `/foo/a.html` after 5 seconds, or perhaps after 20 hours, does the expectation hold?

annethyme: Delayed redirects should, as I see it, be a failure under a different success criterion (see https://auto-wcag.github.io/auto-wcag/rules/SC2-2-1+SC2-2-4-meta-refresh.html). 

What if we limit it to redirects that happen instantly?

kasperisager: Sounds good to me!

annethyme: Now redirects have to be instant, and the redirect note is no longer a note, but part of the expectation. 
Failed testcase added, and Passed redirect testcase now specifies that the redirect is instant.

pages/algorithms/same-resource.md
```diff
@@ -0,0 +1,16 @@
...
+
+Two or more resources can be the same resource even though the URLs for them are different. This can be due to URL parsing, server settings, redirects and DNS aliasing. 
+
+If the [parsed URLs](https://www.w3.org/TR/html52/infrastructure.html#parsing-urls) for two resources are identical, the resources are the same resource. 
+
+Depending on the server, URLs can either be case-sensitive or case-insenstive, meaning that `<a href="page1.html">` and `<a href="Page1.html">` leads to either the same or two different pages.
+
+Fully parsed URLs can be different, but still lead to the same resource after making the HTTP request, due to redirects and DNS aliasing. For example, these URLs are all fully normalised: http://example.com/, http://www.example.com/, https://www.example.com/. The server can however be configured to serve the same site for http and https, and the same site for example.com and www.example.com. This is common, but not guaranteed. 
+
+**Note:** Whereas parsing is very quick to do automatically, resolving HTTP aliases is slow. It can take 5 seconds or longer per request.
```
kasperisager: What was the intention of adding this note? It seems a tad random.

pages/algorithms/equivalent-purpose.md
```diff
@@ -0,0 +1,17 @@
...
+
+Web pages and documents (e.g. PDFs, office formats etc.) may fulfil an equivalent purpose, even if the resources:
+* are located on different URLs, including different domains
+* present different navigation options, e.g. through bread crumbs or local sub menus
+* contain different amounts of information and/or differently worded information
+* use different layouts.
+
+If all resources cover the user's expectations created by e.g. the link name equally well, the resources are considered to serve an equivalent purpose. 
+
+**Note:** If the same content is presented in different formats, the format itself is often part of the link purpose, e.g. an article as both HTML and PDF, or an image in different sizes. If getting the same content in different formats is the purpose of having seperate links, the resources do not serve an equivalent purpose.
```
kasperisager: ...formats or languages.

pages/algorithms/equivalent-purpose.md
```diff
@@ -0,0 +1,17 @@
...
+
```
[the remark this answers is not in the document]
annethyme: Changed the definition to "equivalent resources" as suggested and updated the rule accordingly.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+<a href="/test-assets/6cbcbe/page1.html">Contact us</a>
+<a href="/test-assets/6cbcbe/page3.html">Contact us</a>
+```
+
+#### Passed example 8
+
+Links created via scripting with explicit role of link, but leads to the same resource: 
+
+```html
+<span role="link"
```
audreymaniez: I know that is not the point, but can we just add `tabindex="0"` to every  `span` with `role="link"` :-)

Otherwise, the test case won't be keyboard usable (and since it's not the point of the test, I think it's a little damaging)

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+### Applicability
+
+This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names](#accessible-name) that does not only consist of [whitespace](#whitespace).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name](#accessible-name).
+
+### Expectation
+
+When followed, the links in each set of target elements resolve to the [same resource](#same-resource) or to [equivalent resources](#equivalent-purpose).
+Resolving the links includes potential redirects, if the redirects happen instantly.
```
audreymaniez: I'm interesting in understanding how a redirection that is not instant can impact the evaluation of how the ressources are the same or not ? 

I see [the discussion here](https://github.com/act-rules/act-rules.github.io/pull/220#discussion_r271721303), and I think it's fair on a user impact side, but i would failed the redirection itself with SC 3.2.5 not on link purpose.

kasperisager: Continuing from that example, if `/foo/b.html` redirects to `/foo/a.html` after 10 seconds, a user would actually get to experience and possibly interact with `/foo/b.html` before being redirected to `/foo/a.html`. As such, `/foo/b.html` is experienced as distinct from `/foo/a.html` and the two are therefore clearly not the same resource. If, on the other hand, `/foo/b.html` redirects instantly to `/foo/a.html`, the two are experienced as the same resource, namely `/foo/a.html`.

I absolutely agree that such a delayed redirect should fail the WCAG, but that is not the concern of this rule. As you point out, it would also fail this rule which may or may not be appropriate; I don't know enough about the WCAG to have an opinion on that though.

pages/algorithms/equivalent-resource.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Equivalent resource 
+key: equivalent-resource 
+---
+
+Non-identical resources can still be equivalent by equally complying to the expectation formed by the user when navigating to them, thus serving an equivalent purpose. This would usually involve that the advertised key content is the same.  
+
+Web pages and documents (e.g. PDFs, office formats etc.) may fulfil be equivalent resources, even if the resources:
```
DagfinnRomen: There may be an extra word in this sentence, "may fulfil be equivalent resources".

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Assumptions
+
+This rule assumes that the purpose of the links with identical names would not be ambiguous to users in general when seen in context on the web page, as users of assistive technologies in this case is not at a disadvantage when viewing the link out of context, e.g. on a list of links in a screen reader.
```
carlosapaduarte: The second part of the sentence is not easy to read and understand. Since I'm not sure about what you meant, I'm not able to offer a suggestion :-(

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+URLs differ due to trailing slashes, but resolves to the same resource after redirects caused by user agent:
+
+```html
+<a href="/test-assets/6cbcbe/">Contact us</a> 
+<a href="/test-assets/6cbcbe">Contact us</a>
+```
+
+#### Passed example 6
+
+Pages contain different amounts of information and/or differently worded information, but fulfils the same purpose in relation to the link:
```
carlosapaduarte: fulfils > fulfil

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+Pages have the same advertised key content but use different layouts:
+
+```html
+<a href="/test-assets/6cbcbe/page1.html">Contact us</a>
+<a href="/test-assets/6cbcbe/page3.html">Contact us</a>
+```
+
+#### Passed example 8
+
+Links created via scripting with explicit role of link, but leads to the same resource:
```
carlosapaduarte: leads > lead

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+Same link text used for links going to web pages that are similar, but have different information in their content:
+
+```html
+<a href="/test-assets/6cbcbe/about/contact.html">Contact us</a> 
+<a href="/test-assets/6cbcbe/admissions/contact.html">Contact us</a>
+```
+
+#### Failed example 3
+
+Links created via scripting with explicit role of link, but leads to different resources that offer different content:
```
carlosapaduarte: leads > lead

Additionally, if you drop the first "different" it might improve readability

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+Same accessible name used for image links going to different resources:
+
+```html
+<a href="http://facebook.com"><img src="facebook.jpg" alt="Follow us" /></a> 
+<a href="http://twitter.com"><img src="twitter.jpg" alt="Follow us" /></a>
+```
+
+#### Failed example 5
+
+A set of two SVG `<a>` elements have the same accessible name but links to different resources:
```
carlosapaduarte: links > link

_rules/6cbcbe.md
```diff
@@ -0,0 +1,279 @@
...
+´a´ and ´area´ elements without ´href´ attribute:
+
+```html
+<a>Link text</a>
+<area aria-label="Link text">
+```
+
+#### Inapplicable example 2
+
+No identical link texts:
```
carlosapaduarte: What about
Links with different accessible names

_rules/6cbcbe.md
ShadowBB: "as users of assistive technologies in this case is not at a disadvantage when viewing the link out of context, e.g. on a list of links in a screen reader."

 should be (I think)

". This assumption is made because, in the case described here, users of assistive technologies are not at a disadvantage when viewing the link out of context, e.g. on a list of links in a screen reader"

However that makes little sense because if they are not ambiguous when seen in context that doesn't mean that they are clear without context right?

annethyme: From the Understanding document: 

> The Success Criterion includes an exception for links for which the purpose of the link cannot be determined from the information on the Web page. In this situation, the person with the disability is not at a disadvantage; there is no additional context available to understand the link purpose. However, whatever amount of context is available on the Web page that can be used to interpret the purpose of the link must be made available in the link text to satisfy the Success Criterion.

(https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-link.html)

I will have another look at the wording asap.

annethyme: @ShadowBB 
> However that makes little sense because if they are not ambiguous when seen in context that doesn't mean that they are clear without context right?

I think the point here is that if I created a page with nothing but a link that reads "Click here", it is not an accessibility issue, and thus not a violation of success criterion 2.4.9, because it gives the same (poor) experience to everyone.
And I guess it doesn't just have to be a poor experience, but could also be game-like or artsy.

annethyme: @carlosapaduarte and @ShadowBB, I gave a shot at a rewrite. Please have a look again.

ShadowBB: Oh! I totally interpreted that the other way around. 

You new writeup makes this a lot more clear and I agree with it wholeheartedly now! Nice work!

annethyme:
Sent out to the ACT Rules Community through public-act-r@w3.org:

> This is the FINAL CALL for the new rule "Links with identical accessible names serve equivalent purpose", https://github.com/act-rules/act-rules.github.io/pull/220.
>  
> The final call for this rule will end on May 14. 
> If you have any objections to the rule please raise them before this date.
>  
> We also encourage proceeding with implementations during the final call phase.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,280 @@
+---
+id: b20e66
```
WilcoFiers: Needs to be updated to the new format.

_rules/6cbcbe.md
```diff
@@ -0,0 +1,280 @@
...
+### Applicability
+
+This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names](#accessible-name) that do not only consist of [whitespace](#whitespace).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name](#accessible-name).
+
+### Expectation
+
+When followed, the links in each set of target elements resolve to the [same resource](#same-resource) or to [equivalent resources](#equivalent-purpose).
+Resolving the links includes potential redirects, if the redirects happen instantly.
```
WilcoFiers: This seems like it should be a note. Either add a new line between the two paragraphs, or put them on the same line.

pages/algorithms/equivalent-resource.md
```diff
@@ -0,0 +1,16 @@
...
+
+Non-identical resources can still be equivalent by equally complying to the expectation formed by the user when navigating to them, thus serving an equivalent purpose. This would usually involve that the advertised key content is the same.  
+
+Web pages and documents (e.g. PDFs, office formats etc.) may be equivalent resources, even if the resources:
+* are located on different URLs, including different domains
+* present different navigation options, e.g. through bread crumbs or local sub menus
+* contain different amounts of information and/or differently worded information
+* use different layouts.
+
+If all resources cover the user's expectations created by e.g. the link name equally well, the resources are considered to be equivalent.
```
WilcoFiers: I don't think you can use "e.g." in the middle of a sentence like this. I think this sentence is a little confusing. I've tried to come up with an alternative, but I'm struggling even trying to understand what you are saying here. It sounds like whether or not something is an equivalent resource depends on the link to that resource? I'm not sure that's how it works.

_rules/6cbcbe.md
audreymaniez: Thank you @kasperisager for the explanation. 

@WilcoFiers your comments on putting it as a note means that the redirection is not to be tested within that rule?

_rules/6cbcbe.md
annethyme: Removed the break between the two paragraphs. Is that enough to satisfy your feedback? 

If we move it into a note, it will be informative, and it shouldn't be. It's part of what makes the expectation.

pages/algorithms/equivalent-resource.md
annethyme: Moved the "e.g. the link name" into a note instead: 

> **Note:** The user's expectations for the resource can be formed by different things, e.g. the name of the link leading to the resource, with or without the context around the link. This depends on the accessibility requirement that is tested.

Does this work for you?

_rules/6cbcbe.md
WilcoFiers: In that case yeah, remove the line break. Either have something on the same line, or have an empty line between.

pages/algorithms/equivalent-resource.md
WilcoFiers: LGTM

PR #231, opened 2018-08-23
Rule: SC4-1-2-aria-attr-valid

jkodu:
`aria-*` valid attribute.

Closes issue:
- https://github.com/auto-wcag/auto-wcag/issues/200

## Reviewer checks

- [x] Make sure the pull request is prefixed with `Rule:` or `Algorithm:` or `Chore:` based on work involved.

# How to Review And Approve

### DIFF
--- a/_rules/SC3-1-1-html-xml-lang-match.md
+++ b/_rules/SC3-1-1-html-xml-lang-match.md
@@ -1,5 +1,5 @@
 ---
-name: SC3-1-1-html-xml-lang-match
+name: Matching HTML lang and xml:lang attribute
 
 description: |
  The rule checks that for the `html` element, there is no mismatch between the primary language in non-empty `lang` and `xml:lang` attributes, if both are used.
--- a/_rules/SC3-1-1-html-xml-lang-match.md
+++ b/_rules/SC3-1-1-html-xml-lang-match.md
@@ -1,5 +1,5 @@
 ---
-name: Matching HTML lang and xml:lang attribute
+name: SC3-1-1-html-xml-lang-match
 
 description: |
  The rule checks that for the `html` element, there is no mismatch between the primary language in non-empty `lang` and `xml:lang` attributes, if both are used.
--- a/_rules/SC4-1-1-aria-valid.md
+++ b/_rules/SC4-1-1-aria-valid.md
@@ -1,0 +1,114 @@
+---
+name: Validity of ARIA attribute specified
+
+description: |
+ This rule checks that the `aria-*` attribute specified is valid.
+
+success_criterion:
+- 4.1.1
+
+test_aspects:
+- DOM Tree
+
+authors:
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+Any DOM element within the `body` of a webpage with an `aria-*` attribute.
+
+### Expectation
+
+The `aria-*` attribute specified is a valid attribute.
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [https://www.w3.org/TR/html-aria/#index-aria-global](https://www.w3.org/TR/html-aria/#index-aria-global)
+- [http://www.w3.org/TR/wai-aria/#states_and_properties](http://www.w3.org/TR/wai-aria/#states_and_properties)
+- [http://www.w3.org/TR/WCAG20-TECHS/G108](http://www.w3.org/TR/WCAG20-TECHS/G108)
+- [http://oaa-accessibility.org/examples/](http://oaa-accessibility.org/examples/)
+- [http://www.w3.org/TR/WCAG20/#ensure-compat-rsv](http://www.w3.org/TR/WCAG20/#ensure-compat-rsv)
+-[https://developers.google.com/web/fundamentals/accessibility/semantics-aria/](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
+
+## Test cases
+
+### Passed
+
+#### Pass example 1
+
+Valid attribute `aria-atomic` on element `article`
+
+```html
+<article aria-atomic>This is a decription of something cool...</article>
+```
+
+#### Pass example 2
+
+Valid attribute `aria-modal` on element `div` with role `dialog`
+
+```html
+<div role="dialog" aria-modal>Contains modal content...</div>
+```
+
+#### Pass example 3
+
+Valid attribute `aria-live` on element with role `alert`
+
+```html
+<div role="alert" aria-live="assertive">Your session will expire in 60 seconds.</div>
+```
+
+#### Pass example 4
+
+Valid `aria-*` attributes on element `input` with role `spinbutton`
+
+```html
+<input
+  role="spinbutton"
+  aria-valuemax="100"
+  aria-valuemin="0"
+  aria-valuenow="25"
+  type="number"
+  value="25">
+```
+
+### Failed
+
+#### Fail example 1
+
+`aria-not-checked` is not a valid attribute.
+
+```html
+<li role="menuitemcheckbox" aria-not-checked="true">List Item</li>
+```
+
+#### Fail example 2
+
+`aria-labelled` is not a valid attribute.
+
+```html
+<span id="label">Birthday:</span>
+<div contenteditable role="searchbox" aria-labelled="label" aria-placeholder="MM-DD-YYYY">01-01-2019</div>
+```
+
+### Inapplicable
+
+#### Inapplicable example 1
+
+`aria-rowcount` not applicable to element `canvas`.
+
+```html
+<canvas aria-rowcount="2000">
+<canvas>
+```
--- a/_rules/SC4-1-2-aria-attr-valid.md
+++ b/_rules/SC4-1-2-aria-attr-valid.md
@@ -45,7 +45,7 @@ There are no major accessibility support issues known for this rule.
 
 ### Passed
 
-#### Pass example 1
+#### Passed example 1
 
 A valid ARIA 1.1 attribute `aria-atomic` is used on element `article`.
 
@@ -53,7 +53,7 @@ A valid ARIA 1.1 attribute `aria-atomic` is used on element `article`.
 <article aria-atomic>This is a decription of something cool...</article>
 ```
 
-#### Pass example 2
+#### Passed example 2
 
 A valid ARIA 1.1 attribute `aria-modal` on element `div` with role `dialog`
 
@@ -61,7 +61,7 @@ A valid ARIA 1.1 attribute `aria-modal` on element `div` with role `dialog`
 <div role="dialog" aria-modal>Contains modal content...</div>
 ```
 
-#### Pass example 3
+#### Passed example 3
 
 A valid ARIA 1.1 attribute `aria-live` on element `div` with role `alert`
 
@@ -69,7 +69,7 @@ A valid ARIA 1.1 attribute `aria-live` on element `div` with role `alert`
 <div role="alert" aria-live="assertive">Your session will expire in 60 seconds.</div>
 ```
 
-#### Pass example 4
+#### Passed example 4
 
 Muliple valid ARIA 1.1 attributes `aria-*` are specified on element `input` with role `spinbutton`
 
@@ -85,7 +85,7 @@ Muliple valid ARIA 1.1 attributes `aria-*` are specified on element `input` with
 
 ### Failed
 
-#### Fail example 1
+#### Failed example 1
 
 `aria-not-checked` is not a defined attribute in ARIA 1.1.
 
@@ -93,7 +93,7 @@ Muliple valid ARIA 1.1 attributes `aria-*` are specified on element `input` with
 <li role="menuitemcheckbox" aria-not-checked="true">List Item</li>
 ```
 
-#### Fail example 2
+#### Failed example 2
 
 `aria-labelled` is not a defined attribute in ARIA 1.1.
 
--- a/_rules/SC4-1-2-aria-attr-valid.md
+++ b/_rules/SC4-1-2-aria-attr-valid.md
@@ -1,8 +1,8 @@
 ---
-name: Validity of ARIA attribute
+name: ARIA attribute is valid
 
 description: |
- This rule checks that the `aria-` attribute specified is defined in ARIA 1.1.
+ This rule checks that each aria- attribute specified is defined in ARIA 1.1
 
 success_criterion:
 - 4.1.2
@@ -22,14 +22,11 @@ Any attribute that starts with `aria-`.
 
 ### Expectation
 
-Each target attribute which starts with `aria-` is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
+Each target attribute is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
 
 ## Assumptions
 
-`aria-` attributes are used to provide assistive technologies with accessibility information. This rule does not cater to any custom `aria-` attribute that does not fall under the [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) specification.
-
-This 
-If someone decides to make up their own `aria-` attribute and do some scripting to do other stuff with it, this rule won't account for that.
+This rule does not cater to any custom `aria-` attribute that does not fall under the [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) specification.
 
 ## Accessibility Support
 
@@ -37,12 +34,11 @@ There are no major accessibility support issues known for this rule.
 
 ## Background
 
-- [https://www.w3.org/TR/html-aria/#index-aria-global](https://www.w3.org/TR/html-aria/#index-aria-global)
-- [http://www.w3.org/TR/wai-aria/#states_and_properties](http://www.w3.org/TR/wai-aria/#states_and_properties)
-- [http://www.w3.org/TR/WCAG20-TECHS/G108](http://www.w3.org/TR/WCAG20-TECHS/G108)
-- [http://oaa-accessibility.org/examples/](http://oaa-accessibility.org/examples/)
-- [http://www.w3.org/TR/WCAG20/#ensure-compat-rsv](http://www.w3.org/TR/WCAG20/#ensure-compat-rsv)
-- [https://developers.google.com/web/fundamentals/accessibility/semantics-aria/](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
+- [ARIA in HTML](https://www.w3.org/TR/html-aria/#index-aria-global)
+- [WAI ARIA Supported States and Properties](http://www.w3.org/TR/wai-aria/#states_and_properties)
+- [G108: Using markup features to expose the name and role](http://www.w3.org/TR/WCAG20-TECHS/G108)
+- [WCAG 2.0 - Name, Role, Value: Understanding SC 4.1.2](https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html)
+- [Semantics and ARIA](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
 - [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/)
 
 ## Test cases
@@ -67,7 +63,7 @@ A valid ARIA 1.1 attribute `aria-modal` on element `div` with role `dialog`
 
 #### Pass example 3
 
-A valid ARIA 1.1 attribute `aria-live` on element with role `alert`
+A valid ARIA 1.1 attribute `aria-live` on element `div` with role `alert`
 
 ```html
 <div role="alert" aria-live="assertive">Your session will expire in 60 seconds.</div>
@@ -110,9 +106,9 @@ Muliple valid ARIA 1.1 attributes `aria-*` are specified on element `input` with
 
 #### Inapplicable example 1
 
-Element with out `aria-*` attribute.
+Element without `aria-*` attribute.
 
 ```html
 <canvas>
-<canvas>
+</canvas>
 ```
--- a/_rules/SC4-1-2-aria-attr-valid.md
+++ b/_rules/SC4-1-2-aria-attr-valid.md
@@ -18,15 +18,18 @@ authors:
 
 ### Applicability
 
-Any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) with an attribute that starts with `aria-`.
+Any attribute that starts with `aria-`.
 
 ### Expectation
 
-Each target element that has an attribute which starts with `aria-` is defined in ARIA 1.1.
+Each target attribute which starts with `aria-` is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
 
 ## Assumptions
 
-*There are currently no assumptions*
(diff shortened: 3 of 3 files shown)


_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
+---
```
WilcoFiers: I suggest "SC4-1-2-aria-attr-valid.md" ARIA valid is too generic IMO.

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+name: Validity of ARIA attribute specified
```
WilcoFiers: This doesn't seem grammatically correct, could be my English skills though.

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+description: |
+ This rule checks that the `aria-*` attribute specified is valid.
```
WilcoFiers: "is valid" seems a big vague here. I would suggest "is defined in ARIA 1.1"?

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+success_criterion:
+- 4.1.1
```
WilcoFiers: I don't think that's correct. Should be 4.1.2 IMO.

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+authors:
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+Any DOM element within the `body` of a webpage with an `aria-*` attribute.
```
WilcoFiers: Why within the body? I would also be more specific about the `aria-*`, something like "with an attribute that starts with `aria-`".

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Expectation
+
+The `aria-*` attribute specified is a valid attribute.
```
WilcoFiers: This should start with an "Each target element ...". It's also ambiguous about what it means for an aria attribute to be valid. I think what this should say is that all attributes that start with `aria-` should attributes defined in ARIA 1.1.

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+- [http://www.w3.org/TR/WCAG20/#ensure-compat-rsv](http://www.w3.org/TR/WCAG20/#ensure-compat-rsv)
+-[https://developers.google.com/web/fundamentals/accessibility/semantics-aria/](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
+
+## Test cases
+
+### Passed
+
+#### Pass example 1
+
+Valid attribute `aria-atomic` on element `article`
```
WilcoFiers: This description doesn't explain *why* the test case passes.

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+```html
+<article aria-atomic>This is a decription of something cool...</article>
+```
+
+#### Pass example 2
+
+Valid attribute `aria-modal` on element `div` with role `dialog`
```
WilcoFiers: Same, needs a better description. Why is this a test case?

_rules/SC4-1-1-aria-valid.md
```diff
@@ -0,0 +1,114 @@
...
+```html
+<span id="label">Birthday:</span>
+<div contenteditable role="searchbox" aria-labelled="label" aria-placeholder="MM-DD-YYYY">01-01-2019</div>
+```
+
+### Inapplicable
+
+#### Inapplicable example 1
+
+`aria-rowcount` not applicable to element `canvas`.
```
WilcoFiers: Why not? This is an element in the body with an aria-* attribute.

_rules/SC4-1-1-aria-valid.md
jkodu: Note to self:

- Keep it generic, and say any HTML or SVG element (research)

WilcoFiers:
@ShadowBB This rule relies on the assumption that ARIA attributes are used intentionally. That if a content author used an `aria-*` attribute, that they used it to express something, and not just because they were bored or something. If that's the case, if a content author meant to express something in ARIA, and it turned out not to be a valid ARIA attribute, than the logical conclusion is that they failed to express the state / property they meant to express, and so logically that element doesn't have the role/state/prop it should have.

We need to add an assumption to this rule that expresses this thought.

ShadowBB:
@WilcoFiers That is a large assumption (aria might be added in error by copy-pasting), but even if that assumption was true people seem to add redundant aria attributes (not needed to comply to WCAG), if a mistake was made in one of those, it would still not invalidate WCAG correct?

If you add the assumption that "The aria attribute was used with the intention to comply to WCAG" then I won't have any more doubts. Still a large assumption, but then the rule would be correct.

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+Any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) with an attribute that starts with `aria-`.
+
+### Expectation
+
+Each target element that has an attribute which starts with `aria-` is defined in ARIA 1.1.
+
+## Assumptions
+
+*There are currently no assumptions*
```
WilcoFiers: This has a basic assumption that `aria-` attributes are only used to express accessibility information that can be picked up by assistive technologies. If someone decides to make up their own `aria-` attribute and do some scripting to do other stuff with it, this rule won't account for that.

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,114 @@
...
+- [http://www.w3.org/TR/WCAG20/#ensure-compat-rsv](http://www.w3.org/TR/WCAG20/#ensure-compat-rsv)
+-[https://developers.google.com/web/fundamentals/accessibility/semantics-aria/](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
+
+## Test cases
+
+### Passed
+
+#### Pass example 1
+
+A valid ARIA1.1 attribute `aria-atomic` is used on element `article`.
```
WilcoFiers: Should be "ARIA 1.1" instead of "ARIA1.1".

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,114 @@
...
+```html
+<span id="label">Birthday:</span>
+<div contenteditable role="searchbox" aria-labelled="label" aria-placeholder="MM-DD-YYYY">01-01-2019</div>
+```
+
+### Inapplicable
+
+#### Inapplicable example 1
+
+`aria-rowcount` not applicable to element `canvas`, and also `aria-rowcount` needs a context role of `grid`.
```
WilcoFiers: I don't understand why this isn't applicable. It's an element with an attribute that starts with `aria-`

jkodu: You are right. I have added an inapplicable example where there is no `aria-*` attribute.

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any attribute that starts with `aria-`.
+
+### Expectation
+
+Each target attribute which starts with `aria-` is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
```
carlosapaduarte: We could drop the "which starts with aria-" part, since that is implied from the applicability

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Assumptions
+
+`aria-` attributes are used to provide assistive technologies with accessibility information. This rule does not cater to any custom `aria-` attribute that does not fall under the [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) specification.
```
carlosapaduarte: Not sure if we need the first sentence. But it does make it easier to understand the second.

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+
+This
```
carlosapaduarte: remove

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+### Expectation
+
+Each target attribute which starts with `aria-` is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
+
+## Assumptions
+
+`aria-` attributes are used to provide assistive technologies with accessibility information. This rule does not cater to any custom `aria-` attribute that does not fall under the [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) specification.
+
+This 
+If someone decides to make up their own `aria-` attribute and do some scripting to do other stuff with it, this rule won't account for that.
```
carlosapaduarte: This is what is expressed in the previous paragraph. Do we need to repeat it?

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+```html
+<span id="label">Birthday:</span>
+<div contenteditable role="searchbox" aria-labelled="label" aria-placeholder="MM-DD-YYYY">01-01-2019</div>
+```
+
+### Inapplicable
+
+#### Inapplicable example 1
+
+Element with out `aria-*` attribute.
```
carlosapaduarte: r/with out/without

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+
+```html
+<canvas>
+<canvas>
```
carlosapaduarte: shouldn't the second be `</canvas>`

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+
+A valid ARIA 1.1 attribute `aria-modal` on element `div` with role `dialog`
+
+```html
+<div role="dialog" aria-modal>Contains modal content...</div>
+```
+
+#### Pass example 3
+
+A valid ARIA 1.1 attribute `aria-live` on element with role `alert`
```
carlosapaduarte: on element `div`

_rules/SC4-1-2-aria-attr-valid.md
nitedog: I agree with removing the first sentence.

_rules/SC3-1-1-html-xml-lang-match.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: SC3-1-1-html-xml-lang-match
+name: Matching HTML lang and xml:lang attribute
```
annethyme: Why is this here?

I would suggest "HTML lang and xml:lang attribute matches"

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
+---
+name: Validity of ARIA attribute
```
annethyme: I suggest "ARIA attribute is valid"

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+
+description: |
+ This rule checks that the `aria-` attribute specified is defined in ARIA 1.1.
```
annethyme: "This rule checks that each `aria-` attribute specified is defined in ARIA 1.1"

_rules/SC4-1-2-aria-attr-valid.md
annethyme: Quoting @kasperisager from https://github.com/auto-wcag/auto-wcag/pull/233: 
> "If someone happens to be using aria-* attributes for something other than what is defined by WAI-ARIA, the rule would still be correct; whoever is doing such a thing should stop as they're messing with well-defined semantics of the web. Otherwise, every rule should include assumptions like "we assume that the author is using  elements for images, not for expressing custom state used in scripts". That just seems silly"

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+This 
+If someone decides to make up their own `aria-` attribute and do some scripting to do other stuff with it, this rule won't account for that.
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [https://www.w3.org/TR/html-aria/#index-aria-global](https://www.w3.org/TR/html-aria/#index-aria-global)
```
annethyme: Clean up links so that the link texts are not just URLs

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+- [http://oaa-accessibility.org/examples/](http://oaa-accessibility.org/examples/)
+- [http://www.w3.org/TR/WCAG20/#ensure-compat-rsv](http://www.w3.org/TR/WCAG20/#ensure-compat-rsv)
+- [https://developers.google.com/web/fundamentals/accessibility/semantics-aria/](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/)
+- [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/)
+
+## Test cases
+
+### Passed
+
+#### Pass example 1
```
annethyme: "Passed example 1" ?

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,118 @@
...
+---
+
+## Test procedure
+
+### Applicability
+
+Any attribute that starts with `aria-`.
+
+### Expectation
+
```
annethyme: Does this rule only take into account if the attribute exists somewhere in the ARIA spec, or does it also consider whether it is a supported state/property for that role? 
The variation in elements in the test cases seem to indicate the last, but it is not specified in the Expectations.

_rules/SC3-1-1-html-xml-lang-match.md
kasperisager: Ditto. In general, can we please refrain from introducing unrelated changes in pull requests?

_rules/SC4-1-2-aria-attr-valid.md
jkodu: We seem to used `pass` more than `passed` and `fail` more than `failed` in the examples. Refer all the published rules. Need to agree before making any change.

_rules/SC4-1-2-aria-attr-valid.md
nitedog: I think "starts with 'aria-'" is bad practice. It should be "Any attribute defined by <ARIA-spec-location>" or such.

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,114 @@
...
+Each target attribute is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
+
+## Assumptions
+
+This rule does not cater to any custom `aria-` attribute that does not fall under the [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) specification.
```
nitedog: If you provide a specification in the applicability, you will not need this here because you already defined it to exclude custom 'aria-' attributes.

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any attribute that starts with `aria-`.
+
+### Expectation
+
+Each target attribute is defined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
```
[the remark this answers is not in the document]
kasperisager: I would question whether or not this is a WCAG violation as well.

_rules/SC4-1-2-aria-attr-valid.md
carlosapaduarte: In this rule the applicability needs to be "starts with aria-" because the rule is checking if anything that starts with aria- is indeed in the ARIA specification.

_rules/SC4-1-2-aria-attr-valid.md
carlosapaduarte: I'm not sure if I understand this assumption. The rule applies to 'aria-' attributes and checks if they are in the specification. The assumption seems to say that the 'aria-' attributes that are not in the specification are not applicable. Isn't this contradictory?

_rules/SC4-1-2-aria-attr-valid.md
carlosapaduarte: I agree with @annethyme that we need to make it clear if the rule just checks if the attribute exists in the spec, or if it looks also at it being supported for the role (which I don't think if the case for this rule). A note would make it explicit.

_rules/SC4-1-2-aria-attr-valid.md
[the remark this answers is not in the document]
carlosapaduarte: I haven't been able to found anything in a spec supporting that this a violation of WCAG. However, I agree with @WilcoFiers that when someone uses an 'aria-' attribute is because they want to use some ARIA state or property and not because they just felt like creating a new attribute. But this is something that needs to be clearly expressed in the assumptions.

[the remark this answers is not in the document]
annethyme: I suggest changing the link to `[WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def)`

Not either sure that this is an actual requirement violation, but we do have several of these types, e.g. testing that only roles defined in WAI-ARIA are used etc. 

The reason we agreed that we could have these as ACT rules was that by using something ARIA-related, authors indicate that they have identified a need for AT accommodation, and they intend to meet it. If they then don't manage to do it correctly, there will likely be a need for accommodation, that is mistakenly is still unmet. 

So it's kind of like a spell-checking service for WAI-ARIA.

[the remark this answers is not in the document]
annethyme: Actually, thinking about it, was this rule intended for something else, like checking that only supported states and properties are used? 
Then maybe it should be something along these lines: 

Applicability: 
Any [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.

Expectation: 
The target state or property is a [supported state or property](https://www.w3.org/TR/wai-aria-1.1/#supportedState) for the [semantic role](#semantic-role) of the element it is specified on. 
(and then something about elements without a semantic role)

_rules/SC4-1-2-aria-attr-valid.md
annethyme: It's probably because it was in the template at some point... 
I just find it inconsistent...

_rules/SC4-1-2-aria-attr-valid.md
```diff
@@ -0,0 +1,114 @@
...
+
+authors:
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+Any attribute that starts with `aria-`.
```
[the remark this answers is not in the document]
jkodu: I had it previously, but there was an editorial recommendation to make it more concise like so. Think it conveys the meaning very well either-way.

_rules/SC4-1-2-aria-attr-valid.md
jkodu: Updated to `Passed` and `Failed` accordingly.

_rules/SC4-1-2-aria-attr-valid.md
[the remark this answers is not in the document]
jkodu: @annethyme thanks for the explanation that indeed some of our rules are `spell-checking` service for WAI-ARIA.

And nope, this rule is not limited to just states and properties.

_rules/SC4-1-2-aria-attr-valid.md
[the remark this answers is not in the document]
kasperisager: Given that ARIA attributes can also be used on MathML elements, limiting the rule to HTML and SVG elements might also not be appropriate.

_rules/SC4-1-2-aria-attr-valid.md
jkodu: @annethyme - the rule only takes into account that the attribute exists in the ARIA spec.

@nitedog - The currently applicability and expectations is very simple and straight-forward. Your suggestion `any attribute defined by ...` will combine the expectation with the applicability. I would prefer to leave it as is. 

@carlosapaduarte - hope the above two comments answers your concerns too.

_rules/SC4-1-2-aria-attr-valid.md
jkodu: The idea behind this assumption is to handle the edge case, where in custom attributes may be defined and scripted upon.

It would not see it as contradictory, as we emphasise `custom`.

carlosapaduarte: @jkodu so this would mean that if a developer uses consciously a custom aria- attribute then he may ignore the outcome of this rule? Just wondering if we should condone this usage of attributes starting with aria-

jkodu:
@kasperisager @annethyme 

Now I have concerns with the published rule.
- As we use `aria-*` instead of defined. See comment - https://github.com/auto-wcag/auto-wcag/pull/231#discussion_r222700589
- Only seem to target `HTML` and `SVG`.

@nitedog 
We have added this for f2f agenda to reach a consensus.

annethyme:
I don't think we can wait for the f2f to get clarification, if we need to implement it all this month. 

@WilcoFiers and @jkodu, it seems that you have both the "ARIA allowed attribute" and "ARIA valid attribute" implemented in your engine. Can you see any cases where "ARIA valid attribute" would catch something that "ARIA allowed attribute" won't?
Because to me it seems that @kasperisager is right about complete overlap. 

Answer to comments from @jkodu: 
- using "aria-": I do see the point, but can't really see how we are to get around it, if we want to incorporate a bit of a spellchecking service too. @nitedog, any ideas? Or else, I suggest we leave the published rule as is for now, and have a second look in November, when we are under a bit less pressure.
- "element" vs. "HTML or SVG element" - to me it seems that we have always been using "HTML  or SVG element" in rules, except when we forget it. Happy to adjust, though, if there is a reason.

nitedog:
@annethyme @jkodu I think the new "Accessibility Requirements Mapping" section could help me to understand this rule. I do not believe that having a custom aria- property not (yet) defined by in the WAI-ARIA spec is a violation of WCAG 4.1.2 but maybe I'm wrong.

annethyme:
@nitedog: what if it is properties defined in WAI-ARIA, but placed on elements where they ar e not supported?

nitedog:
@annethyme I think only if they lead to the name and role of the user interface component being not programmatically determinable, or if they cause interference for the user.

jkodu:
There is an issue open for updating the aria-* rules - https://github.com/auto-wcag/auto-wcag/issues/335

Issue #176, opened 2018-07-10
SC4-1-2-aria-attribute-has-valid-value


Brynanders:
---
name: ARIA attribute has valid value
description: |
  This rule checks that aria-* attributes have a valid value
  
success_criterion:
- 4.1.2 # Name, Role, Value

test_aspects:
- DOM Tree
- CSS Styling

authors:
- Anne Thyme Nørregaard
---

### Applicability

Any non-empty attribute that is placed on an HTML or SVG element, and where the attribute name starts with ´aria-´.

### Expectation

Each of the applicable attributes have a valid value according to the [value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value) for the given state or property (´aria-*´attribute) as specified by [WAI-ARIA](https://www.w3.org/TR/wai-aria).

**Note:** 
- If the value of the ´aria-*´ attribute is of value type `ID Reference` or `ID Reference List` ensure that an element with the given id exists in the document. 
- If the value of the ´aria-*´ attribute is of value type `URI` ensure that it matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).

## Assumptions

- Checking that linked resources in ´aria-*´ attributes with value type `URI` are available is considered out of scope for this rule. For this rule it is considered sufficient that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).

## Accessibility Support

_There are no major accessibility support issues known for this rule._

## Background

- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
- [Overview of possible ´aria-*´ attribute value types](https://www.w3.org/TR/wai-aria/#propcharacteristic_value)
-  [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)

### Passed

Element has correct value for aria-attribute

```html
<div id="someElementId">...</div>
<select role="combobox" aria-controls="someElementId"></select>
```

### Failed

Element has wrong value type for aria attribute

```html
<select role="combobox" aria-controls=""></select>
```

### Inapplicable

Element does not have any aria-attributes

```html
<div>Some Content</div>
```

Element has ARIA role, but no attributes that start with "aria-"

```html
<div role="button">Some Content</div>
```

annethyme:
Pull request for this: https://github.com/auto-wcag/auto-wcag/pull/233

PR #233, opened 2018-08-27
SC4-1-2-aria-state-or-property-value-valid

annethyme:
<< Describe the changes >>

Closes issue: #176 

## Reviewer checks

- [X] Make sure the pull request is prefixed with `Rule:` or `Algorithm:` or `Chore:` based on work involved.

# How to Review And Approve

### DIFF
--- a/_rules/SC4-1-2-aria-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-aria-attribute-has-valid-value.md
@@ -43,7 +43,7 @@ _There are no major accessibility support issues known for this rule._
 ## Background
 
 - [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
-[ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
+- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
 - [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def)
 - [Overview of possible ´aria-*´ attribute value types](https://www.w3.org/TR/wai-aria/#propcharacteristic_value)
 - [Uniform Resource Identifier (URI): Generic Syntax (RFC 3986)](https://www.ietf.org/rfc/rfc3986.txt)
--- a/_rules/SC4-1-2-aria-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-aria-attribute-has-valid-value.md
@@ -1,8 +1,10 @@
 ---
 name: ARIA attribute has valid value
+test_type: atomic
+
 description: |
-  This rule checks that `aria-*` attributes have a valid value
-  
+   This rule checks that all ARIA 1.1 attributes (`aria-*`) have valid values
+   
 success_criterion:
 - 4.1.2 # Name, Role, Value
 
@@ -11,6 +13,7 @@ test_aspects:
 - CSS Styling
 
 authors:
+- Wilco Fiers
 - Anne Thyme Nørregaard
 ---
 
@@ -18,18 +21,19 @@ authors:
 
 ### Applicability
 
-Any non-empty attribute that is placed on an HTML or SVG element, and where the attribute name starts with `aria-`.
+Any [non-empty](#non-empty) attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies), and where the attribute name starts with `aria-` and where the attribute can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
 
 ### Expectation
 
-Each of the applicable attributes have a valid value according to the [value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value) for the given state or property (`aria-*` attribute) as specified by [WAI-ARIA](https://www.w3.org/TR/wai-aria).
+Each test target has a valid value according to the [value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value) for that `aria-*` attribute as specified by [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
 
 **Note:** 
-- If the value of the `aria-*` attribute is of value type `ID Reference` or `ID Reference List` ensure that an element with the given id exists in the document. 
-- If the value of the `aria-*` attribute is of value type `URI` ensure that it matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document. 
+- For value type `URI` ensure that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt). This rule does not require that the destination URL exists. 
 
 ## Assumptions
 
+- Checking that IDs referenced in value types `ID Reference` and `ID Reference List` is considered out of scope for this rule. 
 - Checking that linked resources in `aria-*` attributes with value type `URI` are available is considered out of scope for this rule. For this rule it is considered sufficient that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
 
 ## Accessibility Support
@@ -38,7 +42,9 @@ _There are no major accessibility support issues known for this rule._
 
 ## Background
 
-- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+[ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
+- [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def)
 - [Overview of possible ´aria-*´ attribute value types](https://www.w3.org/TR/wai-aria/#propcharacteristic_value)
 - [Uniform Resource Identifier (URI): Generic Syntax (RFC 3986)](https://www.ietf.org/rfc/rfc3986.txt)
 
@@ -46,22 +52,171 @@ _There are no major accessibility support issues known for this rule._
 
 ### Passed
 
-Element has correct value for `aria-*` attribute
+#### Passed example 1
 
-```html
-<div id="someElementId">...</div>
-<select role="combobox" aria-controls="someElementId"></select>
+Element with valid true/false value
+
+ ```html
+<div role="textbox" aria-required="true"></div>
+```
+
+ #### Passed example 2
+ 
+ Element with valid true/false/undefined value
+ 
+ ```html
+<div role="button" aria-expanded="undefined"></div>
+```
+
+ #### Passed example 3
+ 
+ Element with valid tristate value
+ 
+ ```html
+<div role="button" aria-pressed="mixed"></div>
+```
+
+ #### Passed example 4
+ 
+ Element with valid ID reference value
+ 
+ ```html
+<div role="textbox" aria-errormessage="my-error"></div>
+```
+
+ #### Passed example 5
+ 
+ Element with valid ID reference list value
+ 
+ ```html
+<div role="combobox" aria-owns="my-textbox my-grid"></div>
+```
+
+ #### Passed example 6
+ 
+ Element with valid integer value
+ 
+ ```html
+<div role="gridcell" aria-rowindex="2">Fred</div>
+```
+
+ #### Passed example 7
+ 
+ Element with valid number value
+ 
+ ```html
+<div role="spinbutton" aria-valuemin="1.0" aria-valuemax="2.0" aria-valuenow="1.5"></div>
+```
+
+ #### Passed example 8
+ 
+ Element with valid string value
+ 
+ ```html
+<div role="searchbox" aria-placeholder="MM-DD-YYYY">MM-DD-YYYY</div>
+```
+
+ #### Passed example 9
+ 
+ Element with valid token value, attribute inappropriate for the role.
+ 
+ ```html
+<div role="button" aria-orientation="horizontal"></div>
+```
+
+ #### Passed example 10
+ 
+ Element with valid token list value
+ 
+ ```html
+<div role="dialog" aria-dropeffect="copy move"></div>
 ```
 
 ### Failed
 
-Element has wrong value type for `aria-*` attribute
+#### Failed example 1
+
+Element with invalid true/false value
 
 ```html
-<select role="combobox" aria-controls=""></select>
+<div role="textbox" aria-required="undefined"></div>
+```
+
+#### Failed example 2
+
+Element with invalid true/false/undefined value
+
+```html
+<div role="button" aria-expanded="mixed"></div>
+```
+
+#### Failed example 3
+
+Element with invalid tristate value
+
+```html
+<div role="button" aria-pressed="horizontal"></div>
+```
+
+#### Failed example 4
+
+Element with invalid ID reference value, space not allowed in a single ID
+
+```html
+<div role="textbox" aria-errormessage="error1 error2"></div>
+```
+
+#### Failed example 5
+
+Element with invalid integer value
+
+```html
+<div role="gridcell" aria-rowindex="2.5">Fred</div>
+```
+
+#### Failed example 6
+
+Element with invalid number value
+
+```html
+<div role="spinbutton" aria-valuemin="one" aria-valuemax="three" aria-valuenow="two"></div>
+```
+
+#### Failed example 7
+
+Element with invalid token value
+
+```html
+<div role="main" aria-live="nope"></div>
+```
+
+#### Failed example 8
+
+Element with invalid token list value
+
+```html
+<div role="dialog" aria-dropeffect="invalid move"></div>
+```
+
+#### Failed example 9
+
+Element with invalid empty aria attribute for token value
+
+```html
+<div role="button" aria-expanded="collapsed" aria-live="off"></div>
+```
+ 
+#### Failed example 10
+ 
+Element with invalid true/false/undefined value for custom element
+ 
+```html
+<my-button role="button" aria-expanded="collapsed"></my-button>
 ```
 
 ### Inapplicable
+
+#### Inapplicable example 1
 
 Element does not have any `aria-*` attributes
 
@@ -69,8 +224,33 @@ Element does not have any `aria-*` attributes
 <div>Some Content</div>
 ```
 
+#### Inapplicable example 2
+
 Element has ARIA role, but no attributes that start with `aria-`
 
 ```html
 <div role="button">Some Content</div>
+```
+
+#### Inapplicable example 3
+
+Element with null ARIA attribute
+```html
+<div role="checkbox" aria-checked></div>
+```
+
+#### Inapplicable example 4
+
+Element with empty ARIA attribute
+
+```html
+<div role="searchbox" aria-labelledby=""></div>
+```
+
+#### Inapplicable example 5
+
+Element with ARIA attribute, where element isn't HTML or SVG
+  
+ ```html
+  <math aria-hidden="true"></math>
 ```
--- a/_rules/SC4-1-2-aria-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-aria-attribute-has-valid-value.md
@@ -21,7 +21,7 @@ authors:
 
 ### Applicability
 
-Any [non-empty](#non-empty) attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies), and where the attribute name starts with `aria-` and where the attribute can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
+Any [non-empty](#non-empty) attribute that is placed on an HTML or SVG element, where the attribute name starts with `aria-` and can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
 
 ### Expectation
 
@@ -33,8 +33,9 @@ Each test target has a valid value according to the [value type](https://www.w3.
 
 ## Assumptions
 
-- Checking that IDs referenced in value types `ID Reference` and `ID Reference List` is considered out of scope for this rule. 
-- Checking that linked resources in `aria-*` attributes with value type `URI` are available is considered out of scope for this rule. For this rule it is considered sufficient that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+- This rule assumes that WAI-ARIA attributes on elements that are not [exposed to assistive technologies](#exposed-to-assistive-technologies)) can still impact users. For example, anything referenced through aria-labelledby does not have to be exposed to assistive technologies in order for it to become part of the accessible name. Therefore this rule is not limited to elements that are exposed to assistive technologies.
+- This rule assumes that the author of the web content is only using `"aria-*"` attributes to express logic defined in the WAI-ARIA specifications. If an someone is using `"aria-*"` attributes to express custom states/properties that they are using in scripts for other purposes, this rule would be incorrect.
+
 
 ## Accessibility Support
 
--- a/_rules/SC4-1-2-aria-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-aria-attribute-has-valid-value.md
@@ -3,7 +3,7 @@ name: ARIA attribute has valid value
 test_type: atomic
 
 description: |
-   This rule checks that all ARIA 1.1 attributes (`aria-*`) have valid values
+   This rule checks that all ARIA 1.1 states and properties have valid values
    
 success_criterion:
 - 4.1.2 # Name, Role, Value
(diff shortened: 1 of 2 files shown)


_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,76 @@
...
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any non-empty attribute that is placed on an HTML or SVG element, and where the attribute name starts with `aria-`.
```
WilcoFiers: This leaves a gap for attributes that aren't specified in ARIA 1.1. It's unclear if they should fail or be inapplicable.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Expectation
+
+Each of the applicable attributes have a valid value according to the [value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value) for the given state or property (`aria-*` attribute) as specified by [WAI-ARIA](https://www.w3.org/TR/wai-aria).
+
+**Note:** 
+- If the value of the `aria-*` attribute is of value type `ID Reference` or `ID Reference List` ensure that an element with the given id exists in the document.
```
WilcoFiers: I don't think this should be required. I've seen things like `aria-errormessage` being used statically, and relying on the presence or absence of an element as a way to toggle the errormessage. There are legitimate use cases for defining non-existing elements for many ARIA idref attributes. The only thing there is that an `ID Reference` isn't allowed to contain spaces.

ShadowBB:
I think the HTML or SVG elements should be exposed to assistive technology to be applicable.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
annethyme: Rule now specifies that attributes should be listed in WAI-ARIA 1.1

_rules/SC4-1-2-aria-attribute-has-valid-value.md
annethyme: Note now says this is out of scope.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,256 @@
...
+authors:
+- Wilco Fiers
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies), and where the attribute name starts with `aria-` and where the attribute can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
```
WilcoFiers: I don't think this meets the "plain language" requirement. We should probably break this up or list it out in some way to make this easier to read.

I don't think that "exposed to assistive technologies" should be part of this. There are a number of ways that ARIA on elements that are not exposed to assistive technologies will impact users. For example, anything referenced through aria-labelledby does not have to be exposed to assistive technologies in order for it to become part of the accessible name.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Expectation
+
+Each test target has a valid value according to the [value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value) for that `aria-*` attribute as specified by [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
+
+**Note:** 
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document. 
+- For value type `URI` ensure that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt). This rule does not require that the destination URL exists. 
+
+## Assumptions
```
WilcoFiers: These aren't really assumptions. Your note is pretty clear on this part. I think these should be removed.

WilcoFiers: One more thing. There is an assumption here that someone is only using aria- attributes to express logic defined in the ARIA specifications. If someone is using aria- attributes to express custom states / properties that they are using in scripts for other purposes, this rule would be incorrect.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,257 @@
...
+authors:
+- Wilco Fiers
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) attribute that is placed on an HTML or SVG element, where the attribute name starts with `aria-` and can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
```
kasperisager: "placed on" -> "specified on".

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,257 @@
...
+
+**Note:** 
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document. 
+- For value type `URI` ensure that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt). This rule does not require that the destination URL exists. 
+
+## Assumptions
+
+- This rule assumes that WAI-ARIA attributes on elements that are not [exposed to assistive technologies](#exposed-to-assistive-technologies)) can still impact users. For example, anything referenced through aria-labelledby does not have to be exposed to assistive technologies in order for it to become part of the accessible name. Therefore this rule is not limited to elements that are exposed to assistive technologies.
+- This rule assumes that the author of the web content is only using `"aria-*"` attributes to express logic defined in the WAI-ARIA specifications. If an someone is using `"aria-*"` attributes to express custom states/properties that they are using in scripts for other purposes, this rule would be incorrect.
```
kasperisager: I don't quite agree with the last assumption. If someone happens to be using `aria-*` attributes for something other than what is defined by WAI-ARIA, the rule would still be correct; whoever is doing such a thing should **stop** as they're messing with well-defined semantics of the web. Otherwise, every rule should include assumptions like "we assume that the author is using `` elements for images, not for expressing custom state used in scripts". That just seems silly 🙈

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,257 @@
...
+### Applicability
+
+Any [non-empty](#non-empty) attribute that is placed on an HTML or SVG element, where the attribute name starts with `aria-` and can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
+
+### Expectation
+
+Each test target has a valid value according to the [value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value) for that `aria-*` attribute as specified by [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
+
+**Note:** 
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document.
```
annethyme: "... unless the property is required by the semantic role of the element on which the property is specified."

_rules/SC4-1-2-aria-attribute-has-valid-value.md
annethyme: Assumption was added, but then removed again after feedback from other side.

kasperisager: For reference: https://github.com/auto-wcag/auto-wcag/pull/233#discussion_r221615564

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,255 @@
...
+authors:
+- Wilco Fiers
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) attribute that is specified on an HTML or SVG element, where the attribute name starts with `aria-` and can be found in [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
```
carlosapaduarte: We don't need to specify that the attribute name starts with `aria-`. All names in the WAI-ARIA definition do, so it is redundant.
On the other hand, I understand if you wish to keep it, since it will help to optimize the applicability decision process.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,255 @@
...
+
+**Note:** 
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document, unless the attribute is a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) for the [semantic role](#semantic-role) of the element on which the property is specified.
+- For value type `URI` ensure that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt). This rule does not require that the destination URL exists. 
+
+## Assumptions
+
+- This rule assumes that WAI-ARIA attributes on elements that are not [exposed to assistive technologies](#exposed-to-assistive-technologies)) can still impact users. For example, anything referenced through aria-labelledby does not have to be exposed to assistive technologies in order for it to become part of the accessible name. Therefore this rule is not limited to elements that are exposed to assistive technologies.
```
carlosapaduarte: Update [exposed to assistive technologies] to [included in the accessibility tree]

_rules/SC4-1-2-aria-attribute-has-valid-value.md
```diff
@@ -0,0 +1,255 @@
...
+
+Element with invalid token list value
+
+```html
+<div role="dialog" aria-dropeffect="invalid move"></div>
+```
+
+#### Failed example 9
+
+Element with invalid empty aria attribute for token value
```
carlosapaduarte: I don't understand this one. It doesn't seem to match with the code provided below.

_rules/SC4-1-2-aria-attribute-has-valid-value.md
nitedog: Thinking of it, I tend to agree with Carlos. Is it not sufficient to say "where attribute is listed in ..."?

annethyme: I think the idea was to exclude role="", that is also an attribute listed in WAI-ARIA 1.1...

_rules/SC4-1-2-aria-attribute-has-valid-value.md
annethyme: @WilcoFiers, I think this example came from your original rule. 
Do you know what was the intention here?  
Or else I will just remove it, since I can't either find out now what it is supposed to illustrate that is not already covered by other test cases...

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,248 @@
...
+### Applicability
+
+Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
+
+### Expectation
+
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
+
+**Note:** 
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document, unless the property is a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) for the [semantic role](#semantic-role) of the element on which the property is specified.
```
nitedog: This sounds like it should part of the expectation and/or applicability rather than an note. Here is a suggestion:

Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value), except for the value types `ID Reference` and `ID Reference List` on a property that is a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) for the [semantic role](#semantic-role) of the element on which the property is specified.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,248 @@
...
+- For value type `URI` ensure that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt). This rule does not require that the destination URL exists.
```
nitedog: Also here the first sentence set the expectation and should not be a note. It needs to move into the expectation paragraph as a sentence like this (removed "ensure" to match the usual editorial style of expectations):

For value type `URI`, a valid value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt)

Then the note is only the second sentence, which is a clarification but not a requirement.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Note:** 
+- For value types `ID Reference` and `ID Reference List` this rule does not require that elements with the given id(s) exists in the document, unless the property is a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) for the [semantic role](#semantic-role) of the element on which the property is specified.
+- For value type `URI` ensure that the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt). This rule does not require that the destination URL exists. 
+
+## Assumptions
+
+- This rule assumes that [WAI-ARIA 1.1 states and properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) on elements that are not [included in the accessibility tree](#exposed-to-assistive-technologies) or [focusable](#focusable) can still impact users. For example, anything referenced through `aria-labelledby` does not have to be included in the accessibility tree in order for it to become part of the [accessible name](#accessible-name). Therefore this rule is not limited to elements that are included in the accessibility tree or focusable.
```
nitedog: I recommend splitting the assumption from the (helpful but) background explanation:

Assumption: All [WAI-ARIA 1.1 states and properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def), including those on elements that are not [included in the accessibility tree](#exposed-to-assistive-technologies) or [focusable](#focusable), are part of the applicability.

Note: Elements that are not [included in the accessibility tree](#exposed-to-assistive-technologies) or [focusable](#focusable) can still impact users. For example, anything referenced through `aria-labelledby` does not have to be included in the accessibility tree in order for it to become part of the [accessible name](#accessible-name). Therefore this rule is not limited to elements that are included in the accessibility tree or focusable.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Applicability
+
+Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
+
+### Expectation
+
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
+
+For value types `ID Reference` and `ID Reference List`, a valid value for a [WAI-ARIA required propertiy](https://www.w3.org/TR/wai-aria-1.1/#requiredState) requires that the element(s) with the given id(s) exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that sepcifies the target attribute. For non-required properties, this is not a requirement.
```
nitedog: sepcifies  -> specifies

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Background
+
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/ARIA5)
+- [WAI-ARIA 1.1, Definitions of States and Properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def)
+- [WAI-ARIA 1.1, Characteristics of States and Properties, Value](https://www.w3.org/TR/wai-aria/#propcharacteristic_value)
+- [Uniform Resource Identifier (URI): Generic Syntax (RFC 3986)](https://www.ietf.org/rfc/rfc3986.txt)
+
+## Test Cases
```
carlosapaduarte: We don't have a test that considers the second point in the expectation. We can add this

#### Passed example 11

`aria-controls` with `ID Reference` that references an element existing in the same document tree.

```html
<div id="content1">Lorem ipsum...</div>
<div role="scrollbar" aria-controls="content1" aria-orientation="vertical" aria-valuemax="100" aria-valuemin="0" aria-valuenow="25"></div>
```

#### Failed example 10

`aria-controls` with `ID Reference` referencing an element that does not exist in the same document tree.

```html
<div role="scrollbar" aria-controls="content1" aria-orientation="vertical" aria-valuemax="100" aria-valuemin="0" aria-valuenow="25"></div>
```

carlosapaduarte: I forgot to add in the test cases description that `aria-controls` is a required property for elements with `role` of `scrollbar`, which is why the element referenced in the `ID Reference` must exist in the document tree. Please update accordingly.

annethyme: Well spotted! And thanks for making it so easy to include comments by just copy/pasting :)

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Applicability
+
+Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
+
+### Expectation
+
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
+
+For value types `ID Reference` and `ID Reference List`, a valid value for a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) requires that the element(s) with the given id(s) exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that specifies the target attribute. For non-required properties, this is not a requirement.
```
WilcoFiers: I disagree that this should be required. There are good use cases for using IDs of elements that don't yet exist. An example I came across pretty recently was one where aria-errormessage had a static value, and by adding or removing the element with that ID would they toggle the error message.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,270 @@
...
+
+For value type `URI`, a valid value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+
+**Note:** 
+For value type `URI`, this rule does not require that the destination URI exists. 
+
+## Assumptions
+
+- All [WAI-ARIA 1.1 states and properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def), including those on elements that are not [included in the accessibility tree](#exposed-to-assistive-technologies) or [focusable](#focusable), are part of the applicability.
```
WilcoFiers: I don't understand how this is an assumption.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,270 @@
...
+```html
+<div role="button" aria-pressed="mixed"></div>
+```
+
+#### Passed example 4
+ 
+`aria-errormessage` property with valid ID reference value
+ 
+```html
+<div role="textbox" aria-errormessage="my-error"></div>
```
WilcoFiers: While I agree this shouldn't fail, the current expectation says something else.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
kasperisager: This is only a requirement for required properties, as also mentioned in the last sentence. If an `aria-controls` property is required, for example, it makes little sense to consider an empty value invalid and then on the other hand consider a value that points to a non-existing ID valid as the two are essentially equivalent.

jkodu:
@annethyme - please address the review comments.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
annethyme: @WilcoFiers: Better?

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
annethyme: Not for this one, since aria-errormessage is not a required property for role=textbox

annethyme:
Expectation changed so that `ID reference list` for required properties only requires one id to exist in the document. Test case Passed example 11 updated to reflect this change. 
Please review again.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Applicability
+
+Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
+
+### Expectation
+
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
+
+For value types `ID Reference` and `ID Reference List`, a valid value for a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) requires that at least one of the elements with the given ids exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that specifies the target attribute. For non-required properties, this is not a requirement.
```
nitedog: The first sentence reads like a statement rather than an expectation. It is also a long and confusing sentence. Maybe say:

> For value types `ID Reference` and `ID Reference List`, at least one of the elements with the given ids exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that specifies the target attribute.

The part "a valid value for a [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) requires that" was removed. If this is needed, then it should be a note like this:

> Note: This behavior is defined in [WAI-ARIA required property](https://www.w3.org/TR/wai-aria-1.1/#requiredState)

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,270 @@
...
+
+For value type `URI`, a valid value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
```
nitedog: Should these be multiple expectations (ie. expectation 1, 2, 3, etc. rather than multiple paragraphs in the same section)?

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,270 @@
...
+
+**Note:** 
+For value type `URI`, this rule does not require that the destination URI exists. 
+
+## Assumptions
+
+- This rule assumes that elements that are not [included in the accessibility tree](#included-in-the-accessibility-tree) or [focusable](#focusable) can still impact users. Therefore the applicability of this rule is not limited to [WAI-ARIA 1.1 states and properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) on elements that are included in the accessibility tree or focusable.
```
nitedog: "included in the accessibility tree or focusable" -> "included in the accessibility tree or are focusable"

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,276 @@
...
+For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that specifies the target attribute. 
+
+**Note:** This expectation only applies to [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState). For non-required properties, there is no requirement that the elements with the given ids actually exists.
+
+### Expectation 3
+
+For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+
+**Note:** 
+For value type `URI`, this rule does not require that the destination URI exists.
```
kasperisager: These expectations are not valid according to ACT as they are not distinct; expectation 2 and 3 are contained within expectation 1.

annethyme: @nitedog, see Kasper's comment for reference as to why I change this back to one big exxpectation.

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
```diff
@@ -0,0 +1,276 @@
...
+authors:
+- Wilco Fiers
+- Anne Thyme Nørregaard
+---
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
```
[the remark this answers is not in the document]
kasperisager: #312

maryjom:
There's an overview of the [browser support of MathML](https://github.com/mathjax/MathJax/wiki/MathML-Support-In-Browsers).

_rules/SC4-1-2-aria-state-or-property-has-valid-value.md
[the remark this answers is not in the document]
annethyme: Will be handled by PR based on issue #312

PR #245, opened 2018-09-04
Rule: aria-required-owned-element

jkodu:
> Note: CI will fail until #255 is merged, as there is a glossary term `owned-by` which is referred in this rule.

New Rule: aria-required-owned-element
Co-authored by @audreymaniez 👍 🎆 

Closes issue: 
- https://github.com/auto-wcag/auto-wcag/issues/201

## Guidance the the PR (pull request) creator

**When creating PR:**
- [x] Make sure the pull request is prefixed with `Rule:` or `Algorithm:` or `Chore:` based on work involved.

**After creating PR:**

# How to Review And Approve


_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
+---
+name: ARIA required owned elements
+
+description: |
+	The rule checks existence of role(s) on required owned elements.
```
annethyme: Suggesting a rewording: The rule checks that for an explicit semantic role all required owned elements exists

ShadowBB:
I am unsure about this rule as a whole. Sure not having the correct required roles is against ARIA specs. But doing something against ARIA specs is not necessarily a WCAG violation is it? I would not fail any of the two failure examples during a WCAG inspection for instance.

_rules/SC1-3-1-aria-required-owned-element.md
Brynanders: "The rule checks that a role has at least one of its required owned elements."

My understanding was that an owned element need only list one of its required owned elements as its role

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+
+- [Required Owned Element](https://www.w3.org/TR/wai-aria-1.1/#mustContain)
+
+## Test Cases
+
+### Passed
+
+#### Pass example 1
+
+Owned `ul` element with role `list` has `span` element with role `listitem`.
```
Brynanders: Here I would say that the `span` role `lisitem` has a required owned element of `list` and `group`

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+	</span>
+	<span>
+		<!-- atleast one element is expected have a role, not mandatory on all children -->
+	</span>
+</ul>
+```
+
+#### Pass example 2
+
+Owned `ol` element with role `tablist` has `li` element with role `tab`.
```
Brynanders: Again, I would say the `li` is owned by the `ol` and not the other way round.

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+
+*There are currently no assumptions/ add any assumptions as necessary*
+
+## Accessibility Support
+
+*There are no major accessibility support issues known for this rule/ or add any support details as necessary*
+
+## Background
+
+- [Required Owned Element](https://www.w3.org/TR/wai-aria-1.1/#mustContain)
```
annethyme: Consider adding: 
* Implicit semantic roles for native HTML elements can be found in [HTML in ARIA](https://www.w3.org/TR/html-aria/)

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+
+Missing role on owned element children.
+
+```html
+<ul role='list'>
+	<span>
+	</span>
+</ul>
+```
+
```
annethyme: Would it make sense to add a test case for a role where the required owned elements use the "containing" notation? (I had to ask Kasper what that was)

E.g. role="grid" has "rowgroup → row" as required owned elements

This could be both passing and failing test case

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+## Background
+
+- [Required Owned Element](https://www.w3.org/TR/wai-aria-1.1/#mustContain)
+
+## Test Cases
+
+### Passed
+
+#### Pass example 1
```
annethyme: Should this be "Passed example 1"? (same for the rest)

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+<ul role='menu'>
+	<li></li> <!-- implicit role -->
+	<li role='menuitem'></li>
+	<li role='menuitemradio'></li>
+</ul>
+```
+
+### Failed
+
+#### Fail example 1
```
annethyme: Should this be "Failed example 1"? (same for the rest)

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+	</span>
+</ul>
+```
+
+#### Pass example 2
+
+Owned `ol` element with role `tablist` has `li` element with role `tab`.
+
+```html
+<ol role='tablist'>
```
annethyme: Maybe let ´ol´and ´li´ be ´div´ or ´span´ to make it clear that we are talking abou the explicit roles, not the native ones?

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+Multiple satisfying roles on owned element(s).
+
+```html
+<ul role='menu'>
+	<li></li> <!-- implicit role -->
+	<li role='menuitem'></li>
+	<li role='menuitemradio'></li>
+</ul>
+```
+
```
annethyme: Suggest adding a test case that highlights that the required owned element does not have to be a child of the role, but could be any descendant, e.g.:

```html
<div role='menu'>
	<div>
		<div>
			<div>Some content</div>
		</div>
	</div>
	<div>
		<div>
			<div role=´menuitem´></div>
		</div>
	</div>
</div>
```

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+### Applicability
+
+The rule applies to any element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) with a [non empty](#non-empty) `role` attribute whose value is applicable to required owned elements.
+
+**Note:** I
+
+### Expectation
+
+Each target element has atleast one owned element with a relevant `role` attribute.
```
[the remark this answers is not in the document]
jkodu: I do not see the need for a separate rule for catering to `aria-busy`. We could add a test case for script execution.

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+```html
+<ol role='tablist'>
+	<li role='tab'>
+	</li>
+</ol>
+```
+
+#### Pass example 3
+
+Owned `ul` element with role `list` has `li` element which has an implcit semantic role.
```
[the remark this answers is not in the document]
jkodu: I do not think this will fail @annethyme. I tested that an `li` item gets an implicitRole of `listitem` irrespective of what is it nested under & the case you have defined, where the div has an explicit role of `list` means, this should pass.

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,150 @@
...
+Owned `ol` element with role `tablist` has `li` element with wrong role `listitem`.
+
+```html
+<ol role='tablist'>
+	<li role='listitem'>
+	</li>
+</ol>
+```
+
+### Inapplicable
```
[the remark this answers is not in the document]
jkodu: Nice one. Question is, would the above pass?

_rules/SC1-3-1-aria-required-owned-element.md
jkodu: I have added a test case for above and a note to the accessibility support section.

WilcoFiers: @audreymaniez is correct. You have to explicitly set `role=presentation` on an element if you want to get not just the children, but any descendants to be considered owned elements. I haven't found any description of this in the ARIA specs, but if you look at the ARIA best practices you'll see this used consistently.

@ShadowBB The reason I believe this is required by WCAG is that without a correct parent / child relationship a number of things can go wrong:

- For combobox, you need the owned elements. Without a textbox and some sort of list to pull suggestions from, that element isn't a combobox. Those owned elements are required because without them it wouldn't be a combobox.
- Everything else is about computing group size, the size of the group can't be correctly expressed to AT if you can't programmatically determine where the group starts and where it ends. It's like instead of having a single list of 5 items, you'd get 5 lists with 1 item each. That kind of stuff is a violation of WCAG.

Can you give an example of when you think this isn't a WCAG 2 conformance issue? If anyone's interested, we could maybe split this rule up into one rule for combobox, and a second for group sizes? Personally I don't see too much value in that, but I suppose it identifies the issue a little more explicitly.

_rules/SC1-3-1-aria-required-owned-element.md
WilcoFiers: We should make sure that these containing notation things are covered and have tests for them.

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,227 @@
...
+
+#### Passed example 5
+
+Multiple types of required owned elements are present.
+
+```html
+<ul role='menu'>
+	<li></li> <!-- implicit role -->
+	<li role='menuitem'></li>
+	<li role='menuitemradio'></li>
```
audreymaniez: Do we have to write only valid examples ?
Here for instance, [`role='menuitemradio'` require the `aria-checked` property](https://www.w3.org/TR/wai-aria-1.1/#menuitemradio). I know that this test doesnt evaluate the presence of required attributes and, at this time "we don't care" about that property, but while writting snippet (and passed examples), don't we have to write valid code ? or it is assume somewhere (like a disclaimer), that the snippets we write just present the elements needed to accomplish the test, and no other ?

jkodu:
@ShadowBB 

With regards to your comment, I wanted to gather a larger perspective.

<blockquote>
I am unsure about this rule as a whole. Sure not having the correct required roles is against ARIA specs. But doing something against ARIA specs is not necessarily a WCAG violation is it? I would not fail any of the two failure examples during a WCAG inspection for instance.
</blockquote>

Personally I feel that we already have written a few ARIA rules like `aria-valid-attr` etc., which could be questioned the same way as above. But with regards to this rules, the SC necessitates role(s) on required owned elements, and hence I see it as a failure if they do not exist etc., 

@annethyme @Brynanders @WilcoFiers - WDYT?

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,227 @@
...
+
+For each test target at least one instance of one [required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [WAI-ARIA](https://www.w3.org/TR/wai-aria) is present as a [semantic role](#semantic-role) for an element [owned](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element) by the test target.
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+When a required owned element is not an immediate child, but rather a nested descendant, certain AT have some issues recognize child in that configuration which can be negated by defining a valid role for intermediate elements.
```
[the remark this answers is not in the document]
audreymaniez: Yes, adding the role="presentation" is like a hack and I'm not sure that role is required by the spec neither that It works with other AT (I just tested with NVDA but not JAWS nor VoiceOver). The ARIA spec does not give any clue about those type of configurations (like @WilcoFiers , I didn't find any information about that).
But maybe in Auto-WCAG we don't have to write that kind of informations about how to come over an AT dysfunction ? It goes beyond the scope, no ? Maybe the accessibility support issue just has to be : "When a required owned element is not an immediate child, but rather a nested descendant, certain AT have issues recognizing the owned element." 

And maybe just add collateral problem :  "If the role involves specific navigation keys, they may be not operational."

_rules/SC1-3-1-aria-required-owned-element.md
audreymaniez: Okay I open an issue#253  to make it a more general discussion. Is that ok ?

_rules/SC1-3-1-aria-required-owned-element.md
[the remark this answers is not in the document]
annethyme: @audreymaniez, I like that wording. I used it for SC1-3-1-aria-required-context-role #255

_rules/SC1-3-1-aria-required-owned-element.md
jkodu: have added tests for this, refer - Passed Example 7

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,227 @@
...
+		</div>
+	<div>
+	<div role="rowgroup">
+		<div role="row">
+			<span>Item 2</span>
+		</div>
+	</div>
+</div>
+```
+
```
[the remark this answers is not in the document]
jkodu: I believe you should add a failure test case in the PR - #255 like so:

#### Failed example 4

Element `div` with no role  contains a required owned element with implicit role of `listitem`, and thus expects a parent with role `list`.

```html
<div>
   <li>
   </li>
</div>
```

And I have added a passed, test case in this rule, refer - Passed example 8

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
...
+### Applicability
+
+The rule applies to any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an explicit [semantic role](#semantic-role) that has [required owned elements](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [WAI-ARIA](https://www.w3.org/TR/wai-aria).
+
+### Expectation
+
+For each test target at least one instance of one [required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [WAI-ARIA](https://www.w3.org/TR/wai-aria) is present as a [semantic role](#semantic-role) for an element [owned](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element) by the test target.
+
+**Note:**
+- When a widget is missing required owned elements due to script execution or loading, authors [MUST]((w3.org/TR/wai-aria-1.1/#mustContain)) mark a containing element with `aria-busy` equal to `true`.
```
nitedog: Is this a note, part of the applicability, or an assumption? I'm never sure where to put what. However, this seems to be quite a substantive limitation so that a note may be too little. If there is a normative MUST then this may even need to be (part of) another test to check appropriate use of WAI-ARIA?

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+When a required owned element is not an immediate child, but rather a nested descendant, certain AT have issues recognizing the owned element.
```
nitedog: Is it correct to say:
[[
This rule relies on assistive technologies that can recognize the owned element when it is a nested descendant, rather than an immediate child.
]]

Just an attempt at making it more specific and less open to interpretation of what "issues" are.

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an explicit [semantic role](#semantic-role) that has [required owned elements](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [WAI-ARIA](https://www.w3.org/TR/wai-aria).
+
+### Expectation
+
+For each test target at least one instance of one [required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [WAI-ARIA](https://www.w3.org/TR/wai-aria) is present as a [semantic role](#semantic-role) for an element [owned](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element) by the test target.
```
WilcoFiers: I've got the same concern here. IMO this thing is difficult to read. If we break the sentence up, things should be easier to read.

_rules/SC1-3-1-aria-required-owned-element.md
WilcoFiers: This doesn't have anything to do with support. This is how its supposed to work. The definition of "owned element" is pretty misleading. An element isn't considered to be part of a group unless its 1) a child, 2) referenced by aria-owns or 3) a descendent, with every element between it and the owner marked as role=presentation/none.

_rules/SC1-3-1-aria-required-owned-element.md
audreymaniez: The end of the sentence is just a repetition (but somehow upside down 🙃) of the beginning ? isn't it?

Proposal : 

> For each test target, at least one instance of one [required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [WAI-ARIA](https://www.w3.org/TR/wai-aria), is present.

That is a sufficient expectation ?

_rules/SC1-3-1-aria-required-owned-element.md
kasperisager: > I haven't found any description of this in the ARIA specs, but if you look at the ARIA best practices you'll see this used consistently.

@WilcoFiers I just had a look at the ARIA Authoring Practices and the one thing they do consistently mention is that when relations are involved, a given child "is either a DOM descendant of or owned by" its parent. The only place where e.g. `role=none` is used is when removing the default semantics of elements that are put between a parent and its owned descendant. All of this fits perfectly with the definition of "owned element" in ARIA:

> An 'owned element' is any DOM descendant of the element, any element specified as a child via aria-owns, or any DOM descendant of the owned child.

If anything, I see this as an accessibility support question as e.g. VoiceOver supports all of this just fine.

_rules/SC1-3-1-aria-required-owned-element.md
kasperisager: Per this https://github.com/auto-wcag/auto-wcag/pull/245#discussion_r222282853, I simply don't think that that is true; it has everything to do with support as it _is_ how it's supposed to work, both according to ARIA itself and the associated authoring practices.

jkodu: Something to discuss during the rule writing session. I will add to agenda.

nitedog: Or maybe on Thursday so that we can complete it earlier?

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
+---
+name: ARIA required owned elements
+
+description: |
+	The rule checks that a role has at least one of its required owned elements.
```
annethyme: --> "This rule checks that each role has at least one of its required owned elements."

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
...
+authors:
+- Audrey Maniez
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+This rule applies to any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) with a role that requires it to own some element(s). This is only applicable to elements that have an explicit [semantic role](#semantic-role). Which roles have required owned elements is defined in [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/).
```
annethyme: Editorial suggestion: 

"This rule applies to any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an explicit [semantic role](#semantic-role) that has [WAI-ARIA required owned elements](https://www.w3.org/TR/wai-aria/#mustContain)."

_rules/SC1-3-1-aria-required-owned-element.md
annethyme: I think we actually discussed if this should be a separate (semiautomated) rule under a composite rule along with this...

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Expectation
+
+For each test target, at least one instance of one [required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) listed for that role in [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/), is present.
```
annethyme: A Shadi-style editorial suggestion: 

"For each test target, at least one instance of one [WAI-ARIA required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) is present.

_rules/SC1-3-1-aria-required-owned-element.md
annethyme: Newest Accessibility Support content from sibling rule https://github.com/auto-wcag/auto-wcag/pull/255: 
"This rule relies on assistive technologies to recognize owned elements, as defined by [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria). This includes when they are nested descendants that are not immediate children. However, some assistive technologies do not recognize owned elements that are not immediate children, unless workarounds are used."

This discussion point is equally relevant for both these rules.

_rules/SC1-3-1-aria-required-owned-element.md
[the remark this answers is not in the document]
annethyme: ... but luckily #255 is limited to explicit semantic roles, like your rule, so this would be non-applicable :)

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,227 @@
...
+name: ARIA required owned elements
+
+description: |
+	The rule checks that a role has at least one of its required owned elements.
+
+success_criterion:
+- 1.3.1
+
+test_aspects:
+- DOM Tree
```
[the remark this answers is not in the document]
jkodu: Sorted

_rules/SC1-3-1-aria-required-owned-element.md
jkodu: @nitedog - In this case the note is a part of the expectation, to me it sits right with expectation, as we are saying, it is expected to be decorated with `aria-busy` under some conditions....

@annethyme - We did discuss this, let us discuss again in the rule writing session, to see if a new semi-automated rule is necessary for this. 

Otherwise, do not see the above comment as a blocker to progress with this rule.

_rules/SC1-3-1-aria-required-owned-element.md
jkodu: Updated to above comment, and have added to agenda for wider discussion.

_rules/SC1-3-1-aria-required-owned-element.md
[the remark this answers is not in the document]
jkodu: Element `div` with `list` role  contains a required owned element with implicit role of `listitem`.

```html
<div role='list'>
   <li>
   </li>
</div>
```

Is a pass case, and not inapplicable.

_rules/SC1-3-1-aria-required-owned-element.md
```diff
@@ -0,0 +1,243 @@
...
+authors:
+- Audrey Maniez
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+This rule applies to any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an explicit [semantic role](#semantic-role) that has [WAI-ARIA required owned elements](https://www.w3.org/TR/wai-aria/#mustContain).
```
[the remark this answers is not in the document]
jkodu: Updated.

_rules/SC1-3-1-aria-required-owned-element.md
jkodu: I believe these comments are on a different tangent as to what the rule addresses. So will consider this resolved.

_rules/SC1-3-1-aria-required-owned-element.md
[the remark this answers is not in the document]
jkodu: @annethyme thoughts?

_rules/SC1-3-1-aria-required-owned-element.md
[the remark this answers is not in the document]
jkodu: @annethyme - thoughts?

_rules/SC1-3-1-aria-required-owned-element.md
WilcoFiers: @jkodu I don't agree that this is resolved, or that it's not relevant. The following test case is completely broken and yet it seems to pass this rule:

```html
<div role=Listbox>
  <div role=alert>
    <div role=listitem>Item 1</div>
    <div role=listitem>Item 2</div>
  </div>
</div>
```

I'm in the process of reverse engineering how browsers actually work out owned items, and if it even matters. It's a bit more complicated than it may seem on the surface.

annethyme:
@WilcoFiers and @jkodu, did you guys skip this for our Year 1 delivery? Because then it would be nice to know...
Plus we have another rule that is very closely connected to this one: https://github.com/auto-wcag/auto-wcag/pull/255

jkodu:
Closing for now, as it is on hold

annethyme:
Opening this one again, since it's on our to do list

_rules/SC1-3-1-aria-required-owned-element.md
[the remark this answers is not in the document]
audreymaniez: An innaplicale example 3 has been added. Thank you @annethyme

_rules/SC1-3-1-aria-required-owned-element.md
audreymaniez: @WilcoFiers There's a ["owned by" definition](https://github.com/auto-wcag/auto-wcag/pull/255/files#diff-e8efc5e044aa522fbb04b4f072890fa3) in writing for the #255. Is the definition enough to clarify that kind of example you give is a fail one ? 

Actual expectation:
> For each test target, at least one instance of one [WAI-ARIA required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain) is present.

Could be rephrase to add the owned by definition link:

> The target element [owned](#owned-by) at least one instance of one of its [WAI-ARIA required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain).

or do you think are more details are needed in the expectation section ? 

And we should add a fail example that demonstrate that (not sure of the description of the example though): 

> #### Failed Example x
> 
> Element with role `list` is not the closest ancestor of element with role `listitem`, there is a role `alert` in between that  break the relationship. The role `list` is considered having no WAI-ARIA required owned element.
> 
````html
<div role="list">
  <div role="alert">
    <div role="listitem">Item 1</div>
    <div role="listitem">Item 2</div>
  </div>
</div>

File: _rules/SC1-3-1-aria-required-owned-element.md [hunk 24]
```diff
@@ -0,0 +1,282 @@
+---
+name: ARIA required owned elements
```
annethyme: Remember to add id before merging

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+## Expectation
+
+The target element [owns](#owned-by) at least one instance of one of its [WAI-ARIA required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain).
+
+**Note:**
+- When a widget is missing required owned elements due to script execution or loading, authors [MUST]((w3.org/TR/wai-aria-1.1/#mustContain)) mark a containing element with `aria-busy` equal to `true`.
+
+## Assumptions
+
+*There are currently no assumptions*
```
ShadowBB: Normally I would ask to provide the assumption "Aria is being used to comply to WCAG" but we should now probably use a less generic text, (See https://github.com/act-rules/act-rules.github.io/issues/549).

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+- DOM Tree
+- CSS Styling
+authors:
+- Audrey Maniez
+- Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an explicit [semantic role](#semantic-role) that has [WAI-ARIA required owned elements](https://www.w3.org/TR/wai-aria/#mustContain), except when the owned element has an implicit semantic role that is identical to its explicit semantic role.
```
Jym77: I do not really understand:
* why we only want to check element with explicit roles, not those with implicit roles.
* why there is the exception at the end of the applicability clause.

(that's more curiosity than real comment)

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Expectation
+
+The target element [owns](#owned-by) at least one instance of one of its [WAI-ARIA required owned element](https://www.w3.org/TR/wai-aria-1.1/#mustContain).
+
+**Note:**
+- When a widget is missing required owned elements due to script execution or loading, authors [MUST]((w3.org/TR/wai-aria-1.1/#mustContain)) mark a containing element with `aria-busy` equal to `true`.
```
Jym77: Shouldn't that note be an actual expectation? eg, "the target element either owns one of its `mustContain` or has `aria-busy=true`".

The way it’s written now, it seems that `<ul role="list" aria-busy=true></ul>` is failing the rule (no list item) while this is perfectly OK thanks to the `aria-busy` attribute.

Maybe we can also add an example with `aria-busy` to illustrate this.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+```html
+<div role='list'>
+   <li>
+   </li>
+</div>
+```
+
+#### Passed example 8
+
+Element with role `menubar` can have multiple required owned element (`menuitem` `menuitemcheckbox` and `menuitmeradio`) but only one is necessary.
```
Jym77: I think it would be a bit better if the example contained 2 of the three possible elements but not all of them.

The reason is that an incorrect tool that would test for the presence of *all* of them would incorrectly fail on a test case with only some of them and thus we could detect such a problem.
It also makes it a bit easier (I think) for humans to understand "we don’t need all of them" if we show them that it's fine to have only some.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+	 <div role='menuitmeradio'>
+   </div>
+</div>
+```
+
+#### Passed example 9
+
+Element with role `list` contains a required `listitem`, the relationship between both elements is made possible through the `aria-owns` properties.
+
+**Note:** This test case follows the definition of [owned by](#owned-by) used in this rule. If implemented differently, this definition could cause differences in outcome for this test case.
```
Jym77: This is the definition of "owned by" made by WAI ARIA, not just "used in this rule" (ie, it is not something we come up ourselves, it is the actual definition and if a browser/AT does not respect this definition, it's the browser/AT fault for not respecting WAI ARIA, not our fault for making weird definition).

I think this should rather link to the definition of "owned by" in WAI ARIA" (same as "Owned element" in the background Section).

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+The element with the semantic role of `list` owns the element with the role `tab` through the `aria-owns` property.
+
+```html
+<div role="list" aria-owns="id2">
+</div>
+
+<div id="id2" role="tab"></div>
+```
+
+**Note:** This test case follows the definition of [owned by](#owned-by) used in this rule. If implemented differently, this definition could cause differences in outcome for this test case.
```
Jym77: Same comments as for the note of Passed example 9.

Jym77: Additionally, the note on Passed example 9 is before the code, and here it's after the code. For the sake of consistency, I suggest we try and keep notes in the same place.

_rules/aria-required-owned-element-bc4a75.md
Jym77: The note does not apply only to widgets. The link you put says " For example, until a page is fully initialized and complete, an author could mark the document element as busy." I think the document is not a widget.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,283 @@
...
+```html
+<ol role='tablist'>
+	<li role='listitem'>
+	</li>
+</ol>
+```
+
+#### Failed example 3
+
+Element with role `grid` is missing a required owned element `rowgroup`, even though the containing required owned element `row` is present.
```
carlosapaduarte: The required owned elements for `grid` are `row` and `rowgroup`. Since the expectation requires just one instance of the required owned elements, this is a pass.

_rules/aria-required-owned-element-bc4a75.md
jkodu: Thanks for the above 2 comments. I was trying to line up everything similar to #255 and that caused some ambiguity. I am updating this to simple language.

_rules/aria-required-owned-element-bc4a75.md
jkodu: Because myself and @annethyme started authoring this rule (owned elements) and the [context rule](https://github.com/act-rules/act-rules.github.io/pull/255), after a long discussion of what definition to use for owned elements which happened here - https://github.com/act-rules/act-rules.github.io/issues/302, a decision was made to author our own. 

The definition is authored here - https://github.com/act-rules/act-rules.github.io/pull/255/files#diff-e8efc5e044aa522fbb04b4f072890fa3

_rules/aria-required-owned-element-bc4a75.md
jkodu: Nice catch, I have updated this to something that will fail.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+#### Passed example 4
+
+Multiple types of required owned elements are present.
+
+```html
+<ul role="menu">
+	<li></li>
+	<!-- implicit role -->
```
carlosapaduarte: Should be in the line above

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,248 @@
...
+
+```html
+<div role="list">
+	<li></li>
+</div>
+```
+
+#### Passed example 8
+
+Element with role `menubar` can own multiple required owned element with role `menuitem`, `menuitemcheckbox` and `menuitmeradio`, but only one is necessary.
```
carlosapaduarte: This description is "strange", since the example shows two "owned roles". With this description, I was expecting just one.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,248 @@
...
+## Background
+
+- [Required Owned Element](https://www.w3.org/TR/wai-aria-1.1/#mustContain)
+- [Owned Element](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element)
+- [HTML in ARIA](https://www.w3.org/TR/html-aria/)
+- [Implicit WAI-ARIA Semantics](https://www.w3.org/TR/wai-aria-1.1/#implicit_semantics)
+
+## Test Cases
+
+### Passed
```
carlosapaduarte: Some suggestions for further pass examples
- One passed example with aria-busy
- One passed example where the top-level element in the example does not have an explicit role (since it is no longer an applicability requirement)
- One passed example where there are elements with roles that meet the owned by requirements, but also other elements that do not

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,248 @@
...
+Element with role `list` contains a required `listitem`, the relationship between both elements is made possible through the `aria-owns` properties.
+
+**Note:** This test case follows the definition of [owned by](#owned-by). If implemented differently, this definition could cause differences in outcome of this test case.
+
+```html
+<div role="list" aria-owns="id1"></div>
+<div id="id1" role="listitem"></div>
+```
+
+### Failed
```
carlosapaduarte: One suggestion for another fail example
- One fail example where the top-level element in the example does not have an explicit role (since it is no longer an applicability requirement)

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,248 @@
...
+
+```html
+<ul role="">
+	<li></li>
+</ul>
+```
+
+#### Inapplicable example 3
+
+Element has no explicit semantic role.
```
carlosapaduarte: This is no longer inapplicable

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,248 @@
...
+The element with the semantic role of `list` owns the element with the role `tab` through the `aria-owns` property.
+
+**Note:** This test case follows the definition of [owned by](#owned-by). If implemented differently, this definition could cause differences in outcome of this test case.
+
+```html
+<div role="list" aria-owns="id2"></div>
+<div id="id2" role="tab"></div>
+```
+
+### Inapplicable
```
carlosapaduarte: One suggestion for an inapplicable example
- one element with a role that does not have required owned roles (e.g. region, progressbar)

_rules/aria-required-owned-element-bc4a75.md
jkodu: Added examples for points 1 & 2, as for 3 believe passed example 1 and 4 convey this already.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,274 @@
...
+			<span>Item 1</span>
+		</div>
+		<div></div>
+	</div>
+</div>
+```
+
+#### Failed example 4
+
+The element with the semantic role of `list` owns the element with the role `tab` through the `aria-owns` property.
```
carlosapaduarte: To make it clearer why the example fails, you can add something like 
", but does not own the required elements with semantic roles `group` or `listitem`" 
to the end of the sentence

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,269 @@
...
+
+**Note:** The combobox role is excluded from this rule, because the design pattern for it as described in ARIA 1.1 has proven problematic. The combobox will be significantly different for ARIA 1.2, where it does not have [required owned elements][].
+
+## Expectation
+
+Each test target only [owns][] elements with a [semantic role][] from the [required owned element][] list for the test target's [semantic role]().
+
+**Note:** If the [required owned element][] list has two roles in one item separated by a "→", elements with the first of these roles, is only a [required owned element][] if all its owned elements have the second of these roles as their [semantic role][]. For example a `list` has `group → listitem`, so an element with the [semantic role][] of `group` only counts as a [required owned element][] for `list`, if all elements [owned by][] the `group` have the [semantic role][] of `listitem`.
```
Jym77: I understand what this means, but I had to dig into the specs to get it, and I don't think that this sentence currently makes sense…

```suggestion
**Note:** Some [required owned elements][] are only valid if they themselves [own][owns] (or "contain") elements with a given [semantic role][]. This is denoted by an arrow (meaning "containing") in the role description. For example, the role `list` has `group → listitem` as one of its [required owned elements][], meaning that elements with a role of `list` may only [owns][] elements with a role of `group` who themselves only [own][] elements with a role of `listitem`.
```

(I think it's important to use the word "contain" to denote the arrow, given that it is how ARIA calls it)

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,269 @@
...
+
+This element with the `grid` role only owns elements with the `row` role, and the element with the `row` role only owns elements with the `cell` role. The `row` role is one of the [required owned elements][] for `grid`, and `cell` is one of the [required owned elements][] for `row`.
+
+```html
+<table role="grid">
+	<tr role="row">
+		<span role="cell">Item 1</span>
+	</tr>
+</table>
+```
```
Jym77: I think that Passed Examples 1, 2, and 3 are very similar (simple good stuff) with 3 being a bit different (nested roles with required ownership).
I suggest we remove either Passed Example 1 or Passed Example 2.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Passed Example 4
+
+This element with the `menu` role only owns elements with the `menuitem`, `menuitemradio` and `menuitemcheckbox` role. These roles are one of the [required owned elements][] for `menu`. The element with the `none` role is not [owned by][] the `menu` because it is not [included in the accessibility tree][].
```
Jym77:
```suggestion
This element with the `menu` role only owns elements with the `menuitem`, `menuitemradio` and `menuitemcheckbox` role. These roles are all [required owned elements][] for `menu`. The element with the `none` role is not [owned by][] the `menu` because it is not [included in the accessibility tree][].
```

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,269 @@
...
+**Note:** This test case follows the definition of [owned by][]. If implemented differently, this definition could cause differences in outcome of this test case.
+
+```html
+<div role="list" aria-owns="id1"></div>
+<div id="id1" role="listitem">Item 1</div>
+```
+
+#### Passed Example 7
+
+This element with the `list` role only owns elements with the `listitem` role, or elements with the `group` role, in which each element has the `listitem` role. The `listitem` role on its own, as well as in a `group` are in the [required owned elements][] for `list`.
```
Jym77:
```suggestion
This element with the `list` role only owns elements with the `listitem` role, or elements with the `group` role, in which each element has the `listitem` role. Both the `listitem` role on its own, and the `group` role (when containing elements with the `listitem` role) are [required owned elements][] for `list`.
```

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,269 @@
...
+		<span role="listitem">Item 3</span>
+	</div>
+</div>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This element with the `list` role owns elements without the `listitem` role. The `listitem` role is one of the [required owned elements][] for `list`.
```
Jym77:
```suggestion
This element with the `list` role does not owns an element without any of its [required owned elements][] (`listitem` or `group` containing `listitem`).
```
Because 
```html
<div role="list">
  <div role="group">
    <li>Foo</li>
  </div>
</div>
```
is matching the description (`list` owns elements without `listitem` role) but passes the rule.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -0,0 +1,269 @@
...
+<div role="list">
+	<span role="listitem">Item 1</span>
+	<div role="group">
+		<span role="listitem">Item 2</span>
+		<span role="listitem">Item 3</span>
+	</div>
+</div>
+```
+
+### Failed
```
Jym77: I think we should add an example with an element owning one good and one bad role, to illustrate the "only" part of the expectation:
```html
<div role="list">
  <li>Item 1</li>
  <span role="link">Item 2</span>
</div>
```

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -153,7 +143,7 @@ This element with the `list` role only owns elements with the `listitem` role, o
 
 #### Failed Example 1
 
-This element with the `list` role owns elements without the `listitem` role. The `listitem` role is one of the [required owned elements][] for `list`.
+This element with the `list` role does not owns an element without any of its [required owned elements][] (`listitem` or `group` containing `listitem`).
```
Jym77: with
```suggestion
This element with the `list` role owns an element without any of its [required owned elements][] (`listitem` or `group` containing `listitem`).
```

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -255,6 +256,15 @@ This element with the `menu` role has attribute an `aria-busy` attribute set to
 </ul>
 ```
 
+#### Inapplicable Example 6
+
+This element with the `combobox` role is accessible without owned elements.
```
Jym77: Does "is accessible" means "is included in the accessibility tree"?

Issue #215, opened 2018-08-15
Valid use of duplicate accessible names


Brynanders:
---
name: Valid use of duplicate accessible names
description: |
	This rule checks that there is a valid reason for having a set of images with identical accessible names.

success_criterion:
- 1.1.1 # Non-text Content

test_aspects
- DOM Tree
- CSS Styling

authors:
- Bryn Anderson
---

### Applicability

The rule applies to any set of HTML or SVG element with the semantic role of `img`, in a document, exposed to assistive technologies, that have an identical [accessible name]().

### Expectation

Each test target has an identical purpose.

**Note**: Leading and trailing whitespace and difference in case sensitivity should be ignored.

## Assumptions

*There are currently no assumptions*

## Accessibility support

There is a known combination of a popular browser and assistive technology that does not by default support `title` as an accessible name.

### Pass example 1

Image A and Image B are different files but have an identical purpose stated in the alt attribute and serve the same purpose linking to the same destination.

```html
<a href="https://twitter.com/brynanders"> <img src="/twitter_largeicon.png" alt="My twitter feed"> </a>
<a href="https://twitter.com/brynanders"> <img src="/twitter_smallicon.png" alt="My twitter feed"> </a>
```

### Pass example 2

Div A and B serve the same purpose but use the different image files to display a different version of the same informative image. Both files have an identical accessible name stated in the aria-label attribute.

```html
<div id="A" role="img" aria-label="blue sky" style="background-image:url('background1.jpg')"> </div>
<div id="B" role="img" aria-label="blue sky" style="background-image:url('background2.jpg')"> </div>
```

### Failure example 1

Image A and Image B are different files, have identical alt attributes but go to different destinations. 

```html
<a href="https://twitter.com/brynanders"> <img src="/twitter_largeicon.png" alt="Twitter"> </a>
<a href="https://twitter.com"> <img src="/twitter_smallicon.png" alt="twitter"> </a>
```

### Inapplicable example 1

Marked decorative with alt attribute
```html
<img src="/background.png" alt="">
```

### Inapplicable example 2

Marked decorative with role="presentation"
```html
<img src="/background.png" role="presentation">
```

Brynanders:
https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

Brynanders:
#251

Issue #216, opened 2018-08-15
New Rule: Valid use of image file name in accessible name


Brynanders:
---
name: accessible name is image file name
description: |
	This rule checks that the accessible name of an image is not its file name.

success_criterion:
- 1.1.1 # Non-text Content

test_aspects
- DOM Tree
- CSS Styling

authors:
- Bryn Anderson
---

### Applicability

The rule applies to any HTML `img` element that is exposed to assistive technologies, where the element has a non-empty `alt` attribute that is equal to the accessible name of the element, and where the `alt` attribute of the element references the file name and extension of the `src` attribute of the element.

### Expectation

Each test target has an `alt` attribute that serves as an appropriate text alternative as described in [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/WCAG20-TECHS/F30.html).

## Assumptions

*There are currently no assumptions*

## Accessibility support

 *There are no major accessibility support issues known for this rule.*

## Background

- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F30)

### Pass example 1

Image depicting a PDF file format, uses the .pdf extension in the accessible name.

```html
<img scr="/example.pdf" alt="an example of a .pdf file">
```

### Failure example 1

The accessible name is the image file name and does not serve as an appropriate text alternative.

```html
<img src="/image.jpg" alt="image.jpg file">
```

### Inapplicable example 1

No reference to a file extension or file path in the accessible name.

```html
<img src="/example.pdf" alt="an example of a PDF file">
```

Brynanders:
Wondering if this rule should be confined to accessible names for images that reference their own image files... or if it should look for any file reference like I have done so far?

kasperisager:
I'm wondering if this note can be left out:

> Leading and trailing whitespace and difference in case sensitivity should be ignored.

Given that some file systems are case sensitive and that file system paths for sure are whitespace sensitive, it seems to be best to compare the `src` and `alt` attributes verbatim.

Brynanders:
Created PR #263

PR #251, opened 2018-09-11
SC1-1-1-placeholder-text-is-valid-accessible-name

Brynanders:
Closes issue: (#215 )

# How to Review And Approve


_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
```diff
@@ -0,0 +1,78 @@
...
+- DOM Tree
+- CSS Styling
+
+authors:
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to any set of HTML or SVG element with the semantic role of `img`, in a document, exposed to assistive technologies, that have an identical [accessible name]().
```
kasperisager: The "in a document" bit seems unnecessary.

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Expectation
+
+Each test target has an identical purpose.
+
+**Note**: Leading and trailing whitespace and difference in case sensitivity should be ignored.
```
kasperisager: Should this note be moved to "Applicability"?

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility support
+
+There is a known combination of a popular browser and assistive technology that does not by default support `title` as an accessible name.
```
kasperisager: This is repeated in at least a few rules. Can we at some point make this known combination of browsers and assistive technologies _known_?

Brynanders: I would be all for that but I was under the impression that we can't or didn't want to call out vendors. @annethyme do you know what the stance is on that?

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
Brynanders: Its a standard accessibility support note that we have used on several rules relating specifically to AT support for the `tittle` attribute. I'm not sure what the point of the Accessibility Support note would be if not for this.

Brynanders:
> I would not consider this case, **duplicate accessible name**, as _a particular test_ but instead, a _particular test case_ of a test that checks the relevance of the accessible name (but that is not yet written in Auto-WCAG, I can't find one but maybe I missed it).

@audreymaniez this rule actually started out as part of a composite rule to address [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/WAI/WCAG21/Techniques/failures/F30). AT some point the decision was taken to create seperate atomic rules instead.

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
WilcoFiers: Correct. The problem is that it only takes one or two tests to figure out something doesn't have wide support. But if you want to report that, it's not fair to the two or three that you tested to only list them as not supporting it, as opposed to all others that probably don't either, but we didn't test that.

We don't have the resources to do comprehensive testing of AT in Auto-WCAG. There are a few promising projects happening that, if they combined resources might be able to do it. I'd love to link to something like that in the future.

WilcoFiers:
@Brynanders I don't think that answers @audreymaniez's concern. It fits in with a larger theme that I don't quite know the answer to: What do we do with "likely a problem" type issues? Should automated tools call them out as issues, knowing they might be wrong? Should they be called out for review, knowing that those aren't the only things that need to be reviewed? Or should automated accessibility tools just ignore these things?

In WCAG 2.0, there is only one place where "probably an accessibility problem" leads to non-conformance. It's in 4.1.1 Parsing. Duplicate IDs **can** cause problems in AT, but they don't always. Despite that we are asked to always treat that as non-conformance. There is nothing else like it in WCAG 2. We're doing a similar thing in Auto-WCAG with the ARIA rules, where we say: If you use invalid ARIA, it's an issue, even though AT won't break if you do that. It's just very likely a content author meant to do something, but they didn't get it right.

We're looking at doing a similar thing with "link texts must be unique in their context" and "form fields must have a unique name in their context". Something like that may actually make a lot more sense for images as well. We'd have to figure out what "context" means for images. Doing an "images have a unique name in its context" is certainly consistent with other things we're looking at. I'm far more comfortable with such a rule for form fields then for links or images though. 

Lets put this on the agenda for the next call, maybe we can come to some sort of solution.

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+### Failure example 1
+
+Image A and Image B are different files, have identical alt attributes but go to different destinations. 
+
+```html
+<a href="https://twitter.com/brynanders"> <img src="/twitter_largeicon.png" alt="Twitter"> </a>
+<a href="https://twitter.com"> <img src="/twitter_smallicon.png" alt="twitter"> </a>
+```
```
audreymaniez: I understand the example, but it mixes two elements : link and image.

If we want to keep that example, I think it would be necessary to add in the description, that the links both have the *same context*, so we can consider the two identical. 

In fact, in that example,  I think we don't evaluate the accessible name of the *image*, but of the *link*, and the accessible name of a link can be evaluate using the context (level AA [2.4.4 Link Purpose (In Context)](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#navigation-mechanisms-refs) : see for example the [H78 technique (using enclosing paragraph)](https://www.w3.org/TR/WCAG-TECHS/H78.html), or [H79 (using table context)](https://www.w3.org/TR/WCAG-TECHS/H79.html) etc...  So, it's not because two links has the same accessible name (made of the alt attribute of the image in our eample) that they are the same, because of the context they may be part of.

audreymaniez:
> this rule actually started out as part of a composite rule to address F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text). AT some point the decision was taken to create seperate atomic rules instead.

Oh, is there a public document or web page that reference that kind of informations (that the rule is going to be part of a composed rule, and which one) ?

@Brynanders  Would it be relevant to add the F30 reference in the **Background** section ? Some rules has the "background" section (like that one : [Rule: HTML Page has a title](https://auto-wcag.github.io/auto-wcag/rules/SC2-4-2-page-has-title.html)), and it's useful (as a reviewer at least) to quickly get the context.

>  Should automated tools call them out as issues, knowing they might be wrong? Should they be called out for review, knowing that those aren't the only things that need to be reviewed?  Or should automated accessibility tools just ignore these things?

@WilcoFiers that was exactly my purpose. Thank you for clarifying it!

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
kasperisager: Since it seems to relate to leading and trailing whitespace and case differences in the accessible names of applicable elements, it doesn't really make sense outside the "Applicability" section.

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
kasperisager: Makes sense 👍 It'd be fantastic to have something similar to https://github.com/mdn/browser-compat-data for information about AT compatibility.

WilcoFiers:
Thinking about it some more, I don't think unique image names in context makes sense. I think what I'd rather us do is rename this. It's kind of hiding its true intent: To catch  "default" alt attributes. Duplicate alt attributes aren't a problem. What is a problem, is if you're leaving the default value of an editor in the image alt. E.g. alt="image description".

Instead of suggesting this rule is looking for inappropriate duplicates, lets be explicit that this is testing for default values from authoring tools. I'd even feel comfortable to just look at `img` elements and `alt` attributes. I don't think this happens with `role=img` elements, or with `aria-label` type stuff.

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
Brynanders: @audreymaniez interesting. Are you suggesting that [functional](https://www.w3.org/WAI/tutorials/images/functional/) images should be non-applicable because they would be covered by #220 ?

That would leave [informative](https://www.w3.org/WAI/tutorials/images/informative/), [complex](https://www.w3.org/WAI/tutorials/images/complex/) and groups of [images](https://www.w3.org/WAI/tutorials/images/groups/) applicable.

audreymaniez: > That would leave informative, complex and groups of images applicable.

Well, in fact, I think that any kind of images you list (informative, complex...) become innaplicable if the image is included in a link. At the moment, an image is included in a link (complex or not), its alt attribute is not used anymore to describe the image but to give the link a name. And even more : an image that can be evaluated as a decorative one, must have an alt text if its included in a link.

 So yes, I think that images, both within a link, with the same alt attribute and the same context, but that convey to different destination is a failed example for the test about identical links #220.

_rules/SC1-1-1-valid-duplicate-accessible-name-for-image.md
annethyme: I have moved the note on comparing accessible names into the definition for Accessible name: https://github.com/auto-wcag/auto-wcag/pull/254
We were using the note in at least 3 rules, so for consistency it should live in one place.

Brynanders:
"placeholder text" is a already defined as text used to pre-populate form input fields or labels so I will need to redefine this... For now I have created a definition for "generic placeholder text" https://github.com/auto-wcag/auto-wcag/issues/262

Brynanders:
Looks like the University of Toronto conducted some tests around this back in 2011 https://www.w3.org/WAI/GL/WCAG20/tests/test6.html

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
+---
+name: Image accessibile name is placeholder text
+description: |
+	This rule checks for images using generic placeholder text.
```
kasperisager: "using" -> "that use".

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+- DOM Tree
+- CSS Styling
+
+authors:
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to any HTML `img` elements, exposed to assistive technologies, with a non-empty `alt` attribute that is equal to the accessible name.
```
kasperisager: I used this for #216:

> The rule applies to any HTML `img` element that is exposed to assistive technologies, where the element has a non-empty `alt` attribute that is equal to the accessible name of the element.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Expectation
+
+Each test target `alt` attribute does not use [placeholder text](#placeholder-text) to describe the purpose of the image.
```
kasperisager: > Each test target has an `alt` attribute that does not use [placeholder text](#placeholder-text) to describe the purpose of the image.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+## Background
+
+[Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/WCAG20-TECHS/F30.html)
+
+## Test Cases
+
+### Pass example 1
+
+alt text acurately describing the the image of a Twitter icon.
+
```
audreymaniez: Just a typo ("the the") and presentation tag (for the alt attribute)  :

`alt` text acurately describing the image of a Twitter icon.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+- CSS Styling
+
+authors:
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to any HTML `img` element that is exposed to assistive technologies, where the element has a non-empty `alt` attribute that is equal to the accessible name of the element.
+
```
audreymaniez: I don't find any trace in the previous conversations, but does this rule shouldn't also apply to ìnput type"image"`  and `area` ?

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
+---
+name: Image accessibile name is placeholder text
```
audreymaniez: A little typo :-) : "accessibile" should be "accessible"

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+description: |
+	This rule checks for images that use generic placeholder text.
```
audreymaniez: Hum, I'm a bit confused :-) 

For example, for a rule that is based on the same failure (F30 #263), the description ask *what the alt should not be* :  "This rule checks that the accessible name of an image is not its file name".

In that rule (placeholder text), we are looking for an equivalent outcome: what the alt should not be. 

Proposal:

> This rule checks that the accessible name of an image is not a generic placeholder text

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
nitedog: Also, why is this limited to 'alt' attribute only? An authoring tool could also produce placeholder 'aria-label' names.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to any HTML `img` element that is exposed to assistive technologies, where the element has a non-empty `alt` attribute that is equal to the accessible name of the element.
+
+### Expectation
+
+Each test target has an `alt` attribute that does not use [placeholder text](#placeholder-text) to describe the purpose of the image.
```
nitedog: It should be "the accessible name is not placeholder text" (not only the alt attribute)

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Assumptions
+
+This rule assumes that an authoring tool has been used to publish image content and that image `alt` attributes of published content can be populated with generic placeholder text when not otherwise specified.
```
nitedog: Why is this important? It could also be a template or an IDE that creates placeholder text. Technically you could argue that these are types of authoring tools but I don't think we need any assumptions here. We refer to the failure for further background anyway.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
nitedog: "exposed to assistive technologies", "non-empty", and "accessible name" need to be linked to their definitions.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+[Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/WCAG20-TECHS/F30.html)
+
+## Test Cases
+
+### Pass example 1
+
+alt text acurately describing the image of a Twitter icon.
+
+```html
+<img src="/twitter_largeicon.png" alt="twitter">
```
nitedog: Are there any branding or copyright concerns with using the Twitter icon (other than for use of Twitter)? How about using content from the hypothetical [Before and After Demo (BAD)](https://www.w3.org/WAI/demos/bad/after/home.html) website instead?

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,69 @@
...
+Marked decorative with alt attribute
+
+```html
+<img src="/background.png" alt="">
+```
+
+### Inapplicable example 2
+
+Marked decorative with role="presentation"
+
```
audreymaniez: Just a typo (code tag) : 
> Marked decorative with `role="presentation"`

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
audreymaniez: It's still about `input type="image"` and `area`.
You rewrite : "with the semantic role of `img`", but  : 
- `input type="image"` has a role `button` so it won't be tested.
- and `area` has a role `link`, so it won't be tested  (https://www.w3.org/TR/html/semantics-embedded-content.html#the-area-element)

But for those two elements, accessible name defined in the alt text, so both can fail (or pass) that test.

I don't really know how to manage this, but I think either we have to create two other tests to handle those two html element (input and area), or we have to rewritte the applicability, but now (and I'm sorry!) I can't find a proposal to make.

kasperisager: @audreymaniez I do think separate rules would make more sense for those cases.

audreymaniez: @kasperisager I don't think either :-)

Maybe it has already been discussed but I can'f find, but can we consider that issue and #263  sort of the same rule ? I mean, maybe we can focus only on "one" and adapt then to the other ? Both issue has relevant reviews for both issues ??

Brynanders:
> input type="image" has a role button so it won't be tested.
and area has a role link, so it won't be tested (https://www.w3.org/TR/html/semantics-embedded-content.html#the-area-element)
But for those two elements, accessible name defined in the alt text, so both can fail (or pass) that test.

@audreymaniez  I'm suggesting that these be covered by other rules.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
Brynanders: @audreymaniez #263 is definitely similar the difference at the moment is 
that this "placeholder text" rule uses the definition of #262 (generic placeholder text) that provides a baseline example for implimenters to follow - in other words placeholder text could be different in French or English so the exact match is an implementation detail.
SC1-1-1-accessible-name-is-image-file-name #263 simply looks for two attributes that match.

Brynanders: @audreymaniez In regards to `input type="image"` and `area` can you confirm that you do or do not think they should be applicable in this rule?

I do not think they should be applicable because they do not have the semantic role of `img` and this rule targets accessible names of image elements and not the accessible names of `link` or a `button` elements.

nitedog: I tend to agree with expanding the scope beyond 'img' alone, to include 'input type="image"' and 'area'. I think the rest of the rule works well with that change, and we do not need to have additional rules that do the same thing.

Brynanders: Kasper, Audrey and Bryn also agree that this rule should include HTML `input type="image"` and `area`

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
+---
+name: Image accessible name is placeholder text
+description: |
+  This rule checks that image elements that use generic placeholder text as their accessible name, do so without loss of infomation to the user.
```
kasperisager: https://github.com/auto-wcag/auto-wcag/pull/263/files#r224338514

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
...
+- DOM Tree
+- CSS Styling
+
+authors:
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to HTML `input type="image" `and `area` elements or any HTML element with the semantic role of `img` [exposed to assistive technologies](#exposed-to-assistive-technologies), where the [accessible name](#accessible-name) exclusively uses [generic placeholder text](#generic-placeholder-text).
```
kasperisager: https://github.com/auto-wcag/auto-wcag/pull/263/files#r224338750
https://github.com/auto-wcag/auto-wcag/pull/263/files#r224338831

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Expectation
+
+Each test target accessible name can be used in place of the elements non-text content without loss of information to the user.
```
kasperisager: https://github.com/auto-wcag/auto-wcag/pull/263/files#r224341007

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -16,11 +16,11 @@ authors:
 
 ### Applicability
 
-The rule applies to HTML `input type="image" `and `area` elements or any HTML element with the semantic role of `img` [exposed to assistive technologies](#exposed-to-assistive-technologies), where the [accessible name](#accessible-name) exclusively uses [generic placeholder text](#generic-placeholder-text).
+The rule applies to HTML `input type="image"` and `area` elements, or any HTML element with the semantic role of `img` [exposed to assistive technologies](#exposed-to-assistive-technologies), where the [accessible name](#accessible-name) exclusively uses [generic placeholder text](#generic-placeholder-text).
```
kasperisager: Are `input type="image"` and `area` elements only applicable if they're exposed to AT? If so, the sentence should be restructured to reflect this:

> The rule applies to any HTML `input type="image"` or `area` element, or any HTML element with a [semantic role](#semantic-role) of `img`, where the element is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [accessible name](#accessible-name) that exclusively uses [generic placeholder text](#generic-placeholder-text).

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
+---
+name: Image accessible name is placeholder text
```
nitedog: Hmm. I think most rule titles state the success condition rather than the failure condition - for example, "Image accessible name is suitable placeholder text".

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
...
+- [Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/WCAG20-TECHS/F30.html)
+
+## Test Cases
+
+### Pass example 1
+
+The element's accessible name uses generic placeholder text which accurately describes the image.
+
+```html
+<img alt="oage"> (need a source file)
```
nitedog: How can we judge that this is generic placeholder text without the surrounding context?

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
...
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to HTML `input type="image"` or `area` elements, or any HTML element with the semantic role of `img`, where the element is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [accessible name](#accessible-name) that exclusively uses [generic placeholder text](#generic-placeholder-text).
+
+### Expectation
+
+Each test target has an accessible name that can be used in place of the [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content) of the element without loss of information to the user.
```
nitedog: I don't know what loss of information means. I suggest using the wording of the success criterion: "The accessible name of each test target provides serves the equivalent purpose of the element" or such.

audreymaniez: The rule follows the same logic as #263 (which is written upon the same failure F30)

I think here (as in #263), the idea is to write a rule that can be converted in a semi-automated test : a semi-auto test could, based on conditions, flag what seems to be generic placeholder text, and wait for a human validation for the relevance of the generic text.

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
...
+```html
+<img alt="oage"> (need a source file)
+```
+
+### Failure example 1
+
+The element's accessible name uses generic placeholder text which does not accurately describes the image.
+
+```html
+<img src="https://www.w3.org/WAI/demos/bad/after/img/teaser_right2.jpg.png" alt="image">
```
audreymaniez: The src ends with two extension (.jpg.png)

Can we just add one (remove the .png for instance) :) 

````html
https://www.w3.org/WAI/demos/bad/after/img/teaser_right2.jpg
````

File: _rules/SC1-1-1-image-accessible-name-is-placeholder-text.md [hunk 24]
```diff
@@ -0,0 +1,78 @@
+---
+name: Image accessible name is placeholder text
+description: |
+  This rule checks that image elements that use generic placeholder text as their accessible name do so without loss of infomation to the user.
```
annethyme: infomation --> information

_rules/SC1-1-1-image-accessible-name-is-placeholder-text.md
```diff
@@ -0,0 +1,78 @@
...
+- DOM Tree
+- CSS Styling
+
+authors:
+- Bryn Anderson
+---
+
+### Applicability
+
+The rule applies to HTML `input type="image"` or `area` elements, or any HTML element with the semantic role of `img`, where the element is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [accessible name](#accessible-name) that exclusively uses [generic placeholder text](#generic-placeholder-text).
```
carlosapaduarte: link semantic role to its definition

carlosapaduarte: Additionally, this should be applicable only to non-decorative images

kasperisager: Ditto on https://github.com/auto-wcag/auto-wcag/pull/263#discussion_r228971390.

Brynanders: I know what you mean but non-decorative is considered subjective and their fore cannot be included as part of the applicability as per Wilco's comment on the ACT rules format https://github.com/auto-wcag/auto-wcag/pull/263#discussion_r253140910

jkodu:
It was agreed on the WCAG call, that, this is not a failure rule, but this is a nice to have or needs review rule, so closing for now.

PR #252, opened 2018-09-12
Update and rename SC2-2-1+SC3-2-5-meta-refresh.md to SC2-2-1+SC2-2-4-…

annethyme:
SC 2.2.4 added, SC 3.2.5 removed

This rule is a mash of two techniques (redirect + refresh). 
Both techniques are failure techniques for 2.2.1 AND 2.2.4 - so I added the last one. 
However, only the refresh part applies to 3.2.5, not the redirect, so I removed that one - is this the right way to go, though? Hope we can discuss.

Closes issue: none

## Guidance the the PR (pull request) creator

**When creating PR:**

**After creating PR:**

# How to Review And Approve


ShadowBB:
You state that you added 2.2.4 but you added 2.2.3. I think this might be a typo. Correct?

annethyme:
@ShadowBB, yes indeed. Thanks! I have corrected it now.

WilcoFiers:
I'm not sure why you think 3.2.5 isn't applicable. If you automatically reload a page, you trigger a change in context that isn't initiated by the user.

ShadowBB:
I actually agree with Wilco. Redirection is explicitly mentioned in the examples of 3.2.5 as something that should be "unnoticeable".

annethyme:
@WilcoFiers, yes I agree that **reloading** a page is in violation of 3.2.5. But this rule also covers **redirects**. And these are mentioned as a sufficient technique, not a failure technique. And as far as I remember, you were the one to argue that only failures should be Auto-WCAG rules?

ShadowBB:
I have no idea why F40 doesn't also mention 3.2.5. A redirect is a chance of context correct? And a redirect in the way this rule tests for is not initiated by a user request correct? So all fail examples should fail 3.2.5 correct?

audreymaniez:
I was playing around with the examples :-) So my little contribution, 
for the Inapplicable example 7 :

````html
<head>
  <meta data-rule-target http-equiv="refresh" content="+5; http://example.com">
</head>
````

In Firefox (62.0)  it causes the page to refresh, even if the code is invalid (but not in Chrome or Edge, the two acts as expected, no refresh). 

Don't know how to treat that... It's a UA dysfunction, but it causes a big problem, it totally prevents access to information. So does this example has to become a failed example or it's better to just add an accessibility support note and keep it innaplicable ?

annethyme:
Now also has 3.2.5 Change on Request as a requirement, because we think that we are more right than WCAG :)

annethyme:
@jkodu, can you help resolving conflicts?

PR #263, opened 2018-09-18
SC1-1-1-filename-is-valid-accessible-name

Brynanders:
<< Describe the changes >>

Closes issue: #216 

**When creating PR:**

**After creating PR:**
- [x] OPTIONAL: If you want anyone in particular to review your pull request, assign them as "Reviewers".

# How to Review And Approve


_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,64 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `img` element that is exposed to assistive technologies, where the element has a non-empty `alt` attribute that is equal to the accessible name of the element, and where the `alt` attribute of the element references the file name and extension of the `src` attribute of the element.
```
carlosapaduarte: "references the file name and extension" is not unambiguous for me. 

I do not understand what makes test case 1 applicable and test case 3 inapplicable. They both "reference" the file name (example) and the extension (.pdf and PDF). Is "reference" only applicable if both file name and extension are used in the same case as the one used in the src attribute? Or does it have to include the dot in the extension?

kasperisager: That's a really good point and something we'll have to clarify.

Brynanders:
> @Brynanders I still don't understand - the title says "check the accessible name" but the applicability only checks for 'alt' - is that a bug or feature?

@nitedog more of an oversight than anything else. I have updated the Title and Description to reflect the applicability.

nitedog:
Hmmm. Why is this limited to the alt-attribute only? A text alternative as per Failure F30 could be interpreted more broadly in my opinion. No?

kasperisager:
@nitedog @Brynanders The original reasoning, and correct me if I'm wrong @WilcoFiers, was that we could reduce the signal/noise ratio by limiting this check (and the one pertaining to placeholder labels) to elements and attributes that we knew certain authoring tools might populate with generic labels, e.g. `` elements with `alt` attributes.

nitedog:
I don't have data to agree or disagree with that but it seems to me that if authoring tools will rarely use other approaches, like aria-label, then the noise ration will hardly increase. But again, others may have better insights on how often this occurs. In any case, if we do limit this, we should state why in a note or in the assumptions section.

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,77 @@
+---
+name: image accessible name contains image file name
+description: |
+  This rule checks that the use of an image's file name in it's accessible name, is appropriate.
```
WilcoFiers: I suggest we put something about authoring tools in the description. Make it clear why this is a thing.

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,77 @@
...
+- DOM Tree
+- CSS Styling
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
```
WilcoFiers: Include a single empty line between every text block, heading, list, etc. (see style guide).

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,77 @@
...
+The rule applies to any HTML `img` element that is exposed to assistive technologies, where the accessible name of the element contains the file name specified in the `src` attribute of the element.
```
WilcoFiers: Bunch of thoughts:

- Link "exposed to assistive technologies". 
- Check the `alt` attribute, not the accessible name. Authoring tools do this to alt attributes, not to other stuff.
- Add an example of what the file name is. So: src="foo/bar.jpg?baz" has the file name `bar.jpg`.
- Add another condition that checks that the src and the alt attribute aren't the same
- IMO, this should look at an exact match between alt and filename, not if it is contained.

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Expectation
+Each test target has an accessible name that serves as an appropriate text alternative as described in [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/WCAG20-TECHS/F30.html).
```
WilcoFiers: I'm unsure if linking to outside tests is allowed in ACT RF. I suggest we make whatever part we want from F30 part of this expectation, rather than us referencing a document that can change at any time.

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,77 @@
...
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F30)
+
+## Test Cases
+
+### Pass example 1
+
+Image accessible name contains the image file name.
+
+```html
+<img scr="/example.pdf" alt="an example of a .pdf file">
```
WilcoFiers: `.pdf` isn't a file name. I think this would be inapplicable.

nitedog:
does anyone have (query-able) evaluation data they can dig into to explore this? it seems that this would benefit from fine-tuning based on real data.

Brynanders:
@nitedog data would be nice but knowing that several accessibility testing vendors already have checks for this (or equivalent) rule , and considering the unlikely use case for accessible names to intentionally use the image file name or source exclusively as the only value... is there any need for data?

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,81 @@
...
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `img` element that is [exposed to assistive technologies](#exposed-to-assistive-technologies), where the [accessible name](#accessible-name) value exactly matches the file name specified in the `src` attribute.
+
+**Note**: URL paramters, case sensitivey, forward and trailing whitespace should be ignored. For example `src="/foo/bar.jpg?baz "` has a file name "bar.jpg".
```
kasperisager: "URL paramters" -> "query string".

nitedog: Should this be a definition for a term like "matching filename"?

_rules/SC1-1-1-accessible-name-is-image-file-name.md
```diff
@@ -0,0 +1,81 @@
+---
+name: image accessible name is image file name
```
nitedog: Is the correct spelling "filename" or "file name"? WCAG uses the former, should we use that for consistency? (throughout the rule)

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+The `img` element's accessible name matches the image's filename which does not acurately describe the image.
+
+```html
+<img src="https://www.w3.org/WAI/demos/bad/after/img/teaser_right2.jpg" alt="teaser_right2.jpg">
+```
+
+### Inapplicable example 1
+
+Mot exposed to assisstive technologies.
```
audreymaniez: Typo :-) 

"Mot" Should be "Not"

and

"assisstive" should be "assistive"

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input type="image" `and `area` elements or any HTML element with the semantic role of `img` [exposed to assistive technologies](#exposed-to-assistive-technologies), where the [accessible name](#accessible-name) value exactly matches the filename specified in the `src` attribute.
+
+**Note**: URL query strings, case sensitivey, forward and trailing whitespace should be ignored. For example `src="/foo/bar.jpg?baz "` has a file name "bar.jpg".
```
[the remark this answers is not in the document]
audreymaniez: And to be consistent, replace "file name" by "filename" in the note

> **Note**: URL query strings, case sensitivey, forward and trailing whitespace should be ignored. For example `src="/foo/bar.jpg?baz "` has a filename "bar.jpg".

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+### Inapplicable example 1
+
+Mot exposed to assisstive technologies.
+
+```html
+<img role="presentation">
+```
+### Inapplicable example 2
+
+Image is not exposed to assisstive technologies.
```
audreymaniez: Typo : "assisstive" should be "assistive"

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+Image accessible name does not contain the image filename.
+
+```html
+<img src="https://www.w3.org/WAI/demos/bad/after/img/teaser_right2.jpg" alt="modanna lily">
+```
+
+### Inapplicable example 4
+
+The `img` `alt` does not accurately describe the image but is overridden by `aria-label` which does accurately describes the image.
```
audreymaniez: I propose to rewrite to emphasized that alt attribute "use the filename"

> The `alt`  attribute uses the filename, but the accessible name is actually calculated with the `aria-label` value, which doesn't use the filename.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+## Background
+
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F30)
+
+## Test Cases
+
+### Pass example 1
+
+The `img` element's accessible name acurately describes the image.
```
audreymaniez: Since the test is about testing the use of the filename, i propose to rewrite : 

> The `img` element's accessible name uses the filename and accurately describes the image.

(typo on "acCurately")

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+```html
+<img scr="https://www.w3.org/WAI/demos/bad/img/w3c.png" alt="w3c">
+```
+
+### Failure example 1
+
+The `img` element's accessible name matches the image's filename which does not acurately describe the image.
```
audreymaniez: I'm certainly not the best english speaker here :D but : we have "the image filename"  and here we have "the image's filename" (with the "'s" possession). I don't know which is the best, I just suggest it has to be consistent within the rule.

_rules/SC1-1-1-accessible-name-is-image-filename.md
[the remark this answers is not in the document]
audreymaniez: And a typo (I think?) : 

"case sensitivey" should be "case sensitive"

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
+---
+name: image accessible name is image filename
+description: |
+  This rule checks that `img` elements that use their source filename as their accessible name, do so without loss of infomation to the user.
```
audreymaniez: It's inconsistent with the applicability (just `img`). 

Proposal, replace 'ìmg`elements by "image" (as it is done in #251) :  

>   This rule checks that image that use their source filename as their accessible name, do so without loss of infomation to the user.

_rules/SC1-1-1-accessible-name-is-image-filename.md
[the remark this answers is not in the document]
Brynanders: Thanks for spotting the typo's and inconsistency @audreymaniez .

In regards to the filename extension, the intention was that the rule checks for an exact match to avoid catching instances of e.g. `src="/foo/Deque.jpg" alt="Deque"` which I would argue should not be flagged by default.
Which would mean....
- `src="/foo/bar_2.jpg" alt="bar_2.jpg"` is applicable.
- `src="/foo/bar_2" alt="bar_2"` is applicable.
- `src="/foo/bar_2.jpg" alt="bar_2"` is NOT applicable.

_rules/SC1-1-1-accessible-name-is-image-filename.md
Brynanders: Changed acurately > accurately.

This is not an "exact match" of the filename based on the example given in the note so I would not reference the filename in the description.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input type="image"` and `area` elements, or any HTML element with the semantic role of `img` [exposed to assistive technologies](#exposed-to-assistive-technologies), where the [accessible name](#accessible-name) value exactly matches the filename specified in the `src` attribute.
```
kasperisager: Ditto on https://github.com/auto-wcag/auto-wcag/pull/251#discussion_r224378186.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F30)
+
+## Test Cases
+
+### Pass example 1
+
+The element's accessible name matches the image filename which accurately describes the image.
+
+```html
+<img scr="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c">
```
nitedog: scr -> src

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Applicability
+
+The rule applies to HTML `input type="image"` or `area` elements, or any HTML element with the semantic role of `img`, where the element is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [accessible name](#accessible-name) that matches the filename specified in the `src` attribute.
+
+**Note**: URL query strings, difference in letter casing, and forward and trailing whitespace should be ignored. For example `src="/foo/bar.jpg?baz "` has a filename `bar.jpg`.
+
+### Expectation
+
+Each test target has an accessible name that can be used in place of the [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content) of the element without loss of information to the user.
```
nitedog: I'm not sure about "loss of information" but don't have a better suggestion. The SC says "serves equivalent purpose".

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input type="image"` or `area` elements, or any HTML element with the semantic role of `img`, where the element is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [accessible name](#accessible-name) that matches the filename specified in the `src` attribute.
```
kasperisager: `input type="image"` is somewhat ambiguous. Should the rule only apply to `input` elements with a `type` attribute that is exactly equal to `"image"`? Or should it really be applicable to all `input` elements that are image buttons, i.e. has a `type` attribute whose value maps to the `Image Button` state?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or `area` elements, or any HTML element with the semantic role of `img`, where the element is [exposed to assistive technologies](#exposed-to-assistive-technologies) and has an [accessible name](#accessible-name) that matches the filename specified in the `src` attribute.
```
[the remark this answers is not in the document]
Brynanders: Changed this and created a definition for filename #330

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or `area` elements, or any HTML element with the semantic role of `img`, where the element is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivilant to the [filename](#filename) specified in the `src` attribute.
```
Rafalchar: I suppose it should be "equivalent" (just spelling).

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Applicability
+
+The rule applies to HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or `area` elements, or any HTML element with the semantic role of `img`, where the element is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute.
+
+**Note**: Difference in letter casing, and forward and trailing whitespace should be ignored. 
+
+### Expectation
+
+Each test target has an accessible name that serves an equivalent purposecan of the [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content).
```
kasperisager: `purposecan` lacks a space.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or `area` elements, or any HTML element with the semantic role of `img`, where the element is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute.
+
+**Note**: Difference in letter casing, and forward and trailing whitespace should be ignored.
```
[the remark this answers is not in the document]
audreymaniez: As it's suggested in the #251 ([see comment](https://github.com/auto-wcag/auto-wcag/pull/251#discussion_r229372629)), this tests should only be applicable to non decorative image ?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+*There are currently no assumptions*
+
+## Accessibility support
+
+ *There are no major accessibility support issues known for this rule.*
+
+## Background
+
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F30)
```
audreymaniez: the link to the F30 is a not stable one (TR/2016/NOTE-WCAG20-TECHS-20161007). Do we better link to techniques WCAG 2.1 document : https://www.w3.org/WAI/WCAG21/Techniques/failures/F30 ?

Brynanders: Nice one! great spot :)

_rules/SC1-1-1-accessible-name-is-image-filename.md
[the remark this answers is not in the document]
audreymaniez: I should have detailled more my though, sorry  :)

In fact, the image that would be tested can have an alt, and so be included in accessibility tree. But, imagine that image is, in the end, a decorative image. That test should not apply, because that image should be evaluate regarding a test that determine if it has a null alt (or whatever that makes it decorative). But the validity of the text of the alt should not be tested.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,88 @@
...
+## Background
+
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/WAI/WCAG21/Techniques/failures/F30)
+
+## Test Cases
+
+### Pass example 1
+
+The `img` element's accessible name uses the filename and accurately describes the image.
```
jkodu: needs updating the `filename` to the glossary/ definition.

_rules/SC1-1-1-accessible-name-is-image-filename.md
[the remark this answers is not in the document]
Brynanders: updated applicability based on @audreymaniez comments in agreement with @jkodu and @annethyme .
Also updated the [decorative](https://github.com/auto-wcag/auto-wcag/blob/Brynanders-patch-1/pages/algorithms/decorative.md) definition to include a note taken from **Image has name** #274 on how to mark an image as 'decorative'. 
Also made a new commit for [**Image has name** here](https://github.com/auto-wcag/auto-wcag/blob/Brynanders-patch-1/_rules/SC1-1-1-image-has-name.md).

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or `area` elements, or any [non-decorative](#decorative) HTML element with the semantic role of `img`, where the element is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored. 
+
+### Expectation
+
+Each test target has an accessible name that serves an equivalent purpose can of the [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content).
```
ShadowBB: I think this sentence got garbled during a recent edit. I think the word "can" should be dropped.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Failure example 1
+
+
+The `img` element's accessible name matches the image filename which does not acurately describe the image.
+
+```html
+<img src="https://www.w3.org/WAI/demos/bad/after/img/teaser_right2.jpg" alt="teaser_right2.jpg">
+```
+
```
ShadowBB: I think these examples would benefit from a failure example that explicitly makes it clear that extensions should never be part of a text alternative because extensions do not serves the same purpose as the image (if my reading of the expectation is correct).

Suggestion:

### Failure example 2
The `img` element's accessible name matches the image filename which has a file extension which doesn't serves the same purpose as the image.

```html
<img src="https://www.w3.org/WAI/demos/bad/img/w3c.png" alt="w3c.png">
```

_rules/SC1-1-1-accessible-name-is-image-filename.md
ShadowBB: The word "equivalent" by itself in ambiguous. It can be interpreted as "each test target has an accessible name that **at least** serves an equivalent purpose of the non-text-content" or it can be interpreted as "each test target has an accessible name that serves **only** an equivalent purpose of the non-text-content".

I suggest the latter.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,88 @@
...
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or `area` elements, or any [non-decorative](#decorative) HTML element with the semantic role of `img`, where the element is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
audreymaniez: `area` can also be decorative. 

But `area` has the semantic role of `link` not `img`so it can't be included in the "or any [non-decorative](#decorative) HTML element with the semantic role of `img`".

I suggest to reorder and to split to make it easier to read  : 

> The rule applies to `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, `area` element, or any HTML element with the semantic role of `img`, where the element is [non-decorative](#decorative) and is [included in the accessibility tree](#included-in-the-accessibility-tree).
The element has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored. 


**Note** : I changed "elements" to the singular "element" (other rule use the singular for the applicability "the X element")

_rules/SC1-1-1-accessible-name-is-image-filename.md
audreymaniez: I do not agree ... Relevance of a text alternative is so unpredictable. 

You don't know if "W3C.png" would never be a conforming text alternative in a particular case, nor if a perfect matching between a src and a alt (including an extension) can't never be a passed test. 

> because extensions do not serves the same purpose as the image

I don't think it can be so systematic. 

Imagine a photo gallery application (like a media library in a content management system), which the only contentS are photos. In that case, for a blind user that just uploaded a photo in the library, the important information is the the name of the filename (to reach the one he has just uploaded). 
The alt attribut could be use to bring that information (and as an expert in that particular context of use, I won't considered that as a failure). 
In that case, the alt is not use to describe the image, but to inform about the name of the file. 

So i think "Each test target has an accessible name that serves an equivalent purpose " is a wide enough to cover each case (describe what is on the image, describe the function of the image or other use case we don't know yet). 

Sorry for having beeing so long, I'm done now :-)

ShadowBB: That is a good point. I guess there are exceptions. I think an example where that is not the case (like my example) is still a good addition but I now agree that my generalizing statement about file extensions never serving the same purpose as the image is incorrect.

_rules/SC1-1-1-accessible-name-is-image-filename.md
Brynanders: Updated to include decorative `area` elements
Updated to use the singular "element" instead of the plural "elements".

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML `area` element, or any HTML element with the semantic role of `img`, where the element is [non-decorative](#decorative) and is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
WilcoFiers: Can you write this so that it isn't one long sentence?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML `area` element, or any HTML element with the semantic role of `img`, where the element is [non-decorative](#decorative) and is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored. 
+
+### Expectation
+
+Each test target has an accessible name that serves only an equivalent purpose of the [non-text content](https://www.w3.org/TR/WCAG21/#dfn-non-text-content).
```
WilcoFiers: @annethyme you had concerns about this being inconsistent with the way the ARIA rules are formulated. Do we need to look at switching this around?

_rules/SC1-1-1-image-has-name.md
```diff
@@ -25,8 +25,6 @@ The rule applies to HTML `img` elements or any HTML element with the [semantic r
 
 Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty) or is marked as [decorative](#decorative).
 
-**NOTE**: An `img` element can be marked as decorative, by using either `role="presentation"`, `role="none"` or an empty alt attribute `alt=""`.
```
WilcoFiers: Why did you remove this note?

_rules/SC1-1-1-accessible-name-is-image-filename.md
WilcoFiers: "decorative" is a subjective term. Subjective terms are not allowed in applicability according to the ACT Rules Format.

Brynanders: @WilcoFiers I'm confused, it's linked to a definition, is that not sufficient?

_rules/SC1-1-1-image-has-name.md
Brynanders: @WilcoFiers I added this information to the [decorative](https://github.com/auto-wcag/auto-wcag/blob/Brynanders-patch-1/pages/algorithms/decorative.md) definition and therefore considered the note duplication of the definition.

_rules/SC1-1-1-accessible-name-is-image-filename.md
Brynanders: Updating the applicability based on both @audreymaniez and @WilcoFiers requests for this section not to read as one long sentence.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML `area` element, or any HTML element with the semantic role of `img`. Applicable elements are [non-decorative](#decorative), [included in the accessibility tree](#included-in-the-accessibility-tree), and have an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
jkodu: semantic role - should be a linked glossary term

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
...
+
+The `img` element's accessible name uses the filename and accurately describes the image.
+
+```html
+<img src="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c">
+```
+
+### Failure example 1
+
+The `img` element's accessible name matches the image filename which does not acurately describe the image.
```
jkodu: acurately > accurately

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML `area` element, or any HTML element with the [semantic role](#semantic-role) of `img`. Applicable elements are [non-decorative](#decorative), [included in the accessibility tree](#included-in-the-accessibility-tree), and have an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
WilcoFiers: This seems to imply taking the `src` attribute from an `area` element. The `area` element does not have an `src` attribute, so that doesn't make sense. I think this rule should ignore `area` elements for that reason.

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
...
+## Accessibility support
+
+ *There are no major accessibility support issues known for this rule.*
+
+## Background
+
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/WAI/WCAG21/Techniques/failures/F30)
+
+## Test Cases
```
WilcoFiers: You should include `input[type=image]` elements in the tests. And if you're keeping `area` elements, those too.

_rules/SC1-1-1-accessible-name-is-image-filename.md
Brynanders: removed `area` elements

_rules/SC1-1-1-accessible-name-is-image-filename.md
Brynanders: added `input[type=image]`

Brynanders:
@WilcoFiers, @annethyme, @jkodu this should be good to go.

pages/glossary/decorative.md
```diff
@@ -3,4 +3,4 @@ title: Decorative
 key: decorative
 ---
 
-Serving only an aesthetic purpose, providing no information, and having no functionality.
\ No newline at end of file
+Serving only an aesthetic purpose, providing no information, and having no functionality. An `img` element can be marked as decorative, by using either `role="presentation"`, `role="none"` or an empty alt attribute `alt=""`.
```
WilcoFiers: I think this definition contradicts itself. An img can have a purpose, even when it's marked up with role=presentation. It shouldn't, it's an accessibility failure, but it can. I don't think you need this "decorative" definition in this rule at all. Why did you add this?

Brynanders: @WilcoFiers I have updated the definition to separate out the implementation detail as a note after discussion with Anne and Jey

Brynanders: https://github.com/auto-wcag/auto-wcag/pull/263#discussion_r249721940

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,95 @@
+---
+name: Image accessible name is filename
```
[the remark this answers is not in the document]
Brynanders: Old commit

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,89 @@
...
+description: |
+  This rule checks that image elements that use their source filename as their accessible name do so without loss of infomation to the user.
```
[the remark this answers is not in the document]
Brynanders: Been fixed, old commit comment

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,111 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML element with the [semantic role](#semantic-role) of `img` that has a `src` attribute. Applicable elements are [non-decorative](#decorative), [included in the accessibility tree](#included-in-the-accessibility-tree), and have an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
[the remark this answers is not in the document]
Brynanders: @WilcoFiers do you have any objections to @annethyme suggestion to create a list here?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,111 @@
...
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/WAI/WCAG21/Techniques/failures/F30)
+
+## Test Cases
+
+### Pass example 1
+
+The `img` element's accessible name uses the filename which accurately describes the image.
+
+```html
+<img src="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c">
```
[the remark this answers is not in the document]
Brynanders: Filenames don't necessarily need to include a file extension e.g. https://stackoverflow.com/questions/7356205/remove-file-extension-from-a-file-name-string.

To be fair someone else suggested this so it didn't come from me but it makes sense so I included it.

pages/glossary/decorative.md
```diff
@@ -3,4 +3,6 @@ title: Decorative
 key: decorative
```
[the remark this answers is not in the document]
Brynanders: @WilcoFiers I have streamlined the applicability, I think you will like it :)

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,117 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML element with the [semantic role](#semantic-role) of `img` that has a `src` attribute. Applicable elements are not marked as [decorative](#decorative), [included in the accessibility tree](#included-in-the-accessibility-tree), and have an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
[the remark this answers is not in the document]
Brynanders: Updated the applicability... <div role="img" src="blaa"> would no longer be applicable as it doesn't have an accessible name.

pages/glossary/decorative.md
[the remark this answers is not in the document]
annethyme: I like it now!

I think the rule is correct now.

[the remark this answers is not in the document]
Brynanders: I initially took out "included in the accessibility tree" but added it again because otherwise elements with display:none would be applicable and they shouldn't be.

goetsu:
Image a situation where user can download multiple format of the same image 
<button></button>
<button></button>
<button></button>

In that case I think using filename is compliant with 1.1 and must not be marked as a failure so this test can be completely automated

Brynanders:
Corb O'Connor comments in the CFC email chain:
> This looks great; thanks for your hard work on this. Am I understanding correctly that the rule requires the extension to be part of the filename? Imagine I have a picture of Barack Obama, saved as “Barack Obama.jpg” but set the ALT to the filename minus the extension, then this rule would not fail…right? Assuming so, then I have no comments!

Brynanders:
> Image a situation where user can download multiple format of the same image
> 
> ```
> <button></button>
> <button></button>
> <button></button>
> ```
> In that case I think using filename is compliant with 1.1 and must not be marked as a failure so this test can be completely automated

@goetsu thanks for your comment.

Your example would not be marked as a failure with this check because the alt "serves an equivalent purpose to the non-text content".

You could write a rule that looks for that exact scenario and fully automate it but I don't think that building rules for a single purpose would be the most efficient approach.

Brynanders:
> Corb O'Connor comments in the CFC email chain:
> 
> > This looks great; thanks for your hard work on this. Am I understanding correctly that the rule requires the extension to be part of the filename? Imagine I have a picture of Barack Obama, saved as “Barack Obama.jpg” but set the ALT to the filename minus the extension, then this rule would not fail…right? Assuming so, then I have no comments!

In this instance, the accessible name of "Barack Obama" does not include the file extension which is part of the filename specified in the `src` attribute so it would be considered inapplicable and not be flagged.

pages/glossary/decorative.md
```diff
@@ -3,4 +3,6 @@ title: Decorative
...
 ---
 
-Serving only an aesthetic purpose, providing no information, and having no functionality.
\ No newline at end of file
+Serving only an aesthetic purpose, providing no information, and having no functionality. 
+
+**Note:** Authors can mark an `img` element as decorative to indicate that it should be ignored by assistive technology by using either `role="presentation"`, `role="none"`, or `alt=""`. An element should only be marked as decorative if removing the element does not cause a loss of information to the user.
```
EmmaJP: Could this also apply to elements other than 'img', such as 'svg'?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,117 @@
...
+
+authors:
+- Bryn Anderson
+---
+
+## Test procedure
+
+### Applicability
+
+The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML element with the [semantic role](#semantic-role) of `img`, is [included in the accessibility tree](#included-in-the-accessibility-tree), and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.
```
carlosapaduarte: editorial: is [included in the accessibility tree] > which is [included in the accessibility tree]

Isn't the last sentence an assumption?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### Pass example 2
+
+The `input` element with a `type` of `image` has an accessible name that uses the filename which accurately describes the image.
+
+```html
+<input type="image" src="https://www.w3.org/WAI/demos/bad/img/w3c.png" alt="W3C">
+```
+
+### Falied
```
carlosapaduarte: Falied > Failed
(here and in all instances below)

pages/glossary/decorative.md
Brynanders: The context of this rule is `img` so the note addresses that specifically and not other elements.

Brynanders: I could change this to say "Authors can mark elements as decorative..." if you think that would make more sense?

_rules/SC1-1-1-accessible-name-is-image-filename.md
```diff
@@ -0,0 +1,117 @@
...
+ *There are no major accessibility support issues known for this rule.*
+
+## Background
+
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+- [F30: Failure of Success Criterion 1.1.1 and 1.2.1 due to using text alternatives that are not alternatives (e.g., filenames or placeholder text)](https://www.w3.org/WAI/WCAG21/Techniques/failures/F30)
+
+## Test Cases
+
+### Passed
```
WilcoFiers: It doesn't seem like any of the passed / failed examples would be applicable. The applicability doesn't say to ignore file extensions or its file path.

Brynanders: The applicability refers to a definition of "filename" https://github.com/auto-wcag/auto-wcag/blob/master/pages/glossary/filename.md which excludes the file path and appended query strings.

Brynanders: @WilcoFiers @kasperisager and I have reworked some of the test cases to fit the applicability. Please can you give this another look.

annethyme:
@Brynanders, I think that since Carlos' permissions changed the other day, the status of this PR is now "changes requested" instead of "approved". I think you will have to dismiss Carlos' (old) review to get things back in order.

annethyme:
@Brynanders, @jkodu, @WilcoFiers, can this one be merged? 
It has survived its two-week CF.

PR #269, opened 2018-09-21
"Included in accessibility tree" instead of "exposed to assistive technologies"

annethyme:
Proposed name change from "exposed to assistive technologies" to "included in the accessibility tree"

To resolve confusion around if including an element in the tab order should be part of the definition of "exposed to assistive technologies". "Included in the accessibility tree" makes it clear what is included in this definition, and it makes the definition clearer in relation to the question of accessibility support in AT.

**When creating PR:**

**After creating PR:**
- [x Add relevant project (e.g. "Q3 2018 Status") to PR
- [x] OPTIONAL: If you want anyone in particular to review your pull request, assign them as "Reviewers".

# How to Review And Approve


pages/algorithms/exposed-to-assistive-technologies.md
```diff
@@ -9,3 +9,6 @@ Elements contained in the accessibility tree as described in one of the followin
 - [HTML Accessibility API Mappings](https://www.w3.org/TR/html-aam)
 - [SVG Accessibility API Mappings](https://www.w3.org/TR/svg-aam)
 
+Elements that are included in the accessibility tree are exposed to assistive technologies. 
+
+The elements that are included in the accessibility tree are however not the only ones that might be exposed to assistive technologies. An example is that elements that are [focusable](#focusable) can still be reached by a screen reader through tabbing.
```
kasperisager: The file will need to be renamed as well.

WilcoFiers:
My main concern with renaming this is around plain language. On the one hand, to anyone who knows what the accessibility tree is, this is a bit clearer. To anyone who doesn't know this I would guess that the existing name works better. My gut feeling says that assistive technologies is a much better understood term.

pages/algorithms/exposed-to-assistive-technologies.md
WilcoFiers: This strongly implies that this definition shouldn't be renamed. I'd prefer that, as the accessibility tree is a far more technical term than assistive technologies.

annethyme:
Problem is that "exposed to assistive technologies" is unclear when it comes to things that has explicitly been removed from the accessibility tree, but are then still reachable for a user of an assistive technology. @WilcoFiers, how do you propose we get around that with the current definition?

annethyme:
Another solution would be to have a term "exposed to assistive technologies" that links out to "included in the accessibility tree" and "focusable". 
But do we want to include "focusable" in the applicability for e.g. "image has name"?
@WilcoFiers, especially looking for your opinion on this.

annethyme:
Wilco, Jey, Kasper, Bryn and Anne just agreed to go with the "included in the accessibility tree" definition. @annethyme will update PR and send out for review again.

annethyme:
@WilcoFiers and @jkodu, do you feel that this matches what we discussed? We have 3 approvals, so I would like to merge, if you guys are good with the new definition and the use of it.

PR #273, opened 2018-09-27
SC4-1-2-role-attribute-has-valid-value

jkodu:
Rule: Role attribute has valid value

Closes issue:
- https://github.com/auto-wcag/auto-wcag/issues/196

**When creating PR:**

**After creating PR:**

# How to Review And Approve

### DIFF
--- a/_rules/SC4-1-2-role-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-role-attribute-has-valid-value.md
@@ -34,7 +34,7 @@ Each test target has at least one non-abstract valid value as specified in the [
 
 Older browsers do not support more than one token in the value for a role attribute. If multiple values are used in the role attribute, the attribute is ignored in these browsers.
  
-**Note:** Having a whitespace separated list of more than one token in the value of the role attribute is used for what is known as `fallback roles`. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth.
+**Note:** Having a whitespace separated list of more than one token in the value of the role attribute is used for what is known as _fallback roles_. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth.
 
 ## Background
 
--- a/_rules/SC4-1-2-role-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-role-attribute-has-valid-value.md
@@ -1,0 +1,114 @@
+---
+name: Role attribute has valid value
+test_type: atomic
+
+description: |
+   This rule checks that, elements with a role attribute have valid values
+   
+success_criterion:
+- 4.1.2 # Name, Role, Value
+
+test_aspects:
+- DOM Tree
+
+authors:
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) `role` attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
+
+### Expectation
+
+Each test target has at least one valid value as specified in the [WAI-ARIA 1.1, roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+- [WAI-ARIA 1.1, roles categories](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)
+- [MDN: Using ARIA: Roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles)
+
+## Test Cases
+
+### Passed
+
+#### Passed example 1
+
+Element with valid `role` value.
+
+ ```html
+<input type="text" role="textbox">
+```
+
+#### Passed example 2
+ 
+Element with multiple valid `role` value.
+ 
+```html
+<span role="button command" ng-click='$ctrl.doSomething()'></span>
+```
+
+#### Passed example 3
+ 
+Element with valid `role` value.
+ 
+```html
+<div role="dialog" aria-busy="true" aria-modal></div>
+```
+
+#### Passed example 2
+ 
+Element with at least one valid `role` value.
+ 
+```html
+<img role="presentation xyz" src="">
+```
+
+
+
+### Failed
+
+#### Failed example 1
+
+Element with invalid `role` value.
+
+```html
+<input role="dialog" value="123">
+```
+
+#### Failed example 2
+
+Element with multiple invalid `role` value.
+
+```html
+<input type="text" role="complementary contentinfo">
+```
+
+### Inapplicable
+
+#### Inapplicable example 1
+
+Element does not have `role` attribute.
+
+```html
+<div>Some Content</div>
+```
+
+#### Inapplicable example 2
+
+Element with null `role` attribute.
+
+```html
+<div role>Some Content</div>
+```
--- a/_rules/SC4-1-2-role-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-role-attribute-has-valid-value.md
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-Any [non-empty](#non-empty) `role` attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
+Any [non-empty](#non-empty) `role` attribute that is specified on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
 
 ### Expectation
 
@@ -32,9 +32,9 @@ Each test target has at least one non-abstract valid value as specified in the [
 
 ## Accessibility Support
 
-Older browsers do not support fallback roles. If multiple values are used in the role attribute, the attribute is ignored.
-
-At the moment, none of the major assistive technologies support all `[WAI-ARIA 1.1 roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)`.
+Older browsers do not support more than one token in the value for a role attribute. If multiple values are used in the role attribute, the attribute is ignored in these browsers.
+ 
+**Note:** Having a whitespace separated list of more than one token in the value of the role attribute is used for what is known as `fallback roles`. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth.
 
 ## Background
 
--- a/_rules/SC4-1-2-role-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-role-attribute-has-valid-value.md
@@ -3,7 +3,7 @@ name: Role attribute has valid value
 test_type: atomic
 
 description: |
-   This rule checks that role attributes have valid values
+   This rule checks that each role attribute has at least one valid value
    
 success_criterion:
 - 4.1.2 # Name, Role, Value
@@ -24,7 +24,7 @@ Any [non-empty](#non-empty) `role` attribute that is placed on an HTML or SVG el
 
 ### Expectation
 
-Each test target has at least one non-abstract valid value as specified in the [WAI-ARIA 1.1, roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
+Each test target has at least one non-abstract valid value as specified in the [WAI-ARIA 1.1 Categorization of Roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
 
 ## Assumptions
 
@@ -34,12 +34,12 @@ Each test target has at least one non-abstract valid value as specified in the [
 
 Older browsers do not support fallback roles. If multiple values are used in the role attribute, the attribute is ignored.
 
-At the moment, none of the major assistive technologies support all roles.
+At the moment, none of the major assistive technologies support all `[WAI-ARIA 1.1 roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)`.
 
 ## Background
 
 - [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
-- [WAI-ARIA 1.1, roles categories](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)
+- [WAI-ARIA 1.1 Categorization of Roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)
 - [MDN: Using ARIA: Roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles)
 
 ## Test Cases
@@ -119,5 +119,5 @@ Element with null `role` attribute.
 Element that is not exposed to assistive technologies.
 
 ```html
-<div style="display:none;" role="banner">Some Content</div>
+<div aria-hidden="true" role="banner">Some Content</div>
 ```
--- a/_rules/SC4-1-2-role-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-role-attribute-has-valid-value.md
@@ -119,5 +119,5 @@ Element with null `role` attribute.
 Element that is not exposed to assistive technologies.
 
 ```html
-<div aria-hidden="true" role="banner">Some Content</div>
+<div style="display:none;" role="banner">Some Content</div>
 ```
--- a/_rules/SC4-1-2-role-attribute-has-valid-value.md
+++ b/_rules/SC4-1-2-role-attribute-has-valid-value.md
@@ -3,13 +3,14 @@ name: Role attribute has valid value
 test_type: atomic
 
 description: |
-   This rule checks that elements with a role attribute have valid values
+   This rule checks that role attributes have valid values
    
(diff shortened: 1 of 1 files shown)


_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
+---
+name: Role attribute has valid value
+test_type: atomic
+
+description: |
+   This rule checks that, elements with a role attribute have valid values
```
WilcoFiers: Lose the comma.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) `role` attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
+
+### Expectation
+
+Each test target has at least one valid value as specified in the [WAI-ARIA 1.1, roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
```
WilcoFiers: Should be "non-abstract valid role". This should also allow roles from the DPUB and Graphics modules. I think maybe we should create a definition for that. Something like "current ARIA specifications". WDYT?

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
```
WilcoFiers: We should note a couple of things here.
1. Older browsers do not support fallback roles. If multiple roles are used in the role attribute, the attribute is ignored.
2. At the moment, none of the major assistive technologies support all roles.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+ ```html
+<input type="text" role="textbox">
+```
+
+#### Passed example 2
+ 
+Element with multiple valid `role` value.
+ 
+```html
+<span role="button command" ng-click='$ctrl.doSomething()'></span>
```
WilcoFiers: What's the Angular doing here? Lets keep it simple. Also, this fallback doesn't make sense, "command button" is more likely a thing.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+#### Passed example 3
+ 
+Element with valid `role` value.
+ 
+```html
+<div role="dialog" aria-busy="true" aria-modal></div>
```
WilcoFiers: Isn't this the same as the first example? Seems redundant.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+#### Passed example 2
```
WilcoFiers: Heading number is wrong.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+ 
+Element with at least one valid `role` value.
+ 
+```html
+<img role="presentation xyz" src="">
+```
+
+
+
```
WilcoFiers: Only one empty line.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+### Failed
+
+#### Failed example 1
+
+Element with invalid `role` value.
+
+```html
+<input role="dialog" value="123">
```
WilcoFiers: Why is this an invalid role? Doesn't seem invalid to me.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+#### Failed example 2
+
+Element with multiple invalid `role` value.
+
+```html
+<input type="text" role="complementary contentinfo">
```
WilcoFiers: Again, what's invalid about these roles?

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+### Inapplicable
```
WilcoFiers: Add an example with just whitespace.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: Should this rule cater to checking if a given role is applicable for a given element type? If that is the case, this will fail because `complementary` and `contentinfo` is not applicable to `input type=text`. 

If the rule just checks if `role` is a valid value, then you are right.

WDYT?

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: See question below. After confirmation, I will change the tests.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: `button` is a subclass of `command`. Have changed it to `link` to keep it simple.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: Good point.

jkodu:
@WilcoFiers 

Few questions for you in the comments, please advise.

_rules/SC4-1-2-role-attribute-has-valid-value.md
annethyme: Should there be something here about the separate values being space-separated?

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+
+Each test target has at least one valid value as specified in the [WAI-ARIA 1.1, roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+Older browsers do not support fallback roles. If multiple roles are used in the role attribute, the attribute is ignored.
```
annethyme: "Older browsers do not support fallback roles" - can we get a link or a definition for "fallback roles"?

"If multiple roles are used in the role attribute, the attribute is ignored." --> "If multiple **values** are used in the role attribute, the attribute is ignored." - suggesting this  to keep it consistent with the wording in the Expectation

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+
+Element with valid `role` value.
+
+ ```html
+<input type="text" role="textbox">
+```
+
+#### Passed example 2
+ 
+Element with multiple valid `role` value.
```
annethyme: value --> values

_rules/SC4-1-2-role-attribute-has-valid-value.md
annethyme: If we want to also check if the role has been correctly applied, I would say we need an Expectation 2 for this.
And then it needs to be clear in the test case descriptions if the test case fails because of the value not existing in the spec or because it has been applied to an element that it is not applicable to.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,114 @@
...
+<div>Some Content</div>
+```
+
+#### Inapplicable example 3
+
+Element with null `role` attribute.
+
+```html
+<div role>Some Content</div>
+```
```
annethyme: Add Inapplicable test case for something not exposed to assistive technologies.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: I agree with you both.

Believe the rule checks for valid value of the role, and should not check if the given value for the role is applicable for the element.

I have updated the test cases accordingly.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: Updated as per suggestion.

I had no luck finding a link for `fallback roles`. The WAI Aria spec - https://www.w3.org/TR/wai-aria-1.1/ talks about fallback in general. 

@annethyme - if you could find a link do share.

But I believe the word fallback is self explanatory and a definition might be an overkill at the moment.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: @annethyme - I do not believe, we need to detail that when two words are separated by space, that they are separate values.

jkodu: Created an issue to create definition - `non-abstract valid role` - https://github.com/auto-wcag/auto-wcag/issues/277, so this PR can progress.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
+---
+name: Role attribute has valid value
+test_type: atomic
+
+description: |
+   This rule checks that role attributes have valid values
```
annethyme: This rule checks that each role attribute has at least one valid value

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) `role` attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
+
+### Expectation
+
+Each test target has at least one non-abstract valid value as specified in the [WAI-ARIA 1.1, roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
```
annethyme: I suggest changing the name of the link to "WAI-ARIA 1.1 Categorization of Roles"

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+Older browsers do not support fallback roles. If multiple values are used in the role attribute, the attribute is ignored.
```
annethyme: I think fallback roles need an explanation here, or a link to one.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+At the moment, none of the major assistive technologies support all roles.
```
annethyme: Suggest making "roles" a bit more explicit, e.g.:
"At the moment, none of the major assistive technologies support all `[WAI-ARIA 1.1 roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).`"

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Background
+
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+- [WAI-ARIA 1.1, roles categories](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)
```
annethyme: Suggesting change of link text:
`[WAI-ARIA 1.1 Categorization of Roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)`

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+```html
+<div role>Some Content</div>
+```
+
+#### Inapplicable example 4
+
+Element that is not exposed to assistive technologies.
+
+```html
+<div style="display:none;" role="banner">Some Content</div>
```
annethyme: Could we do `aria-hidden="true"` instead of `display:none`, so it's clear that it's "exposed to AT" and not "visible" that is the applicability.

_rules/SC4-1-2-role-attribute-has-valid-value.md
jkodu: @annethyme I think I commented on another PR, where you had a similar suggestion, and I had no avail to find a link for `fallback` role. 

Please feel free to recommend.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+authors:
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) `role` attribute that is placed on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
```
annethyme: placed on --> specified on

_rules/SC4-1-2-role-attribute-has-valid-value.md
annethyme: Suggestion:

> "Older browsers do not support more than one token in the value for a role attribute. If multiple values are used in the role attribute, the attribute is ignored in these browsers.
> 
> **Note:** Having a whitespace separated list of more than one token in the value of the role attribute is used for what is known as "fallback roles". If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth."

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+Older browsers do not support more than one token in the value for a role attribute. If multiple values are used in the role attribute, the attribute is ignored in these browsers.
+ 
+**Note:** Having a whitespace separated list of more than one token in the value of the role attribute is used for what is known as `fallback roles`. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth.
```
kasperisager: `` `fallback roles` `` -> `_fallback roles_`

kasperisager:
There's also something not quite right with the phrase "at least one valid value". The `role` attribute, if specified, will always have just a _single_ value. That value, being a set of space-separated tokens, may however consist of multiple tokens and at least one of these tokens has to correspond to a non-abstract role. I'd suggest something similar to this as the expectation:

> Each test target has at least one space-separated token that corresponds to a non-abstract WAI-ARIA role.
>
> **Note:** Which specific roles to consider may depend on the type of content under test, such as web pages or digital publications, and are not limited to those defined by WAI-ARIA alone. Roles defined in separate WAI-ARIA modules, such as the [Digital Publishing Module](https://www.w3.org/TR/dpub-aria/) or the [Graphics Module](https://www.w3.org/TR/graphics-aria/), may also be considered for this rule.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
+---
+name: Role attribute has valid value
+test_type: atomic
+
+description: |
+   This rule checks that each role attribute has at least one valid value
```
WilcoFiers: This doesn't check "has at least one", since you can leave it blank. I suggest just saying "has a valid value"

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Test procedure
+
+### Applicability
+
+Any [non-empty](#non-empty) `role` attribute that is specified on an HTML or SVG element that is [exposed to assistive technologies](#exposed-to-assistive-technologies).
+
+### Expectation
+
+Each test target has at least one non-abstract valid value as specified in the [WAI-ARIA 1.1 Categorization of Roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization).
```
WilcoFiers: This should include roles from aria-dpub and aria-graphics. I suggest we create a separate definition where we keep a list of ARIA core + modules that should be supported.

I'd also suggest we say "valid role" instead of "valid value".

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
+
+## Accessibility Support
+
+Older browsers do not support more than one token in the value for a role attribute. If multiple values are used in the role attribute, the attribute is ignored in these browsers.
+ 
+**Note:** Having a whitespace separated list of more than one token in the value of the role attribute is used for what is known as _fallback roles_. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth.
```
WilcoFiers: I think this note needs to go into applicability.

_rules/SC4-1-2-role-attribute-has-valid-value.md
kasperisager: See also https://github.com/auto-wcag/auto-wcag/pull/273#issuecomment-428152536.

PR #278, opened 2018-09-28
Update "visible" definition

annethyme:
- Removed reference to "on the page" from title since "page" is not well defined
- Removed circular reference
- Removed references to color contrast and visible focus, since that is accessibility requirements themselves, not part of the "visible" definition to be used for rules for other requirements
- Aiming for a non-ambiguous, testable definition of "visible"

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/261

PLEASE NOTE: @kasperisager is working on changes to the definition that will handle edge cases like tracking pixels and text placed outside of the parent element. We will update the pull request with these changes ASAP.

**When creating PR:**

**After creating PR:**

# How to Review And Approve

### DIFF
--- a/pages/algorithms/visible.md
+++ b/pages/algorithms/visible.md
@@ -1,0 +1,8 @@
+---
+title: Visible
+key: visible
+---
+
+Elements or text perceivable through sight.
+
+An element or text is visible if making it fully transparent would make any change in rendered pixels within the scrollable part of the viewport.
--- a/pages/algorithms/visible.md
+++ b/pages/algorithms/visible.md
@@ -5,4 +5,4 @@ key: visible
 
 Elements or text perceivable through sight.
 
-An element or text is visible if making it fully transparent would make any change in rendered pixels within the scrollable part of the viewport.
+An element, or text within an element, is considered visible if making the element or text fully transparent would cause a difference in the pixels rendered within any area of the viewport that can be reached via scrolling.
--- a/pages/algorithms/visible.md
+++ b/pages/algorithms/visible.md
@@ -5,6 +5,6 @@ key: visible
 
 Content perceivable through sight.
 
-Content is considered visible if making it fully transparent would result in a difference in the pixels rendered within any area of the viewport. This includes pixels that are reached by scrolling.
+Content is considered visible if making it fully transparent would result in a difference in the pixels rendered for any part of the document that is currently within the viewport or can be brought into the viewport via scrolling.
 
 [Content is defined in WCAG](https://www.w3.org/TR/WCAG21/#dfn-content).
--- a/pages/algorithms/visible.md
+++ b/pages/algorithms/visible.md
@@ -3,6 +3,8 @@ title: Visible
 key: visible
 ---
 
-Elements or text perceivable through sight.
+Content perceivable through sight.
 
-An element, or text within an element, is considered visible if making the element or text fully transparent would cause a difference in the pixels rendered within any area of the viewport that can be reached via scrolling.
+Content is considered visible, if making it fully transparent would result in a difference in the pixels rendered within any area of the viewport. This included pixels that are reached by scrolling.
+
+[Content is defined in WCAG](https://www.w3.org/TR/WCAG21/#dfn-content).
--- a/pages/algorithms/visible.md
+++ b/pages/algorithms/visible.md
@@ -5,6 +5,6 @@ key: visible
 
 Content perceivable through sight.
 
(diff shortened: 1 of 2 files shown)


annethyme:
@jkodu, @jkodu, @nitedog: This missing definition is somewhat of a blocker for our rule implementation.
 
@kasperisager is working on changes to the definition that will handle edge cases like tracking pixels and text placed outside of the parent element. We will update the pull request with these changes ASAP.

But maybe you guys could have an initial look at it already now and give feedback so we might be able to take a shortcut for the review cycles?

annethyme:
New edge case: Another element placed on top

annethyme:
Investigate how focusable e.g. buttons and images with/without accessibles names behave in AT when excluded from the accessibility tree.
@skotkjerra, could you maybe help out on this?

annethyme:
@carlosapaduarte and @nitedog, do you guys have time to review this one? 
Our next batch of rule implementations is on hold due to this missing definition.

pages/algorithms/visible.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Visible
+key: visible
+---
+
+Elements or text perceivable through sight.
+
+An element, or text within an element, is considered visible if making the element or text fully transparent would cause a difference in the pixels rendered within any area of the viewport that can be reached via scrolling.
```
nitedog: Not sure about "elements or text". Just sounds odd. Would this work?

[[
Content perceived visually

Content is considered visible, if making it fully transparent would result in a difference in the pixels rendered within any area of the viewport. This includes pixels that are reached by scrolling.
]]

[Content is defined in WCAG](https://www.w3.org/TR/WCAG21/#dfn-content)

carlosapaduarte: I agree with @nitedog that "elements or text" sounds odd. "Content" could be ambiguous being aligned with text, images, etc. for some readers, who might leave out widgets like buttons, sliders, etc. But if we link to the WCAG definition then it shouldn't be a problem. However, I don't think that the definition of Visible should be "Content perceived visually" because then you would have to define visually. "Content perceived through sight" sounds good to me.

I've also got some concerns about the definition. If I have content rendered with 99% transparency I cannot perceive it through sight. If I make that content fully transparent there will be a difference in the rendered pixels (and an automated tool will pick it up). But I don't agree the initial 99% transparent content is visible.

kasperisager: I do agree with using "content" over "elements or text". As for how to go about handling edge cases such as content that is 99% transparent I think there's more value to be gained from a having a definition that caters to the far majority of use cases. If authors make content 99% transparent and it's flagged by an Auto-WCAG rule because the rule considers the content "visible", I honestly think it's fair to require that authors make sure that the content is 100% transparent if they did not intend for it to be visible. That seems like a win-win in any case.

Otherwise, our definition would likely have to include something about contrast requirements for the pixels that would be rendered and the pixels that are actually rendered, which just seems overly complicated.

carlosapaduarte: I hadn't looked at it from that perspective. I agree with you that it's fair to require authors to make content 100% transparent if they want to hide it with transparency. I drop my concern about the definition 👍

jkodu:
@annethyme - could you address the review comments.

annethyme:
@carlosapaduarte and @nitedog, I have updated according to your comments. Please review again.

pages/algorithms/visible.md
```diff
@@ -3,6 +3,8 @@ title: Visible
 key: visible
 ---
 
-Elements or text perceivable through sight.
+Content perceivable through sight.
 
-An element, or text within an element, is considered visible if making the element or text fully transparent would cause a difference in the pixels rendered within any area of the viewport that can be reached via scrolling.
+Content is considered visible, if making it fully transparent would result in a difference in the pixels rendered within any area of the viewport. This included pixels that are reached by scrolling.
```
kasperisager: `This included` -> `This includes`

kasperisager: The comma after `visible` seems misplaced.

pages/algorithms/visible.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Visible
+key: visible
+---
+
+Content perceivable through sight.
+
+Content is considered visible if making it fully transparent would result in a difference in the pixels rendered within any area of the viewport. This includes pixels that are reached by scrolling.
```
kasperisager: The last sentence is not super clear. I'd suggest something along the lines of:

> [...] would result in a difference in the pixels rendered for any part of the document that is currently within the viewport or can be brought into the viewport via scrolling.

Issue #280, opened 2018-09-28
Missing "document context" definition referenced from SC4-1-1-unique-id


annethyme:
In https://auto-wcag.github.io/auto-wcag/rules/SC4-1-1-unique-id.html we link to #document-context that does not exist.

annethyme:
Kasper has been working on a definition like this for our internal version of https://auto-wcag.github.io/auto-wcag/rules/SC4-1-1-unique-id.html:  

> The value of the attribute is unique across all other [`id`](#https://www.w3.org/TR/html/dom.html#element-attrdef-global-id) attributes specified on HTML or SVG elements that are [descendants](#https://www.w3.org/TR/dom/#concept-tree-descendant) of the [root](#https://www.w3.org/TR/dom/#concept-tree-root) element of the element on which the applicable [`id`](#https://www.w3.org/TR/html/dom.html#element-attrdef-global-id) attribute is specified.

(still working on the wording)

Could we use this instead of the "document context" definition?

PR #285, opened 2018-10-03
Update SC4-1-1-unique-id with missing information

annethyme:
The rule was referencing a definition of "document context" that did not exist. Now it has been spelled out in the rule itself.

Closes issue: #280 

**When creating PR:**

**After creating PR:**
- [X] Add yourself (and co-authors) as "Assignees" for PR
- [X] Add label to indicate if it's a `Rule`, `Definition` or `Chore` (more to the administrative side)
- [X] Add relevant project (e.g. "Q3 2018 Status") to PR

# How to Review And Approve

### DIFF
--- a/_rules/SC4-1-1-unique-id.md
+++ b/_rules/SC4-1-1-unique-id.md
@@ -22,7 +22,7 @@ authors:
 
 Any `id` attribute specified on an HTML or SVG element.
 
-Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
+**Note:** Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
 ### Expectation
 
--- a/_rules/SC4-1-1-unique-id.md
+++ b/_rules/SC4-1-1-unique-id.md
@@ -22,7 +22,7 @@ authors:
 
 Any `id` attribute specified on an HTML or SVG element.
 
-**Note:** Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
+Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
 ### Expectation
 
--- a/_rules/SC4-1-1-unique-id.md
+++ b/_rules/SC4-1-1-unique-id.md
@@ -20,13 +20,13 @@ authors:
 
 ### Applicability
 
-Any element that has an `id` attribute.
+Any [`id`][id] attribute specified on an HTML or SVG element.
 
 **Note:** Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
 ### Expectation
 
-Each test target has an `id` attribute value that is unique within the [document context](#document-context) of the element.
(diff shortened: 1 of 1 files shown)


_rules/SC4-1-1-unique-id.md
```diff
@@ -20,13 +20,13 @@ authors:
...
 
-Any element that has an `id` attribute.
+Any [`id`][id] attribute specified on an HTML or SVG element.
 
 **Note:** Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
 ### Expectation
 
-Each test target has an `id` attribute value that is unique within the [document context](#document-context) of the element.
+The value of the attribute is unique across all other `id` attributes specified on HTML or SVG elements that exist within the same [tree](#https://www.w3.org/TR/dom/#concept-tree) as the element on which the applicable `id` attribute is specified. These are the elements that are [descendants](#https://www.w3.org/TR/dom/#concept-tree-descendant) of the [root](#https://www.w3.org/TR/dom/#concept-tree-root) of the element on which the applicable `id` attribute is specified.
```
carlosapaduarte: It became a bit convoluted. What about
The value of the attribute is unique across all other `id` attributes specified on HTML or SVG elements that exist within the same [tree](#https://www.w3.org/TR/dom/#concept-tree) as the element with the target attribute.

I'm not sure the second sentence is required. It seems to repeat the same information.

Another note: in the test cases, the expression "document context" is still used

annethyme:
Now using well-defined terms from the DOM 4.1 spec (working draft) for the Expectation.

_rules/SC4-1-1-unique-id.md
```diff
@@ -20,13 +20,13 @@ authors:
 
 ### Applicability
 
-Any element that has an `id` attribute.
+Any `id` attribute specified on an HTML or SVG element.
 
 **Note:** Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
```
nitedog: I think this note should be part of the applicability.

nitedog:
@WilcoFiers I don't understand your comment. You say: "elements can't be reliably referenced if the ID isn't unique" and this rule checks for unique IDs. What would you want tested otherwise?

WilcoFiers:
@nitedog Not having a reliable reference to an element is a problem with the element, not with the attribute. That's why I'm not in favour of this change.

jkodu:
@annethyme - could you do the needy here.

annethyme:
@WilcoFiers, following your argument, the problem will ALWAYS be with the element, if an attribute on it is somehow faulty.
If we always target the elements, we however see cases where we end up testing several attributes on the same element in a rule as a single test target, but when outputting the result, we can only say which element the issue is on, not point to which attribute on the element has the issue. 
For this reason, I would like to see us targeting the most specific unit possible in these cases. 
And for consistency reasons, I suggest we use the narrowest possible target throughout our repository. If what we are testing is the value of an attribute, target the attribute.

_rules/SC4-1-1-unique-id.md
```diff
@@ -20,13 +20,13 @@ authors:
 
 ### Applicability
 
-Any element that has an `id` attribute.
+Any `id` attribute specified on an HTML or SVG element.
 
-**Note:** Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
+Elements that are neither [exposed to assistive technologies](#exposed-to-assistive-technologies) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
```
nitedog: Why was the "Note" identifier removed? Seems appropriate to keep this as a note because it does not define the applicability (it only clarifies the meaning).

annethyme: Because you said so (https://github.com/auto-wcag/auto-wcag/pull/285#discussion_r223695439), and I always trust you :)
Now you can decide if I need to trust past-Shadi or present-Shadi ;)

nitedog: hahahaha - never trust any-shadi!!!

I looked up my past comment. It looked to me like a requirement because it linked to nifty definitions. But reading it more closely, I think it is a clarification and not a requirement. I'm sorry for the confusion.

annethyme: **Note:** identifier is back again.

annethyme:
@WilcoFiers, could you have a look at this one again and see if you can agree now due to my amazing argumentation? :)

PR #286, opened 2018-10-04
chore: front matter for rule (video has captions)

jkodu:
Fixes, yml front matter, to enable page to be rightly parsed for generation.

image

Closes issue: NA


Brynanders:
Guess you do not ned the review anymore @jkodu - look good anyway.

PR #287, opened 2018-10-05
SC2-1-2-no-keyboard-trap

annethyme:
Created new rule

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/260

**When creating PR:**

**After creating PR:**

# How to Review And Approve

### DIFF
--- a/_rules/SC2-1-2-no-keyboard-trap.md
+++ b/_rules/SC2-1-2-no-keyboard-trap.md
@@ -1,0 +1,239 @@
+---
+name: No keyboard trap
+rule_type: composite
+description: |
+  This rule checks if it is possible to use standard keyboard navigation to navigate through all content on a web page without becoming trapped in any element, or that it is possible to use non-standard keyboard navigation to navigate through content where focus is trapped when using standard ways of keyboard navigation.
+
+success_criterion:
+- 2.1.2 # No Keyboard Trap (A)
+
+atomic_rules:
+- [No keyboard trap standard navigation](SC2-1-2-no-keyboard-trap-standard-navigation)
+- [No keyboard trap non-standard navigation](SC2-1-2-no-keyboard-trap-non-standard-navigation)
+
+authors:
+- Anne Thyme Nørregaard
+---
+
+## Aggregation Definition
+
+### Applicability
+
+The rule applies to any HTML or SVG element on a web page that is [focusable](#focusable).
+
+Note: The WCAG 2.0 success criterion 2.1.2 applies to all content, not just HTML or SVG, where focus can be moved to through keyboard navigation.
+
+### Expectation
+
+For each test target, the outcome of one of the following rules is "passed":
+
+- [No keyboard trap standard navigation](SC2-1-2-no-keyboard-trap-standard-navigation)
+- [No keyboard trap non-standard navigation](SC2-1-2-no-keyboard-trap-non-standard-navigation)
+
+## Assumptions
+
+_There are currently no assumptions._
+
+## Accessibility support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 2.1.2: No Keyboard Trap](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html)
+- [G21: Ensuring that users are not trapped in content](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G21)
+
+## Test Cases
+
+### Passed
+
+#### Passed example 1
+
+No trap for keyboard navigation.
+
+```html
+<a href ="#">Link 1</a>
+<button class="target">Button1</button>
+```
+
+#### Passed example 2
+
+Using `tabindex="1"`.
+
+```html
+<div tabindex=“1”>Text</div>
+```
+
+#### Passed example 3
+
+Using `tabindex="-1"`.
+
+```html
+<div tabindex=“-1”>Text</div>
+```
+
+#### Passed example 4
+
+Keyboard trap with help information in a paragraph before, and where the method advised works.
+
+```html
+<script>
+var trapOn = false ;
+</script>
+
+<p>Press the M-key to Exit</p>
+<a id="link1" href="#">Link 1</a>
+<button id="btn1" onblur="(function(e){trapOn=true; document.getElementById('btn2').focus();})(event)">Button 1</button>
+<button id="btn2" class="target" onkeydown="(function(e){ if (e.keyCode === 77){trapOn=false;document.getElementById('link2').focus();}})(event)" onblur="(function(e){ if(trapOn){document.getElementById('btn1').focus();}})(event)">Button 2</button>
+<a id="link2" href="#">Link 2</a>
+```
+
+#### Passed example 5
+
+Keyboard trap with help information within the trap, and where the method advised works.
+
+```html
+<script>
+var trapOn = false ;
+</script>
+
+<a id="link1" href="#">Link 1</a>
+<button id="btn1" onblur="(function(e){trapOn=true; document.getElementById('btn2').focus();})(event)">Button 1</button>
+<p>Press the M-key to Exit</p>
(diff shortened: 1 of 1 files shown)


_rules/SC2-1-2-no-keyboard-trap.md
```diff
@@ -0,0 +1,241 @@
+---
+name: No keyboard trap
+rule_type: composite
+description: |
+  This rule checks if it is possible to use standard keyboard navigation to navigate through all content on a web page without becoming trapped in any element, or that it is possible to use non-standard keyboard navigation to navigate through content where focus is trapped when using standard ways of keyboard navigation.
```
nitedog: Loooong sentence. How about (just an idea):

[[
This rule checks for keyboard traps. This includes use of standard keyboard navigation to navigate through all content without becoming trapped, and information about non-standard keyboard navigation when they are necessary.
]]

_rules/SC2-1-2-no-keyboard-trap.md
```diff
@@ -0,0 +1,241 @@
...
+- Geir Sindre Fossøy
+- Dagfinn Rømen
+- Anne Thyme Nørregaard
+---
+
+## Aggregation Definition
+
+### Applicability
+
+The rule applies to any HTML or SVG element on a web page that is [focusable](#focusable).
```
nitedog: Do we need "on a web page" in the sentence? I think several other rules leave that out.

_rules/SC2-1-2-no-keyboard-trap.md
```diff
@@ -0,0 +1,241 @@
...
+
+Note: The WCAG 2.0 success criterion 2.1.2 applies to all content, not just HTML or SVG, where focus can be moved to through keyboard navigation.
```
nitedog: What does that mean? Maybe you want to state explicitly "This rule is only applies to HTML and SVG. Thus, it is a partial checks for WCAG Success Criterion 2.1.2, which applies to all content."?

nitedog: Recommend dropping the "2.0" and just say "WCAG"

carlosapaduarte: Agree with @nitedog. The note, as it stands, makes me wonder if it is extending the applicability to non HTML or SVG elements. @nitedog suggestion is much clearer. But I'm not sure if removing 2.0 is a good idea (what if future versions of WCAG renumber the success criteria?)

_rules/SC2-1-2-no-keyboard-trap.md
```diff
@@ -0,0 +1,241 @@
...
+## Accessibility support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 2.1.2: No Keyboard Trap](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html)
+- [G21: Ensuring that users are not trapped in content](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G21)
+
+## Test Cases
```
carlosapaduarte: Probably not the right place to ask: do we need to have test cases in composite rules? Can't we just specify the outcome based on the outcome of the atomic rules (which is what the expectation does)?

_rules/SC2-1-2-no-keyboard-trap.md
DagfinnRomen: Good suggestion. However, the user should be able to use non-standard keyboard navigation, and not only be informed that it exists.

Updated suggestion:
[[
This rule checks for keyboard traps. This includes use of both standard and non-standard keyboard navigation to navigate through all content without becoming trapped.
]]

_rules/SC2-1-2-no-keyboard-trap.md
DagfinnRomen: This is based on the the two atomic rules. I suggest we keep it in the composite rule as well.

_rules/SC2-1-2-no-keyboard-trap.md
kasperisager: I'd question this as well. We do end up with a lot of duplicated test cases otherwise.

_rules/SC2-1-2-no-keyboard-trap.md
nitedog: works for me! Very succinct.

_rules/SC2-1-2-no-keyboard-trap.md
nitedog: I suggest removing it from the other two rules. It is an easy fix editorially but would avoid problems and limitations down the line.

_rules/SC2-1-2-no-keyboard-trap.md
nitedog: OK "WCAG 2.1" -> "WCAG 2"

_rules/SC2-1-2-no-keyboard-trap.md
nitedog: The spec says we need them. And I tend to agree. The combination of atomic rules could require completely different test cases for the composite rules. Maybe not in this case but generally they are important to have.

carlosapaduarte: It's still a combination of atomic rules. As I see it, the implementation of a composite requires running the atomic rules on the test case and combine the outcome of the atomic rules. I guess what I mean is that validation of a composite rule is validation that the logic expressed in the expectation was correctly implemented. It's not a validation of any HTML code.

geirsf: We were wondering about this as well. Maybe you should raise a new issue for this? We will keep them in for now, as they are in the current spec.

_rules/SC2-1-2-no-keyboard-trap.md
```diff
@@ -21,9 +22,9 @@ authors:
 
 ### Applicability
 
-The rule applies to any HTML or SVG element on a web page that is [focusable](#focusable).
+The rule applies to any HTML or SVG element that is [focusable](#focusable).
 
-Note: The WCAG 2.0 success criterion 2.1.2 applies to all content, not just HTML or SVG, where focus can be moved to through keyboard navigation.
+**Note**: This rule is only applies to HTML and SVG. Thus, it is a partial check for WCAG 2.0 success criterion 2.1.2, which applies to all content.
```
kasperisager: Given that this is true for _most_ of the rules we write, I'd question the value of this note. For the sake of consistency, it should either be left out or included in _all_ rules where this applies.

DagfinnRomen: We are missing somewhere to document the interpretation of the success criterion that says something about what the success criterion covers, what the test rule(s) cover(s), and whether the test rule(s) only cover(s) part of the success criterion.

For the time being, the applicability section of the test rule is the best place suited. We are in favour of having a separate document where this is described. 

For the sake of consistency, we agree that this should be included in all test rules where this applies.

EmmaJP: Grammar point - "This rule only applies..." or "This rule is only applicable..." 

What other content would WCAG apply to? In a webpage HTML elements like canvas or video or form element are often heavily Javascript related.

_rules/SC2-1-2-no-keyboard-trap.md
geirsf: @kasperisager see issue #297 . We will keep it in for now, until the changes in the ACT format are ready.

annethyme:
@DagfinnRomen, it seems that if you take out the note, approvals are secured for this rule :)

annethyme:
@ShadowBB, which particular concerns do you have about the atomic rules? We might as well get it out sooner rather than later :)

ShadowBB:
> 
> 
> @ShadowBB, which particular concerns do you have about the atomic rules? We might as well get it out sooner rather than later :)

Hmmm. Poor choice of words. "curious" would be better. I am curious about the atomic rules. :) Did I missed them? They are not in the code nor in the PR right?

_rules/SC2-1-2-no-keyboard-trap.md
geirsf: Agreed @EmmaJP

Issue #289, opened 2018-10-08
Exported test cases are incorrect


kasperisager:
There seems to be something wrong with the exported test cases found at https://auto-wcag.github.io/auto-wcag/auto-wcag-testcases/testcases.json. For example, this test fixture https://auto-wcag.github.io/auto-wcag/auto-wcag-testcases/assets/SC4-1-1-unique-id_inapplicable_example_1.html is _not_ "inapplicable example 1" as both the URL and JSON file indicate; it's actually "passed example 2".

kasperisager:
Also, test cases that are not HTML documents should not be exported as such, e.g. SVG documents.

jkodu:
@kasperisager, so the contents of the test case needs to be parsed to figure out of it is an svg, and you would want them to be `filename.svg` files?

kasperisager:
Exactly 👍 Otherwise, browsers will interpret the files as HTML documents and parse them as such.

kasperisager:
Case in point: https://auto-wcag.github.io/auto-wcag/auto-wcag-testcases/assets/SC2-4-2-page-has-title_inapplicable_example_1.html currently fails the associated rule as the document, whose root is an `<html>` element due to it being interpreted as HTML, does not contain a `<title>`.

jkodu:
@kasperisager - PR - https://github.com/auto-wcag/auto-wcag/pull/299

PR #296, opened 2018-10-09
Fix testcase generation

jkodu:
This PR does the below:
- Fixes the issue where certain test cases were not correctly mapped to test case type.
- Fixes already published rules to follow right format.

Closes issue: 
- https://github.com/auto-wcag/auto-wcag/issues/289


kasperisager:
Do note that the second part of #289 is still outstanding.

PR #300, opened 2018-10-10
Mark all SVG documents and add missing namespaces

kasperisager:
Just what it says on the tin.


kasperisager:
@jkodu Could you merge and rebuild?

jkodu:
The re-build is triggered via a `webhook`. I have merged this, so the re-build will be done automatically shortly.

PR #307, opened 2018-10-19
Clarification on "more than 20 hours" requirement for SC2-2-1+SC3-2-5-meta-refresh

annethyme:
Clarification on "more than 20 hours" requirement. Expectation and test cases updated accordingly.

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/290

# How to Review And Approve

### DIFF
--- a/_rules/SC2-2-1+SC3-2-5-meta-refresh.md
+++ b/_rules/SC2-2-1+SC3-2-5-meta-refresh.md
@@ -23,7 +23,7 @@ The rule applies to the first [valid](https://www.w3.org/TR/html/document-metada
 
 ### Expectation
 
-The `time` of the `content` attribute is 0 or bigger than 72000 (20 hours).
+The `time` of the `content` attribute is 0 or greater than 72000 (20 hours).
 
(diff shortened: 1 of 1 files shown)


_rules/SC2-2-1+SC3-2-5-meta-refresh.md
```diff
@@ -110,6 +110,16 @@ First `<meta http-equiv="refresh">` element is not valid, second one redirects a
...
+#### Passed example 4
+
+Redirects after exactly 20 hours.
+
+```html
+<head>
+  <meta data-rule-target http-equiv="refresh" content="72000; http://example.com" />
+</head>
+```
+
```
kasperisager: This is not a pass, but a failure, as the time is neither 0 nor greater than 72000.

annethyme: Haha, it was definitely supposed to be a failure. It was in the right section (Failed test cases), but I forgot to change the heading after copy/pasting it over.
Should be fixed now

Issue #301, opened 2018-10-13
aria-hidden with focusable content: Overlap with several rules


kasperisager:
Per https://github.com/auto-wcag/auto-wcag/issues/281 and https://github.com/auto-wcag/auto-wcag/pull/269#issuecomment-428519520, `SC4-1-2-aria-hidden-focus` currently overlaps with several rules:

- `SC3-3-2-form-field-has-name`
- `SC2-4-4-link-has-name`
- `SC4-1-2-button-has-name`

It is possible for an element to be applicable to and pass any of the above rules, and then fail `SC4-1-2-aria-hidden-focus` for the same reason the element was applicable to the other rules in the first place. Together, these rules are effectively saying that, sure, you must have accessible names for these types of elements if they're focusable, but the accessible names won't ever be announced to AT because, guess what, you haven't included the elements in the accessibility tree. That seems a little backwards.

I'm thinking it would make more sense for `SC4-1-2-aria-hidden-focus` to first flag elements that are focusable, but not included in the accessibility tree, and if authors then do decide to include the elements in the accessibility tree, the other rules should then kick in to ensure that the elements also have accessible names. 

I have similar concerns for these rules, which also apply to focusable elements that aren't included in the accessibility tree:

- `SC1-2-2-audio-captions`
- `SC1-3-5-autocomplete-valid`

annethyme:
Handled by #310. Please review.

annethyme:
Handled by #264. Please review.

Issue #281, opened 2018-10-01
SC4-1-2-aria-hidden-focus: Possible false positives


kasperisager:
For `SC4-1-2-aria-hidden-focus`, we require that subtrees that are marked as `aria-hidden=true` do not contain focusable descendants. It did however strike me that on at least one occasion, I've implemented an interaction for a component that didn't serve users of AT well and was therefore hidden using `aria-hidden=true`. An alternative to the interaction was still available so functionality wasn't lost. A case like this would however be a failure of `SC4-1-2-aria-hidden-focus` and I'm not entirely convinced that it should.

WilcoFiers:
I did a little more testing with this. Focusable items inside aria-hidden=true don't get announced, and their accessible name isn't computed, at least not in Chrome and Firefox. I've asked for someone to do this test on Windows as well, I'll update when I know more. I don't think there's anything in spec that suggests the accessible name of these should be computed either.

A focusable element without an accessible name is a pretty straight forward failure of 4.1.2 IMO.

kasperisager:
If that is indeed the case then I don't have any issues with this rule as-is 👍

WilcoFiers:
Yeah, got the results. JAWS seems pretty buggy around focusable links that aren't in the accessibility tree. NVDA seems surprisingly fine with it. I'd still say that, especially because this isn't spect out in any way, we can safely call this a fail. Feel free to close the issue if you agree.

PR #309, opened 2018-10-22
Remove "focusable" from button-has-name Appl.

annethyme:
Per @kasperisager's comment: 

> Per #281 and #269 (comment), SC4-1-2-aria-hidden-focus currently overlaps with several rules:
> 
> SC3-3-2-form-field-has-name
> SC2-4-4-link-has-name
> SC4-1-2-button-has-name
> It is possible for an element to be applicable to and pass any of the above rules, and then fail SC4-1-2-aria-hidden-focus for the same reason the element was applicable to the other rules in the first place. Together, these rules are effectively saying that, sure, you must have accessible names for these types of elements if they're focusable, but the accessible names won't ever be announced to AT because, guess what, you haven't included the elements in the accessibility tree. That seems a little backwards.
> 
> I'm thinking it would make more sense for SC4-1-2-aria-hidden-focus to first flag elements that are focusable, but not included in the accessibility tree, and if authors then do decide to include the elements in the accessibility tree, the other rules should then kick in to ensure that the elements also have accessible names.
> 
> I have similar concerns for these rules, which also apply to focusable elements that aren't included in the accessibility tree:
> 
> SC1-2-2-audio-captions
> SC1-3-5-autocomplete-valid 

Closes issue: #301

# How to Review And Approve

### DIFF
--- a/_rules/SC4-1-2-button-has-name.md
+++ b/_rules/SC4-1-2-button-has-name.md
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-The rule applies to elements that are [visible on the page](#visible-on-the-page), [included in the accessibility tree](#included-in-the-accessibility-tree), or [focusable](#focusable) with the [semantic role](#semantic-role) of `button`, except for `input` elements of `type="image"`.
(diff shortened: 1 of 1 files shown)


annethyme:
Duplicate

Issue #264, opened 2018-09-18
Discussion: button-name changes


jkodu:
`aria-hidden` button element, should be `inapplicable`.

WilcoFiers:
Taking it off the agenda. We have 19 items, need to cut some stuff off. This needs a proposal first.

annethyme:
To have buttons that are `aria-hidden` be inapplicable, we would have to change the Applicability to not include elements that are `visible`.
Is this something we want to do?

annethyme:
Same applies to links: https://auto-wcag.github.io/auto-wcag/rules/SC2-4-4-link-has-name.html

PR #310, opened 2018-10-22
Remove "focusable" and "is visible" from Applicability in several rules due to consistency concerns

annethyme:
Per @kasperisager's comment in #301: 

> Per #281 and #269 (comment), SC4-1-2-aria-hidden-focus currently overlaps with several rules:
> 
> SC3-3-2-form-field-has-name
> SC2-4-4-link-has-name
> SC4-1-2-button-has-name
> It is possible for an element to be applicable to and pass any of the above rules, and then fail SC4-1-2-aria-hidden-focus for the same reason the element was applicable to the other rules in the first place. Together, these rules are effectively saying that, sure, you must have accessible names for these types of elements if they're focusable, but the accessible names won't ever be announced to AT because, guess what, you haven't included the elements in the accessibility tree. That seems a little backwards.
> 
> I'm thinking it would make more sense for SC4-1-2-aria-hidden-focus to first flag elements that are focusable, but not included in the accessibility tree, and if authors then do decide to include the elements in the accessibility tree, the other rules should then kick in to ensure that the elements also have accessible names.
> 
> I have similar concerns for these rules, which also apply to focusable elements that aren't included in the accessibility tree:
> 
> SC1-2-2-audio-captions
> SC1-3-5-autocomplete-valid 

Closes issue: #301

Per @jkodu's and @annethyme's comments in #264:

@jkodu: 

> aria-hidden button element, should be inapplicable.

@annethyme: 

> To have buttons that are aria-hidden be inapplicable, we would have to change the Applicability to not include elements that are visible.
> Is this something we want to do?
> 
> Same applies to links: https://auto-wcag.github.io/auto-wcag/rules/SC2-4-4-link-has-name.html

Closes issue: #264

# How to Review And Approve


_rules/SC4-1-2-button-has-name.md
```diff
@@ -192,3 +184,11 @@ Inapplicable: role overridden to link for button element.
...
 <button role='link'>take me somewhere</button>
 ```
+
+#### Failed example 4
+
+`aria-hidden` buttons should be tested.
+
+```html
+<button aria-hidden="true"></button>
+```
```
kasperisager: This would not be applicable.

annethyme: Sure would...

ShadowBB:
Took me a while to form an opinion on this, it's rather complicated. But I think the rules should not depend on each other! If a rule is implemented without a different rule it should still make sense.

So I don't agree with Kasper. It does seem a little backwards but I think we should still fail a focusable element without a name, even if it isn't included in the accessibility tree. And I can imagine some AT still announce focusable elements even if they are not included in the accessibility tree.

_rules/SC2-4-4-link-has-name.md
```diff
@@ -216,16 +200,24 @@ Not exposed to assistive technologies.
...
-Non-visible link.
+Not included in the accessibility tree due to `visibility: hidden`.
 
 ```html
 <a href="http://www.w3.org/WAI" style="visibility: hidden;">Some text</a>
 ```
+
+#### Inapplicable example 10
+
+Not included in the accessibility tree due to `aria-hideden="true"`.
```
EmmaJP: typo in code 'aria-hidden="true"'

Issue #295, opened 2018-10-08
ARIA allowed attribute: Undefined behaviour for elements without roles


kasperisager:
It struck me that https://auto-wcag.github.io/auto-wcag/rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.html does not currently define what should happen if an element without a semantic role specifies an `aria-*` attribute. This should either be specified or the applicability limited to `aria-*` attributes specified on elements with semantic roles.

kasperisager:
This might be one for you, @annethyme, when you're back from vacation.

annethyme:
Does anyone (e.g. @WilcoFiers, @nitedog, @carlosapaduarte, @audreymaniez, @kasperisager) have any insight into which WAI-ARIA states and properties are allowed for elements without a semantic role? 

Is it all attributes listed in [WAI-ARIA Definitions of States and Properties (all aria-* attributes)](https://www.w3.org/TR/wai-aria/#state_prop_def)? 
Or is it only [WAI-ARIA 1.1 Global States and Properties](https://www.w3.org/TR/wai-aria-1.1/#global_states)?

audreymaniez:
@annethyme I think It's only the attributes listed in the second ressource you spotted (WAI-ARIA 1.1 Global States and Properties), as in [ARIA in HTML - Allowed ARIA roles, states and properties](https://www.w3.org/TR/html-aria/#aria-table) for the "any" case.

In the "Is it all attributes listed in WAI-ARIA Definitions of States and Properties (all aria-* attributes)?", you can see `aria-activedescendant` which can not be added to an element without a role (and so witout a semantic role).

PR #311, opened 2018-10-22
Update aria-allowed-attribute: No overlap with aria-valid-attribute + elements with no role

annethyme:
Limiting applicability to avoid overlap with #231 by changing applicability from "aria-* attributes" to "WAI-ARIA states or properties" (a well-defined term). Now it only takes attributes as input that has passed the other rule.

Handles issues raised in: #231

ADDED: Expectation updated to handle elements that does not have a semantic role.

Closes issue: #295

# How to Review And Approve

### DIFF
--- a/_rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.md
+++ b/_rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.md
@@ -4,7 +4,6 @@ description: |
   This rule checks that WAI-ARIA states or properties are allowed for the element they are specified on.
 
 success_criterion:
-- 4.1.1 # Parsing (A)
 - 4.1.2 # Name, Role, Value (A)
 
 test_aspects:
--- a/_rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.md
+++ b/_rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.md
@@ -48,7 +48,7 @@ The attribute is either an [inherited](https://www.w3.org/TR/wai-aria/#inherited
 
 #### Passed example 1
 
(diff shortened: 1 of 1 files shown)


_rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.md
```diff
@@ -19,7 +19,7 @@ authors:
 
 ### Applicability
 
-Any attribute that is specified on an HTML or SVG element where the attribute name starts with `aria-`.
+Any [WAI-ARIA state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
```
kasperisager: What about MathML?

annethyme: We haven't included it in any other rule so far. I suggest we limit to html/svg for now and then take an overall discussion on which rules should consider MathML and other techonologies.

kasperisager: That's not the case though; several rules already consider MathML implicitly as also mentioned in https://github.com/auto-wcag/auto-wcag/issues/312.

annethyme:
I agree with @ShadowBB and have removed reference to SC 4.1.1

_rules/SC4-1-1+SC4-1-2-aria-allowed-attribute.md
annethyme: I suggest we merge this one in as is and then fix ALL applicabilities to be consistent when it comes to inclusion/exclusion of HTML, SVG and MathML

PR #321, opened 2018-10-25
chore: deprecate outdated files

jkodu:
Remove outdated rules and drafts

Closes issue: NA


EmmaJP:
I agree with it would be good to keep out-dated work somewhere easy to find and reference, as I think most of it will need to be revisited at some time and move into the correct format. It would be a shame to lose the work/thinking that was done. An archive folder might be useful.

Appreciate some of the outdated stuff is stuff I was working on.

jkodu:
A decision has been made that each of the outdated files will be removed as and when they are moved to the new rule format.

PR #324, opened 2018-10-25
chore: adding filter sort capability for the rules table

jkodu:
This PR adds filtering and sorting capabilities for the rules table.

Screenshots:

image

Closes issue: NA


jkodu:
@nitedog - the filtering is global to all columns, which is provided as an input element at the top of the headers of the table.

image

Issue #328, opened 2018-10-29
New rule: SC4-1-1-unique-attrs


WilcoFiers:
---
name: attributes are not duplicated
rule_type: atomic
description: |
  This rule checks that HTML and SVG starting tags do not contain duplicated attributes.

success_criterion:
- 4.1.1 # Parsing

test_aspects:
- Source code

authors:
- Wilco Fiers
- Emma Pratt Richens
---

## Applicability

The rule applies to any [starting tag](https://www.w3.org/TR/html5/syntax.html#start-tags) in an HTML or SVG file.

**Note:** This rule cannot be tested on the DOM Tree because the browser ignores duplicated attributes.

## Expectation

For each test target, there are no duplicated [attributes](https://www.w3.org/TR/html5/syntax.html#elements-attributes).

## Assumptions

*There are currently no assumptions.*

## Accessibility support

*There are no major accessibility support issues known for this rule.*

## Background

- - [H94: Ensuring that elements do not contain duplicate attributes](https://www.w3.org/TR/WCAG20-TECHS/H94.html)

#### Passed example 1

No attributes are duplicated.

```html
<img src="image.jpg" alt="" />
```

#### Passed example 2

No attributes, therefore no attributes are duplicated.

```html
<br />
```

#### Passed example 3

Empty attributes, no attributes are duplicated.

```html
<input type="checkbox" disabled readonly />
```

#### Passed example 4

SVG, no attributes are duplicated.

```html
<svg>
  <line x1="0" y1="0" x2="200" y2="200" style="stroke-width:2" />
</svg>
```

#### Passed example 5

Script, no attributes are duplicated. HTML or SVG code within a script should be ignored.

```html
<script>
	var foo = '<img src="image.jpg" alt="" alt="image" />'
</script>
```

#### Failed example 1

At least one attribute is duplicated.

```html
<img src="image.jpg" alt="" alt="image" />
```

#### Failed example 2

Empty attributes, at least one attribute is duplicated.

```html
<input type="checkbox" disabled="disabled" disabled readonly />
```

#### Failed example 3

SVG, at least one attribute is duplicated.

```html
<svg>
  <line x1="0" y1="0" x1="200" y1="200" style="stroke-width:2" />
</svg>
```

#### Inapplicable example 1

Code is XML, and not an HTML or SVG document.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<earl:TestResult rdf:about="#result"></earl>
```

[...]

Issue #329, opened 2018-10-29
Remove outdated and redundant SC4-1-1-accesskey draft rule


EmmaJP:
https://github.com/auto-wcag/auto-wcag/blob/master/_drafts/outdated/SC4-1-1-accesskey.md is a redundant draft rule, as WCAG removed the failure technique that is was based on.

The file can be deleted.

EmmaJP:
Resolved by pull request #332

PR #332, opened 2018-10-30
Rule: SC4-1-1-unique-attrs

EmmaJP:
Rule that checks element attributes are not duplicated.

Closes issue: #328 and #329 

**When creating PR:**

**After creating PR:**

# How to Review And Approve


_rules/SC4-1-1-unique-attrs.md
```diff
@@ -0,0 +1,135 @@
...
+<svg>
+  <line x1="0" y1="0" x1="200" y1="200" style="stroke-width:2" />
+</svg>
+```
+
+### Inapplicable
+
+#### Inapplicable example 1
+
+Code is XML, and not an HTML or SVG document.
```
EmmaJP: Code is XML, and not HTML or SVG.

_rules/SC4-1-1-unique-attrs.md
```diff
@@ -0,0 +1,135 @@
...
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<earl:TestResult rdf:about="#result"></earl>
+```
+
+#### Inapplicable example 2
+
+Code is JavaScript, and not an HTML or SVG document.
```
EmmaJP: Code is Javascript, and not HTML or SVG.

_rules/SC4-1-1-unique-attrs.md
```diff
@@ -0,0 +1,135 @@
...
+#### Inapplicable example 1
+
+Code is XML, and not HTML or SVG.
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<earl:TestResult rdf:about="#result"></earl>
+```
+
+#### Inapplicable example 2
```
maryjom: What about hidden elements (e.g. aria-hidden or display:none)?  Should this rule also be inapplicable in those cases?

_rules/SC4-1-1-unique-attrs.md
```diff
@@ -0,0 +1,135 @@
...
+
+- - [H94: Ensuring that elements do not contain duplicate attributes](https://www.w3.org/TR/WCAG20-TECHS/H94.html)
+
+
+## Test Cases
+
+### Passed
+
+#### Passed example 1
+
```
maryjom: Do you need to have test cases repeated...one for each type of attribute for each element type that supports attributes?  What about WAI-ARIA attributes? Is WAI-ARIA out of scope for now?

_rules/SC4-1-1-unique-attrs.md
kasperisager: Given that the aspect tested is the source code, that's not something that can be accounted for.

_rules/SC4-1-1-unique-attrs.md
WilcoFiers: ARIA is irrelevant for this, since that only starts to apply once the DOM is build. This is about the HTML source, not the DOM. Can you clarify what you mean by "each type of attribute"? I suppose we could add attributes with single, double and without quotes into the mix, but that's the only difference I can think of. Can you elaborate what you'd like to see changed (if anything)?

_rules/SC4-1-1-unique-attrs.md
```diff
@@ -0,0 +1,135 @@
...
+authors:
+- Wilco Fiers
+- Emma Pratt Richens
+---
+
+
+## Applicability
+
+The rule applies to any [starting tag](https://www.w3.org/TR/html5/syntax.html#start-tags) in an HTML or SVG file.
```
kasperisager: While HTML and SVG _documents_ can certainly be stored as _files_ on a _file system_, there's also other ways to store/generate such documents. For that reason, I'd use the term "document" instead of "file".

_rules/SC4-1-1-unique-attrs.md
```diff
@@ -0,0 +1,135 @@
...
+
+**Note:** This rule cannot be tested on the DOM Tree because the browser ignores duplicated attributes.
```
kasperisager: Per the HTML syntax specification, if an attribute name that is already present on a start/end tag is encountered, the attribute is dropped (http://w3c.github.io/html/syntax.html#attribute-name-state). As such, only the first encountered attribute is used if multiple attributes with the same name are defined. That's not quite the same as _ignoring_ duplicated attributes if one regards the term "duplicate" as encompassing the entire set of things that are equal.

_rules/SC4-1-1-unique-attrs.md
dd8: Isn't ARIA relevant if there are duplicate ARIA attributes? Only one of the duplicates ends up in the DOM and A11y Tree

```
<input type='text' aria-label='First Name' aria-label='Last Name' />
```

dd8: And does this apply to all attributes? Or only those with an accessibility impact?

So this is definitely a problem:

```
<img alt='One' alt='Two' src='one.png'> <!-- only one alt attribute ends up in DOM -->
```

but what about this:

```
<img alt='One' alt='One' src='one.png'> <!-- only one alt='One' attribute ends up in DOM -->
```

or this:

```
<meta name='generator' content='Frontweaver 8.2' content='Hugo 0.7'>
```

Issue #330, opened 2018-10-29
Define "filename"


Brynanders:
(Needed for #263 )

## Filename

A filename is a text string that identifies an electronically stored file. It is commonly used as part of a URL path to identify one piece of content from another. For example the `src` attribute specifies a URL path of `src="/foo/bar.jpg?baz "` which contains the filename `bar.jpg`.

Brynanders:
@ShadowBB, @jkodu, @kasperisager can I get a review from you on this one?

PR #334, opened 2018-10-30
filename.md

Brynanders:
Closes issue: #330 
Needed for #263

### DIFF
--- a/pages/glossary/filename.md
+++ b/pages/glossary/filename.md
@@ -3,4 +3,4 @@ title: Filename
 key: filename
 ---
 
-A filename is a text string that identifies an electronically stored file. It is commonly used as part of a URL path to identify one piece of content from another. For example the `src` attribute specifies a URL path of `src="/foo/bar.jpg?baz "` which contains the filename `bar.jpg`.
+A filename is a text string that identifies an electronically stored file located at the end of a URL path excluding any query strings. For example the `src` attribute specifies a URL path of `src="/foo/bar.jpg?baz "` which contains the filename `bar.jpg`.
--- a/pages/glossary/filename.md
+++ b/pages/glossary/filename.md
@@ -3,4 +3,4 @@ title: Filename
 key: filename
 ---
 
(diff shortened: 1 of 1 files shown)


pages/glossary/filename.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Filename
+key: filename
+---
+
+A filename is a text string that identifies an electronically stored file. It is commonly used as part of a URL path to identify one piece of content from another. For example the `src` attribute specifies a URL path of `src="/foo/bar.jpg?baz "` which contains the filename `bar.jpg`.
```
kasperisager: This doesn't actually define how to determine the filename given a URL.

EmmaJP: What is the purpose of the definition?

pages/glossary/filename.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Filename
+key: filename
+---
+
+A filename is a text string that identifies an electronically stored file located at the end of a URL path excluding any query strings. For example the `src` attribute specifies a URL path of `src="/foo/bar.jpg?baz "` which contains the filename `bar.jpg`.
```
EmmaJP: I still think it would be good to differentiate between what a filename is, and where to find it in a URL. Perhaps:
"A filename is a text string that identifies an electronically stored file. In a URL it is located at the end of the path, after the last slash and before any query strings. For example ..." 
or something like that.

Rafalchar: I think it's a good idea. This will make the definition more clear.

ShadowBB: > I still think it would be good to differentiate between what a filename is, and where to find it in a URL. Perhaps:
> "A filename is a text string that identifies an electronically stored file. In a URL it is located at the end of the path, after the last slash and before any query strings. For example ..."
> or something like that.

Emma her suggestion solves the minor issue that this definition would only work in the context of this group. I would approve both implementations though.

PR #358, opened 2018-11-30
chore: security vulnerability

jkodu:
Update `rack` dependency, to fix security vulnerability.

Closes issue: NA


jkodu:
@annethyme has a concern about development PR's and the approval system. She will debrief further.

PR #361, opened 2018-12-01
chore: consume EARL reports and represent implementation data per rule

jkodu:
This PR does the below:
- Consume EARL reports of various test tools from https://github.com/w3c/earl/tree/master/earl-reports
- Use the data to populate `Rules` status and `Implementation` Status in the website.
- Remove `ruby` script that was previously used for implementation data generation
- Added prettier for developer experience (for just js files).

Screenshots:
image

image

Closes issue: NA


jkodu:
@kasperisager 

A choice was made to do this at client side for the time being, rather than one other `npm script` or `ruby` plugin to generate relevant data at build time.

Given that we are looking to move away from jekyll and perhaps use other static site generators, I would keep this as is.

jkodu:
@kasperisager - thoughts?

PR #378, opened 2018-12-19
Fix links on "Iframe has accessible name"

annethyme:
Links in Background were broken due to use of pipe character. 
While fixing the links, I updated them to link to WCAG 2.1 Understanding documents instead of WCAG 2.0. 

Closes issue: none, but fixes bug found on https://auto-wcag.github.io/auto-wcag/rules/SC4-1-2-iframe-has-name.html

# How to Review And Approve


annethyme:
There is a link to the Understanding document for success criterion 2.4.1 Bypass Blocks. This rule is a sufficient technique for SC 2.4.1, but doesn't list the SC as an accessibility requirements, since there is no fail-on-fail relationship between the rule and SC 2.4.1.
Should we then list the 2.4.1 Understanding document in Background at all, when it is not clear what the relation is? Or should we add some words explaining why it is there?

PR #384, opened 2019-01-03
fix:  update test case for rule "aria-hidden-focus"

jkodu:
This PR does the below:
- Update one of the test cases for the rule `aria-hidden-focus`.

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/346


_rules/SC4-1-2-aria-hidden-focus.md
```diff
@@ -84,6 +86,18 @@ Content made unfocusable through disabled.
...
 ```
 
+#### Passed example 5
+
+`aria-hidden` can't be reset once you've set it to true on an ancestor.
+
+```html
+<div aria-hidden="true">
+    <div aria-hidden="false">
+        <button>Some button</button>
```
WilcoFiers: Should have tabindex-1 on it.

WilcoFiers:
@annethyme If you have any comments please leave them before the end of the week, otherwise we'll merge the PR.

PR #386, opened 2019-01-03
fix: rule "HTML has lang" with webpage definition

jkodu:
Update rule with link to `webpage` definition.

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/363


WilcoFiers:
@annethyme @nitedog @carlosapaduarte could one of you please review this?

PR #396, opened 2019-01-04
chore: update name to wcag-act-rules-cg

jkodu:
A lot of refactor to rename group to WCAG ACT RULES CG

Closes issue: NA


jkodu:
@WilcoFiers 

I have updated all the URL references in the project to come `package.json` which will be served into jekyll site data.

https://github.com/auto-wcag/auto-wcag/pull/396/files#diff-1b5be57188c61865bcdc09338f10a4d3R19

So we will have to update these to new URL's when we are ready to merge to the new repo.

PR #400, opened 2019-01-09
New rule: Heading has non-empty accessible name (ffd0e9)

annethyme:
At least in some screen readers, heading elements that only contain breaks or separator characters will show up as empty headings, confusing the user experience. 

I personally think this is a violation of 1.3.1 Info and Relationships. 
Do others agree? 

UPDATE: Ready for final review, including editorials, correctness of test cases etc.

---
# How to Review And Approve


annethyme:
Check if the "separator characters" should in fact be space characters instead: https://www.w3.org/TR/html/infrastructure.html#space-characters

annethyme:
@WilcoFiers, 

1. 
a. Siteimprove already has a check similar to this and it has been decided to try to convert it to an ACT Rule since this is one of many things that clutters and confuses the user experience for screen reader users. When a page has empty headings it can be hard to get an overview of where things belong in the structure of the page, and it can be hard to judge if the heading has been created accidentally or if it's a real heading whose content is for some reason not accessible. 
b. Testing showed lots of differences between different screen readers and browsers, but they all treated native heading elements and role="heading" excactly the same. That's why the applicability is "with the semantic role of `heading`", since this covers both.

2. Will have a look at your proposal along with the results from further testing in different screen readers and browsers.

annethyme:
For @WilcoFiers' point 2: Changes implemented

EmmaJP:
I wasn't part of the conversation on this, but I would also consider break and separator characters on their own, with no other alphanumeric characters, to be an empty heading.

ShadowBB:
This rule sort of feels weird to me.

1. The assumption asks us whether or not this is a WCAG violation which is something these rules should answer, not ask right? I advice dropping the assumption and therefore boldy stating it is a WCAG violation according to this rule.
2. The accessibility support tells us that the rule should be dropped if accessibility support is improved instead of a rule being adopted when accessibilty is improved. This feels backwards but I don't really see a problem with it as long as implementers (and us) somehow are reminded to keep an eye on this rule and deprecate it if/when AT starts ignoring these empty headings.

I am inclined to approve this because the rule itself is solid but I would love for my above two concerns to be shortly discussed in the next AUTO-WCAG call before this rule is merged.

_rules/SC1-3-1-heading-name-does-not-only-contain-spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Expectation
+
+None of the target elements has an [accessible name](#accessible-name) that contains only spaces.
+
+**Note:** In the [Accessible Name and Description Computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) all carriage returns, newlines, tabs, and form-feeds are replaced with a single space.
+
+## Assumptions
+
+This rule assumes that having an element that unintentionally shows up programatically to a user of assistive technologies as a heading element, but is not shown visually as a heading on the page, is a violation of WCAG success criterion 1.3.1 Info and Relationships.
```
danistr: this should be 'programmatically'. Also at 'Background'

_rules/SC1-3-1-heading-name-does-not-only-contain-spaces.md
```diff
@@ -0,0 +1,155 @@
...
+```html
+<h2><br /></h2>
+```
+
+#### Failed example 2
+
+`h2` contains `&nbsp;`(no break space character) as only content.
+
+```html
+<h2>&nbsp;</h2>
```
WilcoFiers: I don't think this would fail, going by what is and isn't replaced with "space" characters. May need to clarify exactly which chars should fail.

_rules/SC1-3-1-heading-name-does-not-only-contain-spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Applicability
+
+This rule applies to any HTML element with the [semantic role](#semantic-role) of `heading` that is [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+### Expectation
+
+None of the target elements has an [accessible name](#accessible-name) that contains only spaces.
+
+**Note:** In the [Accessible Name and Description Computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) all carriage returns, newlines, tabs, and form-feeds are replaced with a single space.
```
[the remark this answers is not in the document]
annethyme: Great questions. The list of "carriage returns, newlines, tabs, and form-feeds" is taken directly from the accessible name calculation. It is not further specified.

After talking to Kasper, it seems that the accessibility issues uncovered in testing with different AT/browser combinations are not fully covered by the transformations done in the accessible name calculation. 
Decision from our side is to try to revert it to the first draft of the rule, that was looking for `br`, `wbr` and characters in the `seperator` unicode categories. Will make the change and send out for review again.

jkodu:
@annethyme - We did some AT on the test cases, to start with an implementation of this rule in a few weeks.

Please see results here - https://github.com/dequelabs/axe-core/issues/1419#issuecomment-470966037

Based on the outcome, looks like some test cases have to be changed.

annethyme:
@jkodu, as mentioned on our call previously, our test results for these items DID show up as headings in IE + JAWS: 

```html
<h2> </h2>
```
```html
<h2>&#32;</h2>
```
```html
<h2><span><br /></span></h2>
```
... Though I realise now, that the last test case in our testing was actually replaced with this one (not that it should do any difference, logically speaking):
```html
<h2><span>&nbsp;</span></h2>
```

The versions used were IE 11 and JAWS 18.
Did your team by any chance use other versions?

annethyme:
@jkodu, one additional comment: 
When testing, we found that there can be differences between which headings shows up depending on the strategy used to reach them, e.g.:
- reading through the whole page
- pulling up a list of headings
- using keyboard short cuts to reach first/next heading

jkodu:
Thanks @annethyme. Will verify and update soon.

jkodu:
@annethyme - as discussed over the call, the findings from AT testing remain the same when using IE 11 and JAWS 18 for my testing & therefore the above concern - https://github.com/auto-wcag/auto-wcag/pull/400#issuecomment-470984711 still exists.

Next steps:
- to do more testing on your end.

annethyme:
@WilcoFiers and @jkodu, our testing showed that in one AT/browser combination (Talkback/Chrome) ALL of the test cases we could come up with for this showed an empty heading when navigating using headings as the granularity level.
Based on this, I would suggest going back to the expectation we previously had: "Each target element has an accessible name that is not only whitespace."
Could you guys agree to this?
If you agree, I will change the rule ASAP, so ww can start implementing soon. (@kasperisager, FYI)

_rules/SC1-3-1-heading-name-does-not-only-contain-spaces.md
```diff
@@ -0,0 +1,190 @@
+---
```
jkodu: filename needs to change to have unique id as a part of it & lose the sc number

jkodu:
@annethyme @Jym77 @Brynanders 

Before changing the applicability to what is suggested here https://github.com/act-rules/act-rules.github.io/pull/400#issuecomment-499834542, we need the below:

- a list of all whitespace characters.
- for each whitespace character used in as a heading, check if AT ignores the same.

@annethyme I document you shared with the extensive testing cannot be accessed anymore.

_rules/SC1-3-1-heading-name-does-not-only-contain-spaces.md
annethyme: Fixed

annethyme:
@jkodu, putting in writing what we talked about on the call a few days ago:  

> Before changing the applicability to what is suggested here #400 (comment), we need the below:
> 
> a list of all whitespace characters.

See existing and much-used definition: https://act-rules.github.io/glossary/#whitespace

> for each whitespace character used in as a heading, check if AT ignores the same.

It does in our testing with Chrome / Talkback in all test cases when we navigate with the granularity set to "headings". For us, this is enough evidence that an applicability of "Each target element has an accessible name that is not only whitespace." is feasible, because *some* users will experience issues - with a big fat note in the Accessibility Support section that this is treated hugely inconsistent across user agents. 

> @annethyme also the document you shared with the extensive testing cannot be accessed anymore.

Re-shared.

_rules/heading-name-does-not-only-contain-spaces-ffd0e9.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Applicability
+
+This rule applies to any HTML element with the [semantic role](#semantic-role) of `heading` that is [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+### Expectation
+
+The target element either:
+- contains [text nodes](https://www.w3.org/TR/dom/#text) that do not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153), or 
+- does not contain any HTML `<br>` elements, or
```
ShadowBB: This should not be enough to pass the expectation on it's own.

This would (for example) pass "failed example 2"

ShadowBB: > 
> 
> This should not be enough to pass the expectation on it's own.
> 
> This would pass "failed example 2" 3, 4, 5, 6, 7, and 9.

jkodu:
@annethyme I have shared a google sheet with findings of tests run against various AT's.

Please go through it and that should help make a decision on moving this issue forward.

annethyme:
Waiting for access to @jkodu's testing document.

Meanwhile documenting the last proposal, that was to take the rule back to roughly how it was in the second iteration:

# Rule suggestion: Heading name does not only contain spaces

--
description: This rule checks that each heading does not have an accessible name that do not only consist of whitespace.
--

### Applicability

This rule applies to any HTML element with the [semantic role](#semantic-role) of `heading` that is [included in the accessibility tree].

### Expectation

None of the target elements has an [accessible name](#accessible-name) that only consists of [whitespace](#whitespace).

**Note:** In the [Accessible Name and Description Computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) all carriage returns, newlines, tabs, and form-feeds are replaced with a single space.

jkodu:
@jkodu's testing document was provided.

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,187 @@
...
+- Kasper Isager
+---
+
+## Applicability
+
+This rule applies to any HTML element with the [semantic role](#semantic-role) of `heading` that is [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+The [accessible name](#accessible-name), if there is one, for each of the target elements is not only [whitespace](#whitespace).
```
kasperisager: Suggestion:

> The [accessible name](#accessible-name) of the test target, if there is an accessible name, is not only [whitespace](#whitespace).

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+
+**Note:** In the [Accessible Name and Description Computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) all carriage returns, newlines, tabs, and form-feeds are replaced with a single space.
+
+## Assumptions
+
+This rule assumes that having an element that unintentionally shows up programmatically to a user of assistive technologies as a heading element, but is not shown visually as a heading on the page, is a violation of WCAG Success Criterion 1.3.1 Info and Relationships.
+
+## Accessibility Support
+
+Handling of headings containing only whitespace characters, carriage returns, newlines, tabs, and form-feeds varies between different assistive technologies and browsers. This means that even though the outcome of this rule is *failed*, users of certain assistive technology and browser combinations might not experience an issue.
```
annethyme: align with current expectation. Different what is considered an empty heading...

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Background
+
+- In some screen reader and browser combinations, headings containing only whitespace characters, carriage returns, newlines, tabs, and form-feeds will show up as empty headings, confusing the user experience. For a screen reader user it will be hard to get an overview of the heading structure of the page if "empty" headings are included in the heading structure. Indeed, it can be hard to know if the heading is just whitespace that has accidentally been marked up as a heading, or if it is an actual heading that for some reason doesn't have an accessible name. Since this is a case where the programmatically determinable structure of the page doesn't match the visual presentation, this is a violation of success criterion [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).
```
annethyme: Remove first sentence, since already covered by accessibility support section.

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Expectation
+
+The [accessible name](#accessible-name) of the test target, if there is an accessible name, is not only [whitespace](#whitespace).
+
+**Note:** In the [Accessible Name and Description Computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) all carriage returns, newlines, tabs, and form-feeds are replaced with a single space.
+
+## Assumptions
+
+This rule assumes that having an element that unintentionally shows up programmatically to a user of assistive technologies as a heading element, but is not shown visually as a heading on the page, is a violation of WCAG Success Criterion 1.3.1 Info and Relationships.
```
annethyme: Link to SC

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
annethyme: Capitalize Success Criterion

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+- In some screen reader and browser combinations, headings containing only whitespace characters, carriage returns, newlines, tabs, and form-feeds will show up as empty headings, confusing the user experience. For a screen reader user it will be hard to get an overview of the heading structure of the page if "empty" headings are included in the heading structure. Indeed, it can be hard to know if the heading is just whitespace that has accidentally been marked up as a heading, or if it is an actual heading that for some reason doesn't have an accessible name. Since this is a case where the programmatically determinable structure of the page doesn't match the visual presentation, this is a violation of success criterion [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).
+- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+`h2` element with no accessible name.
```
annethyme: Insert links in TC descriptions

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+### Failed
+
+#### Failed Example 1
+
+`h2` element contains `br` element as only content. This is translated into a single space in the [accessible name](#accessible-name) computation, which gives an [accessible name](#accessible-name) that is only whitespace.
+
+```html
+<h2><br /></h2>
```
WilcoFiers: This example doesn't fail the rule. The accessible name of this heading is an empty string (""), exactly the same as passed example 1. The accessible name is always trimmed space characters before its returned. For example, take a look at this:

```html
<element1 id="el1" aria-labelledby="el3" />
<element2 id="el2" aria-labelledby="el1" />
<element3 id="el3"> hello </element3>
```

From: https://www.w3.org/TR/accname-1.1/#ex-1-example-1-element1-id-el1-aria-labelledby-el3-element2-id-el2-aria-labelledby-el1-element3-id-el3-hello-element3
> element1's accessible name is "hello" because this is a first traversal of its aria-labelledby, leading to element3.

This isn't well documented by any means, but every implementation of the accessible name computation algorithm  I've researched does this. This example also clearly shows it. There are also others in the accName test cases https://www.w3.org/wiki/AccName_1.1_Testable_Statements

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+```html
+<h2>' </h2>
+```
+
+#### Passed Example 4
+
+`h2` element has an [accessible name](#accessible-name) through `aria-label` that is not only whitespace.
+
+```html
+<h2 aria-label="Orange harvesting season"> </h2>
```
WilcoFiers: This is not a great test case. Without aria-label this would still pass.

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+#### Passed Example 5
+
+`h2` element has an [accessible name](#accessible-name) through the `alt` attribute that is not only whitespace.
+
+```html
+<h2><img src="#" alt="Orange harvesting season"> </h2>
```
WilcoFiers: This is not a great test case. Without the img this would still pass.

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
```diff
@@ -0,0 +1,185 @@
...
+- Kasper Isager
+---
+
+## Applicability
+
+This rule applies to any HTML element with the [semantic role](#semantic-role) of `heading` that is [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+The [accessible name](#accessible-name) of the test target, if there is an accessible name, is not only [whitespace](#whitespace).
```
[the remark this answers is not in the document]
Jym77: I agree that an accessible name always exists (and is sometimes empty). This is a consequence of "Step 1: Initialize" of the [accessible name computation algorithm](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) which sets the `total accumulated text` to the empty string; and the conclusion of same algorithm that returns the `total accumulated text`.

_rules/heading-name-is-not-only-whitespace-ffd0e9.md
Jym77: I do not see which part of the algorithm is supposed to do that trimming 😞 
Especially, steps 2F.iii and 2F.iv (name from content) and 2G (text node) are just appending text, not trimming it of whitespace.

At the same time, I do see this happening in tests I do… 😕 

---
Also, the algorithm is supposed to return a flat string (newlines and such converted to spaces, adjacent spaces reduced to a single space) as per the first paragraph of https://www.w3.org/TR/accname-1.1/#mapping_additional_nd
> The output is a flat, unstructured string

but I do not see which part of the algorithm is flattening the output (except for step 2D which handle `title` and similar attributes. Notably, step 2G (text node) does not flatten the string, which is weird.

---
In any case, if accessible name is always trimmed of spaces, then it can never be only whitespace and this rule is pointless. Many other rules also have weird formulation of "having an accessible name which is not only whitespace" and need to be updated. I'll make an issue on that…

---
Running some tests, with headers whose content is various empty whitespace (`<h1>something</h1>`), is "fun", as always…
* When `something` is (nothing), ` ` (1 space), `  ` (2 spaces), (newline), `<br />`: both Chrome, Firefox and Edge give it an empty accessible name (actually, Edge gives no accessible name and Firefox gives an accessible name of `null` 🙄 ).
* When `something` is `&nbsp;`, Firefox and Edge handle it as the other cases, but Chrome gives it an accessible name of `" "` (which I'm not sure if it's a non breaking space or a normal space, my quick test shows a normal space but that might just be my copy/paste changing it…)
* When `something` is `&nbsp;` or `<br />`, both Chrome, Firefox and Edge render a large vertical space (the space needed for `h1`) but for the other cases, none of them render any extra space. Not really related to the current rule or the accessible name, but still "fun"…

Jym77: > In any case, if accessible name is always trimmed of spaces, then it can never be only whitespace and this rule is pointless. Many other rules also have weird formulation of "having an accessible name which is not only whitespace" and need to be updated. I'll make an issue on that…

See #976

Jym77:
# Proposal 1
- Change rule to "heading has non-empty accessible name"
- Add Accessibility Support that headings with empty accessible name might be ignored by AT and thus won't necessarily cause problem.
- Add usual caveat that "empty accessible name" is not consistent over UA (_e.g._ trimming non breaking spaces or not). <= that's already in def of accessible name.

_Problem_: we'll fail `<h1></h1>` (or `<h1> </h1>`) which seems to be consistently ignored by AT. I don't see that as such a big problem. I feel it's rather similar to failing 4.1.1 Parsing but UA is closing tags for you so it is not actually exposed to users. In other words, an empty heading is still a bad idea (thus we flag it), but it's actually so trivially bad that every UA is going to fix it for you (but you should still fix it yourself).

# Proposal 2
- Change rule to "heading does not have non-empty invisible accessible name" (so, accept empty accessible name but reject accessible name with only invisible characters, _e.g._ weird spaces or line break)
- Add usual caveat that "empty accessible name" is not consistent over UA (_e.g._ trimming non breaking spaces or not). <= that's already in def of accessible name.

_Problems_: 
- trickier to express;
- annoying to find good failed examples, _e.g._ `<h1>&nbsp;</h1>` has empty accessible name in some UA;
- we'll accept `<h1><br /></h1>` where trimming of newline in accessible name seems to happen rather consistently, but that is visually rendered with a "heading sized" gap that make it feels like 1.3.1 is failing. Not a huge deal (rule is not perfect for SC), but annoying given that it is precisely the kind of stuff we try to find with this rule.

Jym77:
For CG call:
The proposal in its current form is not viable. Proposition to get back on our feet are in https://github.com/act-rules/act-rules.github.io/pull/400#issuecomment-551060000 Other possibility is to drop this rule.

WilcoFiers:
@Jym77 In my opinion, proposal 1 isn't a failure of WCAG, so I wouldn't be in favour of that. Proposal 2 I have no objection with, but defining exactly what types of whitespace don't get ignored by assistive technologies is the problem why this rule got stoled in the first place.

Jym77:
> Proposal 2 I have no objection with, but defining exactly what types of whitespace don't get ignored by assistive technologies is the problem why this rule got stoled in the first place.

If input aspect is the accessibility tree, I think that might make it manageable.

Jym77:
Decision on Nov 14: proposal 1.

EmmaJP:
> Decision on Nov 14: proposal 1.

I think because of WCAG 2.4.6 as well as 1.3.1 ... an empty heading does not label or describe the following content, and if used for something visual/decorative a non-semantic element should be used.

Jym77:
Rule has gone through heavy simplification after CG call yesterday. Please review again 😄

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,146 @@
...
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `h1` element has a non-empty [accessible name][].
+
+```html
+<h1 aria-level="1">ACT rules</h1>
```
EmmaJP: I don't think the aria-level attribute is needed on an `<h1>` element. Like `role='heading'` it is redundant.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,146 @@
...
+<h1 aria-level="1" aria-labelledby="h-name"></h1>
+```
+
+#### Passed Example 4
+
+This `h1` element has an non-empty [accessible name][] given by the `alt` attribute of its content.
+
+```html
+<h1 aria-level="1"><img src="#" alt="ACT rules" /></h1>
+```
```
EmmaJP: As before, 'aria-level' not required.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,146 @@
...
+```html
+<h1 aria-level="1"></h1>
+```
+
+#### Failed Example 2
+
+This `div` element with a [semantic role][] of `heading` has an empty [accessible name][].
+
+```html
+<div aria-level="1"></div>
```
EmmaJP: This example does not have a semantic role of heading ... but apparently it should.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,146 @@
+---
+id: ffd0e9
+name: Heading has accessible name
+rule_type: atomic
+description: |
+  This rule checks that each heading has an accessible name.
+accessibility_requirements:
+  wcag20:1.3.1: # Info and Relationships (A)
```
EmmaJP: I think this simplified version would be relevant for 2.4.6 as well as 1.3.1, as a heading cannot describe something if it is empty.

_rules/heading-accessible-name-ffd0e9.md
Jym77: Yes, my bad 😖

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,153 @@
...
+```html
+<div role="heading" aria-level="1"></div>
+```
+
+#### Failed Example 3
+
+Because the `img` element is marked as [decorative][] through its [semantic role][] of `presentation`, it does not provides an [accessible name][] to the `h1` element. Hence the `h1` element has an empty [accessible name][].
+
+```html
+<h1><img src="#" alt="ACT rules" role="presentation" /></h1>
```
EmmaJP: Don't know if an example with image and empty alt attribute is needed as an additional failed example, but mentioning it just in case, as it's the same reason but not the same code. Could maybe add:

`<h2></h2>`

Jym77: I've also thought about it and decided against it. As you say, it's the same reason and we don't need to test all possible cases of a heading having an empty accessible name.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,153 @@
...
+rule_type: atomic
+description: |
+  This rule checks that each heading has an accessible name.
+accessibility_requirements:
+  wcag20:1.3.1: # Info and Relationships (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+  wcag20:2.4.6: # Headings and Labels (AA)
```
carlosapaduarte: I don't think this rule is testing this SC. You're just checking that there is an accessible name, not that it describes the topic or purpose of the section it heads.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,153 @@
...
+```html
+<h1>ACT rules</h1>
+```
+
+#### Passed Example 2
+
+This `div` element has a [semantic role][] of heading and a non-empty [accessible name][].
+
+```html
+<div role="heading" aria-level="1">ACT rules</div>
```
carlosapaduarte: The aria-level attribute is not required in an example for this rule.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,153 @@
...
+```html
+<h1></h1>
+```
+
+#### Failed Example 2
+
+This `div` element with a [semantic role][] of `heading` has an empty [accessible name][].
+
+```html
+<div role="heading" aria-level="1"></div>
```
carlosapaduarte: The aria-level attribute is not required in an example for this rule.

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+#### Failed Example 3
+
+Because the `img` element is marked as [decorative][] through its [semantic role][] of `presentation`, it does not provides an [accessible name][] to the `h1` element. Hence the `h1` element has an empty [accessible name][].
```
carlosapaduarte: provides > provide

_rules/heading-accessible-name-ffd0e9.md
Jym77: @EmmaJP's argument was that if a heading has empty name, it cannot describe its topic. Hence failing this rule (having an empty heading) also fails 2.4.6 (the heading is not descriptive). The fail-on-fail is enough to be listed as requirement. As always, **passing** this rule does not mean anything for the SC.

https://github.com/act-rules/act-rules.github.io/pull/400#discussion_r346930882

_rules/heading-accessible-name-ffd0e9.md
Jym77: I've been followimg @WilcoFiers's advice on this:
https://github.com/act-rules/act-rules.github.io/pull/400#discussion_r321177593

I think the idea is that passing examples shouldn't be failing the SC and that a `div role="heading">Foo</div>` does pass the rule but fails the SC for lack of full structure information (i.e. heading level).

_rules/heading-accessible-name-ffd0e9.md
Jym77: Less needed in that case (because it is a failed example), but probably better to keep for consistency (and emphasising the difference that makes the example pass/fail the rule by not having unrelated differences).

_rules/heading-accessible-name-ffd0e9.md
Jym77: Less needed in that case (because it is a failed example), but probably better to keep for consistency (and emphasising the difference that makes the example pass/fail the rule by not having unrelated differences).

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `h1` element has an empty [accessible name][].
+
+```html
+<h1></h1>
```
ShadowBB: I don't think I would fail this during an actual audit. I have now read the debate that transformed this rule and the accessibility support section. But do we know of even a single mayor AT-UA combination that doesn't simply ignore this? A single example would ease my mind and convince me to fail this in future audits.

I have the same reservations about failed example 2.

_rules/heading-accessible-name-ffd0e9.md
carlosapaduarte: That's fair enough

_rules/heading-accessible-name-ffd0e9.md
Jym77: @ShadowBB will you be present on the next CG call (Nov. 28) to discuss this?

_rules/heading-accessible-name-ffd0e9.md
EmmaJP: Yep, this rule could be an atomic part of a composite rule for 2.4.6

_rules/heading-accessible-name-ffd0e9.md
EmmaJP: What is different about failed example 3 that you don't have reservations about it? And if there are no other failed examples, should there be a rule?

What if failed example 1 were:
`<h1><span style="opacity: 0;" aria-hidden="true">Hidden content</span></h1>`

_rules/heading-accessible-name-ffd0e9.md
nitedog: Unless use of  inside headings changes the rendering (ie. not ignored by UAs/ATs anymore but actually rendered as empty heading)?

_rules/heading-accessible-name-ffd0e9.md
DagfinnRomen: In the test procedure of H42 Using h1-h6 to identify headings, the test procedure step 2 says to "check that heading markup is not used when content is not a heading."

https://www.w3.org/WAI/WCAG21/Techniques/html/H42

Jym77:
@carlosapaduarte @DagfinnRomen @EmmaJP:
I've updated a bit the Accessibility Support and Background, with link to H42 (as pointed by @DagfinnRomen).
I've changed a bit the Failed Examples to make them into actual Failure of the SC (I think).

I think that is small enough changes that it doesn't require to dismiss your reviews. But I'm nonetheless giving you time to get another look if you want 😄 If nothing happens, I'll put the rule to Final Call at the end of the week…

Jym77:
Final Call jumps over the end-of-year break and will end on 
January 17, 2020.

_rules/heading-accessible-name-ffd0e9.md
ShadowBB: @EmmaJP Now that you mention it, I am not sure I would fail failed example 3 either. Nor the example you posted. Unless someone can show me an example of a AT-UA combination that doesn't simply ignore this.

@DagfinnRomen You could interpret that as "There is no content. So it doesn't fail this technique." But even if it failed the technique. It still would not be a problem since it could succeed a different technique and still comply with this SC.

This is all a rather minor gripe. If the consensus of the A11Y community seems to be that we should fail this then I can get on board and fail this, but it would really help me explain this to clients if I had a single example of a AT-UA combination that actually mentions this header. And I can't find one.

Jym77:

PR #401, opened 2019-01-09
SC-1-1-1+SC1-3-1-heading-element-with-image-content-has-an-accessible-name

annethyme:
Not so sure about this rule. Will go and do some testing with screen readers before proceeding.

---
# How to Review And Approve

### DIFF
--- a/SC-1-1-1+SC1-3-1-heading-element-with-image-content-has-an-accessible-name.md
+++ b/SC-1-1-1+SC1-3-1-heading-element-with-image-content-has-an-accessible-name.md
@@ -1,0 +1,98 @@
+---
+name: Heading element with image content has an accessible name
+rule_type: atomic
+description: |
+  This rule checks that headings that has images as their only content also has an accessible name
+
+success_criterion: 
+- 1.1.1 # Non-Text Content (A)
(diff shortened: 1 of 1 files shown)


SC-1-1-1+SC1-3-1-heading-element-with-image-content-has-an-accessible-name.md
```diff
@@ -0,0 +1,98 @@
...
+- does not have any non-seperator unicode characters as its content, 
+- does not have descendants that contain any non-seperator unicode characters as its content.
+
+### Expectation 1
+
+Each target element has an [accessible name](#accessible-name), or has a descendant that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an accessible name.
+
+## Assumptions
+
+_There are no assumptions for this rule._
```
WilcoFiers: I think you are making an assumption in this rule that headings elements with content are always intended as headings.

annethyme:
In further testing I couldn't find any use cases that weren't already covered by other rules (when my screen reader decided to actually look at the whole test page), so dropping this proposal again.

PR #408, opened 2019-01-17
Rule: Page has no orientation lock specified using CSS media queries (b33eff)

jkodu:
**Rule:** 
Page has no orientation lock specified using CSS media queries

**Description:** 
This rule checks that page content is not locked to any single display orientation using CSS media queries.

Closes issue: 
- https://github.com/auto-wcag/auto-wcag/issues/403

**When creating PR:**

**After creating PR:**

# How to Review And Approve


WilcoFiers:
The implementation we have today fails SC 1.3.4 for having a media query CSS lock at all. Instead of trying to test for a method to switch, we make the assumption that such a switch doesn't exist. I think that's a fairly reasonable assumption. @annethyme what are your thoughts about this?

_rules/Kj3fQk.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any `element` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree) of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s).
+
+### Expectation
+
+Each target element does not have any `css` applied, that locks the page content to a preferred orientation.
```
annethyme: "preferred orientation" is the term normally used about what the user prefers. 

I would suggest using "specific orientation" here instead.

_rules/8e8G5z.md
```diff
@@ -0,0 +1,210 @@
...
+authors:
+- Audrey Maniez
+- Jey Nandakumar
+---
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any `element` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree) of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s).
```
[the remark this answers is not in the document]
jkodu: You are right, page is a redundant addition. 
The thing I wanted to do emphasis was that this rule affect the page viewport, but I think this is straight forward.

_rules/8e8G5z.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Expectation
+
+Each target `element` can be viewed and operated, in the user preferred orientation, where the content can be made operable either seamlessly or by activating a control in the page.
```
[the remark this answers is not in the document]
jkodu: The applicability `included in accessibility tree` should cover this?

_rules/Kj3fQk.md
```diff
@@ -0,0 +1,151 @@
...
+authors:
+- Audrey Maniez
+- Jey Nandakumar
+---
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any `element` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree) of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s).
```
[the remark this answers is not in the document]
jkodu: The applicability can be for individual element as well, for example imagine a tetris game, where only the canvas on which the blocks are built are locked to a orientation, but the controls change location.

_rules/8e8G5z.md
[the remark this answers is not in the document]
annethyme: @jkodu, I was referring to "or by activating a control in the page." - does this control have to live up to any accessibility requirements?

_rules/css-orientation-lock-media-queries.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any `element` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+### Expectation
+
+Each target element does not have any `CSS` media queries applied, that locks the page content to a specific orientation.
```
carlosapaduarte: Mixing elements and page in the sentence makes it somewhat fuzzy. Given that the rule targets elements of the page, does it fail if I have an element of the page locked to a specific orientation, but the rest of the page is not locked?
My interpretation of the Success Criterion is that the locking restriction applies to the whole page and not to specific elements. 
The expectation is not clear because it reads "locks the page content". Does this equals to "locks the page" or to "locks some of the page's content"?

_rules/css-orientation-lock-media-queries.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Assumptions
+
+- The rule does not take consider existence of any control on the page that can handle or change the orientation on demand.
+
+- The rule assumes that a specific display orientation is not [essential](https://www.w3.org/TR/WCAG21/#dfn-essential), which is listed as a valid exception to [Success Criterion 1.3.4 Orientation]((https://www.w3.org/WAI/WCAG21/Understanding/orientation.html)). Examples where a particular display orientation may be essential are a bank check, a piano application, slides for a projector or television, or virtual reality content where binary display orientation is not applicable.
```
EmmaJP: So, while a pass is a pass, this rule could also return a fail for something that would not be a fail of the WCAG guideline. "Valid exceptions" do make testing a little tricky, and for this rule not passing might require a manual check warning or something similar.

audreymaniez:
Part of the problems @nitedog adressed might be resolved by rewriting that rule, I think. 

Actually we have : 

> name: Page has no orientation lock specified using CSS media queries

Which is incorrect for a conforming rule, as @nitedog says, it's not because its locked that is a failure. 
What if the rule is (something like) :

> name: Orientation lock specified using CSS media queries doesn't prevent the content to be operable in the user prefered orientation.

The rest of the rule (the description and the applicability) should be rewritten that way, but does that reformulation change something regarding the accuracy of the rule ?   or not all ? :-)

_rules/css-orientation-lock-media-queries.md
kasperisager: Yeah, I also think this assumption is a bit too restrictive and should really make it's way into the expectation, i.e. if an orientation lock _is_ used then it _must_ be essential.

_rules/css-orientation-lock-media-queries.md
jkodu: Note:

Expand on the expectation - move the essential assumption into here and add a note with examples where orientation is essential

_rules/css-orientation-lock-media-queries.md
```diff
@@ -0,0 +1,156 @@
...
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Test Procedure
+
+### Applicability
+
+The rule applies to any `element` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
```
jkodu: Note to self:

- remove included in accessibility tree

_rules/css-orientation-lock-media-queries.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+#### Failed example 2
+
+A page, where orientation is locked via `style` specified on the `head`.
+
+```html
+<html lang="en">
+  <head>
+    <!--
```
carlosapaduarte: You can remove this content since there is no external stylesheet in this example

_rules/css-orientation-lock-media-queries.md
```diff
@@ -0,0 +1,215 @@
...
+      }
+    </style>
+  </head>
+  <body>
+    Page Content
+  </body>
+</html>
+```
+
+#### Inapplicable example 3
```
carlosapaduarte: I don't think this one is needed. It is testing the same as example 4 which is thorough

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,198 @@
...
+- DOM Tree
+- CSS Styling
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any `element` that is [visible](#visible) and has CSS media query styles targeting the display orientation.
```
WilcoFiers: I don't think this is specific enough:

> and has CSS media query styles targeting the display orientation.

I suggest:

> and has a CSS `transform` style with a [rotate](https://drafts.csswg.org/css-transforms/#funcdef-transform-rotate) filter, set from a CSS media query with an [orientation](https://drafts.csswg.org/mediaqueries-3/#orientation) media feature (`horizontal` or `vertical`).

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,198 @@
...
+      @media (orientation: portrait) {
+        body {
+          transform: rotate(90deg);
+        }
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <!-- piano application -->
```
WilcoFiers: I hate to do this, because of the implications of this. But this isn't a proper test. Test cases have to actually pass or fail the rule. We need an actual test case of the orientation being essential, otherwise this rule is invalid.

That's kind of my concern with including the exception in the rule at all. We don't have any examples of it.

_rules/css-orientation-lock-media-queries-b33eff.md
ShadowBB: The "visible" link does not appear to be working.

_rules/css-orientation-lock-media-queries-b33eff.md
ShadowBB: We can create a simple example by having an image that is quite wide (wider than the height of most screens) and then simply stating in the text that for the full effect of the artwork to work, you should view it all at once. That should make the orientation essential.

jkodu: @WilcoFiers you are right. I cannot seem to come up with a passed example where we can emphasise that the orientation is essential. But having removed this example & the not in expectation, I believe this rule conveys that orientation if provided and discerned as lock will be a failure, in other cases we either flag as inapplicable or pass accordingly.

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: I ensured that there is a visible glossary/ definition & that this works.

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: @ShadowBB even having a large image as you suggested does not solve for `We need an actual test case of the orientation being essential`.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,175 @@
...
+- DOM Tree
+- CSS Styling
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any `element` that is [visible](#visible) and has a CSS `transform` style with a [rotate](https://drafts.csswg.org/css-transforms/#funcdef-transform-rotate) filter, set from a CSS media query with an [orientation](https://drafts.csswg.org/mediaqueries-3/#orientation) media feature (`horizontal` or `vertical`).
```
carlosapaduarte: `transform` should be introduced as a property, not a style
`rotate` should be introduced as a transform-function, not a filter
`orientation` values are `landscape` and `portrait`, not `horizontal` and `vertical`

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,175 @@
...
+    </main>
+  </body>
+</html>
+```
+
+### Failed
+
+#### Failed Example 1
+
+A page, where orientation is locked by applying `transform` style on `portrait` media query (coming from a relative stylesheet).
```
carlosapaduarte: style > property

_rules/css-orientation-lock-media-queries-b33eff.md
ShadowBB: > 
> 
> @ShadowBB even having a large image as you suggested does not solve for `We need an actual test case of the orientation being essential`.

I still would say that is possible by simply stating that the artist of the images intended it to be viewed all at once. Even if the image isn't particularly better if viewed all at once, the statement would make the orientation essential for "being able to verify that claim".

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,175 @@
...
+- Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any `element` that is [visible](#visible) and has a CSS `transform` property with a [rotate](https://drafts.csswg.org/css-transforms/#funcdef-transform-rotate) transform-function, set from a CSS media query with an [orientation](https://drafts.csswg.org/mediaqueries-3/#orientation) media feature (`landscape` or `portrait`).
+
+## Expectation
+
+Each target element does not restrict its view and operation to a single display orientation, unless a specific display orientation is [essential](https://www.w3.org/TR/WCAG21/#dfn-essential).
```
WilcoFiers: > Each target element does not restrict its view and operation to a single display orientation

This is ambiguous. That's not allowed in the ACT Rules Format. You're going to have to define these terms if you want to use them.

jkodu:
@carlosapaduarte @Jym77 as for the comment here - https://github.com/act-rules/act-rules.github.io/pull/408#discussion_r300367273

I am seeking some suggestions from you both. I have researched and read quite a bit on how best to define `view` and `operation` in this context & I am running out of ideas.

Any alternative takes on the expectation is much welcome.

jkodu:
** Note to all ** 

Having done a lot of reading on how best to describe the expectation, after reading this [comment](https://github.com/act-rules/act-rules.github.io/pull/408#discussion_r300367273), it made sense to be very specific. 

In this case it is best to describe what is an expected value on the CSS property that confines the given target element to a single display orientation.

It also made sense to remove the ambiguity of essential from this rule, to reduce the number of edge cases.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,172 @@
...
+
+```html
+<html lang="en">
+  <body>
+    I am a page with no styles
+  </body>
+</html>
+```
+
+#### Inapplicable Example 2
```
carlosapaduarte: For the purposes of this rule, there is no difference between this and the first inapplicable example

jkodu: I have modified this slightly. Good find though.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,172 @@
+---
+id: b33eff
+name: Page has no orientation lock specified using CSS media queries
+rule_type: atomic
+description: |
+  This rule checks that page content is not locked to any specific display orientation using CSS media queries, and the content is operable in all display orientations.
+accessibility_requirements:
+ wcag20:1.3.4: # Orientation
+  forConformance: true
+  failed: not satisfied
```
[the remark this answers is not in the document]
jkodu: @carlosapaduarte 

Based on this comment - https://github.com/act-rules/act-rules.github.io/pull/408#issuecomment-512799828

The rule is scoped as it reads without bringing in the ambiguity of `essential`. In that respect, I think even as an assumption adding `essential` will make this rule hard to interpret & hence leaving it as is would be better for the rule.

jkodu:
@carlosapaduarte after considering your above [comment](https://github.com/act-rules/act-rules.github.io/pull/408#pullrequestreview-266061540), it made sense to add the below assumption

> - The rule does not consider elements for which a particular display orientation is [essential](https://www.w3.org/TR/WCAG21/#dfn-essential).

For various reasons already discussed, there are a number of reasons we did not want to consider `essential` within the scope of this rule. Hope this clears your concerns & we can move forward.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,199 @@
...
+  </head>
+  <body>
+    <main>
+      Page Content
+    </main>
+  </body>
+</html>
+```
+
+#### Passed Example 2
```
carlosapaduarte: I've tried this on different browsers on iOS and I wasn't able to get the content right side up on portrait orientation. It was always flipped upside down. It's true that the content is not orientation locked (I can see it fine in landscape), but this is not a nice, accessible example.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,199 @@
...
+- DOM Tree
+- CSS Styling
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any element that is [visible](#visible) and has a CSS `transform` property with a [rotate](https://drafts.csswg.org/css-transforms/#funcdef-transform-rotate) transform-function, set from a CSS media query with an [orientation](https://drafts.csswg.org/mediaqueries-3/#orientation) media feature (`landscape` or `portrait`).
```
kasperisager: This does not account for 3D transformations: https://drafts.csswg.org/css-transforms-2/#transform-functions

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Expectation 1
+
+Each target element is not restricted to a single display orientation using CSS rotate transform-function, whose value is a multiple of 90 degrees but not a multiple of 180 degrees.
```
annethyme: Should "single display orientation" somehow be defined? I am not sure I get why it can be locked to 180 degrees (etc.), while not to 90 degrees (etc.).

_rules/css-orientation-lock-media-queries-b33eff.md
annethyme: As a non-CSS-pro, there is something about the wording (and probably styling with/without backticks) of this, that doesn't really give me a clue about what kind of code we are looking for: 
"has a CSS `transform` property with a [rotate](https://drafts.csswg.org/css-transforms/#funcdef-transform-rotate) transform-function, set from a CSS media query with an [orientation](https://drafts.csswg.org/mediaqueries-3/#orientation) media feature (`landscape` or `portrait`)."

I mean, I can get an idea from the examples, but it would be great also being able to make sense out of the applicability itself :)

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,199 @@
...
+    <main>
+      Page Content
+    </main>
+  </body>
+</html>
+```
+
+#### Passed Example 2
+
+A page, where CSS media query styles does not lock orientation. The CSS rotate transform-function value from the stylesheet is overridden to a value of 180 degrees.
```
annethyme: does not --> do not

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: @kasperisager the applicability specifies [rotate](https://drafts.csswg.org/css-transforms/#funcdef-transform-rotate) which only refer to **2D Transform Functions**.  So this rule does not cater to 3D transforms.

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: @annethyme I believe that given two orientations `landscape` and `portrait` as possible values, it is very straight forward to understand what single display orientation means.

As for why only multiples of 90deg, here is a scenario. Say you are looking at an application on your phone in `portrait` mode, then you flip the phone to `landscape` orientation, if there is now a lock specified, then the app will rotate by 90 and go back to only `portrait` mode. Where any value that is in multiples of 180 does not have actually lock the content. Hope that helps.

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: Ha ha, i re-read it many times now and it makes sense. We are simply suggesting look for a certain CSS with a specific property set.

kasperisager: That's exactly my point; it doesn't, and I absolutely think that it should.

kasperisager: Looking at this again, the rule really needs to look at the resulting transformation matrix (https://drafts.csswg.org/css-transforms/#transformation-matrix) instead of the value of the `transform` property. There are several reasons for this:

1. A transformation matrix can be passed explicitly using the `matrix()` function.
2. A 2D rotation (`rotate()`) is really just a 3D rotation (`rotate3d()`) with a fixed direction (`[0, 0, 1]`).
3. The value of the `transform` property isn't necessarily valid. Consider this:
  ```css
  { transform: rotate(90deg) doh; }
  ```

_rules/css-orientation-lock-media-queries-b33eff.md
WilcoFiers: Yeah, I question whether "is not restricted to a single display orientation" actually adds anything to the expectation. It seems to me the following would work just as well, and is more straight forward:

> None of the target elements has a CSS rotate transform-function, whose value is a multiple of 90 degrees but not a multiple of 180 degrees.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,209 @@
...
+- [rotateY transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateY)
+- [rotateZ transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateZ)
+- [matrix transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)
+- [matrix3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix3d)
+
+which are set as CSS [media query](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries) with an [orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Media_features) with a value of `landsapce` or `portrait`.
+
+## Expectation
+
+None of the target element has a [current transformation matrix](https://drafts.csswg.org/css-transforms/#current-transformation-matrix) that is restricted to a single display orientation.
```
jkodu: Todo: 
- define single display orientation
- add test cases for matrix transform

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,209 @@
...
+
+- [rotate transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)
+- [rotate3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate3d)
+- [rotateX transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateX)
+- [rotateY transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateY)
+- [rotateZ transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateZ)
+- [matrix transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)
+- [matrix3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix3d)
+
+which are set as CSS [media query](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries) with an [orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Media_features) with a value of `landsapce` or `portrait`.
```
kasperisager: > [...] which are set as CSS [media query](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries) [...]

This should be clarified as there are several uses of media queries in CSS. One is through `@media` rules as exemplified in the test cases, but `@import` rules can also be conditional:

```css
@import "./lock-orientation.css" (orientation: portrait);
```

The `<link>` element can also be made conditional:

```html
<link rel="stylesheet" media="(orientation: portrait)" href="./lock-orientation.css">
```

Perhaps something along the lines of:

> that are applied conditionally based on the [orientation](https://www.w3.org/TR/mediaqueries/#orientation) media feature.

No yet entirely unambiguous, but I think it's a start.

jkodu: This is true. Not emphasising just via media queries is important.

jkodu:
@kasperisager let me know if you are 👍 for applicability and expectation. I am updating examples in the mean time.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,209 @@
...
+
+The rule applies to any element that is [visible](#visible) and has a CSS [transform](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) property with any of the below [transform-functions](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function)
+
+- [rotate transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)
+- [rotate3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate3d)
+- [rotateX transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateX)
+- [rotateY transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateY)
+- [rotateZ transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateZ)
+- [matrix transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)
+- [matrix3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix3d)
```
kasperisager: I'd suggest listing these as:

- [`rotate()`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)
- [`rotate3d()`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate3d)
- ...

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,209 @@
...
+- DOM Tree
+- CSS Styling
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any element that is [visible](#visible) and has a CSS [transform](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) property with any of the below [transform-functions](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function)
```
kasperisager: "transform-functions" -> "transformation functions".

_rules/css-orientation-lock-media-queries-b33eff.md
kasperisager: Maybe also add a note explaining why these functions are relevant? Something along the lines of:

> These specific transformation functions are of interest to this rule as they have the potential to affect the rotation of a given element.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,209 @@
...
+
+- [rotate transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)
+- [rotate3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate3d)
+- [rotateX transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateX)
+- [rotateY transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateY)
+- [rotateZ transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateZ)
+- [matrix transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)
+- [matrix3d transform-function](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix3d)
+
+that are applied conditionally on the [orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Media_features) with a value of `landsapce` or `portrait`.
```
kasperisager: This bit can be left out:

> [...] with a value of `landsapce` or `portrait`.

...as these are the only valid values of the orientation media feature.

_rules/css-orientation-lock-media-queries-b33eff.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Expectation
+
+The target element has a [current transformation matrix](https://drafts.csswg.org/css-transforms/#current-transformation-matrix) that does not restrict the element to only `landspace` or `portrait` orientation.
```
kasperisager: I'd maybe consider a slightly more abstract formulation:

> The target element is not rotated around the Z axis at an angle corresponding to +-90 degrees relative to the position of the element prior to transformation.

One way to determine the rotation of an element would of course be to look at the current transformation matrix, but I do think that's an implementation detail. 

Maybe @Jym77 has some input as well?

_rules/css-orientation-lock-media-queries-b33eff.md
kasperisager: Come to think of it, we probably shouldn't assume that the CSS is valid, so I'll backpedal on my suggestion 🙈

kasperisager: `landsapce` should be `landscape` though.

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: I do not think the **note** is important, the fact that we are listing these functions and describing what we need in the expectation is plenty in my opinion.

Also the word `rotation` in your above suggestion will make it ambiguous.

_rules/css-orientation-lock-media-queries-b33eff.md
jkodu: Once again, adding `rotation` makes it ambiguous. And adding a number +-90deg is implementation specific I would argue (Also 180/ 360 is allowed rotation values), so again very implementation specific.

I prefer to leave this as is, because the linked definition of current transformation matrix, in terms looks into coordinate systems, which should detail all things that are axis related.

_rules/css-orientation-lock-media-queries-b33eff.md
kasperisager: > the fact that we are listing these functions and describing what we need in the expectation is plenty in my opinion.

For people who are intimately familiar with the CSS transforms specification, sure. For people who are not there's a missing link though. As for making the term "rotation" unambiguous, I'd just reference https://drafts.csswg.org/css-transforms-2/#Rotate3dDefined.

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,224 @@
...
+
+- [rotate](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)
+- [rotate3d](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate3d)
+- [rotateX](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateX)
+- [rotateY](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateY)
+- [rotateZ](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotateZ)
+- [matrix](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)
+- [matrix3d](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix3d)
+
+that are applied conditionally on the [orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Media_features) with a value of `landsapce` or `portrait`.
```
kasperisager: `landsapce` should be `landscape`.

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,224 @@
...
+
+**Note:** These specific [transformation functions](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function) are of interest to this rule as they have the potential to affect the [rotation](https://drafts.csswg.org/css-transforms-2/#Rotate3dDefined) of a given element.
+
+## Expectation
+
+The target element has a [current transformation matrix](https://drafts.csswg.org/css-transforms/#current-transformation-matrix) that does not restrict the element to either `landspace` or `portrait` orientation.
```
kasperisager: The expectation does not currently make sense. The current transformation matrix is  unrelated to the `landscape` and `portrait` orientations, which are allowed values of the `orientation` media feature. Once again, I'll bring up my suggested alternative:

> The target element is not rotated around the Z axis at an angle corresponding to +-90 degrees relative to the position of the element prior to transformation.

This could very well be put into a definition of what an "orientation lock" actually is. Regardless of what that definition ends up looking like, it has to be unambiguous. And also, the term "rotated" is not ambiguous; it is well defined in terms of a transformation matrix.

_rules/css-orientation-lock-media-queries-b33eff.md
kasperisager: I disagree on all accounts. Let's continue the discussion here: https://github.com/act-rules/act-rules.github.io/pull/408#discussion_r318907852

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -30,13 +30,13 @@ The rule applies to any element that is [visible](#visible) and has a CSS [trans
...
+that are applied conditionally on the [orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Media_features) with a value of `landscape` or `portrait`.
 
 **Note:** These specific [transformation functions](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function) are of interest to this rule as they have the potential to affect the [rotation](https://drafts.csswg.org/css-transforms-2/#Rotate3dDefined) of a given element.
 
 ## Expectation
 
-The target element has a [current transformation matrix](https://drafts.csswg.org/css-transforms/#current-transformation-matrix) that does not restrict the element to either `landspace` or `portrait` orientation.
+The target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative to the position of the element prior to transformation.
```
kasperisager: We're getting there! 🙌 There are however still some issues we need to iron out though. Consider this bit of CSS for starters:

```css
body {
  transform: rotate(90deg);
}

@media (orientation: portrait) {
  body {
    transform: rotate(90deg);
  }
}
```

This would currently fail the rule but I'm having a hard time justifying why it should. The conditional rule makes no difference in practice; the `<body>` element is always rotated 90 degrees regardless of the orientation. I'm thinking we need to consider the rotation of an element at a given orientation relative to the rotation of that element at the opposite orientation. For the example above, the `<body>` element is effectively rotated 0 degrees in portrait orientation relative to its rotation in landscape rotation and vice versa.

One possible solution to this could be to use two expectations:

> ## Expectation 1
>
> In portrait orientation, the target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative to the position of the element in landscape orientation.
>
> ## Expectation 2
>
> In landscape orientation, the target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative to the position of the element in portrait orientation.

It does seem a little excessive but might be enough to get some ideas flowing.

kasperisager: My suggestion above would also make it clear that testers actually need to look at both orientations to determine if there's an orientation lock or not, which makes perfect sense. Consider even a simple case such as this:

```css
@media (orientation: portrait) {
  body {
    transform: rotate(90deg);
  }
}
```

If this rule is only ever evaluated in landscape mode by a tester, the expectation would not fail.

jkodu: These are valid points & captures the idea of [essential](https://www.w3.org/WAI/WCAG21/Understanding/orientation.html#dfn-essential) very well.

I have made the changes to expectation & removed the assumption on `essential` as well as added a test case to showcase above case. Ta

kasperisager: My points have nothing to do with whether the orientation lock is essential or not though; it is still assumed that it isn't.

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,255 @@
...
+
+**Note:** These specific [transformation functions](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function) are of interest to this rule as they have the potential to affect the [rotation](https://drafts.csswg.org/css-transforms-2/#Rotate3dDefined) of a given element.
+
+## Expectation 1
+
+In `portrait` orientation, the target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative to the position of the element in `landscape` orientation.
+
+## Expectation 2
+
+In `landscape` orientation, the target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative to the position of the element in `portrait` orientation.
```
kasperisager: I'm only just now realising that these two expectations are equivalent 🙈 Could you sanity check this, @Jym77?

kasperisager: Rephrasing the expectation, nevermind!

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,255 @@
...
+*There are no major accessibility support issues known for this rule.*
+
+## Background
+
+- [Understanding Success Criterion 1.3.4: Orientation](https://www.w3.org/WAI/WCAG21/Understanding/orientation.html)
+- [CSS Transforms Module Level 1](https://www.w3.org/TR/css-transforms/#funcdef-transform-matrix)
+- [CSS Transforms Module Level 2](https://drafts.csswg.org/css-transforms-2)
+- [CSS3 Media Queries](https://www.w3.org/TR/css3-mediaqueries/)
+- [Managing screen orientation](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Managing_screen_orientation)
+- [MDN Orientation](https://www.w3.org/TR/css3-mediaqueries/#orientation)
```
kasperisager: ~~MDN~~.

jkodu:
Final call ends 24th September

jamespizzurro:
Do we have an example of a live website or webapp that explicitly rotates elements like this when there's an orientation change as opposed to letting the viewport do the work? I don't think I've ever observed this "solution" in the wild.

kasperisager:
@jamespizzurro This is one source: https://css-tricks.com/snippets/css/orientation-lock/. Not that I've seen it used in the wild either 🙈

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,258 @@
...
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'title-require'
```
WilcoFiers: Maybe we should just turn this off by default

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,258 @@
...
+
+- [rotate](https://www.w3.org/TR/css-transforms/#funcdef-transform-rotate)
+- [rotate3d](https://drafts.csswg.org/css-transforms-2/#funcdef-rotate3d)
+- [rotateX](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatex)
+- [rotateY](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatey)
+- [rotateZ](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatez)
+- [matrix](https://www.w3.org/TR/css-transforms/#funcdef-transform-matrix)
+- [matrix3d](https://drafts.csswg.org/css-transforms-2/#funcdef-matrix3d)
+
+that are applied conditionally on the [orientation](https://www.w3.org/TR/css3-mediaqueries/#orientation) [media feature](https://www.w3.org/TR/css3-mediaqueries/#media1) with a value of `landscape` or `portrait`.
```
WilcoFiers: Can you rephrase this, so that you don't have  a list in the middle of a sentence?

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,258 @@
...
+  - Audrey Maniez
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'title-require'
+---
+
+## Applicability
+
+The rule applies to any element that is [visible](#visible) and has a CSS [transform](https://www.w3.org/TR/css-transforms/#propdef-transform) property with any of the below [transformation functions](https://www.w3.org/TR/css-transforms/#transform-functions)
```
WilcoFiers: "any element" is pretty open ended. Should we limit this to any HTML and SVG element? Also definitely shouldn't only have HTML test cases, unless you limit this rule to HTML elements only (which you could also do).

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,258 @@
...
+**Note:** These specific [transformation functions](https://www.w3.org/TR/css-transforms/#transform-functions) are of interest to this rule as they have the potential to affect the [rotation](https://drafts.csswg.org/css-transforms-2/#Rotate3dDefined) of a given element.
+
+## Expectation
+
+The target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative from the position of the element in `landscape` orientation to the position of the element in `portrait` orientation, and vice versa.
+
+## Assumptions
+
+- The rule does not consider elements for which a particular display orientation is [essential](https://www.w3.org/TR/WCAG21/#dfn-essential).
+- The rule does not consider the existence of any control on the page that can change the orientation on demand.
```
WilcoFiers: Can you put in an explanation of what these assumptions mean for the rule?

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,258 @@
...
+
+The rule applies to any element that is [visible](#visible) and has a CSS [transform](https://www.w3.org/TR/css-transforms/#propdef-transform) property with any of the below [transformation functions](https://www.w3.org/TR/css-transforms/#transform-functions)
+
+- [rotate](https://www.w3.org/TR/css-transforms/#funcdef-transform-rotate)
+- [rotate3d](https://drafts.csswg.org/css-transforms-2/#funcdef-rotate3d)
+- [rotateX](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatex)
+- [rotateY](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatey)
+- [rotateZ](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatez)
+- [matrix](https://www.w3.org/TR/css-transforms/#funcdef-transform-matrix)
+- [matrix3d](https://drafts.csswg.org/css-transforms-2/#funcdef-matrix3d)
```
WilcoFiers: Most of these properties don't have tests for them. Since we're explicitly naming them, we should have tests for them.

_rules/css-restrict-orientation-b33eff.md
kasperisager: I'm thinking that limiting this to HTML would make sense.

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,263 @@
...
+  - DOM Tree
+  - CSS Styling
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that is [visible](#visible) and has a CSS [transform](https://www.w3.org/TR/css-transforms/#propdef-transform) property that are applied conditionally on the [orientation](https://www.w3.org/TR/css3-mediaqueries/#orientation) [media feature](https://www.w3.org/TR/css3-mediaqueries/#media1) with a value of `landscape` or `portrait`, where the CSS [transform](https://www.w3.org/TR/css-transforms/#propdef-transform) property has any of the below [transformation functions](https://www.w3.org/TR/css-transforms/#transform-functions):
```
WilcoFiers: "that are applied" -> "that is applied"

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,263 @@
...
+		<main>
+			Page Content
+		</main>
+	</body>
+</html>
+```
+
+#### Passed Example 3
+
+A page where CSS [transform](https://www.w3.org/TR/css-transforms/#propdef-transform) property has [matrix](https://www.w3.org/TR/css-transforms/#funcdef-transform-matrix) [transform function](https://www.w3.org/TR/css-transforms/#transform-functions) conditionally applied on the [orientation](https://www.w3.org/TR/css3-mediaqueries/#orientation) [media feature](https://www.w3.org/TR/css3-mediaqueries/#media1) which does not restricts the element to either `portrait` or `landscape` orientation.
```
WilcoFiers: "does not restricts" -> "does not restrict"

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,263 @@
...
+
+- [rotate](https://www.w3.org/TR/css-transforms/#funcdef-transform-rotate)
+- [rotate3d](https://drafts.csswg.org/css-transforms-2/#funcdef-rotate3d)
+- [rotateZ](https://drafts.csswg.org/css-transforms-2/#funcdef-rotatez)
+- [matrix](https://www.w3.org/TR/css-transforms/#funcdef-transform-matrix)
+- [matrix3d](https://drafts.csswg.org/css-transforms-2/#funcdef-matrix3d)
+
+**Note:** These specific [transformation functions](https://www.w3.org/TR/css-transforms/#transform-functions) are of interest to this rule as they have the potential to affect the [rotation](https://drafts.csswg.org/css-transforms-2/#Rotate3dDefined) of a given element.
```
WilcoFiers: Suggest adding another note:

> **Note:** The `rotate3d`, `rotateZ` and `matrix3d` are currently part of a W3C Editor's Draft.

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,265 @@
...
+		Page Content
+	</body>
+</html>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+A page where there are CSS styles.
```
ShadowBB: Missing a "no".

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -0,0 +1,265 @@
...
+			}
+		</style>
+	</head>
+	<body style="display:none;">
+		<main>
+			Page Content
+		</main>
+	</body>
+</html>
+```
```
ShadowBB: I am missing an inapplicable example where the transform property is used to rotate the page 90 degrees but not based on the current landscape or portrait orientation.

PR #410, opened 2019-01-18
fix: applicability for unique id rule

jkodu:
Update applicability

Closes issue: 
https://github.com/auto-wcag/auto-wcag/issues/349

**When creating PR:**

**After creating PR:**

# How to Review And Approve

### DIFF
--- a/_rules/SC4-1-1-unique-id.md
+++ b/_rules/SC4-1-1-unique-id.md
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-Any [non-empty](#non-empty) `id` attribute specified on an HTML or SVG element.
+Any `id` attribute which is not an empty string, specified on an HTML or SVG element.
 
 **Note:** Elements that are neither [included in the accessibility tree](#included-in-the-accessibility-tree) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
--- a/_rules/SC4-1-1-unique-id.md
+++ b/_rules/SC4-1-1-unique-id.md
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-Any `id` attribute which is not an empty string, specified on an HTML or SVG element.
+Any `id` attribute which is not the empty string (""), specified on an HTML or SVG element.
 
 **Note:** Elements that are neither [included in the accessibility tree](#included-in-the-accessibility-tree) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
--- a/_rules/SC4-1-1-unique-id.md
+++ b/_rules/SC4-1-1-unique-id.md
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-Any `id` attribute specified on an HTML or SVG element.
+Any [non-empty](#non-empty) `id` attribute specified on an HTML or SVG element.
 
 **Note:** Elements that are neither [included in the accessibility tree](#included-in-the-accessibility-tree) nor [visible on the page](#visible-on-the-page) are still considered for this rule.
 
--- a/pages/glossary/non-empty.md
+++ b/pages/glossary/non-empty.md
@@ -3,15 +3,4 @@ title: Non-empty text string
 key: non-empty
 ---
 
-A string of characters (text) is considered *"non-empty"* if it contains 1 or more characters that are contained within any of the following unicode categories:
-
-- **Ll**: Letter, Lowercase
-- **Lu**: Letter, Uppercase
-- **Lt**: Letter, Titlecase
-- **Lo**: Letter, Other
-- **Lm**: Letter, Modifier
-- **Nd**: Number, Decimal Digit
-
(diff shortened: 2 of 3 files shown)


_rules/SC4-1-1-unique-id.md
```diff
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-Any `id` attribute specified on an HTML or SVG element.
+Any [non-empty](#non-empty) `id` attribute specified on an HTML or SVG element.
```
WilcoFiers: Non-empty isn't the right definition here. I can't really tell what you were trying to do here because you didn't add any test cases. But I doubt `id="-"` should be ignored by this, which is what "non-empty" does.

annethyme: We discussed whether it would make sense to split the "non-empty" definition out into two different definitions: 

1. Non-empty: A string of characters is considered non-empty if the string contains characters that are not space characters. (or even have the definition be of "empty" instead)

2. Textual content: A string of characters is considered textual content if it contains one or more characters from the letter or number categories of Unicode. This definition applies to the string as a whole, regardless of any other characters the string might contain.

I think this would ease some of the confusion around what the "non-empty" definition means.

I know this change would require us to go through already published rules and see which of the definitions it should be using. But I am guessing that such a second look could be beneficial.

WilcoFiers: @jkodu are you going to pick that up? What's the plan for this?

pages/glossary/non-empty.md
```diff
@@ -3,15 +3,4 @@ title: Non-empty text string
...
-- **Lt**: Letter, Titlecase
-- **Lo**: Letter, Other
-- **Lm**: Letter, Modifier
-- **Nd**: Number, Decimal Digit
-
-For more details on unicode categories, check out [www.fileformat.info/info/unicode/category/](http://www.fileformat.info/info/unicode/category/index.htm)
-
-  
\ No newline at end of file
+A string of characters is considered *non-empty* if the string contains characters that are not space characters.
```
carlosapaduarte: To ensure that space characters are perceived unambiguously perhaps you could use the 'Separator, Space' category of Unicode (and provide the link) similarly to what you do in the textual content definition.

dd8: The HTML spec allows any character in IDs except ASCII whitespace:
https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute
https://infra.spec.whatwg.org/#ascii-whitespace

So non-ASCII whitespace characters like these are valid in IDs
U+00A0 | NO-BREAK SPACE
U+200B | ZERO WIDTH SPACE

I think the shared definition is problematic because it's used in different contexts where the definition of space is different in each context:
- what is treated as space/empty by the HTML parser
- what is treated as space/empty and displayed as blank or zero width by fonts
- what is treated as space/empty and not voiced by assistive tech

pages/glossary/non-empty.md
```diff
@@ -3,15 +3,4 @@ title: Non-empty text string
 key: non-empty
```
WilcoFiers: You can't just change this definition. Changing this changes the meaning of a whole bunch of existing rules. If you want to change the meaning of this definition you'll need to create a different definition for the rules that currently use this. Please do that in a separate PR so that we can have a record of it.

pages/glossary/non-empty.md
WilcoFiers: Agreed, "space characters" is insufficient. Please specify which characters are considered space characters.

pages/glossary/textual-content.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Textual content
+key: textual-content
+---
+
+A string of characters is considered *textual content* if it contains one or more characters from the letter or number categories of Unicode. This definition applies to the string as a whole, regardless of any other characters the string might contain.
```
WilcoFiers: Specify the ranges.

jkodu:
There are some definition changes, that anne has picked up https://github.com/auto-wcag/auto-wcag/pull/430/files, hence rolling back definition changes on this PR & keeping a simple change to applicability

annethyme:
@Brynanders , I am assigning this rule to you since you are the owner of this rule inside the WAI-Tools project. It has a deadline of March 31.

_rules/SC4-1-1-unique-id.md
```diff
@@ -20,7 +20,7 @@ authors:
 
 ### Applicability
 
-Any `id` attribute specified on an HTML or SVG element.
+Any `id` attribute which is not an empty string, specified on an HTML or SVG element.
```
annethyme: > which is not an empty string

---> 

> which is not the empty string (""),

annethyme: ... to align perfectly with https://github.com/auto-wcag/auto-wcag/pull/430/files

Issue #348, opened 2018-11-09
Update Rule: Valid body lang attribute


jkodu:
Rule: https://auto-wcag.github.io/auto-wcag/rules/SC3-1-2-lang-valid.html

![image](https://user-images.githubusercontent.com/20978252/48265183-c0891e00-e423-11e8-87d0-5ca3817b5d46.png)

Like in the scenario above, where the language text contains only one word, it is not a WCAG failure.

The rule needs to be updated to not flag single words.

dd8:
Word boundaries are hard to identify in languages like Japanese that don't use spaces between words:
https://en.wikipedia.org/wiki/Japanese_punctuation#Space

For example, 日米関係 means 'Japan / US relations'

And the concept of single words doesn't map well to Japanese:

日 means 'sun' 
本 means 'origin'
and '日本' means 'Japan' (sun origin = land of the rising sun)

jkodu:
@ShadowBB - can I get your input on the above comment?

ShadowBB:
@jkodu - I am unsure whether or not these two Chinese characters would constitute a "phrase" or not. I interpreted it as a single word because all the other links in the list where single words but I could be wrong.

The SC states "3.1.2 Language of Parts: The human language of each passage or phrase in the content can be programmatically determined..."

In our interpretation a "phrase" cannot be a single world outside of the context of a sentence structure but is a sequence of 2 or more words that constitute a single "unit" of meaning within a sentence.

kasperisager:
From the Merriam-Webster definition of "phrase":

> a word or group of words forming a syntactic constituent with a single grammatical function

https://www.merriam-webster.com/dictionary/phrase

dd8:
I think problem is the definition of "word" - which varies a lot by language, even in closely related languages:

>Not all languages delimit words expressly. Mandarin Chinese is a very analytic 
>language (with few inflectional affixes), making it unnecessary to delimit words 
>orthographically. However, there are many multiple-morpheme compounds in 
>Mandarin, as well as a variety of bound morphemes that make it difficult to 
>clearly determine what constitutes a word.

https://en.wikipedia.org/wiki/Word#Word_boundaries

>Words differ greatly from language to language. In fact, not all languages
>traditionally have a word for 'word', which means that while words can be
>defined objectively according to certain criteria, they don't necessarily match
>something that is recognised by native speakers.

https://www.quora.com/Do-all-languages-consist-of-words#MoreAnswers

[...]

PR #413, opened 2019-01-29
Added assumption for SC3-1-2-lang-valid

Brynanders:
Based on discussion in #348

### DIFF
--- a/_rules/SC3-1-2-lang-valid.md
+++ b/_rules/SC3-1-2-lang-valid.md
@@ -27,7 +27,7 @@ The `lang` and `xml:lang` attributes have a [valid language subtag](#valid-langu
 
 ## Assumptions
 
-*There are currently no assumptions*
+The `lang` or `xml:lang` attribute is being used to comply to WCAG.
 
 ## Background
 
--- a/_rules/SC3-1-2-lang-valid.md
+++ b/_rules/SC3-1-2-lang-valid.md
@@ -2,7 +2,7 @@
 name: Valid body lang attribute
 
 description: |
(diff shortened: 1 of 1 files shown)


annethyme:
We discussed in #348 if the language attribute was applied to a single word, if that was a WCAG violation.

With the now added assumption, we cover both this case and also cases where a part of the content has been marked up with an incorrect language attribute, but where a language attribute is not needed at all because the language of the part is the same as the language of the whole page.

_rules/SC3-1-2-lang-valid.md
```diff
@@ -27,7 +27,7 @@ The `lang` and `xml:lang` attributes have a [valid language subtag](#valid-langu
 
 ## Assumptions
 
-*There are currently no assumptions*
+The `lang` or `xml:lang` attribute is being used to comply to WCAG.
```
WilcoFiers: I don't think this explains anything. Can you explain?

Brynanders: @WilcoFiers fair point. It relates to this conversation and comment https://github.com/auto-wcag/auto-wcag/issues/348#issuecomment-456379859.

How about the following.

> This rule assumes that the presence of a `lang` or `xml:lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.

annethyme:
@WilcoFiers, do you think an update like this has to go into CFC stage, or can we just merge as is?

jkodu:
Discussed on the call, that we are happy to merge this.

PR #415, opened 2019-01-30
Rule idea: Focusable element has visible focus (bd5de3)

annethyme:
Looking for input on whether this rule is actually testing a WCAG requirement + if people agree on the general direction of the current draft.

Closes issue: none

# How to Review And Approve


annethyme:
Only just realised that @EmmaJP and @john-urbilog has already started up a composite rule like this in https://github.com/auto-wcag/auto-wcag/issues/339.

I don't think it is necessary to do it as a composite rule, but let's get some input :)

_rules/bd5de3.md
```diff
@@ -0,0 +1,37 @@
...
+- Anne Thyme Nørregaard
+---
+
+## Applicability
+
+This rule applies to all elements that can receive focus.
+
+## Expectation
+
+When the target element's focus state is triggered, a [visible](#visible) change in text style, border, box-shadow, background-image or color occurs for the element.
```
WilcoFiers: I don't think you can use the "visible" definition here. I'm also not sure these are all the CSS properties to look for, but I think this is a good direction. You may want to consider leaving just allowing any style change to indicate focus. Not sure what the best way to do this is.

ShadowBB: I think it is better to allow any style change to indicate focus, not just these specific examples. For instance a menu could show the focused item by making the text larger or indenting the menu item that has received focus, making it stand out from it's vertically aligned siblings.

_rules/bd5de3.md
```diff
@@ -0,0 +1,37 @@
...
+- DOM Tree
+- CSS Styling
+
+ authors:
+- Anne Thyme Nørregaard
+---
+
+## Applicability
+
+This rule applies to all elements that can receive focus.
```
ShadowBB: This needs to specify somehow that elements with a tabindex="-1" which can still receive focus with the help of javascript are still applicable. I am unsure whether or not such an applicability can be tested automatically.

kasperisager:
An interesting case to consider is "custom" radio and check boxes as popularized by Bootstrap: https://getbootstrap.com/docs/4.3/components/forms/#custom-forms. These consist of a visually hidden `<input>` element and an adjacent `<label>` element that is highlighted when the `<input>` element receives focus. As such, the element itself has no visible focus but an adjacent element has.

pages/glossary/change-in-appearance.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Change in appearance
+key: change-in-appearance
+---
+
+A change in appearance for an element means that when comparing the element before and after the possible change has happened, the rendered pixels have changed within an area that includes the element plus a 50 pixel margin to each side.
```
WilcoFiers: You should probably consider timing for this as well. Is this change in appearance is supposed to happen in the first frame paint after the change? If not, should there be a delay?

ShadowBB:
Oh, and for clarity sake: I do think this tests a WCAG requirement and I agree with the general direction.

_rules/bd5de3.md
```diff
@@ -0,0 +1,39 @@
+
+---
+id: edb7b7
+name: Focusable element has visible focus
+test_type: atomic
+
+ description: |
+  This rule checks that all elements that can receive focus has a focus styling that is visibly different from the normal styling.
```
jkodu: editorial: visibly > visually

_rules/bd5de3.md
```diff
@@ -0,0 +1,39 @@
+
+---
+id: edb7b7
+name: Focusable element has visible focus
+test_type: atomic
```
jkodu: change to `rule_type`

_rules/bd5de3.md
```diff
@@ -0,0 +1,39 @@
...
+
+ description: |
+  This rule checks that all elements that can receive focus has a focus styling that is visibly different from the normal styling.
+  
+ success_criterion: 
+- 2.4.7 # Focus Visible
+
+ test_aspects:
```
jkodu: change to `input aspects`

_rules/bd5de3.md
```diff
@@ -0,0 +1,39 @@
...
+---
+id: edb7b7
+name: Focusable element has visible focus
+test_type: atomic
+
+ description: |
+  This rule checks that all elements that can receive focus has a focus styling that is visibly different from the normal styling.
+  
+ success_criterion: 
+- 2.4.7 # Focus Visible
```
jkodu: believe this is `accessibility_requirements` now

Jym77:
Note to self:
Some elements might auto-steal focus of some others. Typically, a modal windows does not really prevent background elements to receive focus but steal it as soon as focus is given to something out of the modal. Thus, rule has to trigger on having focus, not on receiving it.

jkodu:
based on ACT meeting on 17 October 2019, it was agreed that this PR will be closed until we have an active person to work on it.

PR #412, opened 2019-01-24
Rule: Headers attribute only refers to cells in the same table (a25f45)

jkodu:
Rule: `headers` attribute only refers to `cells` in the same `table`

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/406

Note:
For clear intents and purposes, the scope of this rule is kept to be fairly simple.
Intent is to author more rules for various other scenarios that involves this SC.

PR #417, opened 2019-01-31
Rule: Table header cells have corresponding data cells (d0f69e)

jkodu:
Rule - All table header cells refer to data cells

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/406


_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to any element with [semantic role](#semantic-role) of `columnheader` or `rowheader` within the `table` element that is either [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree).
```
WilcoFiers: If you're talking about `table` elements, this should be limited to `th` elements IMO. I also think you should ignore any th that is not in the accessibility tree.

Another suggestion here would be to limit this rule to `th` elements in a table that has non-empty `td` elements: https://github.com/dequelabs/axe-core/issues/741#issuecomment-367957486

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Expectation
+- Each target element with the `headers` attribute refers to other `cells` of the same `table`.
+
+Each target element refers to data `cells` of the same `table`.
```
WilcoFiers: Can you explain why there are two sentences here? Is that intentional?

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Assumptions
+
+*There are currently no assumptions*
```
WilcoFiers: You're assuming that the table has data that the th is supposed to refer to.

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Accessibility support
+
+*There are no major accessibility support issues known for this rule.*
```
WilcoFiers: There are significant accessibility support issues with tables. Last time I tested (which is a few years ago, admittedly) none of the AT/UA fully implemented all table header mechanisms (header, scope, auto-scope)

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Background
+
+- [Understanding Success Criterion 1.3.1: Information and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+
+
+## Test Cases
+
+### Passed
```
WilcoFiers: Add examples for `scope` attribute, and for auto-scoped headers (no scope or headers)

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+The `columnheader` refers to `cells` within the same `table`.
+
+```html
+<table>
+  <tr> 
+    <td aria-labelledby="header1">
+      Description
+    </td>
+    <th id="header1">
```
WilcoFiers: This th fails the rule. It's not a pass.

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### Failed example 1
+
+The `columnheader` do not refer to `cells` within the `table`.
+
+```html
+<table>
+  <tr>
+    <th id="header1">Projects</th>
+    <th id="header2">Exams</th>
```
WilcoFiers: These headers should pass, since they are auto-scoped with the td.

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,160 @@
...
+
+*There are no major accessibility support issues known for this rule.*
+
+## Background
+
+- [Understanding Success Criterion 1.3.1: Information and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+
+
+## Test Cases
```
WilcoFiers: You didn't test visibility and accessibility tree.

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,245 @@
...
+---
+
+## Test Procedure
+
+### Applicability
+
+This rule applies to `th` elements in a `table`, that has [non-empty](#non-empty) `td` elements, which are [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+### Expectation
+- Each target element with the `headers` attribute refers to other `cells` of the same `table`.
```
carlosapaduarte: Wouldn't this expectation subsume #412? I'm having a hard time understanding any difference between this and #412 that would justify the need for both rules.
The expectation includes part of the applicability (the `headers` attribute). If we rewrite the applicability we get: 
This rule applies to `th` elements with a `headers` attribute in a `table` that has [non-empty](#non-empty) `td` elements, which are [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree).
#412 applies to `td` elements in addition
The expectation becomes: 
Each target element refers to other `cells` of the same `table`.
#412 refers only to `th` elements, not all `cells` 
I'm pretty sure we can accommodate these checks in a single rule, don't you think?

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,245 @@
...
+- [Understanding Success Criterion 1.3.1: Information and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+
+## Test Cases
+
+### Passed
+
+#### Passed example 1
+
+The `columnheader` refers to `cells` within the same `table`.
```
carlosapaduarte: Shouldn't it be the element with the `headers` attribute that refers to the `columnheader` cells?

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,245 @@
...
+    <tr>
+      <td colspan="2" headers="header1 header2">15%</td>
+    </tr>
+  </tbody>
+</table>
+```
+
+#### Passed example 2
+
+The `columnheader` refers to `cells` within the same `table`.
```
carlosapaduarte: Shouldn't it be the element with the `headers` attribute that refers to the `columnheader` cell?

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,245 @@
...
+    <th id="header1">
+      Projects
+    </th> 
+  </tr>
+</table>
+```
+
+#### Passed example 3
+
+ The `th` element defines a cell as header of a group of table cells by usage of `scope` attribute.
```
carlosapaduarte: There is no target element with a `headers` attribute, so shouldn't this be inapplicable?

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,245 @@
...
+     <tr>
+        <th scope="row">Severus Snape</th>
+        <td>Slytherin</td>
+    </tr>
+</table>
+ ```
+
+#### Passed example 4
+
+Header(s) for table cells are `auto` scoped.
```
carlosapaduarte: There is no target element with a `headers` attribute, so shouldn't this be inapplicable?

_rules/table-header-cells-refer-to-data-cells.md
```diff
@@ -0,0 +1,245 @@
...
+    <tr>
+      <td colspan="2" headers="header3 header2">15%</td>
+    </tr>
+  </tbody>
+</table>
+```
+
+#### Failed example 2
+
+The `headers` attribute is missing.
```
carlosapaduarte: If this is a fail then the expectation should be split in two: first, there is a `headers` attribute; second, it refers to some `cell` in the same table.

_rules/table-header-cells-refer-to-data-cells.md
jkodu: @carlosapaduarte 

I have updated the rule. This rule only does one thing now.
`Each table header in a data table refers to data cells` & does not confuse the same with `headers` attribute.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,230 @@
...
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements in a `table` which are [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+Each target element refers to data cells of the same `table`
```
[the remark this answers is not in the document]
jkodu: @carlosapaduarte I agree, after reading through https://www.w3.org/TR/html/tabular-data.html#forming-relationships-between-data-cells-and-header-cells, believe `has corresponding` offers a better verbiage than `refers to`.

I have updated the rule accordingly.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,258 @@
...
+  </tr>
+  <tr> 
+    <td>12:00</td> 
+  </tr>
+</table>
+```
+
+#### Inapplicable Example 4
+
+The rule only applies to `table` element that is included in the accessibility tree.
```
carlosapaduarte: Both this and the previous examples address the same applicability issue. It would be more interesting to have one with a table that is not visible to cover all applicability issues.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,258 @@
...
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements within a `table`, that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+Each target element has corresponding data cells of the same `table`
```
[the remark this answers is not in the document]
jkodu: I went with 

```
Each target element is either the row header or the column header for data cells of the same `table`.
```

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,258 @@
...
+	<tr>
+		<th scope="row">Beth</th>
+		<td>8</td>
+	</tr>
+</table>
+```
+
+#### Failed Example 2
+
+`th` has corresponding empty `cell`.
```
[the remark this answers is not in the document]
jkodu: Non-empty was deprecated in this PR - https://github.com/act-rules/act-rules.github.io/pull/430/files, but I have updated the expectation accordingly.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
+---
+id: d0f69e
+name: All table header cells has corresponding data cells
```
carlosapaduarte: has > have

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements within a `table` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+Each target element is either the row header or the column header for data cells that is not empty (""), within the same `table`.
```
carlosapaduarte: (second) is > are

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+    </th> 
+  </tr>
+</table>
+```
+
+#### Passed Example 4
+
+The `columnheader` has corresponding `cells` within the same `table`, by usage of `aria-labelledby` attribute.
+
+```html
```
carlosapaduarte: You might want to correct tab spaces in this one

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+      <td>Slytherin</td>
+  </tr>
+</table>
+```
+
+### Failed
+
+#### Failed Example 1
+
+There are no corresponding `cells` for each `th` element within the `table`.
```
carlosapaduarte: I believe "Not all `th` elements within the `table` have corresponding `cells`" is easier to understand

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+    <td></td> 
+  </tr>
+</table>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+The rule only applies only to `table > th` element. The `table` has no `th` elements.
```
carlosapaduarte: Remove the second only

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: This seems fine in the IDE. I have noticed GitHub diff view having issues with highlighting spacing correctly.

image

carlosapaduarte: Yep, same here. But on Github diff it does look out of place.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+input_aspects:
+- DOM Tree
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements within a `table` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
```
ShadowBB: This applicability can be read in two ways:

- This rule applies to th elements within (a table that is visible and included in the accessibility tree).
- This rule applies to (th elements within a table) that is visible and included in the accessibility tree.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+  </tr>
+  <tr> 
+    <td>12:00</td> 
+  </tr>
+</table>
+```
+
+#### Inapplicable Example 4
+
+The rule only applies to `table` element that is visible.
```
ShadowBB: I think this example needs to change (if we keep the applicability like it is) because in this example the table isn't visible and(!) isn't included in the accessibility tree. I think an example that visually hides the table but keeps it in the accessibility tree is best.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,244 @@
...
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements within a `table` that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+Each target element is either the row header or the column header for data cells that are not empty (""), within the same `table`.
```
ShadowBB: I think the empty clause should be dropped. I would not fail example 2 under 1.3.1 (because there is no information, structure or relationships that are conveyed by presentation that cannot be programmatically determined).

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: The first way of reading the applicability is what is intended @ShadowBB. I have changed this to:

> This rule applies to `th` elements within a `table`, where the `table` is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).

Wdyt?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: Alternatively changing the description to below can help to,

> The rule only applies to `table` element that is both visible and included in the accessibility tree.

jkodu:
@ShadowBB I have made a set of changes to this rule based on your suggestions. Good catch on those.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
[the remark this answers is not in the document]
ShadowBB: This has not yet been fixed with the new applicability.

Why are we excluding tables that are invisible (but still included in the accessibility tree)? They would bother screen-readers users if they where not implemented according to the expectation correct?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements within a `table`, where the `table` is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
+
+## Expectation
+
+Each target element is either the row header or the column header for data cells within the same `table`.
```
Jym77: Maybe "row header" and "column header" could be links to some HTML spec.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Assumptions
+
+- Tables are [well-formed, according to the HTML5.1 specification](https://www.w3.org/TR/html51/tabular-data.html#forming-a-table).
```
Jym77: Do we need to have an assumption that the HTML is well-formed? Isn't that a global assumption that all rules only apply to well-formed HTML?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells](https://www.w3.org/TR/html/tabular-data.html#forming-relationships-between-data-cells-and-header-cells)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+`th` has corresponding `cells` within the same `table`.
```
Jym77: Use of `cells` with the backticks is a bit inaccurate since this is not actual `cells` elements, but `td` elements (who do act as cells).

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+	</tr>
+	<tr>
+		<td>05:41</td>
+	</tr>
+</table>
+```
+
+#### Passed Example 2
+
+`th` has corresponding `cells` within the same `table`. Here the cells span multiple columns.
```
Jym77: Same remark on `cells`. Even weirder here since it appears once with backticks and once without.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+    <tr>
+      <td colspan="2">15%</td>
+    </tr>
+  </tbody>
+</table>
+```
+
+#### Passed Example 3
+
+The `columnheader` has corresponding `cells` within the same `table`, by usage of `headers` attribute.
```
Jym77: Again, on `cells`.

Also, there is no `columnheader`. I guess you refer to the `th` element.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+    <th id="header1">
+      Projects
+    </th> 
+  </tr>
+</table>
+```
+
+#### Passed Example 4
+
+The `columnheader` has corresponding `cells` within the same `table`, by usage of `aria-labelledby` attribute.
```
Jym77: Same on `cells` and `columnheader`.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+	<tr>
+    <th id="hd2">Time</th>
+    <td aria-labelledby="hd2">05:44</td>
+	</tr>
+</table>
+```
+
+#### Passed example 5
+
+Each `th` element has corresponding `cells` within the same `table`, by usage of `scope` attribute.
```
Jym77: `cells` .

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+      <td>Slytherin</td>
+  </tr>
+</table>
+```
+
+### Failed
+
+#### Failed Example 1
+
+`th` elements within the `table` do not have corresponding `cells`.
```
Jym77: `cells`

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+	</tr>
+	<tr>
+		<th>Header 2</th>
+	</tr>
+</table>
+```
+
+#### Failed Example 2
+
+`th` does not have `cell`.
```
Jym77: `cell`.

Plus, there is no `th` element in that table, only a `td` element with a role of `columnheader`. Per current applicability, this is inapplicable.
* Do you mean that applicability should be "any element with a role of `columnheader`"?
* You should also add passed examples with a role of `columnheader` (and no `th`) to illustrate this.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+  <tr>
+    <td>Jane</td>
+    <td>Doe</td>
+  </tr>
+</table>
+```
+
+#### Inapplicable Example 2
+
+The rule only applies to `table` & not layout table.
```
Jym77: I think it will be clearer to say "The rule only applies to the `table` element, not to other elements with a role of `table`"

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+
+```html
+<div role="table">
+ <div role="row">
+  <div class="columnheader">Projects</div>
```
Jym77: Should it rather be `<div role="columnheader">`?
At least to avoid the reference to a non-existing class.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+ <div role="row">
+  <div role="cell">15%</div>
+  <div role="cell">15%</div>
+ </div>
+</div>
+```
+
+#### Inapplicable Example 3
+
+The rule only applies to `table` element that is included in the accessibility tree. The `table` is marked as `role=presentation`.
```
Jym77: "included in the accessibility tree" should be link to glossary.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+  </tr>
+  <tr> 
+    <td>12:00</td> 
+  </tr>
+</table>
+```
+
+#### Inapplicable Example 4
+
+The rule only applies to `table` element that is both visible and included in the accessibility tree.
```
Jym77: "visible" and "included in the accessibility tree" should be links to glossary.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
carlosapaduarte: Yes, the `th` in this example does not have the role `columnheader`. It has the role `rowheader`.
If you refer only to the `th` element will be easier to understand.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
carlosapaduarte: Layout table should be linked to something. Better yet, can't we have the applicability stating that this rule does not apply to layout tables?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+      <th scope="row">Cho Chang</th>
+      <td>Ravenclaw</td>
+  </tr>
+    <tr>
+      <th scope="row">Severus Snape</th>
+      <td>Slytherin</td>
+  </tr>
+</table>
+```
+
```
carlosapaduarte: I think it would be nice to have an example with `th` both as column and row headers but without explicit roles or the use of the `scope` attribute. Something like:
```html
<table>
    <caption>Hogwarts Houses stars</caption>
    <tr>
        <th>Name</th>
        <th>House</th>
    </tr>
    <tr>
        <th>Harry Potter</th>
        <td>Gryffindor</td>
    </tr>
    <tr>
        <th>Cedric Diggory</th>
        <td>Hufflepuff</td>
    </tr>
    <tr>
        <th>Cho Chang</th>
        <td>Ravenclaw</td>
    </tr>
    <tr>
        <th>Severus Snape</th>
        <td>Slytherin</td>
    </tr>
</table>
```

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: Linking to glossary that will get added from here - https://github.com/act-rules/act-rules.github.io/pull/412/files#diff-9b3c2a3e3623d28760ab820df60816dc

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: You are right @Jym77, this would be inapplicable. I have removed it completely as it matched Inapplicable example 1 (except having a role, I guess I was trying to show that just having a role on td does not suffice, but it only adds ambiguity).

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,236 @@
...
+			<td>Ravenclaw</td>
+	</tr>
+	<tr>
+			<th>Severus Snape</th>
+			<td>Slytherin</td>
+	</tr>
+</table>
+```
+
+### Failed
```
carlosapaduarte: I think we could use another failed example, with a bit more "complexity", i.e., with `td` cells also. Something like:

```html
<table>
    <tr>
        <th>Header</th>
        <td>Data 1</td>
    </tr>
    <tr>
        <td>Data 2</td>
        <td>Data 3</td>
    </tr>
</table>
```

jkodu: The rule only looks at `th` elements, and for the above example it satisfies the expectation, so this would not be a fail.

carlosapaduarte: Yes, you're completely right!

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,236 @@
...
+
+```html
+<table>
+	<tr>
+		<th>Header 1</th>
+	</tr>
+	<tr>
+		<th>Header 2</th>
+	</tr>
+</table>
```
kasperisager: @jkodu If I recall correctly, didn't we discuss restricting the applicability of this rule to tables that actually have data cells in order for this case to pass? What are your thoughts on this, @ShadowBB? Would you regard an empty table consisting only of headers to be an accessibility issue and, perhaps more importantly, a failure of 1.3.1?

ShadowBB: I would fail that on 1.3.1.

1.3.1 demands that relationships are communicated. If we imply the existence of relationships that are not actually there we devalue the information about correct relationships. I also fail non-table heading elements without associated content.

But I agree that it is a stretch because no actual wrong relationship has been communicated, only implied. (Communicating a relationship would need the involvement of at least 2 elements instead of the incorrect role on just a single one).

jkodu: Discussed with @kasperisager and based on above comment from @ShadowBB, makes sense to leave this as is and without any change.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,236 @@
...
+input_aspects:
+- DOM Tree
+authors:
+- Jey Nandakumar
+- Audrey Maniez
+---
+
+## Applicability
+
+This rule applies to `th` elements within a `table`, where the `table` is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree).
```
annethyme: "within a `table`" is a bit ambiguous. It is something we have worked a lot with in other rules (therefore some of them say stuff like "descendant in the flat tree").

annethyme: Suggestion for applicability: 

```
This rule applies to each `th` element that
- is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree)
- is a [descendant]() of a `table` element that is [visible](#visible) and [included in the accessibility tree](#included-in-the-accessibility-tree), and
- is either a [column header](https://html.spec.whatwg.org/#column-header) or [row header](https://html.spec.whatwg.org/#row-header).
```

jkodu:
@kasperisager - I have spoken to @WilcoFiers about comment - https://github.com/act-rules/act-rules.github.io/pull/417#pullrequestreview-275284929 & we both agree that this rule should also look at rules implemented using ARIA. I will be updating the rule accordingly.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
ShadowBB: We discussed this in the face to face meeting and Wilco convinced me that my explanation above is incorrect.

We fail non-table-headings without content because it messes up the page structure(!) not it's relationships. And table headings don't communicate page structure.

So this example should become not applicable and we should change the applicability of this rule to reflect this.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,188 @@
...
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+Each column header element has corresponding cells, within the same table element.
```
Jym77: "Each" is a bit weird given that there is only one. Same for "cells" (plural) but there is only one.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,188 @@
...
+The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), and is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the below [semantic roles](#semantic-role):
+
+- [cell](https://www.w3.org/TR/wai-aria-1.1/#cell)
+- [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell)
+- [rowheader](https://www.w3.org/TR/html-aria/#index-aria-rowheader)
+- [columnheader](https://www.w3.org/TR/html-aria/#index-aria-columnheader)
+
+## Expectation
+
+Each target element is either the row or column header for cells that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the same element as the target element.
```
Jym77: "is descendant of the same element" is not precise enough. It should be "descendant of the same `table` or `grid` element", at least.
(otherwise, they could be in two `table` but descendant of the same `body` element, and associated through `headers` attribute)

And even that might be not enough if there is nested tables like:
```html
<table>
  <table>
    <th id="foo>Foo</th>
  </table>
  <table>
    <td headers="foo">Bar</td>
  </table>
</table>
```
The `th` and `td` are descendant of the same `table` element (the topmost one), but the rule should probably fail on that case (I guess).

Jym77: I am not sure that "is the row/column header" is precise enough.
Do you mean that there is at least one cell in the table that is assigned the test target as header cell?
https://html.spec.whatwg.org/#header-and-data-cell-semantics
https://html.spec.whatwg.org/#algorithm-for-assigning-header-cells

(if so, example descriptions also need to be updated with correct vocabulary: "there is at least one cell assigned to …")

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: "table element" => "`table` element"

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,188 @@
...
+	</tr>
+	<tr>
+		<td>05:41</td>
+	</tr>
+</table>
+```
+
+#### Passed Example 2
+
+Each column header element has corresponding cells, within the same element having a [semantic role](#semantic-role) of `table`.
```
Jym77: "within the same element" => "within the same `table` element"

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,188 @@
...
+		</tr>
+	</tbody>
+</table>
+```
+
+### Failed
+
+#### Failed Example 1
+
+Not all column headers have corresponding cells within the same table element.
```
Jym77: "Not all" => point out which one is failing. Technically, this example has 2 test targets and one is passing while the other fails.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,188 @@
...
+	</tr>
+	<tr>
+		<td>12:00</td>
+	</tr>
+</table>
+```
+
+#### Inapplicable Example 2
+
+The rule does not apply to table element that is not [visible](#visible).
```
Jym77: Make it not visible but still in the accessibility tree (position out of screen, usually) to emphasize that "not visible" is enough to make the rule not applicable.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,188 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), and is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the below [semantic roles](#semantic-role):
```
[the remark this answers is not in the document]
jkodu: I tried that initially and the flow of sentence did not feel great to me. Leaving this as is. Thanks for the suggestion.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: I like the suggestion. I have done my best. 
I am unsure how to tackle the nested tables.

jkodu: I wanted to keep this generic, and not dip into algorithm for assigning header cells, or anything that is too specific to `table` element as this rule covers `ARIA` table/ grid too.

I will read up the spec, and let me see what other reviewers have to say before I make a change, if necessary.

jkodu:
@carlosapaduarte and @DagfinnRomen I would appreciate a re-read of this rule, given it has changed a little.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -110,7 +110,7 @@ Each column header element has corresponding cells within the same table element
 
 #### Passed Example 4
 
-Each row and column header element has corresponding cells, within the same element having a [semantic role](#semantic-role) of `grid`.
+Each row and column header element has corresponding cells, within the same element having a [semantic role](#semantic-role) of `grid` .
```
Jym77: Remove space before the final period.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -19,7 +19,7 @@ authors:
 
 ## Applicability
 
-The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), and is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the below [semantic roles](#semantic-role):
+The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), and is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the following [semantic roles](#semantic-role):
```
Jym77: Would the rule also be good for `treegrid` roles? In which case applicability can be "`table` or a role than inherit from `table`" (currently `table`, `grid`, and `treegrid`).

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -28,7 +28,7 @@ The rule applies to any HTML element that is a [descendant](https://dom.spec.wha
 
 ## Expectation
 
-Each target element is either the row or column header for cells that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the same element as the target element.
+The target element is either the row or column header for cells that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the same [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid) element, as the target element.
```
Jym77: I still don't think that "this element is the row header for this cell" is properly defined. And I think it gets even worse for column headers.

The algorithm for matching in HTML `table` elements is somewhat tricky (because of colspan and rowspan). I guess it is very similar for pure "role" tables. This also gets tricky when there are multiple overlapping headers or just when the `header` attribute is used.

Typically:
```html
<table>
  <tr> <th id="foo">Foo</th> <th id="bar">Bar</th> </tr>
  <tr> <td>0</td> </tr>
  <tr> <td headers="bar">1 </td> </tr>
</table>
```
should be passing the rule, but
```html
<table>
  <tr> <th id="foo">Foo</th> <th id="bar">Bar</th> </tr>
  <tr> <td>0</td> <td headers="foo">1 </td> </tr>
</table>
```
should be failing it.

Then, it might be enough to add a note that assigning headers cells to data cell is done as per algorithm something (the one currently in the background).

Jym77: I kinda guess that the correct "matching" of elements is that their closest ancestor with a role of `table` or `grid` is the same.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,200 @@
...
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 1.3.1: Information and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics)
+
+## Test Cases
```
Jym77: We should probably add examples (both passed and failed) with `headers` attribute messing up the "natural" table structure.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: Discussed with @Jym77, and after reading up about `treegrid`, decided not to cloud this rule further, so leaving `treegrid` out of applicability.

jkodu:
I clicked "Update Branch" button, which errored, and I clicked again, but this button seemed to merge the PR by mistake :/. Think I was too fast before the re-render.

I have revered the change & looking into how to open this PR - https://github.com/act-rules/act-rules.github.io/pull/978

Issue #411, opened 2019-01-22
Link has name rule: What about empty links?


annethyme:
For this published rule: https://auto-wcag.github.io/auto-wcag/rules/SC2-4-4-link-has-name.html

If the only content of the link is an image, shouldn't the rule also be for 1.1.1 Non-Text Content? 
And if yes, would that mean splitting up the rule in two rules, like we have done for buttons and image buttons?

And what is the link doesn't have any content at all? Is it then an accessibility issue that the link doesn't have an accessible name? 
Some accessibility testing tools flag "empty links". Ours doesn't, because we can't see how that is an accessibility issue.

annethyme:
Changed my mind about the first part. If there is an image inside a link, the image itself will be applicable for rules under 1.1.1. 

I still would like to discuss the case where a link is completely empty. Is that a WCAG violation?

annethyme:
... and even if empty links are not an issue, could links with breaks or spaces as their only content cause problems, similar to https://github.com/auto-wcag/auto-wcag/pull/400?

annethyme:
Decided on Auto-WCAG call that it is a WCAG failure to have empty links, since they are still part of the focus order. 
@annethyme will add a test case for this to the rule.

annethyme:
Handling as part of https://github.com/auto-wcag/auto-wcag/pull/430

Issue #354, opened 2018-11-19
Add 'area' test cases to link-has-name rule


annethyme:
'area' with a 'href' has the semantic role of 'link', and is thus covered by the link-has-name rule. 
Make sure test cases reflect that.

annethyme:
Including fix in PR https://github.com/auto-wcag/auto-wcag/pull/430

PR #380, opened 2018-12-20
New rule: Document title is descriptive

annethyme:
Created new rule: Document title is descriptive

Closes issue: none

# How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

PR #430, opened 2019-03-07
Glossary and rule update: Cleaning up "non-empty"

annethyme:
"Non-empty text string" definition has been removed. 

Rules previously using "non-empty" definition have been updated to either use 
- (Outdated: "letters or numbers" definition), "that doed not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153)",
- "that is not the empty string ("")", or 
- "contains [text nodes](https://www.w3.org/TR/dom/#text) that do not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).", 
... depending on what the rule should actually be looking for. 

Also closes these issues:
- #411 (Added Failed test case for completely empty link in Link has name rule)
- #354 (Add `area` test cases to Link has name rule)

# How to Review And Approve


annethyme:
@WilcoFiers, I will work more tomorrow on figuring out something for handling Kasper's feedback. 
But could I maybe have you take an initial look at the PR too, and see if you can agree to the general direction that this is taking?

pages/glossary/text-content.md
```diff
@@ -3,7 +3,6 @@ title: Text content
...
+A string of characters is considered text content if it contains one or more characters from the [letter or number categories of Unicode](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153). This definition applies to the string as a whole, regardless of any other characters the string might contain.
 
-[rendered text]: rendered-text.html
-[text alternative computation]: text-alternative.html
\ No newline at end of file
+**Note:** This definition is a heuristic implementation of the [WCAG definition of `text`](https://www.w3.org/TR/WCAG21/#dfn-text). However, this definition differs from the WCAG `text` definition that also requires that "the sequence is expressing something in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s)", which is not part of this definition.
```
annethyme: Add something about differing also on the point of "programmatically determinable" vs. unicode categories.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -25,7 +25,7 @@ The rule applies to any HTML element with the [semantic role](#semantic-role) of
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty).
+Each target element has an [accessible name](#accessible-name) that contains [letters or numbers](#letters-or-numnbers).
```
jkodu: Typo: #letters-or-numnbers

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -25,7 +25,7 @@ The root element of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if
 
 ### Expectation
 
-The test target has a [non-empty](#non-empty) `lang` or `xml:lang` attribute.
+The test target has an `lang` or `xml:lang` attribute that is not the empty string ("").
```
jkodu: change: has an `lang`... -> has a `lang`...

pages/glossary/character.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Character
+key: character
```
jkodu: `character` seems like a generic definition that we can refer to elsewhere? Perhaps - https://developer.mozilla.org/en-US/docs/Glossary/Character

I see you drawing some distinction away from `grapheme`, then may be this should not be called `character`?

kasperisager: The issue is that computers don't really deal in graphemes. A language such as JavaScript, for example, represents strings of "characters" as sequences of UTF-16 code points, while other languages, such as C and Go, represent strings of "characters" as sequences of arbitrary bytes. The term "character" therefore has lots of different meanings in different contexts.

annethyme:
Updated to use "that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153)." instead of "contains [letters and numbers]" after feedback on https://github.com/auto-wcag/auto-wcag/pull/440

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -125,11 +125,19 @@ When `link` is off screen.
...
   <body>
-		<a class="offScreenLink" href="http://www.w3.org/WAI"> Web Accessibility Initiative (WAI) </a>
+		<a class="offScreenLink" href="http://www.w3.org/WAI">Web Accessibility Initiative (WAI)</a>
   </body>
 </html>
 ```
 
+#### Passed example 10
+
+`Area` element with `href` attribute has accessible name.
```
kasperisager: `Area` -> `area`.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -197,6 +205,22 @@ Not included in the accessibility tree.
...
+
+Link is completely empty, but still shows up in focus order, so it should have an accessible name.
+
+```html
+<a href="http://www.w3.org/WAI"></a>
+```
+
+#### Failed example 10
+
+`Area` element with `href` attribute does not have accessible name.
```
kasperisager: `Area` -> `area`.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -223,10 +247,18 @@ Not included in the accessibility tree due to `visibility: hidden`.
...
-Not included in the accessibility tree due to `aria-hideden="true"`.
+Not included in the accessibility tree due to `aria-hidden="true"`.
 
 ```html
 <a aria-hidden="true" href="http://www.w3.org/WAI"> Web Accessibility Initiative (WAI) </a>
 ```
+
+#### Inapplicable example 5
+
+`Area` element without `href` attribute does not have role of `link`.
```
kasperisager: `Area` -> `area`.

_rules/image-button-has-name.md
```diff
@@ -1,3 +1,4 @@
+
```
kasperisager: Rogue newline?

annethyme: If I change it, I am not allowed to commit just that change. Is this something we can live with?

annethyme:
@kasperisager, @ShadowBB, @DagfinnRomen. You guys already approved the PR, but I have now also added test cases to the updated rules to reflect the changes. Could you guys have a look again? 
I am going to dismiss your reviews, since the test cases could potentially both be off, but also highlight things in the changes, that there might be disagreement about.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -97,6 +97,14 @@ HTML `img` element marked as [decorative](#decorative) through `role="none"`
 <img role="none" />
 ```
 
+#### Passed example 6
+
+HTML `img` element has accessible name that does not only consist of seperator characters
```
carlosapaduarte: has accessible name > has an accessible name
seperator > separator

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -18,12 +18,12 @@ authors:
 
 ### Applicability
 
-The rule applies to any HTML `input`, `select` and `textarea` element with a [non-empty](#non-empty) `autocomplete` attribute, except if one of the following is true:
+The rule applies to any HTML `input`, `select` and `textarea` element with a `autocomplete` attribute, that is not the empty string (""), except if one of the following is true:
```
carlosapaduarte: a autocomplete > an autocomplete

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -18,12 +18,12 @@ authors:
...
 
 - The element is not [visible on the page](#visible-on-the-page), and not [included in the accessibility tree](#included-in-the-accessibility-tree)
 - The element is an `input` element with a `type` property of `hidden`, `button`, `submit` or `reset`
 - The element has a `disabled` or `aria-disabled="true"` attribute
-- The element has `tabindex="-1"` and has a [semantic role](#semantic-role) that is not a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles)
+- The element does not participate in [sequential focus navigation](https://www.w3.org/TR/html/editing.html#sec-sequential-focus-navigation) and has a [semantic role](#semantic-role) that is not a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles).
```
carlosapaduarte: widget > widget role

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -123,19 +123,44 @@ Full length autocomplete terms.
 <input autocomplete="section-primary shipping work email" />
 ```
 
+#### Passed example 8
+
+The `input` element does not have a semantic role that is a widget, but still participates in sequential focus navigation, and has a single autocomplete term.
```
carlosapaduarte: widget > widget role

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -123,19 +123,44 @@ Full length autocomplete terms.
...
+
+```html
+<input role="none" autocomplete="username" />
+```
+
+#### Passed example 9
+
+The `input` element does not participates in sequential focus navigation, but still has a semantic role that is a widget, and has a single autocomplete term.
```
carlosapaduarte: widget > widget role

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -123,19 +123,44 @@ Full length autocomplete terms.
...
+
+```html
+<input tabindex="-1" autocomplete="username" />
+```
+
+#### Passed example 8
+
+The `input` element does not have a semantic role that is a widget, but still participates in sequential focus navigation since the [`tabindex` attribute](https://www.w3.org/TR/html/editing.html#the-tabindex-attribute) value is not a [valid integer](https://www.w3.org/TR/html/infrastructure.html#valid-integer), and has a single autocomplete term.
```
carlosapaduarte: widget > widget role

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -165,76 +190,92 @@ Autocomplete is inappropriate for the type of field.
 <input type="number" autocomplete="email" />
 ```
 
+#### Failed example 6
+
+Autocomplete is not the empty string, but does not have any terms specified.
```
carlosapaduarte: Autocomplete is not the empty string > The value of autocomplete is not the empty string

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -165,76 +190,92 @@ Autocomplete is inappropriate for the type of field.
...
+Inapplicable element.
 
 ```html
 <button autocomplete="username"></button>
 ```
 
 #### Inapplicable example 2
 
-Empty attribute.
+Autocomplete attribute is the empty string ("").
```
carlosapaduarte: The value of the autocomplete attribute is ...

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -165,76 +190,92 @@ Autocomplete is inappropriate for the type of field.
...
+The element is positioned off screen and hidden to assistive technologies
 
 ```html
 <input autocomplete="username" aria-hidden="true" style="position:absolute; top:-9999em">
 ```
 
 #### Inapplicable example 5
 
-type `input` button.
+The `input` element is in the `button` state.
```
carlosapaduarte: The `input` element's type is `button`.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -106,7 +106,7 @@ Valid `title` provided.
 
 #### Passed example 5
 
-Valid `title` provided.
+First `title` element does not only contain seperator characters.
```
carlosapaduarte: seperator > separator

_rules/SC2-4-2-page-has-title.md
```diff
@@ -119,6 +119,18 @@ Valid `title` provided.
 </html>
 ```
 
+#### Passed example 6
+
+`title` element does not only contain seperator characters.
```
carlosapaduarte: seperator > separator

_rules/SC2-4-2-page-has-title.md
```diff
@@ -166,6 +178,16 @@ Empty first `title`.
 </html>
 ```
 
+#### Failed example 5
+
+`title` only consist of seperator characters.
```
carlosapaduarte: seperator > separator

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -125,11 +125,27 @@ When `link` is off screen.
...
+
+`area` element with `href` attribute has accessible name.
+
+```html
+<area shape="rect" coords="0,0,82,126" href="http://www.w3.org/WAI" alt="Web Accessibility Initiative (WAI)">
+```
+
+#### Passed example 11
+
+`a` element where accessible name does not only consist of seperator characters.
```
carlosapaduarte: seperator > separator

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -175,7 +191,7 @@ Link with image that has empty `aria-labelledby`.
 
 #### Failed example 6
 
-Aria-labelledby references to a non-existing id.
+`aria-labelledby` references to a non-existing id.
```
carlosapaduarte: `aria-labelledby` references a non-existing id.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -197,6 +213,30 @@ Not included in the accessibility tree.
...
+
+`area` element with `href` attribute does not have accessible name.
+
+```html
+<area shape="rect" coords="0,0,82,126" href="http://www.w3.org/WAI">
+```
+
+#### Failed example 11
+
+`a` element where accessible name through content only consist of seperator characters.
```
carlosapaduarte: seperator > separator

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -99,12 +99,36 @@ The `lang` attribute specified has a non-empty value. The rule does not verify t
 
 #### Passed example 7
 
-The `xml:lang` attribute specified has a non-empty value. The rule does not verify the validity of the value specified and checks only for presence of a value.
+The `xml:lang` attribute specifies a value that is not the empty string (""). The rule does not verify the validity of the value specified and checks only for presence of a value.
```
carlosapaduarte: for presence > for the presence

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -99,12 +99,36 @@ The `lang` attribute specified has a non-empty value. The rule does not verify t
...
 
 ```html
 <html xml:lang="xyz">
 ```
 
+#### Passed example 8
+
+The `xml:lang` attribute specifies a value that is not the empty string (""). The rule does not verify the validity of the value specified and checks only for presence of a value.
```
carlosapaduarte: for presence > for the presence

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -99,12 +99,36 @@ The `lang` attribute specified has a non-empty value. The rule does not verify t
...
+
+```html
+<html xml:lang="123">
+```
+
+#### Passed example 9
+
+The `xml:lang` attribute specifies a value that is not the empty string (""). The rule does not verify the validity of the value specified and checks only for presence of a value.
```
carlosapaduarte: for presence > for the presence

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -99,12 +99,36 @@ The `lang` attribute specified has a non-empty value. The rule does not verify t
...
+
+```html
+<html xml:lang="#!">
+```
+
+#### Passed example 9
+
+The `xml:lang` attribute specifies a value that is not the empty string (""). The rule does not verify the validity of the value specified and checks only for presence of a value.
```
carlosapaduarte: for presence > for the presence

_rules/SC3-1-2-lang-valid.md
```diff
@@ -78,7 +82,7 @@ The `lang` attribute specified has a non-empty value & a valid primary language
 
 #### Passed example 4
 
-The `lang` and `xml:lang` attribute specified has a non-empty value & a valid primary language subtag.
+The `lang` and `xml:lang` attributes values that are not the empty string ("") and has valid primary language subtags.
```
carlosapaduarte: The `lang` and `xml:lang` attributes values are not the empty string ("") and have valid primary language subtags.

_rules/SC3-1-2-lang-valid.md
```diff
@@ -88,6 +92,18 @@ The `lang` and `xml:lang` attribute specified has a non-empty value & a valid pr
 </html>
 ```
 
+#### Passed example 5
+
+The `lang` and `xml:lang` attributes values are either the empty string ("") or has a valid primary language subtag.
```
carlosapaduarte: The `lang` and `xml:lang` attributes values are either the empty string ("") or have a valid primary language subtag.

_rules/SC4-1-2-button-has-name.md
```diff
@@ -24,7 +24,7 @@ The rule applies to elements that are [included in the accessibility tree](#incl
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty)
+Each target element has an [accessible name](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
carlosapaduarte: consist > consists

_rules/SC4-1-2-button-has-name.md
```diff
@@ -111,6 +111,14 @@ Off screen elements should be tested.
 </html>
 ```
 
+#### Passed example 8
+
+Button has accessible name that does not only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/SC4-1-2-button-has-name.md
```diff
@@ -148,6 +156,14 @@ Off screen element without an accessible name.
 </html>
 ```
 
+#### Failed example 4
+
+Button has an accessible name that only consists of seperator characters.
```
carlosapaduarte: seperator > separator

_rules/SC4-1-2-form-field-has-name.md
```diff
@@ -27,7 +27,7 @@ This rule applies to any element that is [included in the accessibility tree](#i
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty).
+Each target element has an [accessible name](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
carlosapaduarte: consist > consists

_rules/SC4-1-2-form-field-has-name.md
```diff
@@ -95,6 +95,17 @@ Explicit role.
 <div aria-label="country" role="combobox" aria-disabled="true">England</div>
 ```
 
+#### Passed example 6
+
+Accessible name does not only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/SC4-1-2-form-field-has-name.md
```diff
@@ -149,6 +160,14 @@ Explicit label not supported on div elements.
 <div role="textbox" id="lastname"></div>
 ```
 
+#### Failed example 7
+
+Accessible name only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/SC4-1-2-iframe-has-name.md
```diff
@@ -25,7 +25,7 @@ The rule applies to `iframe` elements that are [included in the accessibility tr
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty).
+Each target element has an [accessible name](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
carlosapaduarte: consist > consists

_rules/SC4-1-2-iframe-has-name.md
```diff
@@ -74,6 +74,15 @@ Usage of `aria-labelledby` attribute to describe the `iframe` content.
 </iframe>
 ```
 
+#### Passed example 4
+
+Accessible name does not only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/SC4-1-2-iframe-has-name.md
```diff
@@ -130,6 +139,15 @@ Usage of `alt` attribute to describe content is not valid.
 </iframe>
 ```
 
+#### Failed example 7
+
+Accessible name only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/SC4-1-2-role-has-required-states-and-properties.md
```diff
@@ -107,7 +107,7 @@ Element does not list required states and properties (´aria-controls´ is requi
 
 #### Failed example 2
 
-Required states and properties does not have non-empty value
+Required states and properties does not have a value that is not the empty string ("")
```
carlosapaduarte: does > do

_rules/image-button-has-name.md
```diff
@@ -22,11 +22,11 @@ authors:
 
 The rule applies to any HTML `input` element with a `type` attribute in the `Image Button` state, that is [included in the accessibility tree](#included-in-the-accessibility-tree).
 
-**Note:** The specification of the `[type](https://www.w3.org/TR/html/sec-forms.html#element-attrdef-input-type)` attribute describes in detail how to map the value of the attribute to its corresponding state.
+**Note:** The specification of the [`type`](https://www.w3.org/TR/html/sec-forms.html#element-attrdef-input-type) attribute describes in detail how to map the value of the attribute to its corresponding state.
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty).
+Each target element has an [accessible name](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
carlosapaduarte: consist > consists

_rules/image-button-has-name.md
```diff
@@ -77,6 +77,22 @@ Image button element with accessible name through `aria-labelledby`
 <div id="id1">Submit</div>
 ```
 
+#### Passed example 5
+
+Accessible name does not only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/image-button-has-name.md
```diff
@@ -103,6 +119,14 @@ Image button with aria-labelledby that does not reference an id that exists in t
 <input type="image" name="submit" src="button.gif" aria-labelledby="id1" />
 ```
 
+#### Failed example 4
+
+Accessible name only consist of seperator characters.
```
carlosapaduarte: consist > consists
seperator > separator

_rules/SC2-4-2-page-has-title.md
```diff
@@ -34,7 +34,7 @@ The page contains at least one `title` element.
 
 ### Expectation 2
 
-The first `title` element contains [non-empty text](#non-empty).
+The first `title` element contains [text nodes](https://www.w3.org/TR/dom/#text) that do not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
WilcoFiers: "test nodes that do not only consist of unicode separator characters" is far from plain language. Please have a look at some of my previous suggestions for ideas.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -25,7 +25,7 @@ The rule applies to any HTML element with the [semantic role](#semantic-role) of
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty).
+Each target element has an [accessible name](#accessible-name) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
WilcoFiers: As before. This is not plain language.

_rules/SC3-1-1-html-has-lang.md
WilcoFiers: `attribute specifies a value that is not the empty string ("")` -> `attribute is not empty ("")`. The same should be done to all other uses of "is not the empty string"

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -25,7 +25,7 @@ The root element of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if
 
 ### Expectation
 
-The test target has a [non-empty](#non-empty) `lang` or `xml:lang` attribute.
+The test target has a `lang` or `xml:lang` attribute that is not the empty string ("").
```
WilcoFiers: `that is not the empty string ("")` -> `that is not empty ("")`

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -25,7 +25,7 @@ The root element of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if
 
 ### Expectation
 
-The test target has a [non-empty](#non-empty) `lang` or `xml:lang` attribute.
+The test target has a `lang` or `xml:lang` attribute that is not empty ("").
```
annethyme: @kasperisager and @WilcoFiers, can you both agree to this change? 

> the empty string ("") --> empty ("")

If yes, I will just go ahead and make this change to all rules that uses that wording.

_rules/SC1-1-1-image-has-name.md
```diff
@@ -23,9 +23,9 @@ The rule applies to HTML `img` elements or any HTML element with the [semantic r
 
 ### Expectation
 
-Each target element has an [accessible name](#accessible-name) that is [non-empty](#non-empty) or is marked as [decorative](#decorative).
+Each target element has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace), or is marked as [decorative](#decorative).
```
annethyme: @kasperisager and @WilcoFiers, can you both agree to this change? 

> does not only consist of [unicode separator characters]()

--> 

> is not only [whitespace]() (+ definition of "whitespace")

If yes, I will go ahead and make this change in all rules that uses this wording.

_rules/SC1-3-5-autocomplete-valid.md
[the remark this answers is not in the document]
kasperisager: Sure, it simplifies things, but it's also incorrect. Tab indices are only a fairly small part of determining whether something participates in sequential focus navigation or not.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -34,7 +34,7 @@ The page contains at least one `title` element.
 
 ### Expectation 2
```
annethyme: This rule is also being worked over in: https://github.com/auto-wcag/auto-wcag/pull/440

_rules/SC1-3-5-autocomplete-valid.md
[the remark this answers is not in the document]
WilcoFiers: We don't need _all_ elements in sequential navigation, we just need to exclude elements from the applicability that aren't. There are two ways to do that, disable or negative tabindex. Both are covered.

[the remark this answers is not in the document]
kasperisager: ...or by having the subtree in which the element resides marked inert due an open `<dialog>`. The point being that by describing what we actually _mean_ rather than indications thereof we hopefully end up with more robust rules.

pages/glossary/whitespace.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Whitespace
+key: whitespace
+---
+
+Whitespace are characters in the [Unicode separator categories](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
kasperisager: Seeing how this will be used in relation to things that aren't accessible names, the definition must also encompass e.g. control characters that are whitespace, such as tabs, newlines, carriage returns etc.

kasperisager: We might actually be able to simply rely on the Unicode `White_Space` property: https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt

_rules/SC1-3-5-autocomplete-valid.md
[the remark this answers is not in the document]
annethyme: @kasperisager, @WilcoFiers and I have discussed and agreed to go ahead with the "sequential focus navigation".

pages/glossary/whitespace.md
annethyme: @WilcoFiers, @kasperisager and I discussed and agreed to go ahead with @kasperisager's suggestion + examples of categories/characters.

annethyme: @kasperisager and @WilcoFiers, the whitespace definition has been updated as agreed and is ready for you guys to have a look at it.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -34,7 +34,7 @@ The page contains at least one `title` element.
...
 
-The first `title` element contains [non-empty text](#non-empty).
+The first `title` element has [text nodes](https://www.w3.org/TR/dom/#text) that does not only have [whitespace](#whitespace).
```
carlosapaduarte: does > do

_rules/SC2-4-2-page-has-title.md
```diff
@@ -166,6 +178,16 @@ Empty first `title`.
 </html>
 ```
 
+#### Failed example 5
+
+`title` only consist of whitespace.
```
carlosapaduarte: consist > consists

_rules/SC3-1-2-lang-valid.md
```diff
@@ -78,7 +82,7 @@ The `lang` attribute specified has a non-empty value & a valid primary language
 
 #### Passed example 4
 
-The `lang` and `xml:lang` attribute specified has a non-empty value & a valid primary language subtag.
+The `lang` and `xml:lang` attributes values that are not empty ("") and has valid primary language subtags.
```
carlosapaduarte: remove "that"
has > has a

annethyme:
"Final Call" started on April 11 and ends on April 25.

pages/glossary/whitespace.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Whitespace
+key: whitespace
+---
+
+Whitespace are characters that have the Unicode "White_Space" property in the [Unicode properties list](https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt).
```
EmmaJP: Perhaps "Whitespace comprises characters..."

"Whitespace are characters..." feels like incorrect grammar.

EmmaJP: Perhaps "Whitespace comprises characters..."

"Whitespace are characters..." feels like incorrect grammar.

EmmaJP: Perhaps "Whitespace comprises characters..."

"Whitespace are characters..." feels like incorrect grammar.

EmmaJP: Perhaps "Whitespace comprises characters..."

"Whitespace are characters..." feels like incorrect grammar.

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -165,76 +190,92 @@ Autocomplete is inappropriate for the type of field.
 <input type="number" autocomplete="email" />
 ```
 
+#### Failed example 6
+
+Autocomplete is not empty, but does not have any terms specified.
+
+```html
+<input autocomplete=" " />
```
WilcoFiers: I don't think this should fail. I think you should put back the "not only whitespace" thing in this rule.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -34,7 +34,7 @@ The page contains at least one `title` element.
 
 ### Expectation 2
 
-The first `title` element contains [non-empty text](#non-empty).
+The first `title` element has [text nodes](https://www.w3.org/TR/dom/#text) that do not only have [whitespace](#whitespace).
```
WilcoFiers: I think talking about "text nodes" is confusing. As best I can tell, a `title` element can not have more than one text node, similar to `script` or `style`.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -119,6 +119,18 @@ Valid `title` provided.
...
 ```
 
+#### Passed example 6
+
+`title` element does not only contain whitespace.
+
+```html
+<html>
+  <body>
+    <title>:-)</title>
```
WilcoFiers: This should fail WCAG, I don't think we should put that into a "passed" example.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -125,11 +125,27 @@ When `link` is off screen.
...
   </body>
 </html>
 ```
 
+#### Passed example 10
+
+`area` element with `href` attribute has accessible name.
+
+```html
+<area shape="rect" coords="0,0,82,126" href="http://www.w3.org/WAI" alt="Web Accessibility Initiative (WAI)">
```
WilcoFiers: This shouldn't be applicable. An `area` element isn't focusable unless it has an image associated with it. We should maybe exclude area from this, since it should be applicable to 1.1.1.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -125,11 +125,27 @@ When `link` is off screen.
...
+```
+
+#### Passed example 11
+
+`a` element where accessible name does not only consist of whitespace.
+
+```html
+<a href="http://www.w3.org/WAI">:-)</a>
```
WilcoFiers: As before, this fails WCAG, we shouldn't put it in a "passed" example.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
```diff
@@ -197,6 +213,30 @@ Not included in the accessibility tree.
...
+```html
+<a href="http://www.w3.org/WAI"></a>
+```
+
+#### Failed example 10
+
+`area` element with `href` attribute does not have accessible name.
+
+```html
+<area shape="rect" coords="0,0,82,126" href="http://www.w3.org/WAI">
```
WilcoFiers: As above, this element can't be reached by AT if it doesn't map to an image.

_rules/SC3-1-1-html-has-lang.md
```diff
@@ -99,12 +99,36 @@ The `lang` attribute specified has a non-empty value. The rule does not verify t
...
 #### Passed example 7
 
-The `xml:lang` attribute specified has a non-empty value. The rule does not verify the validity of the value specified and checks only for presence of a value.
+The `xml:lang` attribute specifies a value that is not empty (""). The rule does not verify the validity of the value specified and checks only for presence of a value.
 
 ```html
 <html xml:lang="xyz">
 ```
 
+#### Passed example 8
```
WilcoFiers: I don't think we should put these new examples in. These are still failures of the SC.

_rules/SC3-1-2-lang-valid.md
```diff
@@ -19,15 +19,19 @@ authors:
 
 ### Applicability
 
-Any DOM element, within the `body` of a webpage with a [non-empty](#non-empty) `lang` or `xml:lang` attribute.
+This rules applies to each HTML or SVG element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the `body` element and has a `lang` or `xml:lang` attribute that is not empty ("").
```
WilcoFiers: This is the wrong type of "within the `body`". This has to be a descendant in the composed tree, not in the light DOM tree.

_rules/SC4-1-2-button-has-name.md
```diff
@@ -111,6 +111,14 @@ Off screen elements should be tested.
 </html>
 ```
 
+#### Passed example 8
+
+Button has accessible name that is not only whitespace.
+
+```html
+<button>:-)</button>
```
WilcoFiers: As before, passing example shouldn't be SC failures IMO.

_rules/SC4-1-2-form-field-has-name.md
```diff
@@ -95,6 +95,17 @@ Explicit role.
 <div aria-label="country" role="combobox" aria-disabled="true">England</div>
 ```
 
+#### Passed example 6
+
+The accessible name is not only whitespace.
+
+```html
+<label>
+  :-)
```
WilcoFiers: As before, passing example shouldn't be SC failures IMO.

_rules/SC4-1-2-iframe-has-name.md
```diff
@@ -74,6 +74,15 @@ Usage of `aria-labelledby` attribute to describe the `iframe` content.
 </iframe>
 ```
 
+#### Passed example 4
+
+Accessible name is not only whitespace.
+
+```html
+<iframe title=":-)" src="../test-assets/SC4-1-2-frame-doc.html">
```
WilcoFiers: As before, passing example shouldn't be SC failures IMO.

_rules/SC4-1-2-role-attribute-has-valid-value.md
```diff
@@ -90,14 +90,30 @@ Element with multiple invalid `role` value.
 <input type="text" role="invalid role">
 ```
 
+#### Failed example 3
+
+Element with role attribute that is not empty (""), neither a valid `role` value.
+
+```html
+<input type="text" role=" ">
```
WilcoFiers: I'd make this inapplicable, allowing "only whitespace" in here. I don't think we should be this strict about it.

_rules/image-button-has-name.md
```diff
@@ -77,6 +77,22 @@ Image button element with accessible name through `aria-labelledby`
 <div id="id1">Submit</div>
 ```
 
+#### Passed example 5
+
+Accessible name is not only whitespace.
+
+```html
+<input type="image" name="submit" src="button.gif" alt=":-)" />
```
WilcoFiers: As before, passing example shouldn't be SC failures IMO.

pages/glossary/whitespace.md
WilcoFiers: Agreed with Emma's 4 posts here :-)

pages/glossary/letters-or-numbers.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Letters or numbers
+key: letters-or-numbers
```
WilcoFiers: Where is this definition used? I couldn't find it. If this isn't used, please remove.

pages/glossary/whitespace.md
EmmaJP: Not sure why that posted 4 times ... sorry, wasn't trying to drive home a point or anything.

EmmaJP: Duplicates removed

_rules/SC1-1-1-image-has-name.md
```diff
@@ -51,83 +51,99 @@ There is a known combination of a popular browser and assistive technology that
...
 ```html
 <img role="none" />
 ```
 
+#### Passed example 6
+
+The HTML `img` element has an accessible name that does not only consist of whitespace
+
+```html
+<img alt=":-)" />
```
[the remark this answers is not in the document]
EmmaJP: I'm curious about this. I can think of situations where this could be a suitable alt attribute, though perhaps not the best alt attribute value. And correct punctuation should be used in alt attributes. The only aspect of SC1.1.1 this might not meet is H86, which is a technique rather than a success criteria (for my understanding).

That said, I do think a better example should be used.

_rules/SC2-4-2-page-has-title.md
kasperisager: Not quite: https://github.com/act-rules/act-rules.github.io/pull/380#discussion_r275313664

_rules/SC1-3-5-autocomplete-valid.md
kasperisager: Why should the above not fail if something like the following does?

```html
<input autocomplete="." />
```

Both were obviously attempts at giving the attribute a non-zero-length value.

_rules/SC3-1-2-lang-valid.md
kasperisager: *Flat tree, yes.

_rules/SC4-1-2-role-attribute-has-valid-value.md
kasperisager: https://github.com/act-rules/act-rules.github.io/pull/430#discussion_r275361440

pages/glossary/whitespace.md
annethyme: @EmmaJP and @WilcoFiers , agree that it could need some improvement.

However, I had to look up the word "comprise" to make sure what it meant, and it's on the "Don't say" list for plain language on https://www.plainlanguage.gov/guidelines/words/use-simple-words-phrases/.

Could we use e.g. is made up of, consist of, includes, or another synonym (e.g. from https://www.thesaurus.com/browse/comprise)?

Maybe native English speaker @Brynanders also has an opinion on this?

pages/glossary/letters-or-numbers.md
annethyme: Didn't realise it was still there. Will remove.

_rules/SC1-3-5-autocomplete-valid.md
```diff
@@ -18,12 +18,12 @@ authors:
...
 ### Applicability
 
-The rule applies to any HTML `input`, `select` and `textarea` element with a [non-empty](#non-empty) `autocomplete` attribute, except if one of the following is true:
+The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` attribute that is not empty (""), except if one of the following is true:
 
 - The element is not [visible on the page](#visible-on-the-page), and not [included in the accessibility tree](#included-in-the-accessibility-tree)
 - The element is an `input` element with a `type` property of `hidden`, `button`, `submit` or `reset`
 - The element has a `disabled` or `aria-disabled="true"` attribute
-- The element has `tabindex="-1"` and has a [semantic role](#semantic-role) that is not a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles)
+- The element does not participate in [sequential focus navigation](https://www.w3.org/TR/html/editing.html#sec-sequential-focus-navigation) and has a [semantic role](#semantic-role) that is not a [widget role](https://www.w3.org/TR/wai-aria-1.1/#widget_roles).
```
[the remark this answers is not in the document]
annethyme: So removing `disabled` attribute, because that means it's not participating in sequential focus navigation and does not have a widget role?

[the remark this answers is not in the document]
annethyme: @WilcoFiers, I don't remember why the "does not participate in sequential focus navigation" should have been removed...

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
annethyme: @WilcoFiers, The rule is applicable to "any HTML element with the semantic role of link that is included in the accessibility tree." It doesn't have to be focusable. 
Since it's about accessible names, it makes sense to align the rule with what is presented to a screen reader. 
And according to https://www.w3.org/TR/html-aria/ an "area with a href" has the implicit ARIA semantics of role=link.

I would say that something that is an non-text content but has the role of a link should fall under both 1.1.1 and 2.4.4/2.4.9.

_rules/SC1-1-1-image-has-name.md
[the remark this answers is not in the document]
annethyme: Please do keep in mind that this rule tests whether there is an accessible name or not.
It does not test whether the accessible name is descriptive. This should be tested by another rule.

I do think we need to document "edge cases" where we might disagree on which rule should be catching a particular WCAG violation. 
Or else we will be back in the mud where one tool/method says a rule is passed, while another says it is not applicable.

annethyme:
AGENDA ITEM for CG call on April 25:

Should we allow Passed test cases that fails the success criterion? 
See discussion on https://github.com/act-rules/act-rules.github.io/pull/430/files#r275256380 (and other comments from Wilco on the same thing further down in the PR for more examples).

_rules/SC1-3-5-autocomplete-valid.md
annethyme: @WilcoFiers, @kasperisager and @annethyme agreed on a call to try to go with the current wording, since it's not all types of whitespace that will be trimmed in attributes like this, and differentiating between the different types of whitespace will complicate the rule, which we don't really know is necessary. 
However, if we, when running the rule in real life settings, see a lot of cases where we are catching spaces and control characters, that are not really causing any accessibility issues, we might have to re-evaluate this decision.

_rules/SC4-1-2-role-attribute-has-valid-value.md
annethyme: See https://github.com/act-rules/act-rules.github.io/pull/430/files#r278572430

pages/glossary/whitespace.md
annethyme: @brynanders suggested just shortening it to "Characters that have the Unicode "White_Space" property", which I think is much cleaner, so I'll go with that until someone objects :)

_rules/SC2-4-2-page-has-title.md
annethyme: Agreed in https://github.com/act-rules/act-rules.github.io/pull/380#discussion_r275313664 to leave it as it is.

_rules/SC1-1-1-image-has-name.md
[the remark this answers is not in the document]
annethyme: @WilcoFiers, @kasperisager and @annethyme agreed on a call to write up a proposal that states that Passed rules should also be passing the accessibility requirements (and then we will have to ensure better coverage through Failed test cases instead). 
However, we do currently have many rules where this is not the case, and this will have to be corrected across the board for all rules. 

I suggest that when a proposal on this has been created and accepted, we clean up all rules. 
However, I would like to keep the rules in this PR as they are for now. 
I could go and clean up the newly added test cases to align them with this future proposal, but in many rules there will be existing test cases that also break this proposed requirement, and it will be messy if I clean it up halfway by doing it just for the newly added test cases, but not the old ones, and I will never finish work on this PR if I try to do it for all test cases for all rules in this PR.

Would @WilcoFiers be okay with me merging as is (since this is the only substantial thing that has come in during the Final Call), knowing that new guidelines for test cases might be on their way?

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
WilcoFiers: @annethyme This `area` element isn't included in the accessibility tree.

annethyme: If I just use arrow navigation in VoiceOver I don't have any trouble getting to this element. It is read out as a link with the alt text as it's accessible name. Which is in line with the ARIA in HTML spec that says that the area has the role of link - and not only if it has an associated image.
I vote for keeping it this test case as is.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
annethyme: Related to https://github.com/act-rules/act-rules.github.io/pull/430/files#r280081504:

I can easily navigate to this using arrow navigation in VoiceOver.
It has the role of link, but no accessible name, so I think it is a perfectly valid test case.

_rules/SC2-4-4+2-4-9+4-1-2-link-has-name.md
kasperisager: I'm seeing the same thing as Anne and cannot find anything in neither the Core nor HTML AAM that would cause an `<area>` element with an `href` attribute to _not_ be included in the accessibility tree.
image

annethyme:
Final Call ended on April 25. 
Only smaller changes made during the Final Call phase, so a new Final Call is not required.
Ready for merge.

PR #437, opened 2019-03-12
chore: WCAG ACT RULES CG Website Update

jkodu:
Preview of the website: https://gracious-goodall-6ccd8e.netlify.com

image

It is hard to put into words the amount of changes gone into this PR.

An overview is as below
- introduced a `__deprecated__` folder for all old pages.
- migrated site infrastructure from [Jekyll](https://jekyllrb.com/) to [Gatsby](https://gatsbyjs.org)
- unleashed some creative freedom along the way
- testcases made available at [https://gracious-goodall-6ccd8e.netlify.com/testcases.json](https://gracious-goodall-6ccd8e.netlify.com/testcases.json)

> Note: I have setup netlify to showcase the site in the interim, this will be removed on a later date

Closes issue:
- As a consequence of updating the site, fixes - https://github.com/auto-wcag/auto-wcag/issues/414


audreymaniez:
Hey!

I have only some general comments on html structuration and accessibility issue (not on functionnality). I hope it's usefull. 

- the lang attribut is missigin on the html element : `<html lang="en">`
- in the [page rule list](https://gracious-goodall-6ccd8e.netlify.com/rules/) every rule is in a `<main>` tag, but only one `<main>` tag is allowed. Maybe structured the whole list with a ul  and li elements ? (or role)list and listitem ?)
- In the rule description markdown , quotes to mark code (<code>\`...\`</code> ) are not converted to `<code>` tag 
- There's a contrast issue on buttons on [that page](https://gracious-goodall-6ccd8e.netlify.com/pages/contribute/ the blue/white combination is not enough). But I think you know it :) 
- Suggestion: In a page rule, the right column could be a little more structured like adding headings (elements with class=heading could be html headings) ?
-  Suggestion: In page rule, code blocks are on one unqiue line (no break return in code example). Is it wanted ? Maybe  remove the  `white-space: unset;` property make the code block more readable ?instead beeing in one single line ?
pre[class*="language-"] {
    white-space: unset;
} 

When zooming in (graphical zoom), the menu is hidden in the left (the responsive view is taking place), there's multiple issues : 

awcag2


- The menu is made visible when hovering (mouse) it, but when menu item receive tab focus (keyboard), the menu is not visible (keyboard user don't know where they are), but it should be visible : can you make all the menu visible when one item has the focus ? 
- The menu can't be scroll down, depending on size of the sreen, only the first item is visible
- there's a z-index issue with code block (class="gatsby-highlight") : the code block is upon the menu.

Screenshot_2019-03-13 https gracious-goodall-6ccd8e netlify com

jkodu:
Thanks @audreymaniez for taking the time to look at this. I am aware of some of the issues you listed above, & will put in some time to fix the same.

nitedog:
Great work @jkodu, many thanks!

Critical before launch:
- Add footer along the lines of: [ACT Rules CG](https://www.w3.org/community/act-rules) is a [Community Group](https://www.w3.org/community/) of the [World Wide Web Consortium (W3C)](https://www.w3.org/). [License](license.html). Developed with the support of the EU-funded [WAI-Tools Project](https://www.w3.org/WAI/about/projects/wai-tools/).
- Add W3C and WAI logos (maybe in the footer or on the top-right, like on the [WCAG-EM Report Tool](https://www.w3.org/WAI/eval/report-tool/))

Future improvements to consider:
- Provide filter/search functionality to browse published rules
- Provide a list of rules in development and deprecated rules
- Provide a mechanism to provide feedback on published rules
- Show existing implementations for each published rule
- Revise all documentation and information architecture

jkodu:
We discussed this on the call, and the group is happy to move to the new website.

There will be improvements coming both in functionality and documentation.

annethyme:
@jkodu, would it maybe be possible to find a font for the website that has better support for the Scandinavian letters? 
As you can see in the screenshot in the names under "Authors", "ø" is shown in a funny way in the current font:

Screenshot 2019-03-21 at 16 20 51

If you use any type of standard font (verdana, arial, times new roman, georgia etc.) the letters should look fine.

EmmaJP:
The site does not seem to respect browser default font size changes. Maybe don't set the body font-size with pixels. Use a relative unit instead.

Brynanders:
There is a zoom issue which prevents scrolling in the sidebar menu https://www.screencast.com/t/vyqY5qAwz.

annethyme:
Could we consider finding colors and logo shapes for the ACT Rules CG website that does not seem to be directly derived from the aXe-Core logo? 
image

WilcoFiers:
We need to remove the logo and deal with that separately.

WilcoFiers:
@kasperisager We'll get to Anne's comments when Jey is back. Regarding Gatsby over Jekyll. Mostly because Ruby is a bottleneck to who can contribute to the website. We also need the ability to dynamically load in data, rather than have it on build-time. We'd need to build a new framework on top of the site, so rather than have Jekyll + React or Vue, and still have this dev bottleneck, Jey wanted to try and rebuild with Gatsby.

kasperisager:
Sounds great! 💯

jkodu:
@annethyme 
- The logo is removed now, it was inspired from A for Accessibility anyway.
- The color(s) were inspired from WAI tools website https://www.w3.org/WAI/ 
image, I have updated them again to closely align with WAI tools.
- I have updated the font pairings to [Roboto](https://fonts.google.com/specimen/Roboto?selection.family=Roboto) for headings, and [Lora](https://fonts.google.com/specimen/Lora?selection.family=Lora) for body, this seems to have a good character set coverage and nice legibility based on this article - https://www.reliablepsd.com/ultimate-google-font-pairings/ 

@EmmaJP 
- Thank you, I have updated the font-size to relative units.

@nitedog 
- Added the requested critical information from here - https://github.com/auto-wcag/auto-wcag/pull/437#issuecomment-474864111
image

@kasperisager 
To add to @WilcoFiers explanation here - https://github.com/auto-wcag/auto-wcag/pull/437#issuecomment-478628857, GatsbyJs allows us to go truly static site. I have listed a few benefits.

1. Ruby plugins were hard to enhance and manage with Jekyll. We had plugins for the below actions: 
 - Generating test cases json & assets
 - Converting code snippet markdown to embedded iframes within rules page
 - Generating list of definitions used at the bottom of rule page
 - Crosslinking definition with list of usages in rules
Now all of the above are done in a better architectural pattern, than mangling either markdown or generated html, as Gatsby allows for a component based approach.
2. Everything is truly static generated now at build time, including fetching implementation data and generating pages. This fulfils something that we have discussed in the past - https://github.com/auto-wcag/auto-wcag/pull/361#pullrequestreview-180637652
3. We now have a better handle on layouts and being able to template different views with out multiple liquid if else tags (as in jekyll), as Gatsby allows for component based templating.
4. Everything is JavaScript based, and we had limited Ruby know-how.
5. Asides above, there are a ton of features that Gatsby offers - https://medium.freecodecamp.org/why-you-should-use-gatsbyjs-to-build-static-sites-4f90eb6d1a7b, which I am sure you are aware of, but my favourite here is a potential for having Gatsby CMS in the future, which will make our life much easier to update pages.

jkodu:
**Note:** Upon merging, this will refresh the website on `gh-pages` branch.

EmmaJP:
Just taking a closer look at the new website preview. Noting things I've noticed in this comment.

- The rules have a column on the right that lists a "Table of contents", some list items have the text nested in a p element and others don't, and it seems to be affecting the indentation. It is also unavailable for smaller screen views and when zoomed in, when jump-links to at least the h2 headings could be helpful.
- I would question the use of all caps for such large quantities of page text. It reduces readability and there are better ways to distinguish different areas of content.
- Is there a reason for completely avoiding header and footer regions and putting that information in the menu? It seems to be complicating things unnecessarily.
- It might be an idea to have proportional padding either side of the main content so that the padding takes up less space on smaller screen sizes. Perhaps a percentage of the main element width rather than 5 rem, possibly with a max or min value at certain break points.
- As someone pointed out, the link hover/focus colour does not contrast sufficiently with the dark blue background in the menu, or the white content background.

Nice to see the font size respecting and adapting to browser preferences. Thank you for the hard work @jkodu.

PR #440, opened 2019-03-14
Rule update: "HTML page has a title"

annethyme:
"contain" --> "descendant" in Expectation 1. 
Editorial cleanup of "Background" links

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/316

# How to Review And Approve

### DIFF
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -47,7 +47,6 @@ _There are no major accessibility support issues known for this rule._
 ## Background
 
 - [Understanding Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
-- [WCAG 2.1 Technique G88: Providing descriptive titles for Web pages](https://www.w3.org/WAI/WCAG21/Techniques/general/G88)
 - [WCAG 2.1 Technique H25: Providing a title using the title element](https://www.w3.org/WAI/WCAG21/Techniques/html/H25)
 
 
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -34,7 +34,7 @@ The [document element](https://www.w3.org/TR/dom/#document-element) has at least
 
 ### Expectation 2
 
-The [child text content]() of the first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
+The [child text content](https://www.w3.org/TR/html/infrastructure.html#child-text-content) of the first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
 
 ## Assumptions
 
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -34,7 +34,7 @@ The [document element](https://www.w3.org/TR/dom/#document-element) has at least
 
 ### Expectation 2
 
-The [child text content](https://www.w3.org/TR/html/infrastructure.html#child-text-content) of the first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
+The [child text content](https://www.w3.org/TR/html/infrastructure.html#child-text-content) of the first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) does not only have [whitespace](#whitespace).
 
 ## Assumptions
 
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -22,13 +22,13 @@ authors:
 
 ### Applicability
 
-The rule applies to any page where the root element is an `html` element, and which is not embedded in another page.
+The rule applies to any page where the [root](https://www.w3.org/TR/dom41/#concept-tree-root) element is an `html` element, and which is not embedded in another page.
 
 **Note**: Pages may be embedded inside other pages through elements such as iframes and object elements.
 
 ### Expectation 1
 
-The root element has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
+The [root](https://www.w3.org/TR/dom41/#concept-tree-root) element has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
 
 **Note**: The `title` element exists in other namespaces such as SVG. These are not HTML `title` elements and should be ignored for this rule.
 
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -34,7 +34,7 @@ The [document element](https://www.w3.org/TR/dom/#document-element) has at least
 
 ### Expectation 2
 
-The first `title` element contains [text nodes](https://www.w3.org/TR/dom/#text) that do not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
+The [child text content]() of the first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
 
 ## Assumptions
 
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -34,7 +34,7 @@ The [document element](https://www.w3.org/TR/dom/#document-element) has at least
 
 ### Expectation 2
 
-The first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) has [children](https://www.w3.org/TR/dom/#concept-tree-child) that are [text nodes] that are not only [whitespace](#whitespace).
+The first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) has [children](https://www.w3.org/TR/dom/#concept-tree-child) that are [text nodes](https://www.w3.org/TR/dom/#text) that are not only [whitespace](#whitespace).
 
 ## Assumptions
 
--- a/_rules/SC2-4-2-page-has-title.md
+++ b/_rules/SC2-4-2-page-has-title.md
@@ -22,9 +22,9 @@ authors:
 
 ### Applicability
 
-The rule applies to any page where the [root](https://www.w3.org/TR/dom41/#concept-tree-root) element is an `html` element, and which is not embedded in another page.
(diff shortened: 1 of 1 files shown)


_rules/SC2-4-2-page-has-title.md
```diff
@@ -28,13 +28,13 @@ The rule applies to any page where the root element is an `html` element, and wh
 
 ### Expectation 1
 
-The page contains at least one `title` element.
+The root element has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
```
kasperisager: "root" -> https://www.w3.org/TR/dom41/#concept-tree-root

_rules/SC2-4-2-page-has-title.md
```diff
@@ -46,11 +46,10 @@ _There are no major accessibility support issues known for this rule._
 
 ## Background
 
-- [https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-title.html](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-title.html)
-- [https://www.w3.org/WAI/WCAG20/quickref/?showtechniques=242#qr-navigation-mechanisms-title](https://www.w3.org/WAI/WCAG20/quickref/?showtechniques=242#qr-navigation-mechanisms-title)
-- [https://www.w3.org/TR/WCAG20-TECHS/G88.html](https://www.w3.org/TR/WCAG20-TECHS/G88.html)
-- [https://www.w3.org/TR/WCAG20-TECHS/H25.html](https://www.w3.org/TR/WCAG20-TECHS/H25.html)
-- The WCAG 2.0 Techniques already contain examples and code snippets to illustrate which content passes or fails the test. Whenever possible {{site.title}} refers to those. Another source for test cases is the W3C Before and After Demonstration.
+- [Understanding Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
+- [WCAG 2.1 Technique G88: Providing descriptive titles for Web pages](https://www.w3.org/WAI/WCAG21/Techniques/general/G88)
```
audreymaniez: Is this technique G88 really relate to that rule ?  It seems the G88 is for the rule you write in the #380 PR ? I just suggest remove the relation for less confusion on the scope of the rule.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -22,19 +22,19 @@ authors:
...
-The page contains at least one `title` element.
+The [root](https://www.w3.org/TR/dom41/#concept-tree-root) element has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
 
-**Note**: The `title` element exists in other namespaces such as SVG. These are not `title` elements for HTML document and should be ignored.
+**Note**: The `title` element exists in other namespaces such as SVG. These are not HTML `title` elements and should be ignored for this rule.
 
 ### Expectation 2
 
-The first `title` element contains [non-empty text](#non-empty).
+The first `title` element contains [letters or numbers](#letters-or-numbers).
```
carlosapaduarte: Just to make sure I understand, this means that the following title would fail, because it only uses punctuation symbols, not letters or numbers?
`<title>#$@&%*!</title>`
And should it fail?

annethyme: @carlosapaduarte, yes, with the current definition it would. The old "non-empty" definition was actually also only looking for letters and numbers, it was just much less obvious. 

I have struggled with the same question for headings. If I create a page about smileys, I would say that ":-)" and ":'-(" would be fine as headings. 

What do you think? Another definition that I have used in some places is "contains text nodes that do not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153)"
So `<title>#$@&%*!</title>` would be a Passed outcome, but `<title> </title>`would be Failed.

Do you think that would be more appropriate?

We do also have another rule that checks if the title is descriptive, so I think I am actually leaning towards this approach...

carlosapaduarte: @annethyme, yes I do prefer the "contains text nodes that do not only consist of Unicode separator characters" definition. 

I don't think that both the `#$@&%*!` title and the `:-)` heading are accessibility violations, so they shouldn't be failed by a rule that checks just a syntactic aspect (in this case, if the page has a title or not).

annethyme: Change has been made and test cases updated to reflect this.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -22,19 +22,19 @@ authors:
 
 ### Applicability
 
-The rule applies to any page where the root element is an `html` element, and which is not embedded in another page.
+The rule applies to any page where the [root](https://www.w3.org/TR/dom41/#concept-tree-root) element is an `html` element, and which is not embedded in another page.
```
kasperisager: The root of a document will always be the document itself so no valid HTML documents will ever match this applicability. The term we're looking for is "document element": https://www.w3.org/TR/dom/#document-element

_rules/SC2-4-2-page-has-title.md
```diff
@@ -166,11 +174,21 @@ Empty first `title`.
 </html>
 ```
 
+#### Failed example 5
+
+The `title` only contains a seperator character.
```
carlosapaduarte: typo: seperator -> separator

Above there is another typo. On passed example 2 "This page give" -> "This page gives"

_rules/SC2-4-2-page-has-title.md
```diff
@@ -22,19 +22,19 @@ authors:
...
-The rule applies to any page where the root element is an `html` element, and which is not embedded in another page.
+The rule applies to any page where the [document element](https://www.w3.org/TR/dom/#document-element) is an `html` element, and where the page is not embedded in another page.
 
-**Note**: Pages may be embedded inside other pages through elements such as iframes and object elements.
+**Note**: Pages may be embedded inside other pages through elements such as `iframe` and `object` elements.
 
 ### Expectation 1
 
-The page contains at least one `title` element.
+The [root](https://www.w3.org/TR/dom41/#concept-tree-root) element has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
```
audreymaniez: Writing considerations  :-) In the #380 you write `<title>` and in that rule `title` without angle brackets.

annethyme: You are right. I have removed the brackets in the other rule, to align with the general style used in Auto-WCAG...

kasperisager: root element -> document element.

_rules/SC2-4-2-page-has-title.md
```diff
@@ -22,19 +22,19 @@ authors:
...
-The page contains at least one `title` element.
+The [document element](https://www.w3.org/TR/dom/#document-element) has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
 
-**Note**: The `title` element exists in other namespaces such as SVG. These are not `title` elements for HTML document and should be ignored.
+**Note**: The `title` element exists in other namespaces such as SVG. These are not HTML `title` elements and should be ignored for this rule.
 
 ### Expectation 2
 
-The first `title` element contains [non-empty text](#non-empty).
+The first `title` element contains [text nodes](https://www.w3.org/TR/dom/#text) that do not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
WilcoFiers: Can we simplify this a little by doing:

> The first `title` element contains [text](https://www.w3.org/TR/dom/#text) that does not only consist of [whitespace (specifically, Unicode separator characters)](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).

annethyme: @WilcoFiers, we don't find your suggestion to be as unambiguous as we would like.

WilcoFiers: Which part of it? Alternative proposal:

> ... contains [text content](https://www.w3.org/TR/dom41/#dom-node-textcontent) that does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153) (i.e. space, line breaks and other whitespace characters)

Text content is what the HTML 5.2 spec says needs to be used to derive the document title from the title element, so it's the right technical term, and it reads better than "text nodes" IMO.

kasperisager: HTML defines "child text content" as a term (https://www.w3.org/TR/html/infrastructure.html#child-text-content), not "text content". In general, I don't really think it's appropriate to refer do DOM attributes in the manner proposed, e.g. use "text content" to mean `node.textContent` or "document title" to mean `document.title`. If referring to DOM attributes or properties, be explicit about it.

WilcoFiers: Fair point. But then shouldn't this simply be using `document.title`?

kasperisager: I don't think it should, nor do I think that rules in general should refer to specific APIs, be they DOM or something else. That's an implementation detail and rules shouldn't dictate the "how" but rather the "what". The "what" in this this case is the concatenation of the data of the text nodes that are children of the `<title>` element, i.e. the "child text content". The "how" _could_ be `document.title`.

WilcoFiers: I don't mean to say to literally access the document.title property. You're right, that's clearly implementation. I mean using "document title", linking to the algorithm for deriving it, the same way we talk about "accessible name" and referencing that algorithm.

kasperisager: That makes perfect sense and for that very same reason I'd also suggest using the term "child text content" similar to how we use the term "accessible name". We could even add "document title" to the glossary and define it as "the child text content of the first `<title>` element that is a descendant of the document".

_rules/SC2-4-2-page-has-title.md
```diff
@@ -22,19 +22,19 @@ authors:
...
-The page contains at least one `title` element.
+The [document element](https://www.w3.org/TR/dom/#document-element) has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
 
-**Note**: The `title` element exists in other namespaces such as SVG. These are not `title` elements for HTML document and should be ignored.
+**Note**: The `title` element exists in other namespaces such as SVG. These are not HTML `title` elements and should be ignored for this rule.
 
 ### Expectation 2
 
-The first `title` element contains [non-empty text](#non-empty).
+The [child text content](https://www.w3.org/TR/html/infrastructure.html#child-text-content) of the first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) does not only consist of [Unicode separator characters](https://www.unicode.org/versions/Unicode11.0.0/ch04.pdf#G134153).
```
WilcoFiers: Please use plain language here. We can replace "non-empty" as a definition, but we shouldn't be inlining complex definitions like this. That makes this very difficult to use for less technical people.

kasperisager: See https://github.com/auto-wcag/auto-wcag/pull/380#discussion_r270832677

_rules/SC2-4-2-page-has-title.md
```diff
@@ -22,19 +22,19 @@ authors:
...
-The page contains at least one `title` element.
+The [document element](https://www.w3.org/TR/dom/#document-element) has at least one [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) that is an HTML `title` element.
 
-**Note**: The `title` element exists in other namespaces such as SVG. These are not `title` elements for HTML document and should be ignored.
+**Note**: The `title` element exists in other namespaces such as SVG. These are not HTML `title` elements and should be ignored for this rule.
 
 ### Expectation 2
 
-The first `title` element contains [non-empty text](#non-empty).
+The first HTML `title` element that is a [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) of the [document element](https://www.w3.org/TR/dom/#document-element) has [children](https://www.w3.org/TR/dom/#concept-tree-child) that are [text nodes](https://www.w3.org/TR/dom/#text) that are not only [whitespace](#whitespace).
```
annethyme: @kasperisager, should it have been specified whether the `title` element we are looking for is a descendant in the flat tree or something else?

kasperisager: That’s actually a really good point. In this, we’re _not_ looking for a descendant in the flat tree, which is not something that has been made explicitly previously. We’ve only made it explicit when we’re specifically looking for either a shadow-including descendant or a descendant in the flat tree. What are people’s thoughts on this? Is an unqualified use of the term “descendant” unambiguous? I’m partial to thinking that it is.

annethyme: @kasperisager and I have agreed among ourselves that we feel the wording is fine. So unless someone else disagrees, we will leave it as is.

Issue #365, opened 2018-12-06
Update Rule: ARIA state or property has valid value


jkodu:
Rule: https://auto-wcag.github.io/auto-wcag/rules/SC4-1-2-aria-state-or-property-has-valid-value.html

In the below scenario, the rule fails SC 4.1.2, but it should fail SC 1.3.1 instead, because 4.1.2 does not tackle relationships between owned and context elements.

Suggestions:
- We add 1.3.1? but there is already 2 other PR for a rule which goes into greater detail of owned and context elements - https://github.com/auto-wcag/auto-wcag/pull/255 and https://github.com/auto-wcag/auto-wcag/pull/245
- Change the rule to only look at STATES and not PROPERTY.

Thoughts welcome.

![image](https://user-images.githubusercontent.com/20978252/49580384-b4bb4980-f946-11e8-82a7-3df536483b0e.png)

jkodu:
Todo:

Go through `aria` spec, and sort the properties into the respective S.C bucket.

jkodu:
@WilcoFiers - see above for tabulation of SC vs states and properties.

annethyme:
I am assigning this rule to myself since I am the owner of this rule inside the WAI-Tools project. It has a deadline of March 31.

annethyme:
Related to https://github.com/auto-wcag/auto-wcag/issues/376

annethyme:
Building further on Jey's analysis...

State/ Property | type | SC | Description
---|---|---

aria-activedescendant | property | []() | Identifies the currently active element when DOM focus is on a composite widget, textbox, group, or application.


aria-atomic | property | [4.1.3](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html) | Indicates whether assistive technologies will present all, or only parts of, the changed region based on the change notifications defined by the aria-relevant attribute.

aria-autocomplete | property | [1.3.5](https://www.w3.org/TR/WCAG21/#identify-input-purpose) | Indicates whether inputting text could trigger display of one or more predictions of the user's intended value for an input and specifies how predictions would be presented if they are made.

aria-busy | state | [1.3.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html) | Indicates an element is being modified and that assistive technologies MAY want to wait until the modifications are complete before exposing them to the user.

[...]

PR #441, opened 2019-03-14
Rule update: ARIA state/property has valid value

annethyme:
Rule limited to only looking at required states and properties that are included in the accessibility tree and does not have a fallback value.

My suggestion is that we keep this rule very strict and limited , since we want to map it to a WCAG success criterion - and then (in time) create another rule that does not mention any WCAG success criteria, but has WAI-ARIA 1.1 as it's requirement, and then let that rule check if everything has a valid value. 

Closes issue: 
- https://github.com/auto-wcag/auto-wcag/issues/376 (Is it possible to limit the applicability of "ARIA state or property has valid value"?) 
- https://github.com/auto-wcag/auto-wcag/issues/365 (Update Rule: ARIA state or property has valid value) 

# How to Review And Approve


_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
```diff
@@ -21,26 +21,27 @@ authors:
...
-**Note:**  For example, anything referenced through `aria-labelledby` does not have to be [included in the accessibility tree](#included-in-the-accessibility-tree) in order for it to become part of the [accessible name](#accessible-name).
-- The ARIA `state` or `property` is being used to comply to WCAG.
+- This rule assumes that elements that are not [included in the accessibility tree](#included-in-the-accessibility-tree) or are [focusable](#focusable) can still impact users if they are used for the [accessible name computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te). Therefore the applicability of this rule also includes `aria-*` attributes that participate in the accessible name computation as well as [WAI-ARIA properties](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that are specified on elements that are included in the accessibility tree or are focusable.
```
audreymaniez: I understand why there's a note on elements that are not included in the accessibility tree, but I don't understand why you point "or are focusable" ?

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
```diff
@@ -21,26 +21,27 @@ authors:
...
 
-### Expectation 1
+### Expectation
 
 Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
 
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that specifies the target attribute. 
+For value types `ID Reference` and `ID Reference List` at least one of the elements with the given ids exists in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) or [shadow tree](https://www.w3.org/TR/dom41/#shadow-trees) as the element that specifies the target attribute. 
 
 For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
```
ShadowBB: This is irrelevant now and can be removed.

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
```diff
@@ -21,26 +21,27 @@ authors:
...
 
-**Note:** Only for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.
-
 **Note:** 
 For value type `URI`, this rule does not require that the destination URI exists.
```
ShadowBB: This is irrelevant now and can be removed.

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
ShadowBB: This is irrelevant now and can be removed.

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
```diff
@@ -21,26 +21,27 @@ authors:
 
 ### Applicability
 
-Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
+Any [WAI-ARIA required state or property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) that 
+- is specified on an HTML or SVG element 
+- is one of the `aria-*` attributes included in the [accessible name computation](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te), or is specified on an element that is [included in the accessibility tree](#included-in-the-accessibility-tree) or is [focusable](#focusable),
```
ShadowBB: Accessible name computation can be removed.

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
```diff
@@ -21,26 +21,27 @@ authors:
 
 ### Applicability
 
-Any [non-empty](#non-empty) [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element.
+Any [WAI-ARIA required state or property](https://www.w3.org/TR/wai-aria-1.1/#requiredState) that
```
ShadowBB: The applicability now only refers to aria-controls and no other property. We could just state that in the applicability and the name of the rule.

annethyme: @ShadowBB , we discussed internally if that was the way forward for this rule. 
While it's the only required property in WAI-ARIA 1.1, there could potentially also be roles and requirements specified in other standards, e.g. EPUB. 
Having "broader" scoped rule could actually catch something in other standards, that maps to a completely different success criterion, I guess.
So I guess I am for your suggestions of limiting the rule to explicitly `aria-controls`, and then we can be completely sure what we are testing...

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
```diff
@@ -1,9 +1,9 @@
 ---
-name: Required ARIA property has valid value
+name: Required aria-conrols has valid value
```
kasperisager: `conrols` -> `controls`.

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
annethyme: aria-controls is required for the roles `scrollbar` and `combobox`. I have now limited the rule to only look for `aria-required` on role `scrollbar`, since `combobox` is a story of its own, see https://github.com/w3c/aria/issues/716.

_rules/SC-1-3-1-required-aria-state-or-property-has-valid-value.md
annethyme: Whole assumption removed now, since scope of rule has been changed once again.

annethyme:
For naming of the rule I am not sure whether to go with:
- aria-controls for scrollbar has valid value
- aria-controls for scrollbar references existing element

I think the latter is more precise, but also a bit longer. 
In the first one it is a bit unclear whether we are just looking for the right format or also the existence of the elements.

ShadowBB:
> 
> 
> For naming of the rule I am not sure whether to go with:
> 
>     * aria-controls for scrollbar has valid value
> 
>     * aria-controls for scrollbar references existing element
> 
> 
> I think the latter is more precise, but also a bit longer.
> In the first one it is a bit unclear whether we are just looking for the right format or also the existence of the elements.

I would go with the second name as it is more precise and only slightly longer.

audreymaniez:
I don't really understand how you get from all the properties to only one. But if I understand the general idea, you want to focus on required properties only, and the role scrollbar seems to have  other required ARIA properties. Does that mean that it will be one rule per required property (and per role) ?

annethyme:
@audreymaniez, great question!
The thing is, that for most required states and properties, a fallback value is specified for that state/property for that role in the table section "Implicit Value for Role:"
For `scrollbar` the only required state/property that does not have an implicit value defined is `aria-controls`.
And through the entire WAI-ARIA 1.1 spec, `aria-controls` for `scrollbar` is the only required state/property that does not have an implicit value defined. And then there is the `combobox`, but here it has to be on a descendent of a specific type, and only when the thing is expanded.

annethyme:
@WilcoFiers, I think it's because I opened the rule, started working on it, and then renamed the file somewhere down the line.
I guess I'll just have to add the original rule file again to make it un-deleted...

annethyme:
@WilcoFiers, and it's back again...

annethyme:
By the way, I don't personally like this rule very much.
There is a lot of code to check for one specific edge case, just because we really want to connect a WAI-ARIA validation issue back to a WCAG success criterion.

This non-WCAG rule checks for all cases of invalid state/property values, but we list any WCAG success criteria as an accessibility requirement, because it could be many different success criteria, or non at all, that are impacted depending on the state/property, and the semantic role of the element that it is specified on: https://github.com/auto-wcag/auto-wcag/pull/464

But is this even a problem?

WilcoFiers:
> Anne wrote:
> By the way, I don't personally like this rule very much.

xD Then why are you writing it? I agree, this seems far too much of an edge case to worry about. As an implementor, I would certainly have my doubts about taking on the overhead for a rule that has such a narrow scope.

@annethyme seriously, write the rules you feel good about writing.

annethyme:

PR #447, opened 2019-03-19
Rule update: Form field label is descriptive (cc0f0a)

annethyme:
## Important context for changes: 
- Adding possibility of letting some of the purpose be determinable from the context of the form field. Closes issue: https://github.com/auto-wcag/auto-wcag/issues/423
- Following the implications of [this comment](https://github.com/auto-wcag/auto-wcag/pull/323#discussion_r229743236), that changed the scope of the rule from labels that are "visible AND included in the accessibility tree" to those "visible OR included in the accessibility tree". If we are including invisible labels, I think ALL accessible names have to be included in the Applicability. Test cases extended to include `aria-label` too.

## How to Review And Approve


_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,37 @@ authors:
 
 ### Applicability
 
-This rule applies to any HTML `label` element or other element referenced by `aria-labelledby` that:
+This rule applies to each element that:
```
kasperisager: Should "each" be "any"?

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,37 @@ authors:
...
-
-**Note**: The list of form field roles is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
-- have a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
+**Note**: The list of applicable [semantic roles](#semantic-roles) is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
+- inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
 - does not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
+
+**Note:**
 - The `option` role is not part of the list of applicable roles, because it does not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). This means [WCAG 2.1](https://www.w3.org/TR/WCAG21/) does not require it to have an accessible name.
```
kasperisager: The `option` role is not part of the list as it has a required context role that inherits from the `select` role. It is therefore not applicable per the bullet point.

annethyme: Please note that after this change, this rule is a bit out of sync with this one: https://auto-wcag.github.io/auto-wcag/rules/SC4-1-2-form-field-has-name.html (published). 
However, since the currently edited rule is more correct, I don't consider this a problem. 
An issue might have to be raised for updating the other rule.

ShadowBB:
Hey Anne,

Really liking what you did with this so far. If I read this correctly it will fail if it has both an invisible associated label and an (for example) aria-label and the associated label is not descriptive. Is this the intention? Because i think the aria-label should win out and the invisible label doesn't need to be descriptive because it is not computed as the accessible name. Correct?

`		<h1>Case 1: This should fail</h1>
			<label for="FirstName">Menu</label>
			<input aria-label="First Name" type="checkbox" id="FirstName">`

`		<h1>Case 2: This should pass</h1>
			<label for="FirstName"><span style="sr-only">Menu</span></label>
			<input aria-label="First Name" type="checkbox" id="FirstName">`

Right?

Currently your rule fails Case 2 if I read it correctly.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -95,13 +107,35 @@ Label is visible, but not included in accessibility tree
...
+Label is included in accessibility tree, but not visible.
 
 ```html
 <p id="label_fname" style="position: absolute; top: -9999px; left: -9999px;">First name:</p>
 <input aria-labelledby="label_fname" type="text" name="fname"/>
 ```
 
+#### Passed example 6
+
+Programatically determinable headings provide a context that together with the labels describe the purpose of the form fields.
```
carlosapaduarte: According to the assumption, this is a pass. But I'm not sure the assumption should be there.

If we keep it, correct both "adress" to "address"

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,39 @@ authors:
...
+**Note:**
+- The `option` role is not part of the list of applicable roles, because it has a required context role that inherits from the `select` role. Furthermore, `option` does not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). This means [WCAG 2.1](https://www.w3.org/TR/WCAG21/) does not require it to have an accessible name.
+
+**Note**: This rule is a partial check for WCAG 2.1 success criterion 2.4.6, which applies to all labels. "Label" in WCAG is used in its general sense and includes text or other components with a text alternative that is presented to a user to identify a component within Web content.
 
-### Expectation
+### Expectation 1
 
-Each target element describes the purpose of the associated form field element.
+The `label` element, if there is any, together with its context, describe the purpose of the associated form field element.
```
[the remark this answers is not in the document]
annethyme: agree that coherence would be nice, but it also has to cover the case where it's not a label, but an accessible name, and accessible names don't have "labeled controls", so it has to be a bit more generic, unless we want to spell out the two different situations on their own...

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,39 @@ authors:
...
+
+**Note:** Context in this case could be e.g. headings, fieldsets and legends, text that is located close by etc.
+
+**Note**: Labels do not need to be lengthy. A word, or even a single character, may suffice.
+
+### Expectation 2
+
+The [accessible name](#accessible-name), if there is any, together with it's context, describe the purpose of the associated form field element.
```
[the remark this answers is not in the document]
annethyme: Great catch.
Can't seem to find the other one. Is it still there?

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,39 @@ authors:
...
+
+**Note:** Context in this case could be e.g. headings, fieldsets and legends, text that is located close by etc.
 
 **Note**: Labels do not need to be lengthy. A word, or even a single character, may suffice.
 
 ## Assumptions
 
-_There are currently no assumptions._
+This rule assumes that while having an important context that is not [programatically determinable](#https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) might be a violation under other [WCAG](https://www.w3.org/TR/WCAG21/) success criteria (e.g. 1.3.1 Info and Relationships, if text that has not been marked up as headings is used to split a form into different sections), this is allowed under success criterion 2.4.6 Headings and Labels: Headings and labels describe topic or purpose.
```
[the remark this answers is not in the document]
annethyme: @ShadowBB, seems that there is a disagreement on your WCAG interpretation.

_rules/SC2-4-6-descriptive-labels.md
annethyme: Typo corrected. We need @ShadowBB in on the discussion on whether to include context or not for this rule, since he is the one who filed an issue on it to be included.

_rules/SC2-4-6-descriptive-labels.md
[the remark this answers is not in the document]
ShadowBB: Hmmm. Failing example 6 feels really wrong to me for some reason. I would say that the labels still describe the topic/purpose even without(!) their heading context. Especially because if these two headings (and their content) would be on different pages I would pass them in a heartbeat.

But I do get your point. Because currently "fill in" would also pass this rule as long as it has a heading "first name":

```html
<h1>First name</h1>
<label for=FirstName>Fill in</label>
<input type="text" id="FirstName">
<h1>Last name</h1>
<label for=LastName>Fill in</label>
<input type="text" id="Lastname">
```

Which I would definitely fail.

So I think Carlos is correct in mentioning the fact that there is something wrong with this assumption. 

I think the context is only allowed to differentiate the topic/purpose of the field from others like it.

So the expectation 1 would then read:

"The `label` element, if there is any describe the purpose of the associated form field element. The context of the element can be used to differentiate the purpose from other fields on the same page."

And the note should then read:

**Note:** In this case, context can be created by headings, fieldsets and legends, text that is near the control, etc. Differentiation by context is usually only needed if there are two fields with similar purpose on the same page. For example if the same page asks for both a shipping and billing address.

Expectation 2 should be handled similarly.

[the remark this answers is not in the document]
ShadowBB: @carlosapaduarte: Would you agree with this solution?

_rules/SC2-4-6-descriptive-labels.md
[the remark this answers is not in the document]
carlosapaduarte: The other one (if you were referring to the "it's") was correct. If you were referring to the labeled control, then you addressed it in another comment.

_rules/SC2-4-6-descriptive-labels.md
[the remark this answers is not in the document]
carlosapaduarte: @ShadowBB Yes, I do agree with your proposal. Labels won't need context to be meaningful (thus failing the "Fill in" example), but one can use context to distinguish between meaningful labels. Super!

Just a minor edit in the text for the expectation: 
"The label element, if there is any, describes the purpose of the associated form field element. The context of the element can be used to differentiate the purpose from other fields on the same page."

annethyme:
@WilcoFiers is thinking about this one.

ShadowBB:
> 
> 
> @WilcoFiers is thinking about this one.

Why does this sound ominous to me. XD

What part is bothering him?

_rules/SC2-4-6-descriptive-labels.md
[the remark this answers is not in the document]
ShadowBB: @carlosapaduarte Awesome. And I agree with your minor edit.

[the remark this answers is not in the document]
annethyme: Thanks for the suggestion! 

I have changed both expectations to match, edited the Passed test case description for the shipping/billing address one and added a failed test case with "fill in"´+ a meaningful context.

_rules/SC2-4-6-descriptive-labels.md
annethyme: Expectations updated to match your edit suggestion, and the description for this test case has been edited accordingly.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,38 @@ authors:
 
 ### Applicability
 
-This rule applies to any HTML `label` element or other element referenced by `aria-labelledby` that:
+This rule applies to any element that:
+- has one of the following [semantic roles](#semantic-role): `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch` and `textbox`, and
+- is [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree), and 
+- has an [accessible name](#accessible-name) or is the [labeled control](https://www.w3.org/TR/html/sec-forms.html#labeled-control) of a `label` element that is either [visible](#visible) or [included-in-the-accessibility-tree](#included in the accessibility tree).
```
carlosapaduarte: [included-in-the-accessibility-tree](#included in the accessibility tree) > [included in the accessibility tree](#included-in-the-accessibility-tree)

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,38 @@ authors:
...
 
-- is either [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree), and
-- is programmatically associated with an HTML element that has one of the listed form field [semantic roles](#semantic-role): `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch` and `textbox`.
-
-**Note**: The list of form field roles is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
-- have a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
+**Note**: The list of applicable [semantic roles](#semantic-roles) is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
+- inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
 - does not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
```
carlosapaduarte: does > do

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,38 @@ authors:
...
 
-- is either [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree), and
-- is programmatically associated with an HTML element that has one of the listed form field [semantic roles](#semantic-role): `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch` and `textbox`.
-
-**Note**: The list of form field roles is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
-- have a [semantic role](#semantic-role) that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
+**Note**: The list of applicable [semantic roles](#semantic-roles) is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
+- inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and
```
carlosapaduarte: inherits > inherit

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,6 +181,23 @@ Label is included in accessibility tree, but not visible, and does not describe
...
+
+Accessible name created through "aria-label" does not describes the purpose of the associated element.
+
+```html
+<input aria-label="Menu" id="fname" type="text" name="fname"/>
+```
+
+#### Failed example 7
+
+The `label` in itself does not describe the purpose of the form field, and while the context can be relied upon to differentiate form fields from each other, it is not sufficient to let rely on the context for describing the entire purpose of the form field.
```
carlosapaduarte: remove "let"

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,6 +181,23 @@ Label is included in accessibility tree, but not visible, and does not describe
...
+
+```html
+<h2>Name</h2>
+<label>Fill in: <input id="name" type="text" name="name"/></label>
+````
+
 ### Inapplicable
```
carlosapaduarte: In the code of the first 3 examples correct 
`style:`
to
`style=`

In inapplicable example 3 I suggest removing the `style` attribute because it detracts from the inapplicability reason that is being highlighted.

annethyme:
Tried splitting it out in two rules where one will only test things that haven't been tested in the other rule, but regretted splitting it up like this after looking into Audrey's comment about defining whether the content is also either visible or included in the accessibility tree. 

Here are the split up (but soon to be reverted) rules for references: 

## Rule no. 1

> name: Heading is descriptive
> description: |
>    This rule checks that visible headings describe the topic or purpose of the content.
> 
> ### Applicability
> 
> This rule applies to any [visible](#visible) element with the [semantic role](#semantic-role) of `heading`.
> 
> **Note:**** Headings with only [whitespace](#whitespace) are not [visible](#visible).
> 
> ### Expectation 1
> 
> The [visible](#visible) heading of each target element describes the topic or purpose of its [section of the content](#section-of-content) in part or in its entirety, or the `heading` has no section of the content.

## Rule no. 2

> name: Heading accessible name is descriptive
> 
> description: |
>    This rule checks that the accessible name for headings describe the topic or purpose of the content.
> 
> ## Test procedure
> 
> ### Applicability
> 
> This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of `heading` that:
> - is [included in the accessibility tree](#included-in-the-accessibility-tree),
> - has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace), and
> - does not have any [visible](#visible) [text nodes](), or where the [visible](#visible) [text nodes]() does not [match](#matching-characters) the [accessible name](#accessible-name).
> 
> ### Expectation 1
> 
> The [accessible name](#accessible-name) of each target element describes the topic or purpose of its [section of the content](#section-of-content) in part or its entirety, or the `heading` has no section of the content.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -21,27 +21,38 @@ authors:
...
-**Note**: This rule is a partial check for WCAG 2.1 success criterion 2.4.6, which applies to all labels. "Label" is used in its general sense and includes text or other components with a text alternative that is presented to a user to identify a component within Web content.
+**Note**: This rule is a partial check for WCAG 2.1 success criterion 2.4.6, which applies to all labels. "Label" in WCAG is used in its general sense and includes text or other components with a text alternative that is presented to a user to identify a component within Web content.
 
-### Expectation
+### Expectation 1
 
-Each target element describes the purpose of the associated form field element.
+The `label` element, if there is any, describes the purpose of the associated form field element. The context of the element can be used to differentiate the purpose from other form fields on the same page.
+
+**Note:** In this case, context can be created by headings, fieldsets and legends, text that is near the control, etc.
```
audreymaniez: In the "context" definition here, I think we missed all the information giving by the Accessible Description: `title` and `aria-describedby`  (https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation)

````html
<p id="shipp" aria-hidden="true">Shipping address</p>
<label>Name<input id="name" type="text" name="name" aria-describedby="shipp"/></label>
<label>Street<input id="street" type="text" name="street" aria-describedby="shipp"/></label>

<p id="bill" aria-hidden="true">Billing address</p>
<label>Name<input id="name" type="text" name="name" aria-describedby="bill"/></label>
<label>Street<input id="street" type="text" name="street" aria-describedby="bill"/></label>
````

For me, this is a passed example, but it's not if I use that rule.

Personnaly, I think the "context"/"use of the context" is a huge topic in itself and would need a seperate rule, but at least, I think a more detailled definition is needed. 

I think it is important to understand in which case the context can be used and when it can't (in particular to define, as is done on a recurrent basis in the rules, whether the context should be visible and included in the accessibility tree?) , detailled a little more what is a "text near the control" (can it be after in the source code? do it have to be visually next to the input?) ...

File: _rules/SC2-4-6-descriptive-labels.md [hunk 13]
```diff
@@ -21,27 +21,38 @@ authors:
 
 ### Applicability
 
-This rule applies to any HTML `label` element or other element referenced by `aria-labelledby` that:
+This rule applies to any element that:
+- has one of the following [semantic roles](#semantic-role): `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch` and `textbox`, and
+- is [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree), and 
+- has an [accessible name](#accessible-name) or is the [labeled control](https://www.w3.org/TR/html/sec-forms.html#labeled-control) of a `label` element that is either [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree).
```
audreymaniez: I think there must a note on the accessible name computation, or a least, a link in the background section : https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,14 +181,31 @@ Label is included in accessibility tree, but not visible, and does not describe
...
+
+#### Failed example 7
+
+The `label` in itself does not describe the purpose of the form field, and while the context can be relied upon to differentiate form fields from each other, it is not sufficient to rely on the context for describing the entire purpose of the form field.
+
+```html
+<h2>Name</h2>
+<label>Fill in: <input id="name" type="text" name="name"/></label>
+````
+
```
audreymaniez: Can we add an example illustrating the accessible name computation, in particular that the aria-label relationship completly delete the label for relation :

<label for="fname">Nom</label>
<input aria-label="Menu" id="fname" type="text" name="fname"/>

_rules/SC2-4-6-descriptive-labels.md
annethyme: @audreymaniez, it is part of the [accessible name](https://act-rules.github.io/glossary/#accessible-name) definition. Isn't that enough?

audreymaniez: han you're totally right! no I think that enough!

_rules/SC2-4-6-descriptive-labels.md
annethyme: Added a Passed example 8 + Failed example 8 for this now :)

_rules/SC2-4-6-descriptive-labels.md
annethyme: @audreymaniez, I appreciate your detailed feedback :) 

Your question has got me thinking, that maybe the first expectation should be limited to cases where the `label` is visible. 

I think your point about the accessible description is very relevant, but only for Expectation 2, that is about the accessible name of the element. 

As I see it, Expectation 1 is there to handle the visible part, and here the accessible description doesn't really matter, only what can be seen.

What do you think?

audreymaniez: With your explanations, the difference between E1 and E2 is now more clear! For me, the only difference between Expectation 1 and 2 was the nature of the element used to give the accessible name to the form element (`label` or other methods like `aria-label` etc.).  I just now understand that by oppose `label` vs `accessible name` you certainly meant "visible label".  And I think you right, otherwise I don't really understand the necessity of two rules.

Do we scope only the `label` tag in the expectation 1? or every text near the input that could be used as a visual name ? If that is so, then I think we don't want to use the `label` in quote, because I think it is really confusing (quotes are usually used fot HTML tags or attributes).

Don't we specify it a little more? I mean, it is not obvious that by `label` we mean `visible` because a `label` tag can be not visible, as you say. 

I just suggest to replace "The `label` element" by something like "The visible label"? or the "The visible name"? 

But that was not my point, I'll try to better explain my idea :-)

First, you right, my comment concerns (now I understand well the rule) the expectation 2 only. 

Then, you write "_**Note:** In this case, context can be created by headings, fieldsets and legends, text that is near the control, etc._". 

But the context is a key element in assessing the conformity of an element, and I think it must be more detailled. The context of a form element can be its accessible description (`aria-describedby`), which is not mentionned in the "context" lists. You cited "fieldset" and "legend", but we can create form group using ARIA `role="group"` with an `aria-label` too... 

I mean, there's a variety of contexts (but also limited), and I think we need to document contexts we recognized as "usable context" as much as we can.

I was looking back to the rule for links with identical name to see if something similar on context was done but not (and I didn't notice it in that rule^^).  
Maybe my suggestion goes beyond what ACT-rules needs to specify?

I don't know if I'm more clear 😁

annethyme: For the things that were not your point, but were super relevant: 

I fixed up a bit of things both in the applicably and expectation 1 to make it more clear that it's about visible labels. 
For this rule, it's only going to consider visible labels that are actually connected to the form element (and accessible names) - at least that was the original scope of this rule. 

I will have a look at the context stuff in next round.

ShadowBB: The note includes "etc", because the list "headings, fieldsets and legends, text that is near the control" is not exhaustive. I would state that things like aria-describedby also add context for AT users. But we could add more things explicitly to this list. I would be fine with that.

annethyme: Now that we start digging more into the context stuff, I start to think that we will have to distinguish between what can be considered the visible context (relevant for expectation 1) and what can be considered context that included in the accessibility tree (relevant for expectation 2). 
As a sighted user, I will not have any use of descriptions created using aria-describedby, and as a screen reader user, I will not have any use of a context that has been marked up with aria-hidden="true". 

However, if we have to consider all of this, it will start to look a lot like proper definitions, which will be very hard to do, because then they have to be very strictly defined... 
I am currently a bit unsure on how to move on with this.

ShadowBB: If context is available in a visual way but not available for AT than that is a 1.3.1 fault and not a 2.4.6 fault and so it would not really matter I think. I think it should still pass this test then.

So the current note would be enough then, I don't think we have to distinguish between visible context and context that is included in the accessibility tree, within this rule.

annethyme: @ShadowBB, I kind of agree, but then what if there is context that is available to screen readers, but not visually? Would it still count as context?

ShadowBB: Good question. It stumped me for a second there. But:

If context is available for AT but not in a visual way than non-disabled users won't know what to fill into the form fields.

For example:

```
<p id="shipp" class="sr-only">Shipping address</p>
<label>Name<input id="name" type="text" name="name" aria-describedby="shipp"/></label>
<label>Street<input id="street" type="text" name="street" aria-describedby="shipp"/></label>

<p id="bill" class="sr-only">Billing address</p>
<label>Name<input id="name" type="text" name="name" aria-describedby="bill"/></label>
<label>Street<input id="street" type="text" name="street" aria-describedby="bill"/></label>
```

The above case is clearly a usability issue, not an accessibility issue. So it is not (nor should it be) a WCAG violation in my opinion.

audreymaniez: @ShadowBB  but back to the specific benefits of the SC 2.4.6 : 
"_Descriptive headings are especially helpful for **users who have disabilities that make reading slow and for people with limited short-term memory**. These people benefit when section titles make it possible to predict what each section contains._ "

I think the scope is broader than blind user or assistive technologies users only, and hidden text give an advantage to blind user, and disadvantage disabled users that do not rely on an AT.

ShadowBB: > 
> 
> @ShadowBB but back to the specific benefits of the SC 2.4.6 :
> "_Descriptive headings are especially helpful for **users who have disabilities that make reading slow and for people with limited short-term memory**. These people benefit when section titles make it possible to predict what each section contains._ "
> 
> I think the scope is broader than blind user or assistive technologies users only, and hidden text give an advantage to blind user, and disadvantage disabled users that do not rely on an AT.

I agree with the fact that the scope is broader than blind users or assistive technology users. However hidden text doesn't give a disadvantage to disabled users that do not rely on an AT _when compared to non-disabled users_.

My example harms all users (apart from those that use screen readers I guess), not just users with disabilities.

annethyme: ### Expectation 1: 
...

**Note:** In this case, context can be created by headings, fieldsets and legends, text that is near the control, etc. Only [visible](#visible) elements should be considered as context for this expectation. 

### Expectation 2
...

**Note:** In this case, context can be created by headings, fieldsets and legends, text that is near the control, etc. For this expectation elements that are [included in the accessibility tree](#included-in-the-accessibility-tree) should be considered for context as well, e.g. elements connected through `aria-describedby`, or `role="group"` with an [accessible name](#accessible-name).

**Note:** If a context that is necessary for differentiating form fields from each other is [visible](#visible), but not [included in the accessibility tree](#included-in-the-accessibility-tree), this might be a violation under other [WCAG](https://www.w3.org/TR/WCAG21/) success criteria (e.g. 1.3.1 Info and Relationships), but this is not considered a failing condition for this rule. For this rule, it is enough that the context is there in some form.

ShadowBB: I think this captures it correctly! Nicely done!

I see no flaw in the wording or grammar.

audreymaniez: Great work @annethyme :-) I think the two expectations are now more understandable :+1:

annethyme: @audreymaniez, so glad to hear that! 
If you have time to review again, it would be much appreciated.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -95,13 +109,44 @@ Label is visible, but not included in accessibility tree
...
+#### Passed example 7
+
+Accessible name created through "aria-label" describes the purpose of the associated element.
+
+```html
+<input aria-label="First name" id="fname" type="text" name="fname"/>
+```
+
+#### Passed example 8
```
annethyme: This test case fails Expectation 1

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -173,13 +244,13 @@ Programatically associated `p` element that is neither visible nor included in t
 The `label` element is associated with an HTML element that does not have a form field semantic role.
 
 ```html
-<label for="fname" style:"display:none;">First name:</label>
+<label for="fname">First name:</label>
 <p id="fname"/>
 ```
 
 #### Inapplicable example 4
```
annethyme: Add Inapplicable test case where a `p` element acts as visible label but is not programatically determinable.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -95,13 +109,44 @@ Label is visible, but not included in accessibility tree
...
+<h2>Shipping address</h2>
+<label>Name<input id="name" type="text" name="name"/></label>
+<label>Street<input id="street" type="text" name="street"/></label>
+
+<h2>Billing address</h2>
+<label>Name<input id="name" type="text" name="name"/></label>
+<label>Street<input id="street" type="text" name="street"/></label>
+````
+
+#### Passed example 7
```
annethyme: Add Passed test case where labels are duplicated on the page but there is a visible context + a different context included in the accessibility tree, and both contexts are descriptive.

pages/glossary/context.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Context
+key: context
+---
+
+Context is information that helps the user understand other information.
```
kasperisager: "the user" -> "a user"?

annethyme:
@carlosapaduarte, @ShadowBB and @audreymaniez, you have been very involved in reviewing this rule previously. 
I have implemented a lot of changes to accommodate all the discussions we have had, and updated the test cases accordingly. 
I hope you want to review again.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -76,35 +89,73 @@ Label that is coded with the `p` element and associated by the aria-labelledby a
...
+```
+
+#### Passed Example 8
+
+Otherwise identically named form fields can be differentiated both visually and in a way that is included in the accessibility tree, though different methods are used for each purpose.
+
+```html
+<h2 aria-hidden="true">Shipping address</h2>
+<p id="shipping" style="position: absolute; top: -9999px; left: -9999px;">(Shipping)</p>
+<label>Name<input id="name" type="text" name="name" aria-describedby="shipping" /></label>(Shipping
```
carlosapaduarte: I believe you don't want the `(Shipping` here

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -76,35 +89,73 @@ Label that is coded with the `p` element and associated by the aria-labelledby a
...
+<label>Street<input id="street" type="text" name="street" aria-describedby="shipping" /></label>
+
+<h2 aria-hidden="true">Billing address</h2>
+<p id="shipping" style="position: absolute; top: -9999px; left: -9999px;">(Billing)</p>
```
carlosapaduarte: the id here should be billing

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -76,35 +89,73 @@ Label that is coded with the `p` element and associated by the aria-labelledby a
...
+<label>Name<input id="name" type="text" name="name" aria-describedby="billing"/></label>
```
carlosapaduarte: And while you're at it, following up on @WilcoFiers points on #546 (unfortunately I couldn't make to today's meeting, so I still don't know what was the decision, but I agree with his point) I think it would be nice to use different ids from the ones already used in the shipping part of the example

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,23 +199,58 @@ Label is included in accessibility tree, but not visible, and does not describe
...
+The `label` in itself does not describe the purpose of the form field, and while the context can be relied upon to differentiate form fields from each other, it is not sufficient to rely on the context for describing the entire purpose of the form field.
+
+```html
+<h2>Name</h2>
+<label>Fill in: <input id="name" type="text" name="name"/></label>
+````
+
+#### Failed Example 8
+
+The descriptive label coded with the `label` element is through the accessible name calculation overwritten by an `aria-label` that is not descriptive.
```
carlosapaduarte: Perhaps we could add here something that makes it clearer why this fails. Something along these lines:
While Expectation 1 is met, Expectation 2 is not, and no expectation can fail.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,23 +199,58 @@ Label is included in accessibility tree, but not visible, and does not describe
...
 ```html
-<label for="fname" style:"display:none;">First name:</label>
+<label for="fname" style="display:none;">First name:</label>
 <input id="fname" type="text" name="fname"/>
 ```
 
-#### Inapplicable example 2
+#### Inapplicable Example 2
 
 Programatically associated `p` element that is neither visible nor included in the accessibility tree.
```
carlosapaduarte: On Chrome the `input` element has an accessible name given by the p#label_fname so this should be applicable

_rules/SC2-4-6-descriptive-labels.md
audreymaniez: Each test case runs on a unique page, so maybe it's not a problem if on the rule page we have identical ID?

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -113,7 +164,7 @@ Label that is coded with the `label` element and does not describe the purpose o
 <input id="fname" type="text" name="fname"/>
 ```
 
-#### Failed example 2
+#### Failed Example 2
 
 Label that is coded with the `p` element and associated by the aria-labelledby attribute. The label does not describe the purpose of the associated element.
```
audreymaniez: Missing quotes on "aria-labelledby"

> Label that is coded with the `p` element and associated by the `aria-labelledby` attribute

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,23 +199,58 @@ Label is included in accessibility tree, but not visible, and does not describe
 <input aria-labelledby="label_fname" type="text" name="fname"/>
 ```
 
+#### Failed Example 6
+
+Accessible name created through "aria-label" does not describes the purpose of the associated element.
```
audreymaniez: Quotes 😄 

> Accessible name created through `aria-label` does not describes the purpose of the associated element.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,23 +199,58 @@ Label is included in accessibility tree, but not visible, and does not describe
...
+
+```html
+<input aria-label="Menu" id="fname" type="text" name="fname"/>
+```
+
+#### Failed Example 7
+
+The `label` in itself does not describe the purpose of the form field, and while the context can be relied upon to differentiate form fields from each other, it is not sufficient to rely on the context for describing the entire purpose of the form field.
```
audreymaniez: Just a suggestion of rewording. In the description of the example you emphasized on the fact that the context is not sufficient for describing the purpose of the form field. But if I understand well your example, the problem is, in the first place, that the label does not describe the purpose at all, and so, the context can't applied on a label that is not descriptive.  

If it's right :-) I suggest: 

> The `label` in itself does not describe the purpose of the form field. While the context could have been relied upon to differentiate form fields from each other, the fact that the label is not descriptive makes the use of this context ineffective.

_rules/SC2-4-6-descriptive-labels.md
audreymaniez: Yes, aria-* relationship are not affected by the visibility of the element (display:none, visibility:hidden have no impact on the calculated label).

_rules/SC2-4-6-descriptive-labels.md
annethyme: Will it work if we do this instead? 

`<p id="shipping" style="display:none">(Shipping)</p>`

I *think* even elements that are excluded from the accessibility tree will still be used for the accessible description computation if references with aria-describedby.

@kasperisager, I did look in https://www.w3.org/TR/accname-1.1/, but am hoping you can confirm my interpretation :)

kasperisager: Can confirm:

> If the current node is hidden and is not directly referenced by `aria-labelledby` or `aria-describedby,` nor directly referenced by a native host language text alternative element (e.g. `label` in HTML) or attribute, return the empty string.

_rules/SC2-4-6-descriptive-labels.md
annethyme: I would prefer if we can reuse the same IDs across test cases, so the test cases are more consistent, and the only thing that changes is what we want to illustrate with that particular test case.
As Audrey says, the test cases should be considered separate pages in themselves.

_rules/SC2-4-6-descriptive-labels.md
carlosapaduarte: :-) 
My issue was not the Shipping in this line `<p id="shipping" style="display:none">(Shipping)</p>`
Your change works but I'm okay with how it was before
My issue was with the (Shipping in the following line `<label>Name<input id="name" type="text" name="name" aria-describedby="shipping" /></label>(Shipping` where it shows up outside the `label` element

_rules/SC2-4-6-descriptive-labels.md
carlosapaduarte: I'm totally for using the same ids in different test cases. 
But in this test case, both `name` and `street` are the ids used in different `input` elements. That was what I was suggesting to avoid because it would fail other rules.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,39 +208,73 @@ Label is included in accessibility tree, but not visible, and does not describe
...
 ```html
-<label for="fname" style:"display:none;">First name:</label>
-<input id="fname" type="text" name="fname"/>
+<h2>Name</h2>
+<label>Fill in: <input id="name" type="text" name="name"/></label>
+````
+
+#### Failed Example 8
+
+While Expectation 1 is met, Expectation 2 fails because the descriptive label coded with the `label` element through the accessible name calculation is overwritten by an `aria-label` that is not descriptive.
```
carlosapaduarte: I'm not sure I would fail this one. While the accessible name is "Fill in:", you get immediately before in the accessibility tree "First name:". 
But, according to the expectation is fails, which means I don't know if we need to work on the expectation. Opinions, anyone?

ShadowBB: I would fail this Carlos. The accessible name should be descriptive without context in this case I would say.

_rules/SC2-4-6-descriptive-labels.md
annethyme: Aha! Looks like a rogue copy/paster leftover. I have removed it now, since it doesn't make sense. But kept the rest, since I am not quite sure how it was before...

_rules/SC2-4-6-descriptive-labels.md
annethyme: Ahh, I see! Fixed it both there and in Passed Example 6

_rules/SC2-4-6-descriptive-labels.md
annethyme: I agree with @ShadowBB. According to the expectation the context can only be used to differentiate similar fields from each other, not to provide the whole description.

Do others have an opinion about this? (e.g. @WilcoFiers, @audreymaniez, @EmmaJP ?)

audreymaniez: I agree too with @ShadowBB, I would fail this one.

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -76,35 +89,82 @@ Label that is coded with the `p` element and associated by the aria-labelledby a
...
+
+<h2 aria-hidden="true">Billing address</h2>
+<p id="billing" style="display:none">(Billing)</p>
+<label>Name<input id="billing-name" type="text" name="name" aria-describedby="billing"/></label>
+<label>Street<input id="billing-street" type="text" name="street" aria-describedby="billing" /></label>
+````
+
+#### Passed Example 9
+
+`p` element that is neither visible nor included in the accessibility tree, but still provides an accessible name to the input field through `aria-describedby`.
```
carlosapaduarte: `aria-describedby` > `aria-labelledby`

_rules/SC2-4-6-descriptive-labels.md
```diff
@@ -148,39 +208,73 @@ Label is included in accessibility tree, but not visible, and does not describe
 <input aria-labelledby="label_fname" type="text" name="fname"/>
 ```
 
-### Inapplicable
+#### Failed Example 6
 
-#### Inapplicable example 1
+Accessible name created through `aria-label` does not describes the purpose of the associated element.
```
carlosapaduarte: describes > describe

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+rule_type: atomic
+description: |
+  This rule checks that labels describe the purpose of form field elements.
+accessibility_requirements: 
+  wcag21:2.4.6: # Headings and labels (AA)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
```
carlosapaduarte: We might want to include "Language" here also, following #671

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Note**: Labels do not need to be lengthy. A word, or even a single character, may suffice.
+
+## Assumptions
+
+This rule assumes that while having an differentiating [context](#context) that is not [programatically determinable](#https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) might be a violation under other [WCAG](https://www.w3.org/TR/WCAG21/) success criteria (e.g. 1.3.1 Info and Relationships, if text that has not been marked up as headings is used to split a form into different sections), this is allowed under success criterion 2.4.6 Headings and Labels: Headings and labels describe topic or purpose.
```
carlosapaduarte: an differentiating > a differentiating

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+```html
+<label for="fname">Fill in:</label>
+<input id="fname" type="text" name="fname" aria-label="First name:"/>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+`Label` that is neither visible to users, nor included in the accessibility tree.
```
carlosapaduarte: `Label` > `label`

_rules/descriptive-labels-f0c5c5.md
annethyme: @carlosapaduarte, I agree, but no decision has been made om #671, and I guess we should at one point stop adding new things to this PR and just see if we can get it closed off soon :) 
I am imagining that #671, if agreed on, will  need to update all the rules that are impacted by the decision, and we might as well do it in one go.

annethyme:
@ShadowBB, the settings of the repository seems to have been changed. I fixed two typos as requested by Carlos, and this dismissed your approval. Can I have you click the button again?
I have raised the problem with even small changes dismissing all reviews, so hopefully we will have a fix for the process later on.

WilcoFiers:
:/ OK so sorry for the tripple submit, Github was being weird.

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Note**: The list of applicable [semantic roles](#semantic-roles) is derived by taking all the [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/) roles that:
+- inherit from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and 
+- do not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
+
+**Note:** The `option` role is not part of the list of applicable roles, because it has a required context role that inherits from the `select` role. Furthermore, `option` does not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). This means [WCAG 2.1](https://www.w3.org/TR/WCAG21/) does not require it to have an accessible name.
+
+**Note**: This rule is a partial check for WCAG 2.1 success criterion 2.4.6, which applies to all labels. "Label" in WCAG is used in its general sense and includes text or other components with a text alternative that is presented to a user to identify a component within web content.
```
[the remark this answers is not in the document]
Jym77: I think it may be good to point out that "label" means the `label` element in this rule while it means something else in WCAG. I agree that stating this rule is a partial check is no the way we usually go and thus should be skipped.

Jym77:
Cancelling Final Call. I will split the rule in two as per @WilcoFiers suggestion.

_rules/SC2-4-6-descriptive-labels.md
Jym77: Similar change on [e086e5](https://act-rules.github.io/rules/e086e5) made in #856 .

Jym77:
@ShadowBB 
> Hey Anne,
> 
> Really liking what you did with this so far. If I read this correctly it will fail if it has both an invisible associated label and an (for example) aria-label and the associated label is not descriptive. Is this the intention? Because i think the aria-label should win out and the invisible label doesn't need to be descriptive because it is not computed as the accessible name. Correct?
> 
> ```
> <h1>Case 1: This should fail</h1>
> <label for="FirstName">Menu</label>
> <input aria-label="First Name" type="checkbox" id="FirstName">
> ```
> 
> ```
> <h1>Case 2: This should pass</h1>
> <label for="FirstName"><span style="sr-only">Menu</span></label>
> <input aria-label="First Name" type="checkbox" id="FirstName">
> ```
> 
> Right?
> 
> Currently your rule fails Case 2 if I read it correctly.

1. The way I understand [computation of accessible name](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te), I agree with you that `aria-label` should win over `label` (in both cases) for the accessible name (because `aria-label` is 2.C while `label` is 2.E, `aria-label` is essentially loosing only to `aria-labelledby`).
2. So, I agree that your second case should pass.
3. And I think it does 😁 Expectation 1 pass (is skipped) because there is no visible label, and Expectation 2 pass because the accessible name is the `aria-label`, not the `label`.

So, I think it's OK. Why do you think that the rule currently fails Case 2?

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+- Dagfinn Rømen
+- Geir Sindre Fossøy
+---
+
+## Applicability
+
+This rule applies to any element that:
+- has one of the following [semantic roles](#semantic-role): `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch` and `textbox`, and
+- is [visible](#visible) or [included in the accessibility tree](#included-in-the-accessibility-tree), and 
+- has an [accessible name](#accessible-name) or is the [labeled control](https://www.w3.org/TR/html/sec-forms.html#labeled-control) of a [visible](#visible) `label` element.
```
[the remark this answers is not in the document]
Jym77: I think this is Passed Example 4, if I read all this correctly:
> ```html
> <p id="label_fname" aria-hidden="true">First name:</p>
> <input aria-labelledby="label_fname" type="text" name="fname"/>
> ```

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+- do not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
+
+**Note:** The `option` role is not part of the list of applicable roles, because it has a required context role that inherits from the `select` role. Furthermore, `option` does not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). This means [WCAG 2.1](https://www.w3.org/TR/WCAG21/) does not require it to have an accessible name.
+
+**Note**: This rule is a partial check for WCAG 2.1 success criterion 2.4.6, which applies to all labels. "Label" in WCAG is used in its general sense and includes text or other components with a text alternative that is presented to a user to identify a component within web content.
+
+## Expectation 1
+
+The [visible](#visible) `label` element, if there is any, describes the purpose of the associated form field element. The [visible](#visible) [context](#context) of the element can be used to differentiate the purpose from other form fields on the same page.
```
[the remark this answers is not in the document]
Jym77: Elements labelled through `aria-labelledby` are taken care of by Expectation 2, ain't they?

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Note:** In this case, [context](#context) can be created by headings, fieldsets and legends, text that is near the control, etc. Only [visible](#visible) elements should be considered as context for Expectation 1.
+
+**Note**: Labels do not need to be lengthy. A word, or even a single character, may suffice.
+
+## Expectation 2
```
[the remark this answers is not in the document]
Jym77: I agree that this rule will be simpler and easier if split in two.

---

> This rule is testing two different conditions, the visible label not matching the visible context, and the accessible name not matching the programatic context.

For the sake of nitpicking 😈 I would argue that ACT-Rules format does not define "condition" (and even puts in in quotation marks which seems to be a way to indicate a blurry term if I judge by the previous sentence of ` what elements, nodes or other "parts" of a test subject are to be tested`) So it's never completely clear what is one or two conditions.

I would further argue that some of the other rules with several expectations are actually testing several conditions, *e.g.*:
* [Video as a media alternative for text](https://act-rules.github.io/rules/ab4d13) (1. the info in the video is contained somewhere else; 2. the video is labelled as an alternative for something (not necessarily the same as in 1.))
* [HTML Page has a title](https://act-rules.github.io/rules/2779a5) (1. the page has at least one `title`; 2. the first `title` has text content)
* …
(and yet these rules would be needlessly complicated if split in two, while this rule would be simplified…)

(as I rule of thumb, I would argue that if a rule has several expectation that do not depend on each other, then it's probably 2 conditions rather than one…)

_rules/descriptive-labels-f0c5c5.md
```diff
@@ -0,0 +1,280 @@
...
+input_aspects:
+- DOM Tree
+- CSS Styling
+authors:
+- Anne Thyme Nørregaard
+- Dagfinn Rømen
+- Geir Sindre Fossøy
+---
+
+## Applicability
```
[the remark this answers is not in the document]
Jym77: I guess this is because this rule is only concerned with form fields. This is a subpart of the SC, but this kind of situation (rule only addressing one small part of the SC) is rather common. We'll need more rules to address the full SC…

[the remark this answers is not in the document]
Jym77: Also, I think this is due to the change from `aria-labelledby` to "accessible name". I think it was quite easy to target "the `label` or `aria-labelledby` element" but it is somewhat difficult to target "the accessible name", hence moving the target to the form field in order to be able to make the rule broader (including `aria-label`, …)

Jym77:
So, following @WilcoFiers good advice:
* I've split the rule in two (+one composite) for visual `label`/accessible name.
* I've split the context definition in two for visual context/programatically determined context

@DagfinnRomen @geirsf you were authors of the original rule, can you check that it is still OK for you?

@audreymaniez @ShadowBB @carlosapaduarte  @DagfinnRomen @jkodu @WilcoFiers you reviewed @annethyme's changes before the (first) Final call, can you recheck it now?

_rules/descriptive-label-cc0f0a.md
```diff
@@ -1,204 +0,0 @@
----
```
WilcoFiers: Please move the file, rather than deleting it and creating a new one. That way we don't lose the git commit history of this file. It also makes reviewing much easier. I'm going to punt on reviewing until this is done.

Jym77: Not easily possible in git, and the way to do it rewrites shared history which is even worse.
https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history

Additionally, there is here a single file which has been copied/renamed into two files because of the split of the rule. Tracking that in git seems to be even less possible.

Jym77: Also, the commit who did the change is actually detecting a rename: https://github.com/act-rules/act-rules.github.io/pull/447/commits/d96505cdac1205d30f7baf545cd4a25d17317ae4

So, I really don't know how to do it better.

…
…
…

OK, I think I figured out what happened (maybe).

- Back then, the file was called `SC2-4-6-descriptive-label.md`, using the old style of naming files (SC + name). This PR was started at that moment.
- Then, the naming convention of files changed (to name + hash). The rule file was moved as part as this commit https://github.com/act-rules/act-rules.github.io/commit/a289b7510c398fc4366f3cf9d6bfc0bbb0c5529f#diff-59490505dddc4c7a2c672f03a7dfa31e which has since been merged.
- Simultaneously, the file was renamed inside the PR as this commit: https://github.com/act-rules/act-rules.github.io/pull/447/commits/d96505cdac1205d30f7baf545cd4a25d17317ae4 
- The hashes are different in both 😢 and anyway the double move makes some ugly merge conflict 😭 

---
Now that I have an idea of what happened, I'll try to fix it but there is a risk that the fix is going to be ugly (likely, all changes made to the file as part of this PR will be squashed in a single commit, which is not that big a deal in the long run since we squash and merge anyway…)

Jym77: @WilcoFiers I tried a couple of different way to make the move working, but it looks like I failed 😞 

As you can see, the latest commit touching these file is https://github.com/act-rules/act-rules.github.io/pull/447/commits/5fcf5890a17fffbb0ff4d9da06047136fb54c865 which show this as a "file was renamed without changes" (I used `git mv`), but the "files changed" tab of the PR still show thus as deleting+creating.

I have no clue on how to improve that and I'm giving up at that point. It seems that it is too late to repare the mess that was created by the parallel rename back in April.

Jym77: Additionally, the "History" tab of these file is listing all the commit including the move ones, from which we can branch to the history of the previous file. So I do not think that history is really lost.

WilcoFiers: The commit history definitely is lost on this. Perhaps @kasperisager can help? `git mv` will do the trick, but you may have to revert a bunch of stuff on the PR.

Jym77: I did use `git mv` to make the move (in each of my attempts). `git mv` is *not* doing that much magic: https://stackoverflow.com/questions/47401843/git-copy-file-as-opposed-to-git-mv

When I'm running the website locally on the branch, I do see the changelog for these files going back to the first draft, so it seems to be OK.

utils/reg-exps.js
```diff
@@ -16,10 +16,10 @@ const regexps = {
 		1. Classical usage => 'this is a link [link](key)'
 		\(([^)]+)\) => one or more non-'(' within parenthesis and remember the content of the parenthesis as first group.
 		2. Internal reference => '[refname]: key "title"'
-		/.*:\s+(.*)\s+/ => drop everything before the ':', remember the space separated first word after it.
+		/.*:\s+(#.*)\s+/ => drop everything before the ':', remember the space separated first word after it if it starts with # (otherwise, it's a link to an external page).
 	*/
 	glossaryKey: /\(([^)]+)\)/,
-	glossaryKeyInDefinition: /.*:\s+(.*)\s+/,
+	glossaryKeyInDefinition: /.*:\s+(#.*)\s+/,
```
jkodu: these changes should be in a different PR.

Jym77: Done in #884

Jym77:
Make PR to move old file to "accessible name"
remove composite rule

Jym77:
Closing this PR. Will reopen 1 PR for just renaming the rule, and then another one for the change. This will ensure the renaming commit is in history and should keep history need for changelog and stuff.

Issue #448, opened 2019-03-19
aria-hidden with focusable content fails the wrong succescriterium


danistr:
```
<div aria-hidden="true">
	<a href="/" >Link</a>
</div>
```

According to the rule this is currently a violation under success criterion 4.1.2. (Name, Role, Value) but should be a violiation on success criterion 1.3.1 (Info and Relationships). For example a link. Visually there is a link visible but for screenreader users this link is no longer perceivable and will not be read.
It can't fail under success criterion 4.1.2. because there is a name and role present and a aria-hidden attribute doesn't remove this.

annethyme:
Handled by https://github.com/auto-wcag/auto-wcag/pull/449

PR #449, opened 2019-03-19
Changed succescriteria from 4.1.2 to 1.3.1

danistr:
<< Describe the changes >>

Closes issue: (#448 )

**When creating PR:**

**After creating PR:**

# How to Review And Approve

### DIFF
--- a/_rules/SC1-3-1-aria-hidden-focus.md
+++ b/_rules/SC1-3-1-aria-hidden-focus.md
@@ -22,7 +22,7 @@ authors:
 
 The rule applies to any element with an `aria-hidden="true"` attribute.
 
-**Note**: Using `aria-hidden="false"` on a decendent of an element with `aria-hidden="true"` **does not** expose that element. `aria-hidden="true"` hides itself and all its content from assistive technologies.
+**Note**: Using `aria-hidden="false"` on a descendent of an element with `aria-hidden="true"` **does not** expose that element. `aria-hidden="true"` hides itself and all its content from assistive technologies.
 
 ### Expectation
 
--- a/_rules/SC1-3-1-aria-hidden-focus.md
+++ b/_rules/SC1-3-1-aria-hidden-focus.md
@@ -44,6 +44,7 @@ By adding `aria-hidden="true"` to an element, content authors ensure that assist
 A focusable element with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making it's state of visible or hidden unclear.
 
 - https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html
+- https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html
 - https://www.w3.org/TR/wai-aria-1.1/#aria-hidden
 - https://www.w3.org/TR/html/editing.html#can-be-focused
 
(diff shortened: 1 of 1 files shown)


annethyme:
@jkodu, I am assigning you to this one, since you are the person responsible for this rule within the WAI-Tools project.

ShadowBB:
> 
> 
> In the applicability's note there's a typo: change decendent to descendent

Done. Thanks!

DagfinnRomen:
The background section should reference the 1.3.1 understanding article, not 4.1.2.
https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html

Also, I suggest a minor update to the description, from "This rule checks `aria-hidden` elements do not contain focusable elements" to "This rule checks that elements with an `aria-hidden` attribute do not contain focusable elements".

ShadowBB:
> 
> 
> The background section should reference the 1.3.1 understanding article, not 4.1.2.
> https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html
> 
> Also, I suggest a minor update to the description, from "This rule checks `aria-hidden` elements do not contain focusable elements" to "This rule checks that elements with an `aria-hidden` attribute do not contain focusable elements".

Well spotted! Both done. Thanks!

ShadowBB:
> 
> 
> I understand the argument for this being a failure of 4.1.2, but I don't quite understand the argument for why this wouldn't also be a failure for 4.1.2. Can someone explain this before we merge? I think at least it needs to be documented somewhere.

Sure. This rule checks that elements with an `aria-hidden` attribute do not contain focusable elements. Elements with an aria-hidden element that contain focusable elements can still have a name and role that can be programmatically determined which is all that is needed to pass 4.1.2 in some cases. So failing this rule is not a guarantee that 4.1.2 also fails.

WilcoFiers:
@ShadowBB Thanks for clarifying. I don't agree with you on this. Putting `aria-hidden=true` on an element removes it from the accessibility tree. Since the name / role / value props are derived from the accessibility tree, any element with `aria-hidden=true` no longer has a name, role, or any props / state. And since it is focusable (thus will be perceived by the user as an interface component), I would say that fails SC 4.1.2.

ShadowBB:
> 
> 
> @ShadowBB Thanks for clarifying. I don't agree with you on this. Putting `aria-hidden=true` on an element removes it from the accessibility tree. Since the name / role / value props are derived from the accessibility tree, any element with `aria-hidden=true` no longer has a name, role, or any props / state. And since it is focusable (thus will be perceived by the user as an interface component), I would say that fails SC 4.1.2.

I have thought about this further and deliberated with a colleague. We came to the conclusion that you are correct. It should indeed fail both SC. Our flaw was in the programmatically determined definition (we forgot that mentioned in needed to be visible to assistive technology too).

Sorry to waste all this time on this. We should probably mention both SC, also in the background.

_rules/SC1-3-1-aria-hidden-focus.md
```diff
@@ -42,7 +43,7 @@ By adding `aria-hidden="true"` to an element, content authors ensure that assist
 
 A focusable element with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making it's state of visible or hidden unclear.
 
-- https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html
```
audreymaniez: If both SCs are concerned, put the 4.1.2 link back in the background section :-)

annethyme:
@jkodu, this was actually out in a 2-week "Final Call" that doesn't end until April 9.

Issue #451, opened 2019-03-19
"label and name from content mismatch" needs to also fail split up labels


ShadowBB:
The "label content name mismatch" rule is now only applicable when there is an aria-label or aria-labelledby attribute. But this means it won't fail when situations where the visible text content and accessible name mismatch when there is no aria involved.

In addition I think it is best to give explicit examples where visible labels are split up by invisible parts of the accessible name. For example:

`<a href="home.html"> link <span style="sr-only">that opens a new screen</span> to the homepage</a>`

In contrast we could limit the scope of the rule to just cases with aria. But if we choose to do so we should definitely only look at non-empty aria-labels. (and we should also make it explicit that if the visible name is contained in the accessible name but with other content in between, it should still fail).

https://auto-wcag.github.io/auto-wcag/rules/SC2-5-3-label-content-name-mismatch.html

annethyme:
... and have something like this as applicable + passed: 

<a href="home.html">Read more<span style="sr-only"> about our pricing</span></a>

annethyme:
Working on it in this pull request: https://github.com/auto-wcag/auto-wcag/pull/447

Jym77:
Reopening issue since (inactive) PR has been closed.

WilcoFiers:
Related to #1619

Issue #377, opened 2018-12-17
Visible label is part of accessible name (2ee8b8): Non-text content is ambiguous


kasperisager:
As https://auto-wcag.github.io/auto-wcag/rules/SC2-5-3-label-content-name-mismatch.html makes use of the term "non-text content" within its applicability and lacks an unambiguous definition of this term, it is not currently a valid ACT rule. We currently have a definition of text content (https://auto-wcag.github.io/auto-wcag/pages/glossary/text-content.html) though given that definition then the following case would be classified as having visible text content:

```html
 <button aria-label="close">X</button>
```

This case is however supposed to be inapplicable to the rule and therefore not be classified as having visible text content.

kasperisager:
Summoning @ShadowBB.

annethyme:
Handling it in https://github.com/auto-wcag/auto-wcag/pull/430

Jym77:
Reopening issue since #452 that was partly tackling it has been closed.

WilcoFiers:
This issue is a duplicate of #1254, and will soon be addressed.

Issue #315, opened 2018-10-23
aria-hidden with focusable content: Expectation is ambiguous


kasperisager:
The expectation for https://auto-wcag.github.io/auto-wcag/rules/SC4-1-2-aria-hidden-focus.html is currently ambiguous as the term _contain_ is never defined. For example, given the following snippet of HTML...

```html
<div aria-hidden="true">
  <button>I'm focusable</button>
</div>

<script>
  {
    const host = document.querySelector("div");
    const root = host.attachShadow({ mode: "open" });

    root.innerHTML = `
      <p>I'm not focusable</p>
    `;
  }
</script>
```

...which elements do the `<div>` _contain_? In well-defined terms, the `<div>` has the following:

- A single [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant): `<button>I'm focusable</button>`.
- Two [shadow-including descendants](https://www.w3.org/TR/dom41/#concept-shadow-including-descendant): `<button>I'm focusable</button>` and `<p>I'm not focusable</p>`.
- A single [descendant](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree): `<p>I'm not focusable</p>`.

If we interpret _contain_ to mean any of the first two options, the snippet will fail the rule, while the third option would actually pass the rule.

annethyme:
@WilcoFiers, could we maybe take this up on the f2f next week?

annethyme:
@kasperisager, can Siteimprove come up with a proposal?

EmmaJP:
So is the issue that a web page is not necessarily a static thing? And that the DOM could change over time? Do ACT Rules need to cover that in some way, or is that up to the tool makers?

Or is the issue that DOM tree, shadow DOM and accessibility tree can contain different things at the same time?

annethyme:
Auto-WCAG conclusion: Go ahead with the definition, but make sure it's readable and not too techy.

annethyme:
Will be handled by https://github.com/auto-wcag/auto-wcag/pull/432

PR #452, opened 2019-03-20
Rule update: Name from content included in accessible name (2ee8b8)

annethyme:
This pull request has been split out from #430.

Closes issue: 
- #377 - "label and name from content mismatch" lacks a definition of non-text content
- #451 - "label and name from content mismatch" needs to also fail split up labels

# How to Review And Approve


annethyme:
Discussion on "character" definition taken from the PR (#430) that this one has been split out from: 
https://github.com/auto-wcag/auto-wcag/pull/430#discussion_r266779011

@jkodu: 

> `character` seems like a generic definition that we can refer to elsewhere? Perhaps - https://developer.mozilla.org/en-US/docs/Glossary/Character
> 
> I see you drawing some distinction away from `grapheme`, then may be this should not be called `character`?

@kasperisager:

> The issue is that computers don't really deal in graphemes. A language such as JavaScript, for example, represents strings of "characters" as sequences of UTF-16 code points, while other languages, such as C and Go, represent strings of "characters" as sequences of arbitrary bytes. The term "character" therefore has lots of different meanings in different contexts.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+
+**Note**: [widget roles](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [support name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent) are: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
+
+### Expectation
+
+The complete [visible](#visible) [text](https://www.w3.org/TR/WCAG21/#dfn-text) of the target element is [included](#included-characters) within its [accessible name](#accessible-name).
+
+## Assumptions
+
+This rule assumes that the [visible](#visible) [text](ttps://www.w3.org/TR/WCAG21/#dfn-text) has the same order as the text in the code, e.g. CSS has not been used to re-arrange pieces of text so that they are presented in a different order.
```
kasperisager: This may not be needed anymore.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Background
+
+- https://www.w3.org/TR/WCAG21/#label-in-name
+
+## Test cases
+
+### Passed
+
+#### Passed example 1
```
ShadowBB: Move Passed example 1 to inapplicable. The visible text and name match.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+
+Visible label and accessible name matches when trailing white spaces are removed.
+
+```html
+<div role="link" aria-label="next page ">next page</div>
+```
+
+#### Passed example 2
```
ShadowBB: Move Passed example 2 to inapplicable. The visible text and name match.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: @kasperisager, You sure? I added it along with another bunch of comments from our discussions. Happy to drop it, though.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: Hm, unexpected side effect of changing the applicability.
Would it be more logical if the applicability was "has visible text, and has an accessible name" so that a full match will still be a pass?

annethyme: @ShadowBB, you were completely right about this, but I would like to have these as passed, so I changed the applicability instead. Could you have a look and see if you think it makes sense?

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: Same as above: 
@ShadowBB, you were completely right about this, but I would like to have these as passed, so I changed the applicability instead. Could you have a look and see if you think it makes sense?

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
ShadowBB: Yeah. That works!

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
ShadowBB: Ditto.

pages/glossary/included-characters.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Included characters
+key: included-characters
+---
+
+A sequence of [encoded characters](#encoded-character) is considered included in another if the latter includes any consecutive sequence of encoded characters that [matches](#matching-characters) the former.
```
WilcoFiers: Wow, you went all in on this definition huh? Can we collapse encoded character / matches into this, so we only have one new definition? I think that's a little more manageable.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+- Jey Nandakumar
+---
+
+## Test procedure
+
+### Applicability
+
+This rule applies to any HTML or SVG element that:
+* has a [semantic role](#semantic-role) that is a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [supports name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent), and 
+* has [visible](#visible) [text](https://www.w3.org/TR/WCAG21/#dfn-text), and
```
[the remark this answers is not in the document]
annethyme: I follow you. Will try to make a definition of "text" that clarifies some of the ambiguity.

pages/glossary/included-characters.md
annethyme: I would like to, but can we (in the current setup, or a planned one?) link to parts of a glossary term?  Because sometimes we need one of them, sometimes the other one.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+      left:-10000px;
+      top:auto;
+      width:1px;
+      height:1px;
+      overflow:hidden;}
+  </style>
+</head>
+
+<body>
+  <a href="home.html">Link <span style="sr-only">that opens a new tab</span> to the homepage</a>
```
carlosapaduarte: style should be class

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+      left:-10000px;
+      top:auto;
+      width:1px;
+      height:1px;
+      overflow:hidden;}
+  </style>
+</head>
+
+<body>
+  <a href="home.html">Read more<span style="sr-only"> about our pricing</span></a>
```
carlosapaduarte: style should be class

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+
+Not all of visible label is included in accessible name.
+
+```html
+<button name="link" aria-label="the full">The full label</button>
+```
+
+#### Failed example 3
+
+The full visible label is split out across the accessible name.
```
carlosapaduarte: Could you please explain why this fails? The SC states that the text of the label should be at the beginning of the accessible name, but that is just a best practice. I don't interpret it as a violation if it does not occur. 

Also, the expectation for this rule does not state that the visible text should be at the beginning of the accessible name or that it cannot be split out across the accessible name. 

I tried this on Chrome with VoiceOver and the AT reads "Link, that opens a new tab Link to the homepage". So, the visible text is there, it's just not at the start of accessible name.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+
+No rendered text in name from content.
+
+```html
+<div role="tooltip" aria-label="OK"></div>
+```
+
+#### Inapplicable example 5
+
+Sequence of characters in name from content is not expressing anything in human language and is therefore not considered [text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
carlosapaduarte: I think this is debatable. X, in this context (and I would argue that human language interpretation is context dependent), expresses something.

pages/glossary/included-characters.md
WilcoFiers: I don't quite understand. Maybe discuss the precise capabilities with Jey on Tuesday.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: I think the argument here is that "X" is only used because it looks like a cross, so it actually means "cross/close", not "X" as the character.

@WilcoFiers and @ShadowBB, can you guys chime in on this discussion?

WilcoFiers: There's definitely a point to be made here. In hindsight, I don't think this can be inapplicable. Applicability has to be objective, and determining whether or not "X" (or things like it) are meant as non-text content or as human language can be subjective. We should adjust this so it passes the rule.

I think that relates to my comment here:
https://github.com/auto-wcag/auto-wcag/pull/452/files#r269770000

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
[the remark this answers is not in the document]
annethyme: @carlosapaduarte, that one just said "[text] that is [visible]" so I figured we might as well write it out...

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,176 @@
...
+---
+
+## Test procedure
+
+### Applicability
+
+This rule applies to any HTML or SVG element that:
+* has a [semantic role](#semantic-role) that is a [widget](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [supports name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent), and 
+* has [visible](#visible) [text](https://www.w3.org/TR/WCAG21/#dfn-text), and
+* has an [accessible name](#accessible-name), even if the name is the empty string ("").
```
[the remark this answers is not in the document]
annethyme: The reason why I started out doing this editing was because we on our side had a validation issue due to the "text" definition. So this problem is from the original rule - the off-screening was added due to additional findings in the validation. I am just trying to solve both issues in one PR...

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
ShadowBB: It is not a matter of starting it at the beginning. This example fails because people that use voice controls might say "click link to the homepage" to their computers because that is what they can see. But there is no control with "link to the homepage" anywhere in there because there is invisible text between "link" and "to".

The word "complete" in the expectation might be ambiguous.  I would say that the complete visible text "link to the homepage" is not in there. There is "link" and there is "to the homepage", and neither are complete.

But you could also argue that "all the visible words are in the name, so the name has the complete visible text".

Maybe we should add the word "uninterrupted".

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
ShadowBB: I agree with Wilco that this is hard to make objective in the applicability. So it should be moved to the expectations so we can subjectively pass it.

Just to be clear: x expresses something. But it does not express something in human language according to the definition in WCAG. It says so in the intent of this SC:

> For the purposes of this SC, text should not be considered a visible label if it is used in a symbolic manner, rather than directly "expressing something in human language" as per the definition of text in WCAG. For example, 1.4.5 Images of Text describes considerations for "symbolic text characters." In the images of text example "B", "I", and "ABC" appear on icons in a text editor, where they are meant to symbolize the functions for Bold, Italics, and Spelling, respectively. In such a case, the accessible name should be the function the button serves (e.g., "Spell check" or "Check spelling"), not the visible symbolic characters. A similar text editor is shown in Figure 1.
Icons for transforming text, including heading, bold, italic, quote, code, and link, along with icons for ordered and unordered lists and other controls
Likewise, where an author has used a greater-than symbol (">") to mimic the appearance of the right-facing arrow, the text does not convey something in human language. It is a symbol, in this scenario likely meant to mimic the icons used for a "Play" button or a "Next" arrow.

pages/glossary/included-characters.md
annethyme: @WilcoFiers, In Alfa we have definitions like this: 

> ##Encoded character
> 
> An encoded character is...
> 
> A string of encoded characters are considered to [match]() each other if...
> 
> A string of characters are considered to be [included in]() another string if...

Where "match" and "included in" are their own definitions that you can link to if they are the ones you need specifically.

I thought it was this that you were going for, but maybe not.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: The definition of "included in" also has this: 

> ... includes any consecutive sequence of encoded characters that [matches](#matching-characters) the former.

So the full accessible name has to be included character by character in the order that the characters are in, according to this rule right now.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: I have tried moving the subjectiveness to the expectation now.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,186 @@
...
+### Applicability
+
+This rule applies to any HTML or SVG element that:
+* has a [semantic role](#semantic-role) that is a [widget role](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [supports name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent), and 
+* has [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text), and
+* has an [accessible name](#accessible-name), even if the name is the empty string ("").
+
+**Note**: [widget roles](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [support name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent) are: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
+
+**Note:** If a text node only has [whitespace](#whitespace), it is not [visible](#visible).
```
carlosapaduarte: Suggested improvement:
[Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Expectation
+
+The [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element are [included](#included-characters) in their entirety within the [accessible name](#accessible-name) of the element, unless the [text nodes](https://www.w3.org/TR/dom/#text) does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore does not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
carlosapaduarte: editorial: does > do (twice)

If the expectation is written using "unless" I fear this could be interpreted as applicability (at least that was my first interpretation). I would suggest replacing "unless" with "or". For me it is easier to interpret (the text nodes are included or they don't express anything in human language)

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,186 @@
...
+</head>
+
+<body>
+  <a href="home.html">Read more<span class="sr-only"> about our pricing</span></a>
+</body>
+```
+
+#### Passed example 5
+
+Text nodes in name from content are not expressing anything in human language and does therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
carlosapaduarte: does > do

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,186 @@
...
+
+```html
+<button aria-label="close">:-)</button>
+```
+
+#### Passed example 6
+
+Text nodes in name from content are not expressing anything in human language and does therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text). In this case, "X" is used because it resembles a "close" icon, not to represent the character "X".
```
carlosapaduarte: does > do

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
carlosapaduarte: It might be a good idea to have this explicit in the comment. Something like:

The full visible label is split out across the accessible name therefore not being [included](#included-characters) in it.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
kasperisager: My bad, the rule of course talks about text as it appears in code now, which we must assume is about the same as it appears on screen 🙈

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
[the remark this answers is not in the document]
annethyme: Wilco has agreed to keeping it in one rule, so I am marking this comment as resolved.

annethyme:
@ShadowBB, thanks for remembering this! I completely forgot it again.
I have added an assumption, with some edits to try to make it clearer (Shadi had some comments about it not being clear what "a sighted user will perceive").

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,187 @@
...
+
+**Note:** [Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).
+
+### Expectation
+
+The [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element are [included](#included-characters) in their entirety within the [accessible name](#accessible-name) of the element, or the [text nodes](https://www.w3.org/TR/dom/#text) do not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore do not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
+
+## Assumptions
+
+- This rule assumes that the [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text) has the same order as the text in the code, e.g. CSS has not been used to re-arrange pieces of text so that they are presented in a different order.
```
carlosapaduarte: [text nodes](https://www.w3.org/TR/dom/#text) has > [text nodes](https://www.w3.org/TR/dom/#text) follow

pages/glossary/included-characters.md
annethyme: @WilcoFiers, I previously merged the two definitions, but now I split them up again, since I realised that the "matching" definition is used in other rules, and for maintenance and consistency reasons, I find it better to link out to it than duplicating it within the "included character" definition.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,187 @@
...
+* has an [accessible name](#accessible-name), even if the name is the empty string ("").
+
+**Note**: [widget roles](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [support name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent) are: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
+
+**Note:** [Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).
+
+### Expectation
+
+The [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element are [included](#included-characters) in their entirety within the [accessible name](#accessible-name) of the element, or the [text nodes](https://www.w3.org/TR/dom/#text) do not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore do not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
DagfinnRomen: The success criterion applies to elements with labels that include text or images of text. Thus, an element with a label that does not include text, is inapplicable. I suggest that the text "or the text nodes do not express anything in human language and therefore do not live up to the WCAG definition of text" is removed from the expectation.

Also - add a bullet to the applicability section to show that the rule applies to elements with visible text?

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,187 @@
...
+</head>
+
+<body>
+  <a href="home.html">Read more<span class="sr-only"> about our pricing</span></a>
+</body>
+```
+
+#### Passed example 5
+
+Text nodes in name from content are not expressing anything in human language and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
DagfinnRomen: This should be moved to inapplicable, as the label does not contain text. The success criterion applies to elements with labels that include text or images of text.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,187 @@
...
+
+```html
+<button aria-label="close">:-)</button>
+```
+
+#### Passed example 6
```
DagfinnRomen: This should be moved to inapplicable, as the label does not contain text. The success criterion applies to elements with labels that include text or images of text.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: @DagfinnRomen, our issue is that the WCAG definition of "text" is ambiguous/subjective and thus cannot be used in the Applicability:

> text: sequence of characters that can be programmatically determined, where the sequence is _expressing something in human language_

(https://www.w3.org/TR/WCAG21/#dfn-text)

Examples where a text node is (probably not in everyone's opinion) expressing something in human language is: 
- X (for "close")
- :-)

Where subjectivity is not allowed in the Applicability, it is allowed in Expectations, so therefore we have had to move the "expresses something in human language" to the Expectation.
Yes, it is a bit weird, that it will then become a "Passed" instead of "Inapplicable" result if the visible label is found to not be expressing something in human language, but we have been back and forth a lot on this, and this is the best option he have been able to come up with.

annethyme: > Also - add a bullet to the applicability section to show that the rule applies to elements with visible text?

We already have this: 

> * has [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text), and

Do you think we need further clarification? 
I guess it would have to be a note then.

DagfinnRomen: I think this should have it covered.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
DagfinnRomen: You can disregard this comment. See response to previous comment.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
DagfinnRomen: You can disregard this comment. See response to previous comment.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: Thanks for getting back on this, @DagfinnRomen :)

mraccess77:
Why is this rule limited to widgets with accessible name from content.  Will there be a separate test for input fields with labels?  The SC 2.5.3 also applies to input, select, and text area fields with visual labels.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,187 @@
...
+## Test procedure
+
+### Applicability
+
+This rule applies to any HTML or SVG element that:
+* has a [semantic role](#semantic-role) that is a [widget role](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [supports name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent), and 
+* has [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text), and
+* has an [accessible name](#accessible-name), even if the name is the empty string ("").
+
+**Note**: [widget roles](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [support name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent) are: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
```
audreymaniez: `searchbox` seems not to be included in roles that support name from content.

annethyme: Thanks for paying attention, @audreymaniez! 

I will remove `searchbox` in the note. 
Because it's only a note and not in the normative section, the change won't require the rule to go into a second Final Call.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
```diff
@@ -0,0 +1,187 @@
...
+<div role="link" aria-label="next page ">next page</div>
+```
+
+#### Passed example 2
+
+Character insensitivity between visible label and accessible name.
+
+```html
+<div role="link" aria-label="Next Page">next page</div>
+```
```
kasperisager: Looking at this again, I noticed that the expectation no longer specifies what to do about differences in letter casing as well as leading and trailing whitespace. Wasn't there a definition of "matching characters" somewhere?

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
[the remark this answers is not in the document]
kasperisager: "the empty string" should probably be just "empty" to align with https://github.com/act-rules/act-rules.github.io/pull/430.

annethyme:
> Why is this rule limited to widgets with accessible name from content. Will there be a separate test for input fields with labels? The SC 2.5.3 also applies to input, select, and text area fields with visual labels.

Hi @mraccess77, 
Thanks for your review. We rarely try to cover a whole success criterion with a single rule. For most success criteria, we expect to see many different rules that each cover a set of particular cases (that have something in common).

This pull request is not writing the rule from scratch, even though it could look like it, since I changed the file name, and Github now considers it a brand new file.
I am only editing an existing rule, https://act-rules.github.io/rules/2ee8b8, to handle two specific issues that was brought up for it: 
- #377 - "label and name from content mismatch" lacks a definition of non-text content
- #451 - "label and name from content mismatch" needs to also fail split up labels

I don't off the top of my head remember the argument for why the applicability for this rule was originally cut out the way it is, but digging into old issues (https://github.com/act-rules/act-rules.github.io/issues/104) it seems like the suggestion for the applicability came from @WilcoFiers.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
annethyme: @kasperisager, it should still be the same. "included characters" definition refers to "matching characters" (at the very bottom of this page), and that says to remove leading and trailing space characters.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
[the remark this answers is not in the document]
annethyme: > "the empty string" should probably be just "empty" to align with #430.

Change done. I consider this change editorial, so therefore it will not impact the final review.

_rules/SC2-5-3-name-from-content-included-in-accessible-name.md
kasperisager: My bad, I was viewing the GitHub render of the file, which does not linkify the internal references 🤦‍♂ All good!

jkodu:
@annethyme - I resolved all the conflicts here (from the develop branch).

annethyme:
This rule was sent out in Final Call on April 29: 

> This in the Final Call for the rule "Name from content included in accessible name", https://github.com/act-rules/act-rules.github.io/pull/452, that has undergone a substantial update since first being published.
>  
> The final call for this rule will end on May 13.
> If you have any objections to the rule please raise them before this date.
>  
> We also encourage proceeding with implementations during the final call phase.

_rules/label-content-name-mismatch-2ee8b8.md
```diff
@@ -90,7 +131,29 @@ Not all of visible label is included in accessible name.
...
+      left:-10000px;
+      top:auto;
+      width:1px;
+      height:1px;
+      overflow:hidden;}
+  </style>
+</head>
+
+<body>
+  <a href="home.html">Link <span class="sr-only">that opens a new tab</span> to the homepage</a>
```
WilcoFiers: This seems like it meets the expectation. There are two visible text nodes, `link ` and ` to the homepage`, both of which are included in the accessible name. If they need to be treated as a single test string, we need to say that.

ShadowBB: Hmmm, I think this is a different interpretation of "in their entirety" than I originally had, but I now see and agree that this is a problem.

I have the feeling that adding the single word "together" in the expectation would solve this, however I am unsure where this should be placed grammatically speaking.

kasperisager: Good catch! The expectation should be akin to "the visible [child text content](https://dom.spec.whatwg.org/#concept-child-text-content)" instead.

annethyme: Changed the linked definition as suggested by @kasperisager, but I didn't really think that was the problem understanding wise for most of us, so I have added a note that clarifies what can also be read out of the used definitions for "included characters" and "matching characters".

@WilcoFiers and @ShadowBB, does this solve your concerns?

ShadowBB: @annethyme That solves it for me.

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Note:** [Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).
+
+## Expectation
+
+The [visible](#visible) [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element either:
+* is [included](#included-characters) in their entirety within the [accessible name](#accessible-name) of the element, or 
+* does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore do not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
+
+**Note:** Due to the definition of [included characters](#included-characters) the whole consequtive sequence of characters in the [visible](#visible) [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) of the target element has to be included character-by-character in the [accessible name](#accessible-name) to meet the expectation of this rule.
```
audreymaniez: Typo "consequtive" should be "consecutive"

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+#### Passed Example 6
+
+Text nodes in name from content are not expressing anything in human language and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text). In this case, "X" is used because it resembles a "close" icon, not to represent the character "X".
+
+```html
+<button aria-label="close">X</button>
+```
+
```
carlosapaduarte: What about
```html
<button aria-label="close">Close :-(</button>
```
With the current expectation, this fails because not all visible text content is included in the accessible name. But what is not included is not expressing something in human language, so I believe it should pass. Is this too much of an edge case, or should we work on the expectation to consider this?

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
```diff
@@ -0,0 +1,192 @@
...
+* has [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text), and
+* has an [accessible name](#accessible-name), even if the name is empty ("").
+
+**Note**: [widget roles](https://www.w3.org/TR/wai-aria-1.1/#widget_roles) that [support name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent) are: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `switch`, `tab`, `treeitem`.
+
+**Note:** [Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).
+
+## Expectation
+
+The [visible](#visible) [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element either:
```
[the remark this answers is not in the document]
annethyme: @carlosapaduarte, I am not pretending to understand this fully, but I hope you might: 

If we remove "in the flat tree", how do we know that the tree we are talking about is after shadow DOM has been applied? 

It comes out of the discussion on https://github.com/act-rules/act-rules.github.io/issues/315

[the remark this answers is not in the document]
annethyme: @kasperisager, do you have an opinion on whether we can safely remove "in the flat tree" here?

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
annethyme: @carlosapaduarte (and @kasperisager), do you think this change to the expectation could cover both the existing test cases + the new edge cases that have come up?

Current Expectation: 

> The [visible](#visible) [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element either:
> * is [included](#included-characters) in its entirety within the [accessible name](#accessible-name) of the element, or 
> * does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore does not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).

Suggested new expectation: 

> The [visible](#visible) [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) that are [descendants](https://www.w3.org/TR/dom41/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the target element is [included](#included-characters) in its entirety within the [accessible name](#accessible-name) of the element, except for parts that do not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore does not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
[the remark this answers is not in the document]
kasperisager: As long as "visible" is included, we're good; if text nodes aren't in the flat tree, they're not visible.

pages/glossary/matching-characters.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Matching characters
+key: matching-characters
+---
+
+A sequence of [characters](#character) is considered to match another if, after removing leading and trailing [space characters](https://www.w3.org/TR/html/infrastructure.html#space-characters) and replacing remaining occurrences of one or more space characters with a single space, the two sequences of characters are equal character-by-character, ignoring any differences in letter casing.
```
jkodu: `character-by-character` why is this hypenated?

annethyme: Fixed

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
annethyme: Another question: How about punctuation and special characters? Does that have to be the same 1:1? 

E.g.:
`<a href="#" aria-label="Press me">Press me!!!!!</a>`

... because in the edge case listed further up, only the "X" can be said to be non-text content, but the parenthesis should not be included in the match either: 
`<a href="#">Close (X) and go back</a>`

mraccess77:
My read of this SC is that the visible name must be in the accessible name -- but the accessible name can include additional text.  So we can't do a direct match but more of a string contains.  I agree we should exclude things that are not meant to be words like "x" and ">>", etc. based on the discussion at https://github.com/w3c/wcag/issues/592 and in the understanding document (https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html).  Discussion should also cover when placeholder is used in addition to another visual label -- otherwise placeholder is the visible label and the fallback accessible name.

Related WCAG 2.1 issues:
https://github.com/w3c/wcag/issues/722
https://github.com/w3c/wcag/issues/598

annethyme:
@mraccess77,

> My read of this SC is that the visible name must be in the accessible name -- but the accessible name can include additional text. So we can't do a direct match but more of a string contains. 

Agree. That is why we are using a definition of "included in" that means that the whole visible name has to be part of the accessible name, but there can be other text too.

> I agree we should exclude things that are not meant to be words like "x" and ">>", etc. based on the discussion at [w3c/wcag#592](https://github.com/w3c/wcag/issues/592) and in the understanding document (https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html). 
> 
> Related WCAG 2.1 issues:
> [w3c/wcag#722](https://github.com/w3c/wcag/issues/722)
> [w3c/wcag#598](https://github.com/w3c/wcag/issues/598)

Thanks for pointing to these great resources! I realise now that the Understanding document has been updated since we started writing this rule, and that clearly states that punctuation should be disregarded when testing for this SC. This will resolve a few of the questions that has popped up recently, though not all.

> Discussion should also cover when placeholder is used in addition to another visual label -- otherwise placeholder is the visible label and the fallback accessible name.

Since we just rely on the accessible name calculation as is, we should be covered for when placeholder acts as an accessible name.

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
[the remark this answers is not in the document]
annethyme: @kasperisager, do you also agree with Carlos that we can let go of "that are descendants of the target element"?

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
annethyme: Expectation changed to the suggestion above to allow for content that is a mix between text and content that does not express anything in human language. 
Punctuation now is explicitly not considered for the comparison due to an update of [the Understanding 2.5.3 Name in Label](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html) document.

annethyme: Test cases added for punctuation (Passed Example 6) and mixed text/non-text (Passed Example 7).

pages/glossary/matching-characters.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Matching characters
+key: matching-characters
+---
+
+A sequence of [characters](#character) is considered to match another if, after removing leading and trailing [space characters](https://www.w3.org/TR/html/infrastructure.html#space-characters) and replacing remaining occurrences of one or more space characters with a single space, the two sequences of characters are equal character by character, ignoring any differences in letter casing.
```
carlosapaduarte: I can't find a glossary entry for #character. Did you mean #encoded-character?

Jym77: The issue was already raised as #783. Last Tuesday. @WilcoFiers said that we don't really need a definition of "character" (and I removed the link from "matching characters"). I'll clean up that link too.

kasperisager: > @WilcoFiers said that we don't really need a definition of "character"

Let me for the record state that I had a chat with @WilcoFiers about this and hopefully convinced him that the term "character" is _extremely_ ambiguous unless properly defined.

jkodu:
@carlosapaduarte see comment - https://github.com/act-rules/act-rules.github.io/pull/452#discussion_r318918777

_rules/name-from-content-included-in-accessible-name-2ee8b8.md
```diff
@@ -0,0 +1,200 @@
...
+      overflow:hidden;}
+  </style>
+</head>
+
+<body>
+  <a href="home.html">Link <span class="sr-only">that opens a new tab</span> to the homepage</a>
+</body>
+```
+
+### Inapplicable
```
Jym77: We need examples for other inapplicable cases:
- no accessible name;
- text content is visible but not accessible (hence not a WCAG label);
- text content is accessible but not visible (hence not a WCAG label).

The later two cases, preferably with the text content not included in the name. _e.g._ 
`<a href="w3.org" aria-label="Home page of the W3C"><span aria-hidden=true>Go to W3C</span></a>`
is **not** a violation of 2.5.3 because there is no label (in the WCAG sense of the term). Thus, it should also be inapplicable for this rule.

Jym77:
A fun example I came up with in #977 
```html
<div role="link" aria-label="next page">
  <span id="next">next</span>
  <span id="space"> </span>
  <span id="bases" style="position: absolute; top: -9999px">All your bases are belong to us!</span>
  <span id="page">page</span>
</div>
```
I think this should pass the rule, but it is interesting to show how the visible text content is built…

Jym77:
Another fun example, related to #976:
```html
<div role="link" aria-label="next 
   
   page">Next page</div>
```
As far as I understand accessible name computation, the extra spaces and newline should be removed from the accessible name when it is flattened and thus this should pass the rule even though the name from content is not included in the `aria-label`.
More elaborate cases with `aria-labelledby` reading several other elements can also be achieved.

jkodu:
Reviewers, I have updated this rule. Please review again.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -22,73 +22,125 @@ acknowledgements:
...
+- [Technique G211: Matching the accessible name to the visible label](https://www.w3.org/WAI/WCAG21/Techniques/general/G211)
 
 ## Test Cases
 
 ### Passed
 
 #### Passed Example 1
 
-[Visible][] label and [accessible name][] matches when trailing white spaces are removed.
+[Visible][] [label][] and [accessible name][] matches when trailing white spaces are removed.
```
carlosapaduarte: matches > match

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -22,73 +22,125 @@ acknowledgements:
...
+
+Text nodes in name from content are not expressing anything in human language and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
+
+```html
+<button aria-label="close">:-)</button>
+```
+
+#### Passed Example 6
+
+[Visible][] [label][] and [accessible name][] matches when punctuation is removed.
```
carlosapaduarte: matches > match

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -22,73 +22,125 @@ acknowledgements:
...
 
 ```html
-<button name="link" aria-label="Next Page in the list">Next Page</button>
+<div role="link" aria-label="Next page">Next page...</div>
+```
+
+#### Passed Example 7
+
+[Visible][label][] and [accessible name][] matches when parts that do not express anything in human language, "X", and punctuation is removed.
```
carlosapaduarte: [Visible][label][] > [Visible][] [label][]

carlosapaduarte: matches > match

carlosapaduarte: is > are

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -22,73 +22,125 @@ acknowledgements:
...
+</head>
+
+<body>
+	<a href="https://github.com/pricing">Read more<span class="sr-only"> about our pricing</span></a>
+</body>
+```
+
+#### Passed Example 5
+
+Text nodes in name from content are not expressing anything in human language and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
carlosapaduarte: Complete the sentence clearly explaining why this passes.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -22,73 +22,125 @@ acknowledgements:
...
+[Visible][] [label][] and [accessible name][] matches when trailing white spaces are removed.
 
 ```html
 <div role="link" aria-label="next page ">next page</div>
 ```
 
 #### Passed Example 2
 
-Character insensitivity between [visible][] label and [accessible name][].
+Character insensitivity between [visible][] [label][] and [accessible name][].
```
carlosapaduarte: Perhaps you can rewrite the sentence in a similar way to the previous example to make it clear why it passes.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -102,6 +154,31 @@ Mathematical symbols cannot be substituted for text as [explicitly mentioned in
...
+</head>
+
+<body>
+	<a href="https://github.com/features/actions"
+		>Link <span class="sr-only">that opens a new tab</span> to the homepage</a
+	>
+</body>
+```
+
 ### Inapplicable
```
carlosapaduarte: Inapplicable example 4 uses a definition that is not used in the applicability (or the expectation). It would be better to replace it with definitions used in the applicability.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
 
 ## Assumptions
 
-_There are currently no assumptions_
+- This rule assumes that the [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text) follow the same order as the text in the code, e.g. CSS has not been used to re-arrange pieces of text so that they are presented in a different order.
+- This rule assumes that the whole set of [visible](#visible) [text nodes](https://www.w3.org/TR/dom/#text) of the element are styled, including placement, in such a way that a sighted user will perceive them all as part of the label of the element. An example where this would not be the case, is where styling is used to split up the visible text nodes in e.g. two parts and place one part further away, or give it a different styling that makes it appear as not connected to the element it labels.
```
Jym77: label => link to WCAG def.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+- [Technique G211: Matching the accessible name to the visible label](https://www.w3.org/WAI/WCAG21/Techniques/general/G211)
 
 ## Test Cases
 
 ### Passed
 
 #### Passed Example 1
 
-[Visible][] label and [accessible name][] matches when trailing white spaces are removed.
+[Visible][] [label][] and [accessible name][] match when trailing white spaces are removed.
```
Jym77:
```suggestion
The [visible][] [descendant text content][] and [accessible name][] match when trailing white spaces are removed.
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
 
 ```html
 <div role="link" aria-label="next page ">next page</div>
```
Jym77: I'd prefer to use an element that doesn't require explicit role to be applicable.

_rules/visible-label-in-accessible-name-2ee8b8.md
Jym77: Removing trailing whitespace is part of the accessible name computation. So visible descendant text content and accessible name are exactly the same in this case.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
 ```
 
 #### Passed Example 2
 
-Character insensitivity between [visible][] label and [accessible name][].
+[Visible][] [label][] and [accessible name][] match when ignoring case sensitivity.
```
Jym77: match => link to def (to make obvious the fact that case sensitivity is ignored when matching).

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
 
 ```html
 <div role="link" aria-label="Next Page">next page</div>
 ```
 
 #### Passed Example 3
 
-Full [visible][] label is contained in the [accessible name][].
+Full [visible][] [label][] is contained in the [accessible name][].
```
Jym77:
```suggestion
The full [visible][] [descendant text content][] is contained in the [accessible name][].
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+
+```html
+<button name="link" aria-label="View next Page in the list">Next Page</button>
+```
+
+#### Passed Example 4
+
+Full [visible][] [label][] is contained in the [accessible name][], though the name contains more text than the [visible][] [label][].
```
Jym77:
```suggestion
The full [visible][] [descendant text content][] is contained in the [accessible name][], though the name contains more text than the [visible][] [descendant text content][].
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+<head>
+	<style>
+		.sr-only {
+			position: absolute;
+			left: -10000px;
+			top: auto;
+			width: 1px;
+			height: 1px;
+			overflow: hidden;
+		}
```
Jym77: I think this style only need its first two lines for the example to work.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+</head>
+
+<body>
+	<a href="https://github.com/pricing">Read more<span class="sr-only"> about our pricing</span></a>
+</body>
+```
+
+#### Passed Example 5
+
+[Descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
Jym77:
```suggestion
The [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+[Visible][] [label][] and [accessible name][] match when punctuation is removed.
 
 ```html
-<button name="link" aria-label="Next Page in the list">Next Page</button>
+<div role="link" aria-label="Next page">Next page...</div>
+```
+
+#### Passed Example 7
+
+[Visible][] [label][] and [accessible name][] match when parts that do not express anything in human language, "X", and punctuation are removed.
```
Jym77:
```suggestion
The [visible][] [descendant text content][] and [accessible name][] match when parts that do not express anything in human language, ("X"), and punctuation are removed.
```
I think it would be better to use a Unicode symbol for cross/closing rather than a letter. That would illustrate more clearly that it is not human language (but some sort of icon).

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+```html
+<div role="link" aria-label="Close and go back">Close (X) and go back</div>
 ```
 
 ### Failed
 
 #### Failed Example 1
 
-[Visible][] label doesn't match [accessible name][].
+[Visible][] [label][] doesn't match [accessible name][].
```
Jym77:
```suggestion
The [visible][] [descendant text content][] doesn't match the [accessible name][].
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
 
 ```html
 <div role="link" aria-label="OK">Next</div>
 ```
 
 #### Failed Example 2
 
-Not all of [visible][] label is included in [accessible name][].
+Not all of [visible][] [label][] is included in [accessible name][].
```
Jym77:
```suggestion
Not all of the [visible][] [descendant text content][] is included in the [accessible name][].
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -102,6 +158,31 @@ Mathematical symbols cannot be substituted for text as [explicitly mentioned in
 <a href="/" aria-label="Proof of two multiplied by two is four">Proof of 2&times;2=4</a>
 ```
 
+#### Failed Example 4
+
+The full [visible][] [label][] is split out across the [accessible name][] and is therefore not [included](#included-characters) in it.
```
Jym77:
```suggestion
The full [visible][] [descendant text content][] is split out across the [accessible name][] and is therefore not [included](#included-characters) in it.
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -102,6 +158,31 @@ Mathematical symbols cannot be substituted for text as [explicitly mentioned in
...
+```html
+<head>
+	<style>
+		.sr-only {
+			position: absolute;
+			left: -10000px;
+			top: auto;
+			width: 1px;
+			height: 1px;
+			overflow: hidden;
```
Jym77: Trim down styling.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -130,7 +211,7 @@ Non-widget role that supports name from content.
 
 #### Inapplicable Example 4
 
-No [rendered text](#rendered-text) in name from content.
+Element has no [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant).
```
Jym77:
```suggestion
This element has no [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant).
```

pages/glossary/encoded-character.md
```diff
@@ -0,0 +1,16 @@
...
+key: encoded-character
+---
+
+An encoded character is a single [Unicode](http://unicode.org/) [scalar value](http://unicode.org/glossary/#unicode_scalar_value).
+
+**Note:** A character is not to be confused with a [grapheme](https://en.wikipedia.org/wiki/Grapheme), that is a single, distinct symbol used in writing, or a [glyph](https://en.wikipedia.org/wiki/Glyph), which is a written representation of a grapheme.
+A character is a [code point](http://unicode.org/glossary/#code_point), specifically an integer, that a computer uses for representing a value within the Unicode system. 
+As such, multiple characters may represent the same glyph and multiple glyphs may represent the same grapheme.
+
+## Examples
```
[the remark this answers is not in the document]
Jym77: Still needs to be done.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,133 @@ acknowledgements:
...
+
+**Note:** [Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).
 
 ## Expectation
 
-The complete [visible text content](#visible-text-content) of the target element either matches or is contained within its [accessible name][].
+After removing all [punctuation](#punctuation) from both the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) and the [accessible name](#accessible-name), the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) of the target element is [included](#included-characters) in its entirety within the [accessible name](#accessible-name) of the element, except for parts of the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) that do not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore does not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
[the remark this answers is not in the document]
Jym77: I've been talking with @kasperisager about this (and my related issues with the definition of visible, see #977).

One way out of it would be to have a definition of "rendered node" (or similar) in the line of:
> a node is rendered if removing it from the DOM tree changes the rendered content.

That way, the `span id="space"` above would contain a rendered text node (with text content `" "`). But the `span id="hidden"` and `span id="off-screen"` would not.

Similarly, an example like https://github.com/act-rules/act-rules.github.io/issues/977#issuecomment-552869788
```html
<p style="background: #white; text-shadow: black 0 0 0px">
  <span style="color: rgba(0,0,0,0); ">Some text in a human language</span>
</p>
```
where the text is technically not visible (only the shadow is) would still have a rendered text node. 

---
The drawbacks with that are:
* Some nodes like above can be invisible but still have rendered text nodes descendants. That's a bit weird but fairly OK (and anyway a consequence of whitespace being invisible).
* If the shadow above is made more blurry, it gets harder and harder to read its content, to the point that it gets hard to argue that the text content should be used (_e.g._ to consider it as a WCAG label).
* Something like `<span style="background: red; color: red;">Foo</span>` has a rendered text node because the `span` stretch for the same width as the text.

(because of these tricky cases and the weird interaction with visible, the definition of rendered can probably use examples…)

All these are pretty edge cases and they can be handled by an assumption in the line of
> This rule assumes that no text node is both rendered and not visible.

(likely needs an explanatory note with examples like spaces)
(essentially, "don't use spaces for layout!")

---
With an assumption like that, this rule can use "visible descendant text nodes" or something similar to grab what is needed.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,130 @@ acknowledgements:
...
+</head>
+
+<body>
+	<a href="https://github.com/pricing">Read more<span class="sr-only"> about our pricing</span></a>
+</body>
+```
+
+#### Passed Example 5
+
+The [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and do therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
carlosapaduarte:
```suggestion
The [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) does not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and does therefore not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -18,77 +18,130 @@ acknowledgements:
...
+
+**Note:** [Text nodes](https://www.w3.org/TR/dom/#text) with only [whitespace](#whitespace) are not [visible](#visible).
 
 ## Expectation
 
-The complete [visible text content](#visible-text-content) of the target element either matches or is contained within its [accessible name][].
+After removing all [punctuation](#punctuation) from both the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) and the [accessible name](#accessible-name), the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) of the target element is [included](#included-characters) in its entirety within the [accessible name](#accessible-name) of the element, except for parts of the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) that do not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore does not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```
carlosapaduarte:
```suggestion
After removing all [punctuation](#punctuation) from both the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) and the [accessible name](#accessible-name), the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) of the target element is [included](#included-characters) in its entirety within the [accessible name](#accessible-name) of the element, except for parts of the [visible](#visible) [descendant text content](https://dom.spec.whatwg.org/#concept-descendant-text-content) that do not express anything in [human language](https://www.w3.org/TR/WCAG21/#dfn-human-language-s) and therefore do not live up to the [WCAG definition of text](https://www.w3.org/TR/WCAG21/#dfn-text).
```

WilcoFiers:
Closing due to inactivity.

PR #454, opened 2019-03-22
New rule: Required state for ARIA widget has valid value

annethyme:
A suggestion for one more rule that could come out of "ARIA state or property has valid value", if we need to split it up over different success criteria.

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/365

# How to Review And Approve


annethyme:

PR #455, opened 2019-03-25
Fix casing of filename

kasperisager:


annethyme:
@WilcoFiers, is it okay if I just go ahead and merge this one? I see it as editorial.

Issue #376, opened 2018-12-17
Is it possible to limit the applicability of "ARIA state or property has valid value" (6a7281)?


kasperisager:
Per the assumption of https://auto-wcag.github.io/auto-wcag/rules/SC4-1-2-aria-state-or-property-has-valid-value.html, invalid states or properties specified on elements that are either included in the accessibility tree or are focusable can still impact accessibility; on this we agree. During validation of the rule by @ShadowBB he however found a case where this assumption did not hold and a false positive was therefore encountered. `aria-invalid="flase"` was specified on an element excluded from the accessibility tree and subsequently flagged, but as `aria-invalid` does in no way participate in the accessible name computation this did not cause any actual issues.

My question is this: Can we in any way limit the applicability of the rule to only consider cases where states and properties excluded form the accessibility tree could cause _actual_ issues? For example, only `aria-valuetext` and `aria-valuenow` seem to participate in the accessible name computation so any other states or properties could be ignored if they where excluded from the accessibility tree.

WilcoFiers:
Discussed on the call, the group doesn't feel strongly that this needs to changes, especially because the case that raised this was actually a problem. If you feel this is a major problem, please put together a proposal.

jkodu:
@annethyme - could you look into this.

annethyme:
Related to https://github.com/auto-wcag/auto-wcag/issues/365

ShadowBB:
In theory this rule (at the time of writing this issue) could have caused a false positive in the following edge case scenario:

Somebody copy-pastes some code:
- That has an invalid aria attribute value
- That is not exposed to the accessibility tree
- That has no influence on the accessible name computation
- and finally also is never exposed to the accessibility tree because the form of interaction that would expose it is never used in this specific version of the code.

However(!) currently the rule has an assumption ("The ARIA state or property is being used to comply to WCAG.") that renders this issue moot.

Seeing as the aria attribute not (and never) being exposed to the accessibility tree also(!) guarantees that  the attribute isn't placed there to comply to WCAG the rule is moot anyway.

So I say we close this issue as it is no longer an issue.

annethyme:
We have tried adding this to the assumption to handle concerns raised here: 

[...]

PR #464, opened 2019-03-27
Non-WCAG rule: ARIA state/property has valid value

annethyme:
This is now a WAI-ARIA validation rule. 
It does not refer to any WCAG success criterion, but to WAI-ARIA 1.1 as its accessibility requirement (the Auto-WCAG template does not allow this, however).

Builds on issues:
- #376 (Is it possible to limit the applicability of "ARIA state or property has valid value"?)
- #365 (Update Rule: ARIA state or property has valid value)

# How to Review And Approve


annethyme:

PR #488, opened 2019-04-24
chore: Update templates to CR format

WilcoFiers:
Updating the template for ACT Rules to follow the new CR format. Following this PR, we'll start updating all of the published rules to use the new format as well.


pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,38 @@
...
 rule_type: atomic
 
 description: |
   This rule checks ...
 
-success_criterion:
-  - x.x.x # (Name of Success Criterion)
+accessibility_requirements:
+  wcag20:x.x.x: # (Name of Success Criterion)
+    forConformance: true
```
carlosapaduarte: shouldn't it be 
true | false

pages/design/composite-template-empty.md
```diff
@@ -1,37 +1,40 @@
...
 name:
 rule_type: composite
 description: |
   This rule checks ...
 
-success_criterion:
-  - x.x.x # (Name of Success Criterion)
+accessibility_requirements:
+  wcag20:x.x.x: # (Name of Success Criterion)
+    forConformance: true
```
carlosapaduarte: Same here
true | false

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,38 @@
...
 name:
 rule_type: atomic
 
 description: |
   This rule checks ...
 
-success_criterion:
-  - x.x.x # (Name of Success Criterion)
+accessibility_requirements:
+  wcag20:x.x.x: # (Name of Success Criterion)
```
annethyme: Should we have other examples, e.g. a technique?
I guess that most rules that test for a SC also test for a technique. So far we have listed it under Background, but it would make sense to have it as structured data, when possible.

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,38 @@
...
 authors:
   -  # (full names as found in `contributors` property in  `package.json` - if not yet listed, please have authors added to the list)
 ---
 
-## Test Procedure
-
-### Applicability
+## Applicability
 
 The rule applies to any (??) element ...
```
annethyme: Can we make it "This rule applies to..." since this is the wording we mostly use.
And should it be a comment instead, since this is not always the wording used?

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,38 @@
...
 
-### Expectation (1)
+## Expectation (1)
 
 Each target element ...
```
annethyme: As above: Should it be a comment instead, since this is not always the wording used?

pages/design/atomic-template-empty.md
```diff
@@ -41,42 +44,42 @@ _There are no major accessibility support issues known for this rule._
 
 ### Passed
 
-#### Passed example 1
+#### Passed Example 1
 
 Description...
```
annethyme: Should this be a comment?

pages/design/atomic-template-empty.md
```diff
@@ -41,42 +44,42 @@ _There are no major accessibility support issues known for this rule._
...
+#### Passed Example 2
 
 ...
 
 ### Failed
 
-#### Failed example 1
+#### Failed Example 1
 
 Description...
```
annethyme: Again: Should this be a comment?

pages/design/atomic-template-empty.md
```diff
@@ -41,42 +44,42 @@ _There are no major accessibility support issues known for this rule._
...
+#### Failed Example 2
 
 ...
 
 ### Inapplicable
 
-#### Inapplicable example 1
+#### Inapplicable Example 1
 
 Description...
```
annethyme: Again: Should this be a comment?

pages/design/composite-template-empty.md
```diff
@@ -1,37 +1,40 @@
...
+    failed: not satisfied
+    passed: satisfied | further testing needed
+    inapplicable: satisfied | further testing needed
 
-atomic_rules:
+input_rules:
   -
```
annethyme: Should there be a comment here, about how we refer to input rules? I would suggest by id

pages/design/composite-template-empty.md
```diff
@@ -1,37 +1,40 @@
...
 authors:
   -  # (full names as found in `contributors` property in  `package.json` - if not yet listed, please have authors added to the list)
 ---
 
-## Aggregation Definition
-
-### Applicability
+## Applicability
 
 The rule applies to any (??) element ...
```
annethyme: Again: I would suggest "This rule applies to..."

pages/design/composite-template-empty.md
```diff
@@ -1,37 +1,40 @@
...
 
-### Expectation
+## Expectation (1)
 
 For each test target, the outcome of (at least one of / all of / any of etc.) the following rules is (passed / failed / etc.):
```
annethyme: in other places the pipe character is used to show that you have to pick one. Do we need more consistency?

pages/design/composite-template-empty.md
```diff
@@ -1,37 +1,40 @@
...
 
 - (list atomic rules that are used in this composite rule)
```
annethyme: Again: Should there be a comment here about how we refer to input rules? I would suggest by id

pages/design/rule-template.md
```diff
@@ -6,41 +6,44 @@ The rule template contains a plain language description of the rule, some backgr
 
 Use the [empty atomic rule template](atomic-template-empty.html) to create new rule. When creating a new rule, first read [rule design](rule-design.html).
 
-## Atomic Rule template
+## Atomic Rule Template
```
annethyme: Would it be possible to embed the template file, so it doesn't have to maintained in two different places? 
They have previously been a bit out of sync, which is confusing.

pages/design/rule-template.md
```diff
@@ -57,90 +60,94 @@ _There are no major accessibility support issues known for this rule._
...
 
-## Composite rules
+## Composite Rules
 
 Composite rules are rules that take results from different rules and through some logic come to a single result. For example: SC 1.2.3 allows video to pass with either a transcript, an audio description, or if it is a media alternative. Each of these would be atomic rules, that are used in a composite rule. The composite rule than describes that at least one of these must pass for the composite rule to pass.
 
 For more about composite rules, see the [ACT Rules Format](https://www.w3.org/TR/act-rules-format/#composed-rules). To create a composite rule, use the [empty composite rule template](./composite-template-empty.html).
 
-### Composite Rule template
+### Composite Rule Template
```
annethyme: Again: Would it be possible to embed the template file, so it doesn't have to maintained in two different places?
They have previously been a bit out of sync, which is confusing.

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,38 @@
...
 
-## Test Procedure
-
-### Applicability
+## Applicability
 
 The rule applies to any (??) element ...
 
-### Expectation (1)
+## Expectation (1)
```
annethyme: What did we agree on here? 
Does expectations still only have to be numbered, if there are multiple of them?

WilcoFiers: Only number them when there is more than one expectation.

pages/design/rule-template.md
WilcoFiers: Probably, that seems separate from this pull request.

pages/design/rule-template.md
WilcoFiers: Probably, that seems separate from this pull request.

pages/design/atomic-template-empty.md
annethyme: ## FOR TEMPLATE UPDATE OF RULES

Does this need to be "indented" so that it is clear that it is "under" the success criterion? 

And should it have list format to not break the table layout that things are shown in as preview?

pages/design/atomic-template-empty.md
annethyme: ## FOR TEMPLATE UPDATE OF RULES

When updating rules to the template it seems that the format "wcag20:x.x.x" breaks the table view in the preview, whereas "wcag20: x.x.x" (with space) doesn't. 
Which one should we go for?

annethyme: Might be needed: Instructions for what to do if no accessibility requirements exist.

https://github.com/act-rules/act-rules.github.io/pull/500#issuecomment-489584544

ShadowBB: I agree! Especially when the rule is part of a composite rule.

annethyme: ## FOR TEMPLATE UPDATE OF RULES

Today we sometimes list the conformance level next to the name of the success criterion, e.g. "# Name, Role, Value (A)". 
Do we want to do this going forward or get rid of the conformance level in all places where it is found like this?

jkodu:
Worth noting that the build of the current site will break after this is merged, because the frontmatter id `success_criterion` will not exist anymore.

pages/design/atomic-template-empty.md
annethyme: Discussed with Wilco. It's important to keep the indentation, even though the table in preview unfortunately breaks.

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,50 @@
...
 rule_type: atomic
 
 description: |
   This rule checks ...
 
-success_criterion:
-  - x.x.x # (Name of Success Criterion)
-
-test_aspects:
+accessibility_requirements: # Remove whatever is not applicable
```
jkodu: @WilcoFiers 
If there are no requirements, like in this PR https://github.com/act-rules/act-rules.github.io/pull/523/files#diff-f1ba9ad105ee9aa5ee46dfd067f72397R7, is it worth adding `accessibility_requirements`?

pages/design/composite-template-empty.md
WilcoFiers: I prefer "the rule applies to". Please open an issue if you'd like to discuss this.

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,50 @@
...
+    passed: satisfied | further testing needed
+    inapplicable: satisfied | further testing needed
+    
+  aria11:some-id: # <Some heading in WAI-ARIA>
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied | further testing needed
+    inapplicable: satisfied | further testing needed
+
+  wcag-technique:H123: # <Technique title>
```
annethyme: Should it be "wcag-technique:XXX:" so that it kind of matches the "wcag20:X.X.X" format?

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,50 @@
...
-
-test_aspects:
+accessibility_requirements: # Remove whatever is not applicable
+  wcag20:x.x.x: # <Name of Success Criterion> (A | AA | AAA)
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied | further testing needed
+    inapplicable: satisfied | further testing needed
+    
+  aria11:some-id: # <Some heading in WAI-ARIA>
```
annethyme: For WAI-ARIA, can we use the id's that are actually used in WAI-ARIA 1.1, e.g. "#region" (https://www.w3.org/TR/wai-aria-1.1/#region) or "#requiredState" (https://www.w3.org/TR/wai-aria-1.1/#requiredState) ?

pages/design/atomic-template-empty.md
annethyme: @carlosapaduarte, I guess the point here is that success criteria can only be listed here, if failing the rule means failing the succes criterion, and in that case the rule will always be required for conformance. 

If it's not required for conformance, a technique could possibly be listed as the accessibility requirement as now shown on line 22.

pages/design/atomic-template-empty.md
```diff
@@ -1,35 +1,50 @@
...
+    failed: not satisfied
+    passed: satisfied | further testing needed
+    inapplicable: satisfied | further testing needed
+
+  wcag-technique:H123: # <Technique title>
+    forConformance: false
+    failed: not satisfied
+    passed: satisfied | further testing needed
+    inapplicable: satisfied | further testing needed
+
```
annethyme: @WilcoFiers, can we add the possibility for if nothing is required as well? 

Or, if it all becomes too much to have in the template, just write a basic template for how to do this, and then have each of the more detailed examples (WCAG/WAI-ARIA/techniques/nothing) in a separate place, maybe on a page on the website that we link out to in a comment in this section.

pages/design/atomic-template-empty.md
annethyme: I would still like to see this changed, if I am to approve this PR.

pages/design/atomic-template-empty.md
annethyme: ... or we can keep it as it is

pages/design/atomic-template-empty.md
jkodu: Don't think so, this is quite obvious in my opinion.

pages/design/atomic-template-empty.md
jkodu: See - https://github.com/act-rules/act-rules.github.io/pull/488/files#r284186950

pages/design/atomic-template-empty.md
jkodu: I think the intention here is to say, write an expectation. It gets conveyed in my opinion, so leave it as is.

pages/design/atomic-template-empty.md
jkodu: `The` is a definite article in English grammar used when the author/ reader knows exactly what is being referred to.

Eg: 
`It is the newest building in the town.`
`Who is the president of Madagascar?`

Given we are in the context of the rule. I believe `the` is the right usage over `this`.

pages/design/atomic-template-empty.md
jkodu: I think the possibility of having no `accessibility_requirements` is something that should be explained elsewhere, perhaps in the documentation and wiki that is put together by @nitedog.

jkodu: Once that page is in place, it can be cross-referred here. This work should not withhold this PR.

pages/design/atomic-template-empty.md
jkodu: Given we are using `x.x.x` everywhere, I think this should use the came convention here. 

It may be worth adding an example for each of the sections, but again this could be explained in this document - https://github.com/act-rules/act-rules.github.io/pull/488/files#r284189403

pages/design/composite-template-empty.md
jkodu: Also see - https://github.com/act-rules/act-rules.github.io/pull/488#discussion_r284188864

WilcoFiers:
@annethyme I'm merging this PR. If you have any further concerns, please open an issue.

Issue #470, opened 2019-04-10
What is the naming convention for rule files?


jkodu:
In some places we give rule files a complete human readable name and in other places we generate a 6 character sha, unsure which one works better.

See - https://github.com/auto-wcag/auto-wcag/pull/416/files#diff-300c950393b60709c0c304906c118f5d for example.

jkodu:
PR has been merged. Closing.

PR #489, opened 2019-04-25
chore: rename files and update associations

jkodu:
- renamed `rules` to follow `{filename}-{unique-id}` format.
- update associations of atomic rules, which are reliant on file name.
- opened a new issue for discussion - https://github.com/act-rules/act-rules.github.io/issues/490 (something to discuss on next agenda).

Closes issue: 
- https://github.com/act-rules/act-rules.github.io/issues/470


jkodu:
@annethyme - please wait for this to be merged, before picking up the rule format changes. Thanks.

_rules/video-only-media-alternative-fd26cf.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: fd26cf
-name: Video-only as a media alternative for text
+name: Video only as a media alternative for text
```
WilcoFiers: This change is incorrect. This isn't a "video, only as a media alternative". It's a "video-only solution, as a media alternative for text". Please revert.

jkodu:
Note: I have merged this, as this is a chore task & does not mandate multiple reviews.

Issue #481, opened 2019-04-19
Add manual reporting download links to rules


WilcoFiers:
To do manual evaluations I am thinking we should use the WCAG-EM Report Tools. We would test each rule separately, rather than test everything at once. This will keep the work less overwhelming.

The WCAG-EM Report Tool has a way to open audits stored in JSON, and a way to save it. So the way this would work is someone would download a rule audit file from the ACT-R website, they would open it and run through it in the report tool, and at the end they would download the JSON file and publish it to the ACT-R repository.

---

To support this, we need to add a link to each rule which will allow someone to download a JSON file that can be opened into the WCAG-EM report tool. It must have the following filled out:

- [x] All test cases are put into the structured sample
- [x] The audit scope is defined
- [x] All success criteria the rule maps to are set to Cannot tell
- [x] All success criteria that the rule does not map to are set to inapplicable

jkodu:
PR has been merged. Closed.

PR #492, opened 2019-04-29
chore: add manual reporting download links to rule

jkodu:
- This PR generates a JSON file for **each** rule in the mapping expected by WCAG EM Report tool - https://www.w3.org/WAI/eval/report-tool/#/, hence allowing for creating manual implementation reports for each rule.

- The test cases json is made available on each rule page.

- I had to run prettier to format some files, so apologies for the number of files that may have spacing changes, this is prettier's doing.

Screenshots:
image
image

Closes issue: 
- https://github.com/act-rules/act-rules.github.io/issues/481


jkodu:
@audreymaniez the changes in the `_rules` files are only formatting related.

jkodu:
Note: merging this as this is a chore & does not require 3 approvals.

PR #495, opened 2019-05-03
Template update - ARIA attribute is valid

annethyme:
Template update

### DIFF
--- a/_rules/aria-attr-valid-5f99a7.md
+++ b/_rules/aria-attr-valid-5f99a7.md
@@ -5,7 +5,7 @@ rule_type: atomic
 description: 
   This rule checks that each aria- attribute specified is defined in ARIA 1.1
 accessibility_requirements: 
-   wcag20: 4.1.2
+   wcag20: 4.1.2 # Name, Role, Value
    forConformance: true
    failed: not satisfied
    passed: further testing needed
--- a/_rules/aria-attr-valid-5f99a7.md
+++ b/_rules/aria-attr-valid-5f99a7.md
@@ -1,23 +1,26 @@
 ---
 id: 5f99a7
-name: ARIA attribute is valid
-description: |
+name: ARIA attribute is valid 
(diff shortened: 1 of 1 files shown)


_rules/aria-attr-valid-5f99a7.md
```diff
@@ -1,23 +1,26 @@
...
-description: |
+name: ARIA attribute is valid 
+rule_type: atomic 
+description: 
   This rule checks that each aria- attribute specified is defined in ARIA 1.1
-success_criterion:
-  - 4.1.2
-test_aspects:
+accessibility_requirements: 
+   wcag20: 4.1.2 # Name, Role, Value
```
annethyme: Points from meeting with Wilco: 
- Indentation is important for this. There has to be 3 levels. And no bullets.
- No space between colon and success criterion number, e.g "wcag20:4.1.2" - even though it breaks the table in the preview.
- After the success criterion number (e.g. "4.1.2") there has to be a colon for the whole thing to become a YAML object (so "4.1.2:"
- For some rules the conformance level (e.g. "(A)") was listed after the success criterion name (in the comment) in the old rules. Wilco thinks we should keep this, which means it will have to be added to all rules. 

> accessibility_requirements:
>   wcag20:x.x.x: # <Name of Success Criterion> (A | AA | AAA)
>     forConformance: true
>     failed: not satisfied
>     passed: satisfied | further testing needed
>     inapplicable: satisfied | further testing needed

_rules/aria-attr-valid-5f99a7.md
```diff
@@ -1,23 +1,26 @@
 ---
 id: 5f99a7
-name: ARIA attribute is valid
-description: |
+name: ARIA attribute is valid 
+rule_type: atomic 
+description:
```
annethyme: missing the pipe character " |" that allows for multiple lines...

PR #497, opened 2019-05-03
Template update - ARIA state or property allowed

annethyme:
Template update


_rules/aria-state-or-property-allowed-5c01ea.md
```diff
@@ -1,23 +1,27 @@
...
 id: 5c01ea
 name: ARIA state or property allowed
+rule_type: atomic
 description: |
   This rule checks that WAI-ARIA states or properties are allowed for the element they are specified on.
-success_criterion:
-  - 4.1.2 # Name, Role, Value (A)
-test_aspects:
+accessibility_requirements:
+  - wcag20: 4.1.2
```
annethyme: Add "# Name, Role, Value".
I don't think the "(A)" part needs to be re-added, since nothing in the template mentions this.

PR #498, opened 2019-05-03
Template update - ARIA state or property has valid value

annethyme:
Template update


_rules/aria-state-or-property-has-valid-value-6a7281.md
```diff
@@ -4,23 +4,26 @@ name: ARIA state or property has valid value
 rule_type: atomic
 description: |
   This rule checks that each ARIA state or property has a valid value
-success_criterion:
-  - 4.1.2 # Name, Role, Value
-test_aspects:
+accessibility_requirements:
+  - wcag20: 4.1.2
```
annethyme: Add "# Name, Role, Value"

PR #499, opened 2019-05-03
Template update - Audio-only as a media alternative for text

annethyme:
Template update

### DIFF
--- a/_rules/audio-as-media-alternative-afb423.md
+++ b/_rules/audio-as-media-alternative-afb423.md
@@ -4,7 +4,9 @@ name: Audio-only as a media alternative for text
 rule_type: atomic
 description: |
   This rule checks `audio` is a media alternative for text on the page.
-test_aspects:
+accessibility_requirements:
+  
+input_aspects:
   - DOM Tree
   - CSS Styling
   - Audio output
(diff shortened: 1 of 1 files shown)


_rules/audio-as-media-alternative-afb423.md
```diff
@@ -4,7 +4,9 @@ name: Audio-only as a media alternative for text
 rule_type: atomic
 description: |
   This rule checks `audio` is a media alternative for text on the page.
-test_aspects:
+accessibility_requirements:
+
```
annethyme: @ShadowBB, should this rule have a technique as it's accessibility requirement?

ShadowBB: > 
> 
> @ShadowBB, should this rule have a technique as it's accessibility requirement?

No. There is no relevant technique that would necessarily fail if this rule fails.

_rules/audio-as-media-alternative-afb423.md
```diff
@@ -3,8 +3,8 @@ id: afb423
 name: Audio-only as a media alternative for text
 rule_type: atomic
 description: |
-  This rule checks `audio` is a media alternative for text on the page.
-test_aspects:
+  This rule checks `audio` is a media alternative for text on the page. 
+input_aspects:
```
annethyme: We need to have the accessibility_requirements section here, even though it might will be empty :)

PR #500, opened 2019-05-03
Template update - audio elements have a transcript

annethyme:
Template update

### DIFF
--- a/_rules/audio-have-transcript-2eb176.md
+++ b/_rules/audio-have-transcript-2eb176.md
@@ -4,7 +4,9 @@ name: audio elements have a transcript
 rule_type: atomic
 description: |
   Non-streaming `audio` elements must have a text alternative for all included auditory information.
-test_aspects:
+accessibility_requirements:
+
+input_aspects:
   - DOM Tree
   - CSS Styling
   - Audio output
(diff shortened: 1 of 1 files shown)


ShadowBB:
Should we think of something to say in the "accessibility requirements" field other than just keeping it blank? Something like "No accessibility requirements fail when this rule fails." or maybe "No accessibility requirements fail when this rule fails, but this rule is used in a composite rule that is necessary for WCAG accessibility requirements"?

ShadowBB:
> 
> 
> @ShadowBB, should this rule have a technique as its accessibility requirement?

Yes. G158 should map onto it.

_rules/audio-have-transcript-2eb176.md
```diff
@@ -4,7 +4,9 @@ name: audio elements have a transcript
 rule_type: atomic
 description: |
   Non-streaming `audio` elements must have a text alternative for all included auditory information.
-test_aspects:
+accessibility_requirements:
+
```
annethyme: @ShadowBB, is the mapping something like this? 

> wcag21-technique: G158 # Providing an alternative for time-based media for audio-only content
>     forConformance: false
>     failed: not satisfied
>     passed: satisfied
>     inapplicable: further testing needed

(I am much in doubt of whether Passed maps to Satisfied or Further testing needed)

@WilcoFiers, what is the format we are using for referring to techniques? 
About the 2.1 versioning: The 2.0 techniques are no longer updated as far as I know and might be out of sync with the newest version.

annethyme: I have agreed with Wilco that adding this mapping will have to come in another round, since this requires much more thorough review than the pure template update project.

PR #514, opened 2019-05-07
Template update - Image button has accessible name

vemolista:
Template updates


vemolista:
Another example of a rule where there are two WCAG thingies and IMO the forConfromance etc. part should be indented or something.

PR #521, opened 2019-05-07
Template update- Links have an accessible name

vemolista:
Template update


_rules/link-has-name-c487ae.md
```diff
@@ -3,25 +3,36 @@ id: c487ae
 name: Links have an accessible name
 description: |
```
annethyme: Missing rule_type: atomic

PR #534, opened 2019-05-07
Template update - Video has audio alternative

vemolista:
Template update


_rules/video-has-audio-alternative-eac66b.md
```diff
@@ -1,26 +1,29 @@
 ---
 id: eac66b
 name: Video has audio alternative
-rule_type: atomic
+rule_type: composite
```
WilcoFiers: Good catch!

PR #543, opened 2019-05-08
Template update: No keyboard Trap

jkodu:
Template update to new format of ACT Rules.

@vemolista - looks like one of the rules was missed out.


_rules/no-keyboard-trap-80af7b.md
```diff
@@ -48,31 +52,31 @@ _There are no major accessibility support issues known for this rule._
```
WilcoFiers: Remove `## Aggregation Definition` and change the heading levels of Applicability / Expectation

PR #545, opened 2019-05-08
chore: fix frontmatter for atomic rules enumeration

jkodu:
Some keys were wrong as a part of the template updation.


jkodu:
`input_rules` is the correct convention. So closing this PR.

Issue #476, opened 2019-04-15
Updating the new website some more


WilcoFiers:
>**Note: (from Jey) I am ticking items I have done & merged**

## Navigation

- [x] Change the name from "ACT RULES CG" to "ACT-Rules Community"
  - [x] Remove its underline
  - [x] Make it look like it isn't just badly wrapped text
- [x] Move the "A community of...", the logos and the EU stuff and a link to the license into an actual page footer, not in the top of the navbar
- [ ] Reorganise the menu:
  - Rules
  - Implementations
  - How to Contribute
  - Documentation
  - About Us
  - Github Repository
  - Move glossary under Documentation
  - Collapse the subsections (see the nav on pattern-library.dequelabs.com/ for an example)
- [x] Don't uppercase menu items, it's hard to read
- [x] Don't show the "collapse menu" button on high resolution (>1280)
- [x] Rename: "DEFINITION OF "DONE" 1.0" to "Definition of Done"
- [ ] Don't scroll the nav after loading a page, it's jarring
- [x] Make the nav collaps button more subtle, have a look at what JIRA did in their design. It hides the button when the thing is open, only showing it on hover/focus, has the button sitting on a line with a color, indicating that there's a menu there to be opened, and changes icons once its open.
- [x] Don't put the logo in the scrollable area, it's small enough that we don't need it to

## Main body

- [x] Make the whitespace above and below the h1 relative to the screen height. That size is fine for a large screen, not for a small one. Same for the other headings, too much whitespace for a small screen.

## Contribute page

- [x] The "Learn more about how to join WCAG-ACT-RULES-CG" button doesn't work
- [x] Join the group links are wrong.

## Rules page

- [x] Don't include the rule count in the h1. It's not important
- [x] Remove the authors column, not important
- [x] Don't upppercase success criterion
- [x] Use the actual SC title, rather than "WCAG2:NAME-ROLE-VALUE", also not uppercase it
- [x] Don't do the outline on hover, it implies that you can interact with something
- [x] Put the atomic in a list under the description rather than in a column that doesn't even fit the ID
- [x] Use the atomic rule titles rather than their IDs
- [x] ~~Remove the numbering. If we use numbers, people might start refering to the numbers, we don't want that.~~ - @WilcoFiers - Keeping this, we want to know how many rules we have, otherwise we cannot keep track, given we are removing total count on the heading.
- [x] Show the number of implementations for each rule (e.g. Implementations: 3)

## Glossary page

[...]

PR #552, opened 2019-05-14
chore: Update site for accessibility and usability

jkodu:
- Several changes which addresses both accessibility and usability of the site.
- Have tackled issues listed in linked issue below. (Ones that are ticket are handled).

Closes issue: 
- https://github.com/act-rules/act-rules.github.io/issues/476


package.json
```diff
@@ -1,8 +1,8 @@
 {
-	"name": "ACT-RULES-CG",
+	"name": "ACT-Rules-Community",
```
WilcoFiers: I don't think NPM allows capitalisation in package names at all, does it? Maybe just pull from package.author, rather than using package.name.

src/pages/glossary.js
```diff
@@ -24,7 +24,7 @@ export default ({ data }) => {
 							<article key={node.id}>
 								<section>
 									<a id={key} href={`#${key}`}>
-										<h2>{frontmatter.title}</h2>
+										<h3>{frontmatter.title}</h3>
```
WilcoFiers: Don't skip heading levels. Do something else if you need the styles changes on these.

src/utils/render-fragments.js
```diff
@@ -129,7 +127,7 @@ export function getGlossaryUsageInRules(usages) {
 	}
 	return (
 		<aside>
-			<h3>Used In Rules:</h3>
+			<h4>Used In Rules:</h4>
```
WilcoFiers: Is this causing a heading-level skip?

PR #444, opened 2019-03-19
[WIP] Rule update: Heading is descriptive (b49b2e)

annethyme:
@rvantonisse wrote in https://github.com/auto-wcag/auto-wcag/issues/424: 

> During several audits with this rule, “Heading is Descriptive”, we encountered some situations in which we were forced by the rule to fail the SC.
> 
> The stituations were:
> 
> - Empty heading with a section of the content.
> - Heading without section of the content
> - Heading that describes only a part of the section of the content.
> 
> We believe these situations should not fail the WCAG SC 2.4.6 and therefore the rule needs some improvements.

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/424

Also updating the rule to:
- in the Applicability use the replacements terms for the "non-empty" definition introduced in this pull request: https://github.com/auto-wcag/auto-wcag/pull/430/files
- in the Expectation evaluate _both_ the accessible name and text nodes of the heading, if the two are different. This was not clear from the rule beforehand.

# How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

PR #555, opened 2019-05-15
New rule: Link is descriptive (5effbb)

carlosapaduarte:
<< Describe the changes >>

Closes issue: (ADD ISSUE NUMBER HERE)

**When creating PR:**

**After creating PR:**

# How to Review And Approve


descriptive-link
```diff
@@ -0,0 +1,72 @@
+---
+id: 
+name:
```
ramitgarg: Add description.

descriptive-link
```diff
@@ -0,0 +1,180 @@
+---
+id:
```
jkodu: I usually use https://www.browserling.com/tools/random-hex for generating the id

descriptive-link
```diff
@@ -0,0 +1,180 @@
...
+
+accessibility_requirements: 
+  wcag20:2.4.4: # Link Purpose (In Context) 
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: further testing needed
+
+input_aspects:
+- DOM Tree
```
jkodu: this is not right YAML. This needs to be indented.

descriptive-link
```diff
@@ -0,0 +1,180 @@
...
+
+authors:
+- Carlos Duarte
```
jkodu: this is not right YAML. These needs to be indented.

descriptive-link
```diff
@@ -0,0 +1,180 @@
...
+- Marie Trudelle 
+- Ramit Garg
+---
+
+## Applicability
+
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `link` that has a [non-empty](#non-empty) [accessible-name](#accessible-name).
```
jkodu: `non-empty` definition has been deprecated, please use an alternative

See - https://github.com/act-rules/act-rules.github.io/pull/430

descriptive-link
```diff
@@ -0,0 +1,180 @@
...
+
+## Expectation 2
+
+Accessible name together with its [context](#context) describes the purpose of the link
+
+## Assumptions
+
+- The rule assumes that all links are [user interface components](https://www.w3.org/TR/WCAG20/#user-interface-componentdef) as defined by WCAG 2.
+
+## Accessibility support
```
jkodu: Change to `Accessibility Support`.

descriptive-link
```diff
@@ -0,0 +1,180 @@
...
+- Ramit Garg
+---
+
+## Applicability
+
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `link` that has a [non-empty](#non-empty) [accessible-name](#accessible-name).
+
+## Expectation 1
+
+Link text together with its [context](#context), if visible, describes the purpose of the link
```
annethyme: Should this be "Accessible name ... described the purpose" instead of "Link text..."? At least it needs to match the applicability, that talks about "accessible name".
Maybe it would make sense to have something both for the accessible name and the visible text of the link? Similar to what we have been trying to for https://github.com/act-rules/act-rules.github.io/pull/444/files (that is currently in debate about whether to split up the rule or keep it as one).

annethyme:
Siteimprove is in general on board with the idea of this rule.

ShadowBB:
ACC sees no problems with this new rule.

descriptive-link
carlosapaduarte: @annethyme, I changed to a single expectation (and added a note) trying to cover both text content and accessible name.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+- Ramit Garg
+---
+
+## Applicability
+
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `link` that has an [accessible-name](#accessible-name) that is not the empty string ("").
+
+## Expectation
+
+Both the [text content](#text-content), if [visible](#visible), and the [accessible-name](#accessible-name) of the target element, together with its [visible](#visible) [context](#context) describe the purpose of the link.
```
ShadowBB: I think WCAG only asks us to have a good link text. Now I am uncertain what "link text" means (it has no definition) but we always interpreted it as the accessible-name of the link.

I would pass <a href="#' aria-label="read more about act rules">read more</a> while this rule would fail that scenario because 'read more' would not describe the purpose of the link on it's own.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+
+Note: Even though elements with the [semantic role](#semantic-role) of `link` [support name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent), their [text content](#text-content) and [accessible-name](#accessible-name) may be different if the content's author decides to. For this reason, this rules checks both [text content](#text-content) and [accessible-name](#accessible-name).
+
+## Assumptions
+
+- The rule assumes that all links are [user interface components](https://www.w3.org/TR/WCAG20/#user-interface-componentdef) as defined by WCAG 2.
```
ShadowBB: Why is this relevant? Isn't this only relevant for SC 4.1.2?

descriptive-link-5effbb.md
annethyme: I think this mirrors the discussion about descriptive headings in https://github.com/act-rules/act-rules.github.io/pull/444.
That rule is kind of stranded, because we can't really agree on what we should be testing (visible/accessible name) and what the context is (visible/included in the accessibility tree).

annethyme: I suggest not using a more general "context" definition, but creating a definition for specifically "programmatically determined link context" in line with the [WCAG definition for this](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determined-link-context): 

> In HTML, information that is programmatically determinable from a link in English includes text that is in the same paragraph, list, or table cell as the link or in a table header cell that is associated with the table cell that contains the link.

This definition is in my experience narrower than the one I think many accessibility experts are using, which often also includes headings etc.

In the ACT-R definition we will probably in addition to the WCAG part of the definiton have to define whether "programmatically determinable link context" is things that are visible or things that are included in the accessibility tree. "Programatically determinable" indicates the latter, in my opinion.

This definition will also be crucial for making a 2.4.4 (A) version of the 2.4.9 (AAA) rule [Links with identical accessible names serve equivalent purpose](https://act-rules.github.io/rules/b20e66)

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+  wcag20:2.4.4: # Link Purpose (In Context) 
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: further testing needed
+
+input_aspects:
+- DOM Tree
+- CSS Styling
+
```
annethyme: Should "Language" be an input aspect here?
Still up for discussion in https://github.com/act-rules/act-rules.github.io/issues/671

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+authors:
+- Carlos Duarte
+- Marie Trudelle 
+- Ramit Garg
+---
+
+## Applicability
+
+The rule applies to any HTML element with the [semantic role](#semantic-role) of `link` that has an [accessible-name](#accessible-name) that is not the empty string ("").
```
annethyme: The rule --> This rule

annethyme: Comparing with https://act-rules.github.io/rules/b20e66, should it be?

The rule applies to any HTML **_or SVG_** element with the [semantic role](#semantic-role) of `link`**_, or a role that inherits from the link role, are  [included in the accessibility tree](#included-in-the-accessibility-tree)_**, and that has an [accessible-name](#accessible-name) that **_do not only consist of [whitespace](#whitespace)_**.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [2.4.4 Link Purpose (In Context)](http://www.w3.org/TR/2008/REC-WCAG20-20081211/#navigation-mechanisms-refs)
+- [G91: Providing link text that describes the purpose of a link](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G91)
+- [G53: Identifying the purpose of a link using link text combined with the text of the enclosing sentence](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G53)
+
```
annethyme: Looking at the Understanding document, it seems that there might be a bunch of other techniques that could be listed as well. How do you choose which ones to include/exclude?

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+```html
+<a href="#main" aria-label="Go to the main content of the page"></a>
+```
+
+#### Passed Example 3
+
+Link text together with its context describes the purpose of the link
+
+```html
+<a href="#desc">Click</a> to see the description of this product
```
annethyme: Should this be in a `<p>` to really match the definition of "programatically determined link context"?

descriptive-link-5effbb.md
annethyme: If we decide to have both visible name and accessible name as part of the expectation, I would prefer to have it at least split out in two expectations, which will allow for talking about which expectation failed/passed if it's not the same for both conditions.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+Link text describes the purpose of the link
+
+```html
+<a href="#desc">See the description of this product</a>
```
annethyme: Wouldn't we have to actually create the content that is being linked to for anyone to be able to evaluate whether they would pass/fail these test cases? 
Similar to https://act-rules.github.io/rules/b20e66.

I know it is a bunch of extra work...

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+
+`<a>` with its role changed from link to another role
+
+```html
+<a href="http://www.w3.org/WAI" role="button">Web Accessibility Initiative (WAI)</a>
+```
+
+#### Inapplicable Example 2
+
+Link that is not visible and does not have an accessible name
```
annethyme: The applicability actually doesn't say anything about these two things.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+
+Link with empty accessible name
+
+```html
+<a href="http://www.w3.org/WAI"><img src="#" alt="" /></a>
+```
+
+#### Inapplicable Example 4
+
+Link with empty accessible name
```
annethyme: Isn't this kind of the same test case as Inapplicable Example 3?

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
...
+<a href="http://www.w3.org/WAI"><img src="#" title=""/></a>
+```
+
+#### Inapplicable Example 5
+
+Link with empty accessible name
+
+```html
+<a href="http://www.w3.org/WAI"><img src="#" aria-labelledby="id1" /></a>
+```
```
annethyme: Just a suggestion: 
Would it be worth adding a test case for an `<a>` element without a `href` attribute? Then it doesn't have the role of `link`.

descriptive-link-5effbb.md
annethyme: @carlosapaduarte, there is a shot at a "programmatically determined link context" definition in this new rule proposal: https://github.com/act-rules/act-rules.github.io/pull/693/files#diff-242a35ac05c129a11753d39e952c744c
(that was all along supposed to come bundled with its 2.4.9 sibling rule, https://act-rules.github.io/rules/b20e66)

If you continue work on this rule, feel free to build on/edit/change the definiton for this rule. I will align with you/this rule, when I pick up further work on the other rule.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,179 @@
+---
+id: 5effbb
+name: Link text is descriptive
```
annethyme: Would it make sense to call it "Link in context is descriptive" or something like that to make it clear that it is for 2.4.4 and not 2.4.9. 
Basically, the 2.4.9 rule will be REALLY easy to write whenever this rule is done. It's just removing the context part. So I think we should already now allow room for creating this other rule.

descriptive-link-5effbb.md
carlosapaduarte: As you said, when we reach a conclusion in #671 we'll probably have to go through all the rules and add Language to those that need it. Agree that this will be one of those.

descriptive-link-5effbb.md
carlosapaduarte: I've changed from 2 expectations to one expectation following your previous [comment](https://github.com/act-rules/act-rules.github.io/pull/555#discussion_r292893335) :-) 
But I'm happy to go back to the 2 expectations solution

carlosapaduarte: After looking at the proposed definition of "programmatically determined link context" I believe we can keep this with a single expectation, and a much simpler one

descriptive-link-5effbb.md
carlosapaduarte: True! I'm removing it.

descriptive-link-5effbb.md
carlosapaduarte: You're right. I've added the sufficient techniques that address the input aspects of the rule

descriptive-link-5effbb.md
carlosapaduarte: I had a go at it

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,205 @@
...
+- Ramit Garg
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [link](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the link role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible-name](#accessible-name) that does not only consist of [whitespace](#whitespace).
+
+## Expectation
+
+The [visible](#visible) [text content](#text-content) of the target element, together with its [programmatically determined link context](#programmatically-determined-link-context) describe the purpose of the link.
```
Jym77: What about the exception for links that are "ambiguous to users in general"?

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,205 @@
...
+- [ARIA7: Using aria-labelledby for link purpose](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA7)
+- [ARIA8: Using aria-label for link purpose](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8)
+- [H77: Identifying the purpose of a link using link text combined with its enclosing list item](https://www.w3.org/WAI/WCAG21/Techniques/html/H77)
+- [H78: Identifying the purpose of a link using link text combined with its enclosing paragraph](https://www.w3.org/WAI/WCAG21/Techniques/html/H78)
+- [H79: Identifying the purpose of a link in a data table using the link text combined with its enclosing table cell and associated table header cells](https://www.w3.org/WAI/WCAG21/Techniques/html/H79)
+- [H81: Identifying the purpose of a link in a nested list using link text combined with the parent list item under which the list is nested](https://www.w3.org/WAI/WCAG21/Techniques/html/H81)
+
+## Test Cases
+
+### Passed
```
Jym77: https://www.w3.org/TR/WCAG21/#dfn-programmatically-determined-link-context also refers to list and table (including header). Should we include examples of these?
Understanding 2.4.4 has this nice (I think) example with several version of the same book (HTML, PDF, MP3) where the title is only said once (start of the list, or table header) and the actual links just say "HTML", "PDF", "MP3", the context making it clear that it is the book in that format. We seem to be missing example where several links share a given context that makes everything clear.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,205 @@
...
+ <p id="main">This is the main content</p>
+</main>
+```
+
+#### Passed Example 3
+
+Link text together with its context describes the purpose of the link
+
+```html
+<p><a href="#desc">Click</a> to see the description of this product</p>
```
Jym77: Given that Understanding 2.4.4 explicitly recommends to put the link at the end of the sentence (after context) rather than at the start, I'd rather have the example showing the best practice.

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,205 @@
...
+
+authors:
+- Carlos Duarte
+- Marie Trudelle 
+- Ramit Garg
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [link](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the link role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible-name](#accessible-name) that does not only consist of [whitespace](#whitespace).
```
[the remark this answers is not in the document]
carlosapaduarte: I don't think so, but I was keeping coherent with #693 
My reasoning (I don't know if it was also @annethyme's reasoning) is that in the future there might be roles that inherit from link

descriptive-link-5effbb.md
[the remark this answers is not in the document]
carlosapaduarte: To the definition that is being worked on #693

descriptive-link-5effbb.md
carlosapaduarte: I've included two further examples: one with a list, another with a table

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,237 @@
...
+
+<p id="desc">This product consists of several web pages</p>
+```
+
+#### Passed Example 4
+
+Both link text together with its context and accessible name describe the purpose of the link
+
+```html
+<p><a href="#" aria-label="Order item">Place item</a><span aria-hidden="true"> in shopping cart</span></p>
```
ShadowBB: This would fail "label-in-name" and we should make sure all our examples pass WCAG 2.1 right?

carlosapaduarte:
All comments addressed. Needs reviews again @jkodu @Jym77 @ShadowBB

descriptive-link-5effbb.md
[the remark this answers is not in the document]
Jym77: As part of #447 (and per @WilcoFiers suggestion), I have been (re)working the definitions of [context](https://github.com/act-rules/act-rules.github.io/blob/update-SC2-4-6-descriptive-label/pages/glossary/context.md) using [visual context](https://github.com/act-rules/act-rules.github.io/blob/update-SC2-4-6-descriptive-label/pages/glossary/visual-context.md) and [programatically determined context](https://github.com/act-rules/act-rules.github.io/blob/update-SC2-4-6-descriptive-label/pages/glossary/programatically-determined-context.md) (not restricted to programatically determined **link** context).

Given that this seems to be also needed for this rule, and for #693 (which I've also inherited from @annethyme), I strongly suggest that the two of us get on a call soonish and come up with defs that we both agree on.
(maybe as separate PR since that would not require a Final Call if it's Glossary only, and we don't want 3 big PR do have mutual dependencies preventing us from merging…)

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [`link`](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the `link` role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible-name](#accessible-name) that does not only consist of [whitespace](#whitespace).
+
+## Expectation
+
+The [visible](#visible) [text content](#text-content) of the target element, together with its [programmatically determined link context](#programmatically-determined-link-context) describe the purpose of the link.
+
+**Note:** There may be situations where the purpose of the link is supposed to be unknown or obscured. In those situations, the success criteria will still be met, even if the link's purpose is not described by it.
```
Jym77: I'd rather see it as an Assumption.
Which would fit a nice pattern of "this rule assumes that we are not considering one of the exceptions allowed by WCAG".

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,237 @@
...
+- [G53: Identifying the purpose of a link using link text combined with the text of the enclosing sentence](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G53)
+- [ARIA7: Using aria-labelledby for link purpose](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA7)
+- [ARIA8: Using aria-label for link purpose](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8)
+- [H77: Identifying the purpose of a link using link text combined with its enclosing list item](https://www.w3.org/WAI/WCAG21/Techniques/html/H77)
+- [H78: Identifying the purpose of a link using link text combined with its enclosing paragraph](https://www.w3.org/WAI/WCAG21/Techniques/html/H78)
+- [H79: Identifying the purpose of a link in a data table using the link text combined with its enclosing table cell and associated table header cells](https://www.w3.org/WAI/WCAG21/Techniques/html/H79)
+- [H81: Identifying the purpose of a link in a nested list using link text combined with the parent list item under which the list is nested](https://www.w3.org/WAI/WCAG21/Techniques/html/H81)
+
+## Test Cases
```
Jym77: Could you make the descriptions of the test cases into actual English sentences (mostly adding a final period and using "the" whenever you reference something precise).

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,237 @@
...
+
+```html
+<span role="link" onclick="document.location+='#desc';return false;">See description of the product</span>
+
+<p id="desc">This product consists of several web pages</p>
+```
+
+#### Passed Example 7
+
+The context afforded by the list and the link text describe the purpose of the links.
```
Jym77: "afforded" => "provided"?

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,237 @@
...
+<ul>
+  <li><a href="https://www.gutenberg.org/files/4300/4300-h/4300-h.htm">Ulysses</a></li>
+  <li><a href="https://www.gutenberg.org/ebooks/4300.epub.images?session_id=04cd710372888de8d8d322215cdfe8ce5b0f8d73">EPUB format</a></li>
+  <li><a href="https://www.gutenberg.org/files/4300/4300-0.txt">Plain text</a></li>
+</ul>
+```
+
+#### Passed Example 8
+
+The context afforded by the table header and the link text describe the purpose of the links.
```
Jym77: "afforded" => "provided"?

descriptive-link-5effbb.md
```diff
@@ -0,0 +1,237 @@
...
+ <p id="main">This is the main content</p>
+</main>
+```
+
+
+### Inapplicable 
+
+#### Inapplicable Example 1
+
+`<a>` with its role changed from `link` to another role
```
Jym77: I think we usually use `a` rather than `<a>`.

Also, for the sake of making nice description, I'd prefer "An `a` element with …"

carlosapaduarte:
Updated rule according to requests
Included the definition of Programmatically determined context as it required by this rule

Please review

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,274 @@
...
+
+<p id="desc">This product consists of several web pages</p>
+```
+
+#### Passed Example 2
+
+The accessible name describes the purpose of the link.
+
+```html
+<a href="#main" aria-label="Go to the main content of the page"></a>
```
Jym77: I'm also a bit worried about these examples (there's a few of them) where the link is essentially empty (nothing inside the `a` element).

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+#### Passed Example 7
+
+The context provided by the list and the link text describe the purpose of the links.
+
+```html
+<ul>
+  <li><a href="https://www.gutenberg.org/files/4300/4300-h/4300-h.htm">Ulysses</a></li>
+  <li><a href="https://www.gutenberg.org/ebooks/4300.epub.images?session_id=04cd710372888de8d8d322215cdfe8ce5b0f8d73">EPUB format</a></li>
```
Jym77: "EPUB format" => "EPUB" 
should be enough (and keep in sync with next example).

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,274 @@
...
+<main>
+  <p id="main">This is the main content</p>
+</main>
+```
+
+### Failed
+
+#### Failed Example 1
+
+The link text, together with the absence of programatically determined link context, does not describe the purpose of the link.
```
Jym77: "programatically" => "progra**mm**atically" (2 'm') (several times).

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,274 @@
...
+```html
+<p><a href="#" aria-label="Place item in cart">Place item</a><span aria-hidden="true"> in shopping cart</span></p>
+```
+
+#### Passed Example 5
+
+The accessible name describes the purpose of the link.
+
+```html
+<div role="link" aria-label="Skip to main content" onclick="document.location+='#main';return false;"></div>
```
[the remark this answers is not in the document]
ShadowBB: Also: Sorry I did not catch this the first time around. ><

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Programmatically determined context
+key: programmatically-determined-context
+---
+
+The programmatically determined context of a piece of information (_e.g._ an element) is all the additional information that has a [programmatically determinable](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) [relationship](https://www.w3.org/TR/WCAG21/#dfn-relationships) to it.
+
+It includes, but is not restricted to, `aria-label` attribute, element that is referred by `aria-labelledby` or `aria-describedby` attributes, the `title` attribute, ancestor `label` elements, `label` elements referring the current element with a `for` attribute, ancestor elements, table header associated with the cell containing the information, headings, text inside the same paragraph (`p` element), sentence, table cell, list item, …
+
+**Note:** Since screen readers interpret punctuation, they can also provide the context from the current sentence, when the focus is on a link in that sentence.
```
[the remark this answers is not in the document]
carlosapaduarte: I took the note from the understanding document (https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html#dfn-programmatically-determined-link-context). I'm not really sure if we can generalize from links to other pieces of information.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,270 @@
...
+authors:
+  - Carlos Duarte
+  - Marie Trudelle 
+  - Ramit Garg
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [`link`](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the `link` role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible-name](#accessible-name) that does not only consist of [whitespace](#whitespace).
+
```
geirsf: Why are you not looking at whitespaces?  Are you planing to do a different rule for this?

carlosapaduarte: Links should have accessible names. If they don't they will fail https://act-rules.github.io/rules/c487ae

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,270 @@
...
+  - Language
+authors:
+  - Carlos Duarte
+  - Marie Trudelle 
+  - Ramit Garg
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [`link`](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the `link` role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible-name](#accessible-name) that does not only consist of [whitespace](#whitespace).
```
Jym77: "[accessible-name]" => "[accessible name]"

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Expectation
```
DagfinnRomen: Consider adding a note to say something about what is sufficient description of the purpose of the link. Such as: "Examples of sufficient description would be the name of the linked document or web application. However, this is not an absolute requirement, as other things may also describe the purpose of the link."

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,270 @@
...
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Failed Example 2
+
+The accessible name without any further programmatically determined link context, does not describe the purpose of the link.
+
+```html
+<a href="#main" aria-label="Go"></a>
```
Jym77: Add some text inside the link (`<a>Go</a>`).

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,270 @@
...
+ <p id="main">This is the main content.</p>
+</main>
+```
+
+#### Failed Example 4
+
+The accessible name without any further programmatically determined link context, does not describe the purpose of the link.
+
+```html
+<div role="link" tabindex="0" aria-labelledby="id1" onclick="document.location+='#main';return false;"></div>
```
Jym77: Add text in the link (`<div>Go</div>`).

carlosapaduarte:
@jkodu The CI tests are failing because of the lack of an alt attribute on an img. However, adding the alt would change the (inapplicable) example. Is there a way to add an exception for this test for this example (or rule)?

carlosapaduarte:
Thanks for the fix @Jym77

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Failed Example 2
+
+The accessible name without any further programmatically determined link context, does not describe the purpose of the link.
+
+```html
+<a href="#main" aria-label="Go">Go</a>
```
WilcoFiers: Aria-label is unnecessary here.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+
+```html
+<div role="link" tabindex="0" onclick="document.location+='#main';return false;">More</div>
+
+<main>
+	<p id="main">This is the main content.</p>
+</main>
+```
+
+#### Failed Example 4
```
WilcoFiers: This test is unnecessary. We don't have to test that someone knows how to compute the accessible name. That's not what this rule is about.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+</main>
+```
+
+#### Failed Example 5
+
+The accessible name without any further programmatically determined link context, does not describe the purpose of the link.
+
+```html
+<svg xmlns="http://www.w3.org/2000/svg" x="0" y="0">
+	<a href="#main" aria-label="Go">
```
WilcoFiers: As above, aria-label seems unnecessary here.

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Programmatically determined context
+key: programmatically-determined-context
+---
+
+The programmatically determined context of a piece of information (_e.g._ an element) is all the additional information that has a [programmatically determinable](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) [relationship](https://www.w3.org/TR/WCAG21/#dfn-relationships) to it.
+
+It includes, but is not restricted to, `aria-label` attribute, element that is referred by `aria-labelledby` or `aria-describedby` attributes, the `title` attribute, ancestor `label` elements, `label` elements referring the current element with a `for` attribute, ancestor elements, table header associated with the cell containing the information, headings, text inside the same paragraph (`p` element), sentence, table cell, list item, …
```
WilcoFiers: I don't think this should be an open-ended definition (e.g. "includes but not restricted to"). There isn't an endless list of programatically determinable content. We should list all the programatically determined context that an element has, and be objective about it. So when you say "text inside the same paragraph", does that mean a child of, a descendant of the DOM tree, of the flat tree? What **exactly** does that mean?

We can be precise here, so I think we should be.

WilcoFiers: I don't think title and aria-label are part of an element's context. I can see an argument for aria-labelledby and for labels, but I disagree with calling attributes part of the context.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Passed Example 7
+
+The context provided by the list and the link text describe the purpose of the links.
+
+```html
+<ul>
+	<li><a href="https://www.gutenberg.org/files/4300/4300-h/4300-h.htm">Ulysses</a></li>
```
WilcoFiers: I don't think one list item gives context to another list item. According to [H81](https://www.w3.org/WAI/WCAG21/Techniques/html/H81.html) you can do it with nested lists, but it seems to me that the reason it is nested is important here. Makes sense too, because with AT you can exit into and out of lists to get to the higher level list item for context.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+- [ARIA7: Using aria-labelledby for link purpose](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA7)
+- [ARIA8: Using aria-label for link purpose](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8)
+- [H77: Identifying the purpose of a link using link text combined with its enclosing list item](https://www.w3.org/WAI/WCAG21/Techniques/html/H77)
+- [H78: Identifying the purpose of a link using link text combined with its enclosing paragraph](https://www.w3.org/WAI/WCAG21/Techniques/html/H78)
+- [H79: Identifying the purpose of a link in a data table using the link text combined with its enclosing table cell and associated table header cells](https://www.w3.org/WAI/WCAG21/Techniques/html/H79)
+- [H81: Identifying the purpose of a link in a nested list using link text combined with the parent list item under which the list is nested](https://www.w3.org/WAI/WCAG21/Techniques/html/H81)
+
+## Test Cases
+
+### Passed
```
WilcoFiers: I would like to see more example of programatic context that passes, as well as examples of things that **aren't** programatic context and so will fail. That's really where the rubber meets the road, defining what is and isn't programatic context.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+  - Ramit Garg
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'alt-require'
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [`link`](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the `link` role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible name](#accessible-name) that does not only consist of [whitespace](#whitespace).
```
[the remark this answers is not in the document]
Jym77: Actually, with three clauses, two of them including a nested subordinate clause, it might be time to move to a bullet list:
> This rule applies to any element which
> - has a role of link or that blabla; and
> - is in the AT; and
> - has a name that blabla.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,270 @@
...
+  - Ramit Garg
+---
+
+## Applicability
+
+This rule applies to any HTML or SVG element with the [semantic role](#semantic-role) of [`link`](https://www.w3.org/TR/wai-aria/#link) or a role that inherits from the `link` role, that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has an [accessible-name](#accessible-name) that does not only consist of [whitespace](#whitespace).
+
+## Expectation
+
+The [visible](#visible) [text content](#text-content) of the target element, together with its [programmatically determined link context](#programmatically-determined-context) describe the purpose of the link.
```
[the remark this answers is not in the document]
Jym77: https://act-rules.github.io/glossary/#visible-text-content ?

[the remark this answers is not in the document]
carlosapaduarte: I'm looking at visible link text plus accessible name (programmatically determined link), per the SC. I think Jean-Yves' suggestion solves the issue.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Passed Example 2
+
+The accessible name describes the purpose of the link.
+
+```html
+<a href="#main" aria-label="Go to the main content of the page">Go</a>
```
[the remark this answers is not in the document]
carlosapaduarte: I would argue it passes SC 2.4.4 because of the programmatically determined link context, which in this example comes from the accessible name. 
There is no assumption of equivalence between link text and accessible name.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,285 @@
...
+
+The context provided by the table header and the link text describe the purpose of the links.
+
+```html
+<table>
+	<tr>
+		<th colspan="3">Ulysses</th>
+	</tr>
+	<tr>
+		<td><a href="https://www.gutenberg.org/files/4300/4300-h/4300-h.htm">HTML</a></td>
```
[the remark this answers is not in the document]
carlosapaduarte: Project Gutenberg has been founded in 1971 by the inventor of ebooks. It would be really sad if they went out of business. I believe the urls for the resources are stable. But let me know if you prefer, I can copy the books as test assets since they are free.

_rules/descriptive-link-5effbb.md
carlosapaduarte: You are correct. And that makes this example equal to the previous one, so I'm removing it.

_rules/descriptive-link-5effbb.md
carlosapaduarte: Correct. But I'll leave this one given that it is a link inside an svg element

_rules/descriptive-link-5effbb.md
carlosapaduarte: Added examples where the non-programmatic context is on a different paragraph, outside the list or in different table cells.

pages/glossary/programmatically-determined-context.md
carlosapaduarte: I don't agree. I think the accessible name should be part of the programmatically determined context of an element.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,323 @@
...
+
+<p><a href="https://www.w3.org/2005/04/FSWS/workshop-report.html">Workshop</a></p>
+```
+
+#### Failed Example 5
+
+The link text does not describe the purpose of the link. The other information available on the page is not [programmatically determined link context][] because it is outside the list where the links are.
+
+```html
+<h3>Ulysses</h3>
```
Jym77: I'd argue that headings are parts of the programmatically determined context.
Keeping the text alone (maybe with some styling to make it look important) would be enough to make me happy 😄

carlosapaduarte: Agreed and updated the example.

pages/glossary/programmatically-determined-context.md
```diff
@@ -8,11 +8,12 @@ The programmatically determined context of a piece of information (_e.g._ an ele
 It includes:
 
 - the element's [accessible name][];
-- elements that are referred by the `aria-describedby` attribute; 
+- elements that are referred by the `aria-describedby` attribute;
 - the `title` attribute;
 - ancestor `label` elements;
 - `label` elements referring the current element with a `for` attribute;
 - table headers associated with the cell containing the information;
+- the header that precedes the element;
```
Jym77: It's probably more than just the one preceding the element, more like the full hierarchy of headers going to it.
```html
<h1>Headers</h1>
<h2>Levels</h2>
<h3>Possible levels</h3>
<h3>Hierachy</h3>
<span id="target">Do not skip levels!</span>
```
The header context of the `span` is the `h1`, the `h2` and the second `h3`.

Jym77: @carlosapaduarte do you think this should pass or fail?
```html
<h1>Hamlet</h1>
  <h2>Act 1</h2>
    <h3>Scene 1</h3>
      <a href="…">HTML</a> <a href="…">PDF</a> <a href="…">DOCX</a>
    <h3>Scene 2</h3>
      <a href="…">HTML</a> <a href="…">PDF</a> <a href="…">DOCX</a>
  <h2>Act 2</h2>
    <h3>Scene 1</h3>
      <a href="…">HTML</a> <a href="…">PDF</a> <a href="…">DOCX</a>
    <h3>Scene 2</h3>
      <a href="…">HTML</a> <a href="…">PDF</a> <a href="…">DOCX</a>
```

I think this passes because the full headers hierarchy is providing context to understand what the links mean.

carlosapaduarte: @Jym77 yes, I agree that passes because of the full hierarchy. I'll update accordingly

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,25 @@
+---
```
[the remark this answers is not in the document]
carlosapaduarte: I've addressed your feedback in earlier commits. I've marked most comments as resolved. Those that could still be open for discussion I've left them open and I'm seeing them in the PR.

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,25 @@
...
+title: Programmatically determined context
+key: programmatically-determined-context
+---
+
+The programmatically determined context of a piece of information (_e.g._ an element) is all the additional information that has a [programmatically determinable](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) [relationship](https://www.w3.org/TR/WCAG21/#dfn-relationships) to it.
+
+It includes:
+
+- the element's [accessible name][];
```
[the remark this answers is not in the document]
carlosapaduarte: For a generic definition of programmatically determined context, I would argue that the accessible name is content (that can be determined programmatically). For this rule, after re-reading the SC, I have to agree that it should not be in the definition of programmatically determined **link** context. Consequently:

- I will move and update the current definition programmatically-determined-link-context in #693 to this PR so that we can move on (since #693 is on hold)
- I will move this definition of programmatically-determined-context to #905 which was waiting for this PR for such a definition

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,25 @@
...
+- elements that are referred by the `aria-describedby` attribute;
+- the `title` attribute;
+- ancestor `label` elements;
+- `label` elements referring the current element with a `for` attribute;
```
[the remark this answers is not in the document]
carlosapaduarte: Agreed. I'll remove it before moving the definition to #905

pages/glossary/programmatically-determined-context.md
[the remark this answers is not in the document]
carlosapaduarte: #905 is closed so I will keep both definitions here in the meantime

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,20 @@
+--
+title: Programmatically Determined Link Context
+key: programmatically-determined-link-context
+--
+
+In HTML, information that is programmatically determinable from a link includes [text nodes][] that are [included in the accessibility tree][] and are
+
+- sibling [phrasing content][] nodes in the same [paragraph][] as the link;
```
WilcoFiers: Phrasing content doesn't take text alternatives into consideration. Nor does it consider pseudo element content. I'm not sure what the best way to go here is. Perhaps you can look at the accessible name of these elements, or if they don't have one the name from contents value ([Step 2F.i in accName 1.1](https://www.w3.org/TR/accname-1.1/#step2F.i)).

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,20 @@
...
+- sibling [phrasing content][] nodes in the same list (HTML element with the semantic role of `listitem`) as the link;
+- sibling [phrasing content][] nodes in the same table cell (HTML element with the semantic role of `cell` or `gridcell`) as the link;
```
WilcoFiers: What if a link is contained inside a list inside a paragraph inside a table cell? Do you get to use all of them as context? I would argue it only counts towards the closest one.

pages/glossary/programatically-determined-link-context.md
WilcoFiers: I don't think paragraph works here. I wager all screen readers look at `p` elements for paragraphs, not this HTML spec definition of it.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,323 @@
...
+
+This rule applies to any HTML or SVG element which
+
+- has the [semantic role][] of ['link'][link] or a [semantic role][] that inherits from the ['link'][link] role; and
+- is [included in the accessibility tree][]; and
+- has an [accessible name][] that does not only consist of [whitespace][].
+
+## Expectation
+
+The [visible text content][] of the target element, together with its [programmatically determined link context][] describe the purpose of the link.
```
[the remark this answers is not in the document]
carlosapaduarte: Thanks for catching this. The accessible name was previously considered in the link context.
Added a new expectation that looks at the accessible name.

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: The definition requires that nodes contributing to the programmatically determined context are text nodes included in the accessibility tree. I'll add to those requirements that they have a non-empty accessible name. With this, it is safe to remove the phrasing content requirement and leave it as "sibling nodes in the same paragraph as the link"

pages/glossary/programmatically-determined-context.md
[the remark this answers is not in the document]
carlosapaduarte: I'll remove it from this PR and keep it for future needs.

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: I don't agree. I think all should be used as context. In the example below don't you think that the table header and the text in the same cell are providing context even if the element is inside a list in the cell?

```html
<table>
  <tr>
    <th colspan="3">Ulysses</th>
  </tr>
  <tr>
    <td>
      Download the book in your preferred format:
      <ul>
        <li>
          <a href="https://www.gutenberg.org/files/4300/4300-h/4300-h.htm">
            HTML
          </a>
        </li>
        <li>
          <a
            href="https://www.gutenberg.org/ebooks/4300.epub.images?session_id=04cd710372888de8d8d322215cdfe8ce5b0f8d73"
          >
            EPUB
          </a>
        </li>
        <li>
          <a href="https://www.gutenberg.org/files/4300/4300-0.txt">
            Plain text
          </a>
        </li>
      </ul>
    </td>
  </tr>
</table>
```

WilcoFiers:
@carlosapaduarte You've added an entire new expectation. Please remember to dismiss reviewers after making any non-editorial changes.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,314 @@
...
+
+This rule applies to any HTML or SVG element which
+
+- has the [semantic role][] of ['link'][link] or a [semantic role][] that inherits from the ['link'][link] role; and
+- is [included in the accessibility tree][]; and
+- has a non-empty ("") [accessible name][].
+
+## Expectation 1
+
+The [visible text content][] of each target element together with its [programmatically determined link context][] describe the purpose of the link.
```
WilcoFiers: This rule is making an assumption that there are two types of link texts, and that both of them have to pass the SC. There is nothing about this in the success criterion or the understanding document, so I'm not all too sure this is accurate. You also haven't addressed my previous comment about visible text content not including images.

My suggestion, drop expectation 1. I don't think there's a way to have it make sense. I get that interpreting "link text" as "the accessible name of a link" is somewhat of a stretch, but I don't think there's anything else you can do that makes more sense than that.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,314 @@
...
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Passed Example 2
+
+The [accessible name][] describes the purpose of the link.
+
+```html
+<a href="#main"><img src="/test-assets/5effbb/main.png" alt="Go to the main content" /></a>
```
WilcoFiers: As mentioned above, this link has no visible text so it doesn't meet expectation one. Expectation 1 seems incorrect to me.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,314 @@
...
+
+<p id="answer">Glasgow was the European Capital of Culture in 1990.</p>
+```
+
+#### Passed Example 5
+
+The link text describes the purpose of the link.
+
+```html
+<span role="link" tabindex="0" onclick="document.location+='#desc';return false;">See description of the product.</span>
```
WilcoFiers: You don't need the `return false` bit.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,314 @@
...
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Failed Example 2
+
+The link text, together with the absence of [programmatically determined link context][], does not describe the purpose of the link.
+
+```html
+<div role="link" tabindex="0" onclick="document.location+='#main';return false;">More</div>
```
WilcoFiers: As above, don't need the `return false;`

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
+--
+title: Programmatically Determined Link Context
+key: programmatically-determined-link-context
+--
+
+In HTML, information that is programmatically determinable from a link includes [text nodes][] that are [included in the accessibility tree][] with a non-empty ("") [accessible name][] and are
+
+- sibling nodes in the same paragraph (HTML `p` element) as the link;
+- sibling nodes in the same list (HTML element with the semantic role of `listitem`) as the link;
+- sibling nodes in the same table cell (HTML element with the semantic role of `cell` or `gridcell`) as the link;
```
WilcoFiers: I don't think sibling nodes are sufficient here. For example:

```html
<p>Go to the <span class="coolStypes">Wikipedia</span> <a href="http://wikipedia.org/">homepage</a></p>
```

In this case the text node "Wikipedia" is not a sibling node of the link. Or alternatively you could have something like this:

```html
<p>Go to the Wikipedia <u><a href="http://wikipedia.org/">homepage</u></a></p>
```

Since the link is wrapped in `u`, the link has no sibling nodes.

_rules/descriptive-link-5effbb.md
carlosapaduarte: I'm okay with that. With links having their accessible name calculated from content this should be a good enough approximation.

Jym77: I'm having mixed feeling about that based on techniques [ARIA7](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA7) and [ARIA8](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8) (both used to pass 2.4.4).

---
ARIA8 has an example with description saying (emphasise mine):
> In some situations, designers may choose to lessen the visual appearance of links on a page by using shorter, repeated _link text_ such as "read more".

and passing example 
```html
<a href="taxhike.html" aria-label="Read more about Seminole tax hike">[Read more...]</a>
```

I do understand that as "link text" being the "name from content" (in that case, "[Read more…]") and the `aria-label` being part of the context.

On the other hand, this is also a sufficient technique for 2.4.9 where context is not taken into account which lead to thing that the "link text" **is** the accessible name (as suggested by @WilcoFiers). This is further supported by the fact that the WCAG def of programatically determined link content does not refer to accessible name…

😕 

---
Same phenomenon goes for ARIA7 where the example description says:
> This example will mean that the link text as shown on screen is then used as the start of the accessible name for the link.

for
```html
<h2 id="headline">Storms hit east coast</h2>

<p>Torrential rain and gale force winds have struck the east coast, causing flooding in many coastal towns.
<a id="p123" href="news.html" aria-labelledby="p123 headline">Read more...</a></p>
```

The description quite clearly separate the "link text" from the accessible name and uses the name from content as "link text"

ARIA7 further says:
> The specified behavior of aria-labelledby is that the associated label text is announced instead of the link text (not in addition to the link text). When the link text itself should be included in the label text, the ID of the link should be referenced as well in the string of IDs forming the value of the aria-labelledby attribute.

which quite strongly pushes for "link text" **not** being the accessible name.

But at the same time, the second example of ARIA7 is called "Concatenating link text from multiple sources" which hints that the "link text" is the result of the `aria-labelledby` computation.

😕 

---
I do not know how to concile the fact that WCAG both seem to call "link text" the name from content (in the sentence `shorter, repeated link text such as "read more"`) and the accessible name (in the example which passes 2.4.9 because the `aria-label` attribute is correct)…

---
Additionally, ARIA8 is sufficient for 2.4.9, meaning that `aria-label` is **not** part of the (programatically determined link) context but part of the link itself; while ARIA7 is **not** sufficient for 2.4.9, meaning that `aria-labelledby` is part of context only.

😕

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
...
+key: programmatically-determined-link-context
+--
+
+In HTML, information that is programmatically determinable from a link includes [text nodes][] that are [included in the accessibility tree][] with a non-empty ("") [accessible name][] and are
+
+- in the same paragraph (HTML `p` element) as the link;
+- in the same list (HTML element with the semantic role of `listitem`) as the link;
+- in the same table cell (HTML element with the semantic role of `cell` or `gridcell`) as the link;
+- in a table header cell (HTML element with the semantic role of `columnheader` or `rowheader`) that is [associated][] with the table cell (HTML element with the semantic role of `cell` or `gridcell`) of which the link is a descendent;
+- nodes of elements referenced by an `aria-describedby` attribute of the link element.
```
Jym77: I know it's how the WCAG def is written, but I'm not super happy about dropping headings from the context.

Reading technique ARIA7, I kinda guess that the idea is to explicitly include them through a `aria-describedby` or `aria-labelled-by` attribute when needed.

Maybe it's worth illustrating this by adding examples to the definition (like in #994).

_rules/descriptive-link-5effbb.md
Jym77: Some more thoughts on that…
I'm still not sure whether "link text" means "text content" or "accessible name" in WCAG. But I do agree that both are an accessibility issue, hence checking whether accessible name is descriptive is a valid and useful check.

I suggest to update the rule's **name** (not just description) accordingly to make it explicit that the rule only check the accessible name.

The question of whether we need or want a rule for checking whether the text content is descriptive or not is completely independant from this PR.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
+--
+title: Programmatically Determined Link Context
+key: programmatically-determined-link-context
+--
+
+In HTML, information that is programmatically determinable from a link includes [text nodes][] that are [included in the accessibility tree][] with a non-empty ("") [accessible name][] and are
```
WilcoFiers: I don't think this should be limited to text nodes. That way you're ignoring images and other things with accessible names through attributes.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
...
+
+- in the same paragraph (HTML `p` element) as the link;
+- in the same list (HTML element with the semantic role of `listitem`) as the link;
```
WilcoFiers: This is confusing. Why not say in the same listitem, if that's the role you're after.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
+--
+title: Programmatically Determined Link Context
+key: programmatically-determined-link-context
+--
+
+In HTML, information that is programmatically determinable from a link includes [text nodes][] that are [included in the accessibility tree][] with a non-empty ("") [accessible name][] and are
+
+- in the same paragraph (HTML `p` element) as the link;
```
WilcoFiers: Need to clarify what "in the same" means. I suspect what you mean is that it has the same `p` element as an ancestor in the flattened tree, correct?

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
...
+- in the same list (HTML element with the semantic role of `listitem`) as the link;
+- in the same table cell (HTML element with the semantic role of `cell` or `gridcell`) as the link;
+- in a table header cell (HTML element with the semantic role of `columnheader` or `rowheader`) that is [associated][] with the table cell (HTML element with the semantic role of `cell` or `gridcell`) of which the link is a descendent;
```
WilcoFiers: Descendent in the flat tree I'm guessing. You may want to create a definition of "descendent" that specifies it's a descendent in the flat tree.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,19 @@
+--
+title: Programmatically Determined Link Context
```
WilcoFiers: I think you're making an assumption here that HTML is valid. Let's take something like this:

```html
<p>
  <h1>ACT Rules Community Group</h1>
  <p><a href="...">learn more</a></p>
</p>
```

I would argue that this should fail the SC, but according to this definition, the heading text is part of the context, and so it passes. I don't know if that's a bad thing necessarily. This doesn't cause false positives, but it doesn't tell the whole story either. I'm OK ignoring this edge case, but if we do, I think there needs to be an assumption.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+id: 5effbb
+name: Link in context is descriptive
+rule_type: atomic
+description: |
+  This rule checks that the accessible name of a link together with its context describe its purpose.
+accessibility_requirements:
+  wcag20:2.4.4: # Link Purpose (In Context)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
```
WilcoFiers: If you add an assumption that all links have the semantic role of link, you could say that passing this rule satisfies the SC. Not sure if that's what you want, but it wouldn't be unreasonable I think.

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: Updated, but I don't think the definition is needed

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: I agree that this should fail the SC. I've updated the definition to look only at the nearest ancestor of the link element, so the heading is no longer part of the context.

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: Headings should still be included in the programmatically determined context definition. It's just that the programmatically determined link context definition is stricter.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+id: 5effbb
+name: Link in context is descriptive
+rule_type: atomic
+description: |
+  This rule checks that the accessible name of a link together with its context describe its purpose.
+accessibility_requirements:
+  wcag20:2.4.4: # Link Purpose (In Context)
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
```
ShadowBB: I don´t think this rule satisfies.

This rule together with https://act-rules.github.io/rules/c487ae does satisfy. But currently a link can have an empty accessible name and be marked inapplicable while it would fail SC 2.4.4.

Either this rule doesn't satisfy completely or this rule needs an assumption that all links have a non-empty accessible link. Which I don't think is a good assumption because a whole rule is written to combat it.

pages/glossary/programatically-determined-link-context.md
WilcoFiers: There's a lot of repetition in this new text. You've also only solved the case of the element being the same, But similarly to the previous example, I could swap out the top level `p` for another element in your list, and you'd have the same problem again:

```html
<li>
  <h1>ACT Rules Community Group</h1>
  <p><a href="...">learn more</a></p>
</li>
```

Additionally, this no longer allow nested list items to be part of context, like you have in passed example 5. This would now fail the rule:


```html
<ul>
  <li> Ulysses <ul>
    <li><a href="...">HTML</a></li>
    <li><a href="...">EPUB</a></li>
    <li><a href="...">Plain text</a></li>
  </ul></li>
</ul>
```

_rules/descriptive-link-5effbb.md
carlosapaduarte: @ShadowBB Agreed. I'll update accordingly.

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: Regarding the first example, I now would say that this counts as context for this SC (i.e., the text from the heading is in the same listitem). It is my interpretation of the understanding document that context comes from the nearest ancestor (otherwise it should have been written "includes text that is in the same paragraph**s**, list**s**, or table cell**s**...")

This interpretation does have the side effect of now failing passed example 5 (nicely spotted). I've updated it to use the aria-describedby attribute to provide the required context.

I don't really know what to do about the text repetition. I wasn't very happy with it, but this was the most succint I managed to be. Any suggestions to improve it, @WilcoFiers?

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,295 @@
...
+
+```html
+<span role="link" tabindex="0" onclick="document.location+='#desc'">See description of the product.</span>
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Passed Example 5
+
+The [programmatically determined link context][] provided by the `aria-describedby` attribute and the [accessible name][] (from the link's text) describe the purpose of the links.
```
WilcoFiers: Apologies, I don't think I made my point clear. I think the previous version of this, where a parent `li` was used as context, was correct. That needs to pass. It just doesn't the way this is now formulated.

carlosapaduarte: OK. Reverted the example. And updated the definition to include the assumption that HTML is used according to spec.

pages/glossary/programatically-determined-link-context.md
carlosapaduarte: Thanks for the suggestion. In it you are listing elements and not relationships, so I've updated it to list relationships.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,24 @@
...
+key: programmatically-determined-link-context
+--
+
+In HTML, information that is programmatically determinable from a link includes HTML elements that are [visible][] and [included in the accessibility tree][] with a non-empty ("") [accessible name][] and
+
+- have as ancestor in the [flat tree][] a `p` element that is [ancestor][] of the link;
```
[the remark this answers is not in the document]
carlosapaduarte: Solved by your suggestion 👍

carlosapaduarte:
Updated definition of programmatically determined link context as discussed with @WilcoFiers

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+name: Link in context is descriptive
+rule_type: atomic
+description: |
+  This rule checks that the accessible name of a link together with its context describe its purpose.
+accessibility_requirements:
+  wcag20:2.4.4: # Link Purpose (In Context)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
```
Jym77: I'm wondering if this rule should also map to 2.4.9 It is not clear what we should do in such case (one SC is a subset of another).
More discussion on the topic: https://github.com/act-rules/act-rules.github.io/pull/833#discussion_r351650002

carlosapaduarte: I see how this works for fail examples of 2.4.4 (they also fail 2.4.9). But what about passing examples of 2.4.4 that fail 2.4.9? It would require a manual interpretation of the result of the rule if you are looking at SC 2.4.9 specifically.
Having it as two rules, each dedicated to a specific SC solves that "ambiguity"

Jym77: As usual, the important is the fail-on-fail relationship.
- Failing this rule fails both 2.4.4 and 2.4.9
- Passing this rule requires further testing for 2.4.9 (and passes 2.4.4 if we assume that all links have a role of `link`).

This is perfectly fine in term of rule design and this is what happen with most rules (passing the rule requires further testing).

I agree that this does collide with our goal that "passed examples must pass the SC"… 🤔

Jym77: I think this rule needs the mapping to 2.4.9
Failing this rule also fails 2.4.9 (if the link in context is not descriptive, then the link alone is not descriptive); this is a consequence of "failing 2.4.4 also fails 2.4.9 which is more precise".

As per [ACT Rules Format](https://www.w3.org/TR/act-rules-format/#accessibility-requirements-mapping):
> the rule must list all accessibility requirements from those documents that are not satisfied when one or more of the outcomes of the rule is failed. 

So, the rule **must** list both 2.4.4 and 2.4.9 as accessibility requirements.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Expectation
+
+The [accessible name][] of each target element together with its [programmatically determined link context][] describe the purpose of the link.
+
+## Assumptions
+
+- This rule assumes that the purpose of the link is not ambiguous to users in general when seen in context on the web page, which is the exception mentioned in success criterion [2.4.4 Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html). If the link is ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the link out of context which makes it more of a general user experience concern than an accessibility issue.
+- This rule assumes that all elements with the [semantic role][] of ['link'][link] are used as links.
+- This rule assumes that all links have the [semantic role][] of ['link'][link].
```
Jym77: I don't think we need that last assumption. If there is a link which does not have the correct semantic role, the rule is going to ignore it and pass/be inapplicable but that's fine.

The previous assumption is needed because `<div role="link">Foo</div>` fails the rule but not the SC. This assumption is not needed because `<a href="contact.html" role="none">Bar</a>` fails the SC but not the rule; but this is not a problem.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+<a href="#main"><img src="/test-assets/5effbb/main.png" alt="Go to the main content"/></a>
+
+<main>
+	<p id="main">This is the main content.</p>
+</main>
+```
+
+#### Passed Example 3
+
+The [accessible name][] (from the link's text) together with its [programmatically determined link context][] available from the text in the same sentence, describe the purpose of the link.
```
Jym77:
```suggestion
The [accessible name][] (from the link's text), together with its [programmatically determined link context][] (available from the text in the closest `p` ancestor), describes the purpose of the link.
```

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+
+```html
+<p>To see the description of this product <a href="#desc">click here</a>.</p>
```
Jym77: Could we make that example with a link text which is not just "click here"? Given that these generic link texts ("click here", "read more", …) are usually frowned upon…

Maybe 
```suggestion
<p>See the description of <a href="#desc">this product</a>.</p>
```

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+
+```html
+<span role="link" tabindex="0" onclick="document.location+='#desc'">See description of the product.</span>
+
+<p id="desc">This product consists of several web pages.</p>
+```
+
+#### Passed Example 5
+
+The [programmatically determined link context][] provided by the list and the [accessible name][] (from the link's text) describe the purpose of the links.
```
Jym77:
```suggestion
The [programmatically determined link context][] (provided by the ancestor with a role of `listitem` and text "Ulysses") and the [accessible name][] (from the link's text) describe the purpose of the links.
```

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+			</li>
+			<li><a href="https://www.gutenberg.org/files/4300/4300-0.txt"> Plain text </a></li>
+		</ul>
+	</li>
+</ul>
+```
+
+#### Passed Example 6
+
+The context provided by the table header and the [accessible name][] (from the link's text) describe the purpose of the links.
```
Jym77:
```suggestion
The [programmatically determined context][] (provided by the table header [assigned](https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-assigning-header-cells) to the cell containing the link) and the [accessible name][] (from the link's text) describe the purpose of the links.
```

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+			c0.444,0.444,1.159,0.444,1.587,0l9.952-9.952c0.444-0.429,0.444-1.143,0-1.571L21.205,5.007z"
+		/>
+	</a>
+</svg>
+
+<main>
+	<p id="main">This is the main content.</p>
+</main>
+```
+
```
Jym77: To cover the def of programmatically determined link context, you need to add passed examples with:
- context being in the same cell
- `aria-describedby` attribute.

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+#### Failed Example 4
+
+The [accessible name][] (from the link's text) does not describe the purpose of the link. The other information available on the page is not [programmatically determined link context][] because it is on a different paragraph.
```
Jym77:
```suggestion
The [accessible name][] (from the link's text) does not describe the purpose of the link. The other information available on the page is not [programmatically determined link context][] because it is in a different `p` element.
```

_rules/descriptive-link-5effbb.md
```diff
@@ -0,0 +1,292 @@
...
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+An `a` element with its [semantic role][] changed from `link` to another role.
+
+```html
+<a href="http://www.w3.org/WAI" role="button">Web Accessibility Initiative (WAI)</a>
```
Jym77: That's typically breaking the assumption "all link have a role of link", and I don't think we should have example that go against assumptions. In that case, I think the assumption should go away.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,20 @@
+--
+title: Programmatically Determined Link Context
+key: programmatically-determined-link-context
+--
+
+The programmatically determined link context is a list of elements. All elements in the list must be [included in the accessibility tree][], and have one of the following relationships to the link:
```
Jym77: I have to point out that "the closest ancestor that is a `p`" is not a relationship _stricto sensu_… (the relationship would be "**is** the closest `p` ancestor"…)

Plus, the definition is currently ambiguous because it doesn't specify the size of the list. So, there is no unicity of PDL context and, typically, the empty list is always a PDL context for any link.


---
The PDL context is the list of all elements which are included in the accessibility tree and are either:
- an ancestor of the link with a role of listitem; or
- the closest ancestor of the link that is a `p`; or
- …

should be good.

---
Replace "serve as row/column header for the cell" by "is assigned to the cell containing the link" + link to https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-assigning-header-cells

_rules/descriptive-link-5effbb.md
WilcoFiers: Posted this in the contrast rule too. I don't think we need to worry about the passed examples passing 2.4.9 here, as long as they all pass 2.4.4. I don't think there's a way around it for these rules where the level AAA one is a subset of the level A variant.

carlosapaduarte:
@Jym77 I've addressed your request for changes. Can you check if you're happy with it?

@WilcoFiers @ShadowBB @DagfinnRomen you had approved this before the changes requested by @Jym77. The changes impacted the assumptions (removal of one assumption), some of the examples' descriptions, two new passed examples, and some rewording of the definition of programmatically determined link context. You might want to do a quick review of the changes.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,21 @@
...
```
Jym77: There is still an issue on how many elements matching the condition are in the Programmatically Determined Link Context, making the definition ambiguous.

```html
<span id="label>Foo 1</span>
<p>Foo 2
<ul>
<li>Foo 3
<table>
<tr><th>Foo 4</th></tr>
<tr><td>Foo 5 <a href="#" aria-described-by="label">Bar</a></td></tr>
</table>
</ul>
</p>
```
What is the programmatically determined link context of this link?
With the current definition, any subset of {Foo 1, Foo 2, Foo 3, Foo 4, Foo 5} (including the empty set!) can be.

---
Also,  "list" kinda suggest ordering of the elements in the context, which is probably not what you want. "set" seems better.

---
```suggestion
The programmatically determined link context is the set of all elements that are [included in the accessibility tree][], and have one of the following relationships to the link:
```

Jym77:
I fixed one of the failing test by renaming a file. The other failing test is related to #1136 and will hopefully be solved before the 2 week Final Call period 😄

carlosapaduarte:
Final call ends February 14

jkodu:
I fixed up the failing tests ;)

PR #559, opened 2019-05-16
New Rule: Create Semantic Focus (a20046)

ramitgarg:
<< Describe the changes >>

Closes issue: (ADD ISSUE NUMBER HERE)

**When creating PR:**

**After creating PR:**

# How to Review And Approve


Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+authors:
+- Carlos Duarte
+- Damien Engels
+- Marie Trudelle 
+- Ramit Garg
+---
+
+## Applicability
+
+The rule applies to all element in the sequential focus navigation excepts scrollable content element.
```
WilcoFiers: - "all elements" is pretty vague. I propose "all HTML and SVG elements"
- "sequential focus navigation" must be defined
- "scrollable content element" must be defined

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+
+## Expectation
+
+Target element has a semantic role.
```
WilcoFiers: You need to define "semantic role".

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+
+## Assumptions
+
+- Giving focus to scrollable content elements is the only way to scroll non interactive content using a keyboard.
```
WilcoFiers: I don't think I understand why this is an assumption. Can you clarify?

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+```
+
+#### Passed Example 4
+
+
+### Failed
+
+#### Failed Example 1
+
+Element in sequential navigation doesnot have semantic role
```
WilcoFiers: Typo: "doesnot"

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+
+```html
+<a href="#desc" tabindex="0" aria-hidden="true">More</a>
```
WilcoFiers: This scenario is addressed by a different role: https://act-rules.github.io/rules/6cfa84 I suggest that we avoid overlap by adding "that is [included in the accessibility tree](#included-in-the-accessibility-tree)" to the applicability.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+```
+
+#### Failed Example 2
+
+
```
WilcoFiers: Remove the duplicate empty line.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+Element in sequential navigation doesnot have semantic role
+
+```html
+<div href="#desc" role="presentation" tabindex="0">More</div>
```
WilcoFiers: If you want this to fail, you should add a second expectation saying that the semantic role isn't "presentation" or "none".

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+Element with tabindex 0 has explicit semantic role 
+
+```html
+<div id="id1" role="button" tabindex="0">Go</div>
+```
+
+#### Passed Example 4
+
+
+### Failed
```
WilcoFiers: Add a test where you override the role of an element that's focusable by default.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+#### Failed Example 2
+
+
+Element in sequential navigation doesnot have semantic role
+
+```html
+<div href="#desc" role="presentation" tabindex="0">More</div>
+```
+
+### Inapplicable
```
WilcoFiers: Add a test where you disable an element to remove it from focus sequence.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+Focusable element have semantic role 
+
+```html
+<a href="xyz">Go to order</a>
```
WilcoFiers: Please put a real URL here.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+
+accessibility_requirements: 
+  wcag20:2.4.4: # Link Purpose (In Context) 
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: further testing needed
+
+input_aspects:
+- DOM Tree
```
WilcoFiers: Should add CSS Styling, the sequential focus navigation is impacted by display:none and the like.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
+---
+id: 
+name: Scemantic Focus 
+rule_type: atomic
+
+description: | 
+ This rule checks that every element part of the sequential focus navigation has a semantic role.
+
+accessibility_requirements: 
+  wcag20:2.4.4: # Link Purpose (In Context)
```
WilcoFiers: I don't believe this is a failure of 2.4.4. More generally, while I'm totally on board with this rule, I don't believe this is a failure of WCAG 2.1 at all. I've had that discussion in the past, this rule looks quite a lot like something we have in axe-core today. Here's a link to that conversation:
https://github.com/dequelabs/axe-core/issues/632

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
+---
+id:
```
jkodu: I usually use browserling.com/tools/random-hex for generating the id

Scemantic Focus
jkodu: This is not valid YAML. This needs to be indented.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
...
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: further testing needed
+
+input_aspects:
+- DOM Tree
+
+authors:
+- Carlos Duarte
```
jkodu: This is not valid YAML. These needs to be indented.

Scemantic Focus
annethyme: `[sequential focus navigation](https://www.w3.org/TR/html/editing.html#sec-sequential-focus-navigation`

annethyme:
Siteimprove doesn't think this rule is testing for anything required by WCAG.

ShadowBB:
Although I think this is an interesting rule that can definitely indicate accessibility failures, it probably doesn't explicitly test anything in WCAG.

@WilcoFiers , I think this also depends on the definition of "user interface components" because if user interface components include things that receive focus, those things will need a role or fail 4.1.2.

Scemantic Focus
```diff
@@ -0,0 +1,110 @@
+---
+id: 
+name: Scemantic Focus
```
EmmaJP: Typo

Scemantic Focus
EmmaJP: Anna suggested 4.1.2

ShadowBB: > 
> 
> Anne suggested 4.1.2

This would also not fail 4.1.2

4.1.2 asks a role of all "user interface components". But if a piece of text with a span on it suddenly receives focus (because of tabindex for example) that doesn't make that piece of text a user interface component because it doesn't control anything.

I agree that it is a good rule (because most of the time it will point to a 4.1.2 violation) but it is currently not necessarily a WCAG failure to fail this rule.

Scemantic Focus
carlosapaduarte: It would have been better to have this as note in the applicability instead of an assumption. But after reviewing the definition of focusable are I don't think this is needed, given that if a scrollable region of an element is focusable, then it should have a semantic role so that users of AT can understand what role that region plays.

semantic-focus-a20046.md
```diff
@@ -0,0 +1,120 @@
+---
+id: a20046
+name: Sequential focus has semantic role
+rule_type: atomic
+
+description: | 
+ This rule checks that every element part of the sequential focus navigation order has a semantic role.
```
EmmaJP: "...every element" that is "part of the sequential..."

semantic-focus-a20046.md
```diff
@@ -0,0 +1,120 @@
+---
+id: a20046
+name: Sequential focus has semantic role
+rule_type: atomic
+
+description: | 
+ This rule checks that every element that is part of the sequential focus navigation order has a semantic role.
+
+accessibility_requirements: 
+  wcag20:4.1.2: # Name, Role, Value (A)
```
ShadowBB: I would (still) not say this fails WCAG 4.1.2. 

A Div with tabindex is not a user interface component and doesn't need a role.

jkodu:
@ramitgarg any updates on these?

carlosapaduarte:
@jkodu we decided last week to drop this one since it does not address any SC. do you want us to keep working on it?

jkodu:
This seems to have "Should we proceed" for a while? Have we reached a conclusion on the same?

carlosapaduarte:
As far as I recall, the last time this was discusses in a meeting, the conclusion was to leave on hold because this is not testing a SC, but it's just a best practice.

PR #567, opened 2019-05-23
Name from content (2ee8b8), add substitute math symbols failure

### LINKED ISSUE #493: Issue with rule "Label and name from content mismatch" (2ee8b8)
@mraccess77 wrote in https://github.com/act-rules/act-rules.github.io/issues/104#issuecomment-488006872: 

> I think "2" and "two" would pass as they would be voiced the same way be users of speech recognition and this is the reason for this test. Some discussion here w3c/wcag#592.

WilcoFiers:
Name from content, add substitute math symbols failure

Closes #493

**When creating PR:**

**After creating PR:**

# How to Review And Approve


_rules/label-content-name-mismatch-2ee8b8.md
```diff
@@ -93,6 +93,14 @@ Not all of visible label is included in accessible name.
 <button name="link" aria-label="the full">The full label</button>
 ```
 
+#### Failed Example 3
+
+Mathematical symbols can not be substitute for text.
+
+```html
+<a href="/" aria-label="Page two">Page 2</a>
```
jkodu: can we add examples with other mathematical operators (+,-,*,/).

WilcoFiers: In the spirit of not overloading the test cases, I'd prefer not to. Anyone else things this is essential? @kasperisager @carlosapaduarte @annethyme?

kasperisager: I do think calling numerals "mathematical symbols" is a bit of a stretch.

carlosapaduarte: Indeed, I have to agree with @kasperisager 
That would be appropriate if the example included a mathematical formula, but here the issue is not related with the interpretation of mathematical symbols, but with potential translation issues.
My suggestion would be to retain this example but with a different description, and add an example with a mathematical formula with this example's description.

ShadowBB: Can we have one example that includes both?

`<a href="/" aria-label="Proof of one plus one is two">Proof of 1+1=2</a>`

jkodu:
Linked PR - https://github.com/act-rules/act-rules.github.io/pull/452

We have to ensure that there are no merge conflicts and or no difference in interpretation of the rule.

jkodu:
@WilcoFiers can you resolve conflicts here and get this going again please.

Jym77:
This is minor change and will only require a 1 week Final Call.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -93,6 +93,14 @@ Not all of [visible][] label is included in [accessible name][].
 <button name="link" aria-label="the full">The full label</button>
 ```
 
+#### Failed Example 3
+
+Mathematical symbol cannot be substitute for text as [explicitly mentioned in WCAG](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name#mathematical-expressions-and-formulae).
```
Brynanders: Mathematical symbol cannot be substitute... > Mathematical symbols cannot be substituted...

audreymaniez: +1 @Brynanders "Mathematical symbols cannot be substituted by text"

Jym77: Can actually go both ways, depending on which one is the "original" and which one is the [substitute](https://en.wiktionary.org/wiki/substitute#Noun) 😄

Brynanders: @Jym77 its grammatically incorrect to use the singular "symbol" and "substitute"
symbol > symbols
substitute > substituted

Jym77: Ah, yes, I missed the requested pluralisation of "symbol" 😞 Makes sense…

Jym77:
Final Call ends on October 29.

bruce-usab:
I like adding the math example, but I think the aria label should be “Proof of two times two equals four” and not “Proof of two multiplied by two is four”.  (A more literal reading might help with translation too.)

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -93,6 +93,14 @@ Not all of [visible][] label is included in [accessible name][].
 <button name="link" aria-label="the full">The full label</button>
 ```
 
+#### Failed Example 3
+
+Mathematical symbols cannot be substituted for text as [explicitly mentioned in WCAG](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name#mathematical-expressions-and-formulae).
+
+```html
+<a href="/" aria-label="Proof of two multiplied by two is four">Proof of 2&times;2=4</a>
```
bruce-usab: I like adding the math example, but aria-label should be “Proof of two times two equals four” and not “Proof of two multiplied by two is four”, as per the guidance in the Understanding page link.   (Also, a more literal aria-label might help with translation too.)

Jym77:
> I like adding the math example, but I think the aria label should be “Proof of two times two equals four” and not “Proof of two multiplied by two is four”. (A more literal reading might help with translation too.)

That's a bit of the point 😃 The example is failing because math symbols cannot be used on one side (accessible name/label) while text is used for the other. WCAG lists as a reason the fact that `×` can be read both at "times" and "multiplied by":
>  For example, making the name "eleven multiplied by three is equivalent to thirty-three" might mean a user who said "eleven times three equals thirty-three" may not match.

Hence, I voluntary chose "multiplied by" (which I consider as being a less common way to pronounce `×`, and visually clashing with `&times;`) in order to illustrate why this is a problem (in addition to linking to the longer explanation in WCAG).

bruce-usab:
Yikes!  I totally missed that this was a failure example!  Thank you @Jym77!

PR #589, opened 2019-06-06
New Rule: Meta viewport do not prevent zoom

audreymaniez:

**When creating PR:**

**After creating PR:**

# How to Review And Approve

### DIFF
--- a/_rules/meta-viewport.md
+++ b/_rules/meta-viewport.md
@@ -1,0 +1,131 @@
+---
+id: 
+name: meta viewport
+rule_type: atomic
+description: |
+  This rule checks that the meta element is not used to block the user agent ability to zoom 
+accessibility_requirements:
+  wcag20:1.4.10: # Reflow (AA)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to the first valid `<meta name="viewport">` element with a `content` attribute in a document.
+
+## Expectation
+
+The `content` attribute of the test target: 
+- do not have the property `user-scalable` or it is set to `yes`
+- do not have the property `maximum-scale` 
+
+## Assumptions
+
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The meta `viewport` do not define the `maximum-scale` and `user-scalable` properties.
+
+````html
+
+<head>
+	<meta name="viewport" content="width=device-width" />
+</head>
+
+````
+
+#### Passed Example 2
+
+The meta `viewport` define the `user-scalable` so the user can still zoom in.
+
+````html
+
+<head>
+  <meta name="viewport" content="width=device-width, user-scalable=yes" />
+</head>
+
+````
+
+#### Passed Example 3
+
+The meta `viewport` with an empty `content` attribute.
+
+````html
+
+<head>
+  <meta name="viewport" content="" />
+</head>
+
+````
+
+### Failed
+
+#### Failed Example 1
+
+The meta `viewport` define the `user-scalable` so the user can't zoom in.
+
+````html
+
+<head>
+  <meta name="viewport" content="width=device-width, user-scalable=no" />
+</head>
+
+````
+
+#### Failed Example 2
+
+The meta `viewport` set the `maximum-scale` to 1 so the user can't zoom in.
+
+````html
+
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=1.0" />
+</head>
+
+````
+
+#### Failed Example 3
+
+The meta `viewport` set the `maximum-scale` to 2 so the user can't zoom in more than twice.
+
+````html
+
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=2" />
+</head>
+
+````
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+The meta `viewport` is not present in the `<head>`.
+
+````html
+
+<head>
+  <meta http-equiv="refresh" content="0; URL='https://github.com'" />
+</head>
+
+````
--- a/_rules/meta-viewport.md
+++ b/_rules/meta-viewport.md
@@ -29,7 +29,6 @@ The `content` attribute of the test target:
 
 ## Assumptions
 
-
 ## Accessibility Support
 
 There are no major accessibility support issues known for this rule.
@@ -47,11 +46,9 @@ There are no major accessibility support issues known for this rule.
 The meta `viewport` do not define the `maximum-scale` and `user-scalable` properties.
 
 ````html
-
 <head>
 	<meta name="viewport" content="width=device-width" />
 </head>
-
 ````
 
 #### Passed Example 2
@@ -59,11 +56,9 @@ The meta `viewport` do not define the `maximum-scale` and `user-scalable` proper
 The meta `viewport` define the `user-scalable` so the user can still zoom in.
 
 ````html
-
 <head>
   <meta name="viewport" content="width=device-width, user-scalable=yes" />
 </head>
-
 ````
 
 #### Passed Example 3
@@ -71,11 +66,9 @@ The meta `viewport` define the `user-scalable` so the user can still zoom in.
 The meta `viewport` with an empty `content` attribute.
 
 ````html
-
 <head>
   <meta name="viewport" content="" />
 </head>
-
 ````
 
 ### Failed
@@ -85,11 +78,9 @@ The meta `viewport` with an empty `content` attribute.
 The meta `viewport` define the `user-scalable` so the user can't zoom in.
 
 ````html
-
 <head>
   <meta name="viewport" content="width=device-width, user-scalable=no" />
 </head>
-
 ````
 
 #### Failed Example 2
@@ -97,11 +88,9 @@ The meta `viewport` define the `user-scalable` so the user can't zoom in.
 The meta `viewport` set the `maximum-scale` to 1 so the user can't zoom in.
 
 ````html
-
 <head>
   <meta name="viewport" content="width=device-width, maximum-scale=1.0" />
 </head>
-
 ````
 
 #### Failed Example 3
@@ -109,11 +98,9 @@ The meta `viewport` set the `maximum-scale` to 1 so the user can't zoom in.
 The meta `viewport` set the `maximum-scale` to 2 so the user can't zoom in more than twice.
 
 ````html
-
 <head>
   <meta name="viewport" content="width=device-width, maximum-scale=2" />
 </head>
-
 ````
 
 ### Inapplicable
@@ -123,9 +110,7 @@ The meta `viewport` set the `maximum-scale` to 2 so the user can't zoom in more 
 The meta `viewport` is not present in the `<head>`.
 
 ````html
-
 <head>
   <meta http-equiv="refresh" content="0; URL='https://github.com'" />
 </head>
-
 ````
--- a/_rules/aria-attr-valid-5f99a7.md
+++ b/_rules/aria-attr-valid-5f99a7.md
@@ -50,7 +50,7 @@ There are no major accessibility support issues known for this rule.
 A valid ARIA 1.1 attribute `aria-atomic` is used on element `article`.
 
 ```html
-<article aria-atomic="true">This is a decription of something cool...</article>
+<article aria-atomic>This is a decription of something cool...</article>
 ```
 
 #### Passed Example 2
@@ -58,7 +58,7 @@ A valid ARIA 1.1 attribute `aria-atomic` is used on element `article`.
 A valid ARIA 1.1 attribute `aria-modal` on element `div` with role `dialog`
 
 ```html
-<div role="dialog" aria-modal="true">Contains modal content...</div>
+<div role="dialog" aria-modal>Contains modal content...</div>
 ```
 
 #### Passed Example 3
--- a/_rules/aria-attr-valid-5f99a7.md
+++ b/_rules/aria-attr-valid-5f99a7.md
@@ -50,7 +50,7 @@ There are no major accessibility support issues known for this rule.
 A valid ARIA 1.1 attribute `aria-atomic` is used on element `article`.
 
 ```html
-<article aria-atomic>This is a decription of something cool...</article>
+<article aria-atomic="true">This is a decription of something cool...</article>
 ```
 
 #### Passed Example 2
@@ -58,7 +58,7 @@ A valid ARIA 1.1 attribute `aria-atomic` is used on element `article`.
 A valid ARIA 1.1 attribute `aria-modal` on element `div` with role `dialog`
 
 ```html
-<div role="dialog" aria-modal>Contains modal content...</div>
+<div role="dialog" aria-modal="true">Contains modal content...</div>
 ```
 
 #### Passed Example 3
--- a/_rules/aria-state-or-property-allowed-5c01ea.md
+++ b/_rules/aria-state-or-property-allowed-5c01ea.md
@@ -51,7 +51,7 @@ _There are no major accessibility support issues known for this rule._
 `aria-pressed` state is supported for role `button`.
 
 ```html
-<div role="button" aria-pressed="false"></div>
+<div role="button" aria-pressed="false">My button</div>
 ```
 
 #### Passed Example 2
@@ -59,7 +59,7 @@ _There are no major accessibility support issues known for this rule._
 `aria-pressed` state is supported for `role=button` that is the implicit role for `button` element.
 
 ```html
-<button aria-pressed="false"></button>
+<button aria-pressed="false">My button</button>
 ```
 
 #### Passed Example 3
@@ -67,7 +67,7 @@ _There are no major accessibility support issues known for this rule._
 Global state that is supported by all base markup elements.
 
 ```html
-<div aria-busy="true"></div>
+<div aria-busy="true">My busy button</div>
 ```
 
 #### Passed Example 4
@@ -83,7 +83,7 @@ Global state that is supported by all base markup elements.
 `aria-checked` state is required for role `aria-checkbox`
 
 ```html
-<div role="checkbox" aria-checked="false"></div>
+<div role="checkbox" aria-checked="false">My checkbox</div>
 ```
 
 #### Passed Example 6
@@ -91,7 +91,7 @@ Global state that is supported by all base markup elements.
 `aria-controls` property is supported for role `combobox`
 
 ```html
-<div role="combobox" aria-controls="id1"></div>
+<div role="combobox" aria-controls="id1">My combobox</div>
 ```
 
 #### Passed Example 7
@@ -99,7 +99,7 @@ Global state that is supported by all base markup elements.
 WAI-ARIA states and properties with empty value are also applicable to this rule
 
 ```html
-<div role="combobox" aria-controls></div>
+<div role="combobox" aria-controls>My combobox</div>
 ```
 
 #### Passed Example 8
@@ -107,7 +107,7 @@ WAI-ARIA states and properties with empty value are also applicable to this rule
 WAI-ARIA states and properties with empty value, specified as an empty string, are also applicable to this rule
 
 ```html
-<div role="combobox" aria-controls=""></div>
+<div role="combobox" aria-controls="">My combobox</div>
 ```
 
 ### Failed
@@ -117,7 +117,7 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 `aria-sort` property is neither inherited, supported, nor required for role `button`.
 
 ```html
-<div role="button" aria-sort=""></div>
+<div role="button" aria-sort="">Sort by year</div>
 ```
 
 #### Failed Example 2
@@ -125,7 +125,7 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 `aria-sort` attribute is neither inherited, supported, nor required for `role=button` that is the implicit role for `button` element.
 
 ```html
-<button aria-sort=""></button>
+<button aria-sort="">Sort by year/button>
 ```
 
 ### Inapplicable
@@ -135,7 +135,7 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 No [WAI-ARIA state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
 
 ```html
-<div role="region"></div>
+<div role="region">A region of content</div>
 ```
 
 #### Inapplicable Example 2
--- a/_rules/aria-state-or-property-allowed-5c01ea.md
+++ b/_rules/aria-state-or-property-allowed-5c01ea.md
@@ -51,7 +51,7 @@ _There are no major accessibility support issues known for this rule._
 `aria-pressed` state is supported for role `button`.
 
 ```html
-<div role="button" aria-pressed="false">My button</div>
+<div role="button" aria-pressed="false"></div>
 ```
 
 #### Passed Example 2
@@ -59,7 +59,7 @@ _There are no major accessibility support issues known for this rule._
 `aria-pressed` state is supported for `role=button` that is the implicit role for `button` element.
 
 ```html
-<button aria-pressed="false">My button</button>
+<button aria-pressed="false"></button>
 ```
 
 #### Passed Example 3
@@ -67,7 +67,7 @@ _There are no major accessibility support issues known for this rule._
 Global state that is supported by all base markup elements.
 
 ```html
-<div aria-busy="true">My busy button</div>
+<div aria-busy="true"></div>
 ```
 
 #### Passed Example 4
@@ -83,7 +83,7 @@ Global state that is supported by all base markup elements.
 `aria-checked` state is required for role `aria-checkbox`
 
 ```html
-<div role="checkbox" aria-checked="false">My checkbox</div>
+<div role="checkbox" aria-checked="false"></div>
 ```
 
 #### Passed Example 6
@@ -91,7 +91,7 @@ Global state that is supported by all base markup elements.
 `aria-controls` property is supported for role `combobox`
 
 ```html
-<div role="combobox" aria-controls="id1">My combobox</div>
+<div role="combobox" aria-controls="id1"></div>
 ```
 
 #### Passed Example 7
@@ -99,7 +99,7 @@ Global state that is supported by all base markup elements.
 WAI-ARIA states and properties with empty value are also applicable to this rule
 
 ```html
-<div role="combobox" aria-controls>My combobox</div>
+<div role="combobox" aria-controls></div>
 ```
 
 #### Passed Example 8
@@ -107,7 +107,7 @@ WAI-ARIA states and properties with empty value are also applicable to this rule
 WAI-ARIA states and properties with empty value, specified as an empty string, are also applicable to this rule
 
 ```html
-<div role="combobox" aria-controls="">My combobox</div>
+<div role="combobox" aria-controls=""></div>
 ```
 
 ### Failed
@@ -117,7 +117,7 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 `aria-sort` property is neither inherited, supported, nor required for role `button`.
 
 ```html
-<div role="button" aria-sort="">Sort by year</div>
+<div role="button" aria-sort=""></div>
 ```
 
 #### Failed Example 2
@@ -125,7 +125,7 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 `aria-sort` attribute is neither inherited, supported, nor required for `role=button` that is the implicit role for `button` element.
 
 ```html
-<button aria-sort="">Sort by year/button>
+<button aria-sort=""></button>
 ```
 
 ### Inapplicable
@@ -135,7 +135,7 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 No [WAI-ARIA state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).
 
 ```html
-<div role="region">A region of content</div>
+<div role="region"></div>
 ```
 
 #### Inapplicable Example 2


_rules/meta-viewport.md
```diff
@@ -0,0 +1,116 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to the first valid `<meta name="viewport">` element with a `content` attribute in a document.
```
jkodu: why only the first valid one? Does having multiple override the previous meta?

audreymaniez: @jkodu **I think you're right, and we need to focus every meta viewport in the document.**

I based the writing of the rule on the meta refresh rule (and that is the applicability for a meta refresh). But I now did some tests for the meta viewport :-) 

And I find two different behaviors depending on properties (`user-scalable` or `maximum-scale`)

````html
	<meta name="viewport" content="width=device-width, user-scalable=no" />
	<meta name="viewport" content="width=device-width, user-scalable=yes" />
````
The `user-scalable=yes` is on the last meta:
- FF/Android -> the user can zoom
- CH/Android -> the user can zoom
- Internet/Android - Safari/iOS  -> the user can zoom

````html
	<meta name="viewport" content="width=device-width, user-scalable=yes" />
	<meta name="viewport" content="width=device-width, user-scalable=no" />
````

The `user-scalable=yes` is on the first meta:
- FF/Android -> the user can't zoom
- CH/Android -> the user can't zoom
- Internet/Android - Safari/iOS  -> the user can zoom

So we have cases where the browser take into account the last declaration of the meta viewport for the `user-scalable` property, and cases where browser just don't care about :-)

**BUT** for the `maximum-scale` property, there are differences. 
In some browsers, whatever it is defined in the first or last meta viewport, at the moment it is defined, the value applies.

````html
	<meta name="viewport" content="width=device-width, user-scalable=yes" />
	<meta name="viewport" content="width=device-width, user-scalable=yes,maximum-scale=1" />
````
The `maximum-scale` is set to 1 in the last viewport
- Android/FF -> The user can't zoom 
- Android CH -> The user can't zoom 
- Android/Internet-Safari/iOS-> The user can zoom

````html
	<meta name="viewport" content="width=device-width, user-scalable=yes,maximum-scale=1" />
	<meta name="viewport" content="width=device-width, user-scalable=yes" />
````

The `maximum-scale` is set to 1 in the first viewport 
- Android/FF -> The user can't zoom 
- Android/CH -> The user **can** zoom 
- Android/Internet-Safari/iOS-> The user can zoom

As the `<title>`, a meta viewport defined in the `<body>` element applied too, and seems to follow the same patten as defined above ... 

````html
<head>    
	<meta name="viewport" content="width=device-width, user-scalable=yes" />
</head>

<body>
	<meta name="viewport" content="width=device-width, user-scalable=no" />
</body>
````
In that case, in FF/Android, the user can't zoom, cause the last meta set the user-scalable to "no" (even if the meta is in the body, not the head).

-----

I didn't find anything in the spec about what to do with multiple meta viewport (or didn't find anything *easily* :-) ), so based on those tests and the multiplicity of behavior, maybe we need to focus on every meta viewport that are present in the head? 


SO! For applicabilty, I would say:  

> The rule applies to every `<meta name="viewport">` elements with a `content` attribute in a document.

File: _rules/meta-viewport.md [hunk 2]
```diff
@@ -0,0 +1,116 @@
+---
+id:
```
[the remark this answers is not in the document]
audreymaniez: Done!

_rules/meta-viewport.md
```diff
@@ -0,0 +1,116 @@
+---
```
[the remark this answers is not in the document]
audreymaniez: Done! Thanks @jkodu!

_rules/meta-viewport.md
jkodu: Nice work, testing these @audreymaniez 

I agree on the applicability:

>The rule applies to every <meta name="viewport"> elements with a content attribute in a document.

And should definitely add notes on `Accessibility Support` about the above found inconsistencies.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+---
+
+## Applicability
+
+The rule applies to every `<meta name="viewport">` elements with a `content` attribute in a document.
+
+## Expectation
+
+The `content` attribute of the test target: 
+- do not have the property `user-scalable` or it is set to `yes`
```
carlosapaduarte: do > does

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+- do not have the property `maximum-scale`
```
carlosapaduarte: do > does

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The meta `viewport` do not define the `maximum-scale` and `user-scalable` properties.
```
carlosapaduarte: do > does

Shouldn't `meta` be written between backticks because you're referring to the HTML element? If so, change in all examples, please.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+````html
+<head>
+	<meta name="viewport" content="width=device-width" />
+</head>
+````
+
+#### Passed Example 2
+
+The meta `viewport` define the `user-scalable` so the user can still zoom in.
```
carlosapaduarte: define > defines
include "property" after `user-scalable`

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+<head>
+  <meta name="viewport" content="" />
+</head>
+````
+
+### Failed
+
+#### Failed Example 1
+
+The meta `viewport` define the `user-scalable` so the user can't zoom in.
```
carlosapaduarte: define > defines
include "property" after `user-scalable`

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+````html
+<head>
+  <meta name="viewport" content="width=device-width, user-scalable=no" />
+</head>
+````
+
+#### Failed Example 2
+
+The meta `viewport` set the `maximum-scale` to 1 so the user can't zoom in.
```
carlosapaduarte: set > sets
include "property" after `maximum-scale`

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+````html
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=1.0" />
+</head>
+````
+
+#### Failed Example 3
+
+The meta `viewport` set the `maximum-scale` to 2 so the user can't zoom in more than twice.
```
carlosapaduarte: set > sets
include "property" after `maximum-scale`

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Expectation
+
+The `content` attribute of the test target: 
+- do not have the property `user-scalable` or it is set to `yes`
+- do not have the property `maximum-scale` 
+
+## Assumptions
+
+- There may be some inconsistencies with how different browser vendors interpret `<meta name="viewport">`.
```
carlosapaduarte: I think this should be in the Accessibility Support section, not in the Assumptions

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=2" />
+</head>
+````
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+The meta `viewport` is not present in the `<head>`.
```
carlosapaduarte: But there might be others in the remainder of the document. The applicability is not restricted to the `<head>`

jkodu: I have made the description generic, but check permitted parents here - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The `<meta name="viewport">` does not defines the `maximum-scale` and `user-scalable` properties.
```
carlosapaduarte: defines > define

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=2" />
+</head>
+````
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+The `<meta name="viewport">` is not present.
```
carlosapaduarte: I checked the permitted parents and `<head>` is the permitted parent only for `<meta charset>` and `<meta http-equiv>`. It doesn't say anything about `<meta name>`. I admit I haven't tested it, but I'm willing to bet that browsers will interpret it even outside `<head>`.

Additionally, because the `<meta>` element is empty it does not have an end tag, so you should update the `/>`to `>` in all examples.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+````html
+<head>
+  <meta name="viewport" content="width=device-width, user-scalable=no" />
+</head>
+````
+
+#### Failed Example 2
+
+The `<meta name="viewport">` sets the `maximum-scale=1.0` to 1 so the user can't zoom in.
```
carlosapaduarte: Don't think you need to repeat the 1. Keep it with the property or outside, but not both.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,118 @@
...
+
+````html
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=1.0" />
+</head>
+````
+
+#### Failed Example 3
+
+The `<meta name="viewport">` sets the `maximum-scale=2` to 2 so the user can't zoom in more than twice.
```
carlosapaduarte: Don't think you need to repeat the 2. Keep it with the property or outside, but not both.

_rules/meta-viewport-b4f0c3.md
jkodu: I agree. I added within the page > `The `<meta name="viewport">` is not present within the page.`. That should add some clarity.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,120 @@
...
+
+_There are currently no assumptions._
+
+## Accessibility Support
+
+- There may be some inconsistencies with how different browser vendors interpret `<meta name="viewport">`.
+
+## Background
+
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
```
ShadowBB: Add a link to the understanding document of the SC.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,120 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to every `<meta name="viewport">` elements with a `content` attribute in a document.
```
Jym77: I think "document" should be `[document](https://www.w3.org/TR/dom/#concept-document)`.
(but that link seems to only be here on very few rules, so I'm not sure about it…)

Jym77: Should it be "applies to **each** `meta` element**[singular]**"?
(as in "the rule applies to each individual element and not to every one of them at the same time")

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Expectation
+
+The `content` attribute of the test target: 
+- does not have the property `user-scalable` or it is set to `yes`and
```
Jym77: To avoid messing up the double logical connector, I would use an Oxford-comma (or rather an Oxford-colon):
"- does not have the property `user-scalable` or it is set to `yes`; and"
Maybe even making it more explicit with an either…or:
"- either does not have the property `user-scalable` or has it set to `yes`; and"

_rules/meta-viewport-b4f0c3.md
Jym77: Agreed. Link to WCAG is needed.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,120 @@
...
+
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The `<meta name="viewport">` does not define the `maximum-scale` and `user-scalable` properties.
```
Jym77: Looking at other rules, it seems we usually write "The `meta name="viewport">` **element**".
(same on all examples)

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,120 @@
...
+
+The `<meta name="viewport">` sets the `maximum-scale=2` so the user can't zoom in more than twice.
+
+````html
+<head>
+  <meta name="viewport" content="width=device-width, maximum-scale=2">
+</head>
+````
+
+### Inapplicable
```
Jym77: There should be an inapplicable example **with** a `meta viewport` element **without** a `content` attribute.

_rules/meta-viewport-b4f0c3.md
carlosapaduarte: Do we really need the "in a document" part? The input aspects already specify that this applies to the DOM tree.

_rules/meta-viewport-b4f0c3.md
carlosapaduarte: Agree with @Jym77 suggestion (I prefer the second wording)

jkodu:
Final call ends on 30th July 2019

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+- either does not have the property `user-scalable` or has it set to yes; and
+- does not have the property `maximum-scale`. 
+
+## Assumptions
+
+_There are currently no assumptions._
+
+## Accessibility Support
+
+- There may be some inconsistencies with how different browser vendors interpret `<meta name="viewport">`.
```
WilcoFiers: Can you be a little more specific? What does it mean for this to be "inconsistent"?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Applicability
+
+The rule applies to each `<meta name="viewport">` element with a `content` attribute.
+
+## Expectation
+
+The `content` attribute of the test target: 
+- either does not have the property `user-scalable` or has it set to yes; and
+- does not have the property `maximum-scale`.
```
WilcoFiers: Clearly `maximum-scale=1000` shouldn't fail this. I suggest putting a limit on this. If you allow `maximum-scale` of 4 or greater, this rule will also be applicable to 1.4.4 Resize text.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
+---
+id: b4f0c3
+name: meta viewport
+rule_type: atomic
+description: |
+  This rule checks that the meta element is not used to block the user agent ability to zoom. 
+accessibility_requirements:
+  wcag20:1.4.10: # Reflow (AA)
```
WilcoFiers: I don't think this is a failure of 1.4.10. Instead I would say this is a failure of 1.4.4, provided you allow max scale of 400% or greater. My reasoning is this, SC 1.4.10 says that when content is presented in this 320px width / height, it doesn't require two-dimensional scrolling. It doesn't actually say that it always has to be possible to display content in this 320px width / height. It certainly implies it, but I would argue that the reason it implies it is because SC 1.4.4 requires it. If you read the intent in the understanding document, it even mentions that 400% zoom. https://www.w3.org/WAI/WCAG21/Understanding/reflow.html

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+
+_There are currently no assumptions._
+
+## Accessibility Support
+
+- There may be some inconsistencies with how different browser vendors interpret `<meta name="viewport">`.
+
+## Background
+
+- [WCAG Success Criterion 1.4.10 Reflow](https://www.w3.org/TR/WCAG21/#reflow)
```
WilcoFiers: You don't need to link to WCAG. The accessibility requirements mapping has that link.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+- [Understanding Success Criterion 1.4.10: Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html)
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
+
+
+## Test Cases
```
WilcoFiers: I think you should add some content to these test pages. Without content, none of these actually fail the success criterion.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Expectation
+
+The `content` attribute of the test target: 
+- either does not have the property `user-scalable` or has it set to yes; and
+- does not have the property `maximum-scale`. 
+
+## Assumptions
+
+_There are currently no assumptions._
```
WilcoFiers: There is, you're making an assumption that there is content in the page that there is content in the page for which two-dimensional scrolling doesn't provide meaning. Or in the case of the reflow SC, that there is text in the page. You can either put that in the assumption, or you can address it in the applicability, either way works for me.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
+---
+id: b4f0c3
+name: meta viewport
```
annethyme: Shouldn't it be "Meta viewport does not prevent zoom"

_rules/meta-viewport-b4f0c3.md
annethyme: Adding to Wilco's point, you might make part of the applicability that there is [visible](#visible) content on the page. (don't know if "content" then also needs a definition...)

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to each `<meta name="viewport">` element with a `content` attribute.
+
+## Expectation
+
+The `content` attribute of the test target:
```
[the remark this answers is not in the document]
jkodu: I have done some testing & both of these options do not prevent zooming in modern phones. Also there seems to be accessibility options that override these options by default. I think some changes to the rules have to be done to either change messaging and or bucket them into `needs review` would be the right outcome. I have left some notes on both tickets.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,135 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to each `<meta name="viewport">` element with a `content` attribute.
```
annethyme: See discussion: https://github.com/act-rules/act-rules.github.io/pull/589/files#r305324362

_rules/meta-viewport-b4f0c3.md
annethyme: This is marked as resolved, but the problematic link is still there? 
Makes it a bit hard to find out if I am reviewing an old version of the rule???

_rules/meta-viewport-b4f0c3.md
annethyme: I do not consider this one resolved?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Expectation
+
+The `content` attribute of the test target: 
+- either does not have the property `user-scalable` or has it set to yes; and
+- does not have the property `maximum-scale`, or has it set to a value that prevents significant zooming.
```
carlosapaduarte: prevents > allows

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Applicability
+
+The rule applies to each `<meta name="viewport">` element with a `content` attribute.
+
+## Expectation
+
+The `content` attribute of the test target: 
+- either does not have the property `user-scalable`, or has it set to yes; and
+- either does not have the property `maximum-scale`, or has it set to a value that allows significant zooming.
```
DagfinnRomen: How would you define "significant zooming"?

jkodu: @DagfinnRomen I have updated this to be something more meaningful.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Applicability
+
+The rule applies to each `<meta name="viewport">` element with a `content` attribute.
+
+## Expectation
+
+The `content` attribute of the test target: 
+- either does not have the property `user-scalable`, or has it set to yes; and
+- either does not have the property `maximum-scale`, or has it set to a value that allows zooming upto 200%.
```
DagfinnRomen: *up to

kasperisager:
Digging into this a little more, it seems that the `@viewport` rule might end up being removed entirely due to performance concerns, see https://github.com/w3c/csswg-drafts/issues/258. However, the `<meta name="viewport">` element might get a normative specification in the CSSOM View module, see https://github.com/w3c/csswg-drafts/issues/331.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,225 @@
...
+## Expectation
+
+The [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
+
+- specify the property `user-scalable` with a value of `yes`; nor
+- specify the property `maximum-scale` with a value of less than 2
+
+**Note:** 
+1) For Safari on the iPhone, the content attribute is mapped to property/value pairs according to the following algorithm: https://drafts.csswg.org/css-device-adapt/#parsing-algorithm
+2) Properties and the values for the content attribute of `<meta name="viewport">` are specified in a detailed tabulation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#Attributes).
```
kasperisager: The anchor "here" will either need a proper name or the link will have to be moved:

> [...] here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#Attributes

jkodu: Unfortunately there is no direct link to the tabulation I wanted to refer & background already refers to `meta` element page. So removing note 2.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Applicability
+
+The rule applies to each [`<meta>`](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element) element with a [`name`](https://www.w3.org/TR/html52/document-metadata.html#element-attrdef-meta-name) attribute whose value is set to [`viewport`](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag#Enter_viewport_meta_tag) and has a [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute.
+
+## Expectation
+
+The [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
+
+- specify the property `user-scalable` with a value of `yes`; nor
```
kasperisager: Brain fart on my end; `yes` should be `no` 🤦‍♂

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,223 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to each [`<meta>`](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element) element with a [`name`](https://www.w3.org/TR/html52/document-metadata.html#element-attrdef-meta-name) attribute whose value is set to [`viewport`](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag#Enter_viewport_meta_tag) and has a [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute.
```
Jym77: Don't cite MDN.

Jym77: Are `viewport`, `user-scalable` and `maximum-scale` case-sensitive?
The way the rule is currently written, they are. But does that reflects reality?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Expectation
+
+The [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
+
+- specify the property `user-scalable` with a value of `no`; nor
+- specify the property `maximum-scale` with a value of less than 2
+
+**Note:** For Safari on the iPhone, the content attribute is mapped to property/value pairs according to the following algorithm: https://drafts.csswg.org/css-device-adapt/#parsing-algorithm
```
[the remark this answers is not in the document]
jkodu: We seem to refer this in many other rules without saying it is WIP. What would you recommend we change this to?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,223 @@
...
+- [The `meta` element](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element)
+- [<meta>: The Document-level Metadata element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The `<meta name="viewport">` element does not define the `maximum-scale` and `user-scalable` values.
```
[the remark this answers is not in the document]
jkodu: Not going to change this @Jym77. Let us try to have a unified consensus here first before making another change. Perhaps make an issue to discuss please.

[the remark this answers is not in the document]
Jym77: I am rather in favour of using `<element>` instead of `element`, but it seems that the huge majority of rules are doing it the other way, with only 4 using `<element>`:
```
 (develop=)$ grep -lr \`\< _rules/
_rules/html-title-is-descriptive-c4a8a4.md
_rules/link-has-name-c487ae.md
_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
_rules/meta-refresh-bc659a.md
```

_rules/meta-viewport-b4f0c3.md
jkodu: Yup I think so - https://www.w3.org/TR/css-device-adapt-1/#meta-properties.

jkodu:
Final Call 24th September

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+#### Passed Example 3
+
+The `<meta name="viewport">` element defines the `maximum-scale=6.0` which allows the user to zoom.
+
+```html
+<html>
+	<head>
+		<meta name="viewport" content="width=device-width, maximum-scalable=6.0" />
```
cblouch: You mean maximum-scale here.

_rules/meta-viewport-b4f0c3.md
Jym77: * Names (_i.e._ `viewport`) seem to be case-insensitive: https://html.spec.whatwg.org/multipage/semantics.html#standard-metadata-names

* I'm having harder time figuring if properties names and values (_i.e._ what's inside `content`) are case sensitive or not. The link you put list them without explicitly mentioning case-sensitivity (I agree that I would assume it by default). It seems that Chrome does not respect meta viewport and I do not have a Safari to test, can you run some tests?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'title-require'
+---
+
+## Applicability
+
+The rule applies to each [`<meta>`](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element) element with a [`name`](https://www.w3.org/TR/html52/document-metadata.html#element-attrdef-meta-name) attribute whose value is set to [`viewport`](https://www.w3.org/TR/css-device-adapt-1/#viewport-meta) and has a [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute.
+
+## Expectation
```
Jym77: There are a bunch of rules on how to interpret "unknown" values (_e.g._ `maximum-scale=no`, `user-scalable=-1`, …)
https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale
https://www.w3.org/TR/css-device-adapt-1/#user-scalable

This should at the very least be listed, maybe just as a note.

kasperisager:
@Jym77 I already covered the `@viewport` rule here: https://github.com/act-rules/act-rules.github.io/pull/589#issuecomment-526063911. Executive summary: The CSS Device Adaption specification, and by extension the `@viewport` rule, has failed to gained traction among browser vendors and will be phased out. Only Chrome has a working implementation (behind a feature flag even) and it will be removed in the future.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+
+The [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
+
+- specify the property `user-scalable` with a value of `no`; nor
+- specify the property `maximum-scale` with a value of less than 2
```
WilcoFiers: I don't think this rule is atomic. I'd much prefer either split this over two expectations, or even have two separate rules. Strong preference for the second, although I'm not going to hold up publication for it.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+  - Audrey Maniez
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'title-require'
+---
+
+## Applicability
+
+The rule applies to each [`<meta>`](https://www.w3.org/TR/html52/document-metadata.html#the-meta-element) element with a [`name`](https://www.w3.org/TR/html52/document-metadata.html#element-attrdef-meta-name) attribute whose value is set to [`viewport`](https://www.w3.org/TR/css-device-adapt-1/#viewport-meta) and has a [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute.
```
WilcoFiers: `meta` should not have angle brackets on it. I also don't think this needs to be linked TBH, same for `name` and `content`.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Expectation
+
+The [`content`](https://www.w3.org/TR/html52/document-metadata.html#dom-htmlmetaelement-content) attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
```
WilcoFiers: Expectations is disconnected from the applicability. That's why expectations start with "Each test target ..."

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+
+- specify the property `user-scalable` with a value of `no`; nor
+- specify the property `maximum-scale` with a value of less than 2
+
+**Note:** For a known popular browser, the content attribute is mapped to property/value pairs according to the following algorithm: https://www.w3.org/TR/css-device-adapt-1/#parsing-algorithm
```
WilcoFiers: What does "for a known popular browser" mean? Is the intent here to say that the value must value must be obtained using this parsing algorithm? It probably should be.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Assumptions
+
+- The [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s) has [visible](#visible) [content](https://www.w3.org/TR/WCAG21/#dfn-content).
+
+## Accessibility Support
+
+- Interpretation of `<meta name="viewport">` may vary with browsers, meaning browsers may override `user-scalable=no` and `maximum-scale=1` values to support gestures like pinch and zoom based interaction on permissible devices, thereby causing inconsistent behavior.
```
WilcoFiers: Can you add an explanation of what this statement means for this rule?

_rules/meta-viewport-b4f0c3.md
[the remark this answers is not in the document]
WilcoFiers: I agree with JY on this one. This goes against convention. This is one we definitely did settle.

_rules/meta-viewport-b4f0c3.md
Jym77: …and yet again…
It seems that the `@viewport` CSS (and the translation from `meta name="viewport"` to it) is basically used only by Safari and that other browsers (Chrome) are doing some other kind of magic to handle the `meta name="viewport"` attribute… In which case "is not translated into" is a very bad formulation.

🤔

jkodu: @Jym77 - see comment - https://github.com/act-rules/act-rules.github.io/pull/589#issuecomment-530292361

_rules/meta-viewport-b4f0c3.md
jkodu: @Jym77 as discussed on a call, I tested various casing locally & it looks like meta tags and their values are case insensitive.

Some relevant links:
- https://support.google.com/webmasters/answer/79812?hl=en
- https://webmasters.stackexchange.com/questions/67760/which-meta-names-are-case-sensitive
- https://html.spec.whatwg.org/multipage/dom.html#case-insensitivity

_rules/meta-viewport-b4f0c3.md
jkodu: Previous note was as below. I do not think this note is relevant anymore.

> **Note:** For Safari on the iPhone, the content attribute is mapped to property/value pairs according to the following algorithm: https://drafts.csswg.org/css-device-adapt/#parsing-algorithm

_rules/meta-viewport-b4f0c3.md
Jym77: Then, we should put case-insensitivity in the rule, _e.g._ "witha `name` attribute is a case-insensitive match for `viewport`" and so on.

_rules/meta-viewport-b4f0c3.md
jkodu: I agree with you. This sentence is too generic & does not cover the core of what this rule tries to establish. I was having it as a caveat to cater for cases where these values are set to achieve other behaviour. I will remove this.

_rules/meta-viewport-b4f0c3.md
kasperisager: @jkodu @WilcoFiers Did you come to a conclusion on this?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,254 @@
...
+  - Audrey Maniez
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'title-require'
+---
+
+## Applicability
+
+The rule applies to each `meta <element>` with a `name` attribute whose value is set to [`viewport`](https://www.w3.org/TR/css-device-adapt-1/#viewport-meta) and has a `content` attribute.
```
kasperisager: "`meta <element>`" should be "`meta` element". Where did the brackets come from?

_rules/meta-viewport-b4f0c3.md
jkodu: @kasperisager yes, we are not splitting the rule, leaving this as is.

jkodu:
@kasperisager the conclusion came from here - https://github.com/act-rules/act-rules.github.io/pull/589#discussion_r322742616

_rules/meta-viewport-b4f0c3.md
jkodu: See comment - https://github.com/act-rules/act-rules.github.io/pull/589#discussion_r322742616

Jym77:
> @jkodu I'm fairly certain that @Jym77's comment was exemplifying the difference between an element name without and with brackets (as in `element` vs. `<element>`) and concluding that we only use the latter notation in 4 places and the former therefore being the convention. I did not read it as him suggesting literally writing `<element>`.

Precisely.
And @WilcoFiers confirmed in https://github.com/act-rules/act-rules.github.io/pull/589#discussion_r323164259 that the ACT-Rules convention is to write elements name without brackets (_i.e._ `title`, not `<title>`, `meta`, not `<meta>`, …)

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,254 @@
...
+The `meta name="viewport" <element>` does not define the `maximum-scale` and `user-scalable` values.
+
+```html
+<html>
+	<head>
+		<meta name="viewport" content="width=device-width" />
+	</head>
+	<body>
+		<p>
+			Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore
```
Jym77: Also, I'm not sure we need that much garbage in the examples (make them bigger than needed). `Lorem ipsum` should be enough.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+#### Passed Example 5
+
+The `meta name="viewport" <element>` sets the `maximum-scale=2` so the user can zoom in more than twice (200%).
+
+```html
+<html>
+	<head>
+		<meta name="viewport" content="width=device-width, maximum-scale=2.0" />
```
Jym77: This is essentially the same example as Passed Example 3 and is not needed.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+#### Passed Example 2
+
+The `meta name="viewport" <element>` defines the `user-scalable=yes` so the user can still zoom in.
+
+```html
+<html>
+	<head>
+		<meta name="viewport" content="width=device-width, user-scalable=yes" />
```
Jym77: Why having `widthe=device-width` all over th place? What value does it add to the rule?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,250 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to each `meta` element with a `name` attribute whose value is set to [`viewport`](https://www.w3.org/TR/css-device-adapt-1/#viewport-meta) and has a `content` attribute.
```
kasperisager: "set" -> "equal".

_rules/meta-viewport-b4f0c3.md
```diff
@@ -0,0 +1,250 @@
...
+input_aspects:
+  - DOM Tree
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to each `meta` element with a `name` attribute whose value is equal to [`viewport`](https://www.w3.org/TR/css-device-adapt-1/#viewport-meta) and has a `content` attribute.
```
Jym77: "whose value is equal to `viewport`" => "whose value is a case-insensitive match for `viewport`".

kasperisager: Good catch 👍 Also, "case-insensitive" -> "[ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive)".

PR #593, opened 2019-06-10
chore: add husky and lint-staged

jkodu:
This PR does the below:
- adds `husky`.
- adds `lint-staged`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,9 @@
 		"showdown": "^1.9.0"
 	},
 	"devDependencies": {
+		"husky": "^2.4.0",
 		"jest": "^24.8.0",
+		"lint-staged": "^8.2.0",
 		"prettier": "^1.17.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -180,5 +182,16 @@
 			"<rootDir>/.cache/",
 			"<rootDir>/.public/"
 		]
+	},
+	"husky": {
+		"hooks": {
+			"pre-commit": "npm test"
```
WilcoFiers: Actually, no I don't think we should run tests before a commit. We have CI for that.

jkodu: I know, but it will help people like me :).

WilcoFiers: > IF machine==jey npm test

WilcoFiers: Fine, I'll try it. If it gets on my nerves, I'm ripping it out though :P

PR #596, opened 2019-06-10
chore: add accessibility requirements to test cases (json)

jkodu:
This PR adds the accessibility requirements to test cases. This is useful for auto mapping of test cases against rules to validate against in test engines.

image


WilcoFiers:
Very good :)

Issue #467, opened 2019-03-28
Negative formulation


audreymaniez:
More a general reflexion about how we write tests etc... 

In RGAA we have a writing rule that is not to write test in negative form. This because writing in a negative way can lead testers (manual tester espacially) answer "no" to a test when it is compliant (precisly because of the syntaxic form of the test). 

This is more a question of linguistic/syntaxic format/norm, even logical format.

Really basic example: to the following test

> "That element doesn't have that property"

You will propbably answer:

> "**NO** it doesn't have" 

and 

> "**YES** it does have" (or even "**NO**  it does have").

**The problem is you can anwer by "YES" to a non compliant test (and "No or yes" to a compliant). And that is not a logicial outcome.**

Real example from Auto-WCAG rules 

> **None of the target elements** participate in sequential focus navigation, **nor do they have** descendants in the flat tree that participate in sequential focus navigation

With the writing rule we have in RGAA, It would become, something like that:

> Target elements and their descendants in the flat tree, if any, are excluded from the sequential focus navigation.

Another example, an applicability section with multiple conditions: 

> Any WAI-ARIA required state or property that
> 
>    - is specified on an HTML or SVG element
>   -  is one of the aria-* attributes included in the accessible name computation, or is specified on an element that is included in the accessibility tree or is focusable,
>   -  **is not the empty string ("")**,
>    - **does not have a default value listed under WAI-ARIA implicit value for role**.

The last two conditions are negative. With the writing rule, It would become, something like that: 

> Any WAI-ARIA required state or property that
> 
>   - is specified on an HTML or SVG element
>   - is one of the aria-* attributes included in the accessible name computation, or is specified on an element that is included in the accessibility tree or is focusable,
>   - **is different from an empty string ("")**,
>   - **has a value different from the default value listed under WAI-ARIA implicit value for role.**

EmmaJP:
One of the agreed naming conventions #250 is to use declarative verbs. I think this should help to achieve positive language in the rule names/titles. 

Perhaps resolving this issue will do a similar thing for the descriptions, which I agree should also use declarative verbs and positive language - where meeting the rule is a pass.

jkodu:
@audreymaniez do you still see this as an issue?

jkodu:
Spoke to @audreymaniez, who confirmed that this issue can be closed.

PR #597, opened 2019-06-11
Positive proposals

audreymaniez:
This PR is a set of proposals regarding the #467 issue.

Try to make some description/expectation more logical using positive formulation.

**When creating PR:**

**After creating PR:**

# How to Review And Approve


_rules/image-has-name-23a2a8.md
```diff
@@ -24,7 +24,7 @@ The rule applies to HTML `img` elements or any HTML element with the [semantic r
 
 ## Expectation
 
-Each target element has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace), or is marked as [decorative](#decorative).
+Each target element has an [accessible name](#accessible-name) that consists of characters other than [whitespace](#whitespace).
```
audreymaniez: For that rule, I suppress the "is marked as decorative" in the expectation because I don't understand why the "marked as decorative" is not part of the applicability? 

> The rule applies to HTML `img` elements or any HTML element with the [semantic role](#semantic-role) of `img` that is [included in the accessibility tree](#included-in-the-accessibility-tree) and that is not marked as [decorative](#decorative)

WilcoFiers: @audreymaniez Because marking an image as decorative is a technique to satisfy the SC. Decorative images are still non-text content, and so they're applicable to the SC 1.1.1. You're right that it doesn't make a big difference, but I think it fits the intent of the SC a little better.

audreymaniez: So I put it back. Thank you for the answer

_rules/image-has-name-23a2a8.md
```diff
@@ -24,7 +24,7 @@ The rule applies to HTML `img` elements or any HTML element with the [semantic r
 
 ## Expectation
 
-Each target element has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace), or is marked as [decorative](#decorative).
+Each target element has an [accessible name](#accessible-name) that consists of characters other than [whitespace](#whitespace), or is marked as [decorative](#decorative).
```
Jym77: That seems to change the meaning. `"   a   "` is not only whitespaces but I'm not sure we can say that it "consists of characters other than whitespace" (since there are also whitespaces in it).
And `"A picture of a dolphin"` does contain whitespaces but is definitely a perfectly valid name for an image.

Something in the line of "contains at least one character which is not a whitespace" would be better (and keep the same meaning as the original one), but that still contains a negative, only pushed a bit further…

_rules/link-has-name-c487ae.md
```diff
@@ -34,7 +34,7 @@ The rule applies to any HTML element with the [semantic role](#semantic-role) of
 
 ## Expectation
 
-Each target element has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace).
+Each target element has an [accessible name](#accessible-name) that consists of characters other than [whitespace](#whitespace).
```
Jym77: Same remark.

_rules/image-has-name-23a2a8.md
ShadowBB: I agree that "consists of characters other than whitespace" is ambiguous. It could mean "this string only consists of characters other than whitespace" or it could mean "this string consists of at least 1 character other than whitespace".

This ambiguity should be resolved.

jkodu:
@audreymaniez any update on this???

_rules/image-has-name-23a2a8.md
audreymaniez: I agree that in those cases, trying to reverse the negative in a positive way do not help the reading, more, it can make it harder to understand and leads to mistakes.

audreymaniez:
So, 

The proposal that were made for the "HTML lang and xml:lang match" [is already apply on the rule description](https://github.com/act-rules/act-rules.github.io/blob/develop/_rules/html-xml-lang-match-5b7ae0.md).

Then, the two other modifications that were proposed on the "_that is not only [whitespace](#whitespace)_" seems not satisfying and add some more complexity to the description.


For those reasons, the PR is closed.

PR #600, opened 2019-06-12
chore: Update license

jkodu:
Update license to match https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document


pages/license.md
```diff
@@ -2,60 +2,24 @@
...
-        8.  12.8.8. shall not be considered accepted by an implementer who manifests an intent not to accept the terms of the W3C CLA RF Licensing Requirements license as offered by the licensor.
-        9.  12.8.9. The RF license conforming to the requirements in this policy shall be made available by the licensor as long as the Specification is in effect. The term of such license shall be for the life of the patents in question.
-
-        I am encouraged to provide a contact from which licensing information can be obtained and other relevant licensing information. Any such information will be made publicly available.
-
-    9.  12.9. You or Your. "You," "you," or "your" means any person or entity who exercises copyright or patent rights granted under this CLA, and any person or entity that person or entity controls.
+## W3C SOFTWARE AND DOCUMENT NOTICE AND LICENSE
```
WilcoFiers: Could you remove the upper case from this? Upper case text is problematic with screen readers.

pages/license.md
```diff
@@ -2,60 +2,24 @@
...
+
+- The full text of this **NOTICE** in a location viewable to users of the redistributed or derivative work.
+- Any pre-existing intellectual property disclaimers, notices, or terms and conditions. If none exist, the [W3C](https://www.w3.org/) Software and Document Short Notice should be included.
+- Notice of any changes or modifications, through a copyright statement on the new code or document such as "This software or document includes material copied from or derived from [title and URI of the W3C document]. Copyright © [YEAR] W3C® (MIT, ERCIM, Keio, Beihang)."
+
+### Disclaimer
+
+THIS WORK IS PROVIDED "AS IS," AND COPYRIGHT HOLDERS MAKE NO REPRESENTATIONS OR WARRANTIES, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO, WARRANTIES OF MERCHANTABILITY OR FITNESS FOR ANY PARTICULAR PURPOSE OR THAT THE USE OF THE SOFTWARE OR DOCUMENT WILL NOT INFRINGE ANY THIRD PARTY PATENTS, COPYRIGHTS, TRADEMARKS OR OTHER RIGHTS.
```
WilcoFiers: Same here. I suspect this needs to be upper case, but then let's use CSS for that.

jkodu:
@nitedog this is not just a link, I have copied over the text.

nitedog:
@jkodu sorry if I was not clear -- I'm proposing to simply link to the license rather than to copy and paste the text. What is the reason for copying the license text? This creates confusion and headache in my view (people have to actually compare if it is the same license text etc.).

jkodu:
@nitedog 

The reason for copy/ pasting the text is that, if and when the original license changes, it does not change the license on this repo automagically. 

Any license changes will be done via a PR, which will help notify and communicate of such changes.

PR #607, opened 2019-06-13
Map ARIA rules to ARIA instead of WCAG 4.1.2

WilcoFiers:
This proposal was discussed during today's ACT-R call. In order to help the ACT Task Force complete their exit criteria on time, and to resolve one of our long standing discussions, I'm updating our three ARIA rules, map them to the ARIA specification, instead of mapping them to WCAG SC 4.1.2.

Because this change is needed by the ACT Task Force, we will are going to skip the standard review stage and go directly to a **shortened** CFC. Please let me know if you have any concerns. This CFC closes on Friday June 21st 2019.

I am working on a proposal for new ARIA rules that will map to WCAG, which will have fewer assumptions.

---

**When creating PR:**

**After creating PR:**

# How to Review And Approve


_rules/aria-attr-valid-5f99a7.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
 description: |
   This rule checks that each aria- attribute specified is defined in ARIA 1.1
 accessibility_requirements:
-  wcag20:4.1.2: # Name, Role, Value (A)
+  aria11:host_general_attrs:
```
annethyme: I don't understand why we would link this rule to a section that has requirements for host languages. I think we need to find requirements for authors...

_rules/aria-state-or-property-allowed-5c01ea.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
 description: |
   This rule checks that WAI-ARIA states or properties are allowed for the element they are specified on.
 accessibility_requirements:
-  wcag20:4.1.2: # Name, Role, Value (A)
+  aria11:supportedState:
+    title: ARIA 1.1, 5.2.3 Supported States and Properties
```
annethyme: When looking at the rule, it tests whether the state/property is EITHER [inherited](https://www.w3.org/TR/wai-aria/#inheritedattributes), [supported](https://www.w3.org/TR/wai-aria/#supportedState), or [required](https://www.w3.org/TR/wai-aria/#requiredState), so I don't think it's enough to link to one of them as the requirement...

mraccess77:
Just some thoughts -- not voting.  Some issues are WCAG violation and some issues are not -- you could say that most do not follow the ARIA 1.1 spec  So while in general I agree mapping to ARIA 1.1 is better -- I believe we would be missing an opportunity to not fail WCAG for cases where the issue is also a WCAG violation.  I guess it would be better to have atomic rules or contextual rules that had more granular mappings.

_rules/aria-attr-valid-5f99a7.md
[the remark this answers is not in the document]
WilcoFiers: No, we don't. These are local keys that we can define to be whatever we need them to be. If in 300 years we need to change them, we can. It's just a shorthand to the ARIA 1.1 URL.

_rules/aria-attr-valid-5f99a7.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
...
+    title: ARIA 1.1, 7.2 State and Property Attributes
```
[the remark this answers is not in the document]
WilcoFiers: We can infer that information from the identifier. In the spirit of not asking for information we already know, I'm not in favour. That just seems like another way someone can get things wrong and slow things down. If you want to discuss, please open a separate issue.

_rules/aria-state-or-property-allowed-5c01ea.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
 description: |
```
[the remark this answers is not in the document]
WilcoFiers: This is out of scope for this PR. Can you open a separate issue or PR please?

_rules/aria-state-or-property-allowed-5c01ea.md
WilcoFiers: I disagree. It seems to me that inherited properties, and required properties are a subset of supported properties. If you have a better suggestion that would be appreciated.

_rules/aria-state-or-property-has-valid-value-6a7281.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
 description: |
```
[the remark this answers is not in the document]
WilcoFiers: Please open separate issues or PRs for these.

_rules/aria-state-or-property-has-valid-value-6a7281.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
...
   This rule checks that each ARIA state or property has a valid value
 accessibility_requirements:
-  wcag20:4.1.2: # Name, Role, Value (A)
+  aria11:propcharacteristic_value:
+    title: ARIA 1.1, 6.2.4 Value
```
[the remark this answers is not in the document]
WilcoFiers: This is unrelated to the PR, please open an issue or PR for this.

_rules/aria-attr-valid-5f99a7.md
WilcoFiers: ARIA doesn't neatly spell out its requirements like WCAG does. Nor does it have a clear distinction between what is a requirement for a host language, a content author or a "conformance checker" (which I believe is an ARIA test tool). The section this references is about using ARIA attributes in a language, which I think maps quite well to what this rule is about. If you have an alternative I'm happy to change it.

_rules/aria-attr-valid-5f99a7.md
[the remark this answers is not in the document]
annethyme: As a reviewer, I might not know how to infer that information. It requires a lot of background knowledge that not all might have, especially if we start talking about other specs. E.g. how do I find the most recent version of RGAA?

_rules/aria-state-or-property-allowed-5c01ea.md
annethyme: @WilcoFiers , @kasperisager says you are right.

annethyme: ... but we strongly suggest we link to "https://www.w3.org/TR/wai-aria-1.1/#states_and_properties" instead, where the whole thing is defined.

_rules/aria-state-or-property-has-valid-value-6a7281.md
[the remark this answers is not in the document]
annethyme: Link to this one instead: https://www.w3.org/TR/wai-aria-1.1/#state_prop_values

And we need the assumption (see above) for the mapping to make sense.

_rules/aria-attr-valid-5f99a7.md
annethyme: (Comment from @kasperisager is on its way)

_rules/aria-attr-valid-5f99a7.md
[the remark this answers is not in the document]
annethyme: Will open seperate issue for this.

_rules/aria-state-or-property-allowed-5c01ea.md
kasperisager: Come to think of it, I'd actually suggest we associate the rule with https://www.w3.org/TR/wai-aria-1.1/#state_property_processing given that it defines a "**MUST**" author conformance requirement:

> However, authors **MUST** only use non-global states and properties on elements with a role supporting the state or property; either defined as an explicit WAI-ARIA role, or as defined by the host language implicit WAI-ARIA semantic matching an appropriate WAI-ARIA role.

_rules/aria-attr-valid-5f99a7.md
kasperisager: So I did some digging and perhaps a simplified example might illustrate my findings and concerns. Imagine that ARIA defines the concept of tools and the process of using a screwdriver to put in a screw. What the rule in question then tries to do is to consider it an error for authors to use a hammer to put in a nail because ARIA doesn't specify "hammer" as a tool. However, ARIA also doesn't specify that using tools other than the ones defined by ARIA is an issue. Yes, it is quite a glaring issue if one attempts to use a flemgrestel for putting in a dulygoo, or use something that looks like a screwdriver, but isn't quite, for putting in a screw. 

The fact remains though that nowhere in ARIA is stated as a "**MUST**" author conformance requirement that only `aria-*` attributes defined by ARIA be used. In fact, Core AAM even specifies that user agents **SHOULD** allow authors to use `aria-*` attributes not defined by ARIA and map these to accessibility APIs regardless, for the purpose of forward compatibility (https://www.w3.org/TR/core-aam-1.1/#statePropertyMappingGeneralRules).

In summary, I don't see how we can list ARIA as an accessibility requirement for this rule. It's not that I don't want us to; I simply cannot justify it. For a rule such as "ARIA state or property allowed" it's more straight forward in light of https://github.com/act-rules/act-rules.github.io/pull/607#discussion_r298135912.

Also, this...

> Nor does it have a clear distinction between what is a requirement for a host language, a content author or a "conformance checker" (which I believe is an ARIA test tool).

...I don't quite agree with. ARIA has user agent, host language, and author conformance requirements littered throughout ("host languages **MUST**", "user agents **MUST**", and "authors **MUST**") and makes it clear that conformance checkers must issue errors when author requirements aren't met.

jkodu:
@WilcoFiers, a dependent PR for actually showing these PR's on the site is here - https://github.com/act-rules/act-rules.github.io/pull/639

I branched from this branch & updated the site code to adapt to these mappings. 

Both of these will have to be merged in succession.

_rules/aria-state-or-property-allowed-5c01ea.md
```diff
@@ -5,11 +5,12 @@ rule_type: atomic
...
   This rule checks that WAI-ARIA states or properties are allowed for the element they are specified on.
 accessibility_requirements:
-  wcag20:4.1.2: # Name, Role, Value (A)
+  aria11:state_property_processing:
+    title: ARIA 1.1, 7.6 State and Property Attribute Processing
```
annethyme: Approve that the following from the references WAI-ARIA section sets the requirement for this rule: 

"Global states and properties are supported on any element in the host language. However, authors MUST only use non-global states and properties on elements with a role supporting the state or property; either defined as an explicit WAI-ARIA role, or as defined by the host language implicit WAI-ARIA semantic matching an appropriate WAI-ARIA role."

_rules/aria-state-or-property-has-valid-value-6a7281.md
```diff
@@ -37,9 +38,7 @@ For value type `URI`, this rule does not require that the destination URI exists
...
-  **Note:** For example, anything referenced through `aria-labelledby` does not have to be [included in the accessibility tree](#included-in-the-accessibility-tree) in order for it to become part of the [accessible name](#accessible-name).
-- The ARIA `state` or `property` is being used to comply to WCAG.
+This rule catches values that are undefined in WAI-ARIA, and where the resulting behaviour in user agents are also undefined in WAI-ARIA. This might lead to accessibility issues, if the intention was to use behaviour defined in WAI-ARIA. When values are used that do not have a defined behaviour in WAI-ARIA, the HTML/SVG specification decides what default values should be used, since it is defined here what should happen when an invalid value is used for an attribute. If the default value for invalid attribute values happens to match the author's intention for the value, there will not be an accessibility issue.
```
annethyme: Don't know if this has the right format for an assumption. I am always told by others to rewrite my assumptions in a more assumption-like manner, so I don't even know what advice to give for changes...

_rules/aria-attr-valid-5f99a7.md
WilcoFiers: I created an issue here. You've convinced me. I think we can solve this another way. I'll take this file change from the PR.

_rules/aria-state-or-property-has-valid-value-6a7281.md
WilcoFiers: I think this is fine. TBH, I think our assumptions aren't all that great. That's what me and Emma met about last week. They are hard to do well.

PR #586, opened 2019-06-04
fix: Make "manual" rules consistent with their SC

WilcoFiers:
This work is part of ##546, to show that we can update our test cases without much loss of coverage, to be consistent with WCAG success criteria.

## Guidance for the PR (pull request) creator

**When creating PR:**
- [ ] Make sure you requesting to **pull a issue/feature/bugfix branch** (right side) to the master branch (left side).

**After creating PR:**
- [ ] Add yourself (and co-authors) as "Assignees" for PR
- [ ] Add label to indicate if it's a `Rule`, `Definition` or `Chore` (more to the administrative side)
- [ ] Add relevant project (e.g. "Q3 2018 Status") to PR
- [ ] OPTIONAL: If you want anyone in particular to review your pull request, assign them as "Reviewers".
- [ ] Close the issue that the PR resolves (and make sure the issue is referenced in the top of this comment)

# How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

Issue #546, opened 2019-05-08
Passing test cases need to meet the accessibility requirement


WilcoFiers:
As part of #486, I've worked on creating a clearer definition of what an implementation is. One problem I came across when looking at this is that some of our test cases don't work if I set myself the following requirement:

> If `rule A` is a superset of `rule B`, and `implementation A` is an implementation of `rule A`, than `implementation A` is also an implementation of `rule B`.

This is important, because it makes the scope of a test irrelevant to whether or not a rule has been implemented. This is particularly common with expert evaluations. An export does not test whether or not a page has a language attribute independent of them testing if the language attribute exists, or if it is correct. More often than not this is captured in a single test procedure.

Our test cases today don't work like that. For example, a `pass` test case for "HTML has lang attribute" is this:

```html
<html xml:lang="xyz"></html>
```

Even though this test case passes the rule, it fails the accessibility requirement. I think these kinds of test cases need to be removed, possibly inverted as failures, or moved to other rules entirely. More generally, all `passed` and `inapplicable` test cases have to satisfy the accessibility requirements.

Additionally, I think we need to change some of the `failed` test cases so that the only thing in the example that doesn't satisfy the accessibility requirement is something that also fails the rule. For example the following failed example for "ARIA state or property has valid value":

```html
<div role="button" aria-expanded="mixed"></div>
```

Has to be changed to give the button an accessible name:

```html
<div role="button" aria-expanded="mixed">Expand</div>
```

### Pros

1. This makes implementations about whether or not an implementor has the same interpretation of the accessibility requirement, rather than it being about having the exact same scope in it rules.

2. It means adopting rules could be far less work. Rather than having to implement all new rules, more likely is that existing rules can be updated.

3. It makes for better examples, as we are never saying something "passed", that doesn't satisfy the underlying accessibility requirement

### Cons

1. The test cases will be more complex, and less atomic

2. There are fewer guarantees that an implementor isn't doing "additional" testing in its implementation. This can partially be mitigated by adding more failure test cases.

Jym77:
I agree that giving "wrong" examples is not really a good idea. In addition to implementations, we can have developers looking at the ACT-Rules (*e.g.* because their tool spotted a failure and pointed here for documentation) and believing that what they see here is "the thing to do" even if it's not.

[...]

PR #615, opened 2019-06-18
fix: Rule 6a7281 - Change tests cases to be consistent with their SC

audreymaniez:
This work is part of #546, to change examples and make then fit the SC, not only the rule. See related PR #586, #587 for similar changes. 

**When creating PR:**

**After creating PR:**

# How to Review And Approve


_rules/aria-state-or-property-has-valid-value-6a7281.md
```diff
@@ -123,26 +124,18 @@ _There are no major accessibility support issues known for this rule._
 `aria-placeholder` property with valid string value
 
 ```html
-<div role="searchbox" aria-placeholder="MM-DD-YYYY">MM-DD-YYYY</div>
+<div role="searchbox" aria-placeholder="MM-DD-YYYY" aria-label="Your birthdate">MM-DD-YYYY</div>
 ```
 
 #### Passed Example 9
```
audreymaniez: Delete the former example 9 because the attribute is not allowed on the element

_rules/aria-state-or-property-has-valid-value-6a7281.md
```diff
@@ -275,15 +269,7 @@ Element has ARIA role, but no ARIA states or properties
 `aria-checked` state with empty value
 
 ```html
```
audreymaniez: Remove the former inapplicable 4 because it has no accessible name

_rules/aria-state-or-property-has-valid-value-6a7281.md
WilcoFiers: Should we just change the role then?

audreymaniez: The test case 9 was illustrating the fact that an attribute not allowed with a certain role, doesn't make the rule failed. Changing the role will juste give us a similar example to what we already have in the rule I think.

PR #616, opened 2019-06-18
chore: add test to validate html code snippets & examples

jkodu:
> Note: looking for opinion here.

Added a test that checks all example code snippets on each rule against a HTML validator. Believe this will help writing valid test cases.

See https://github.com/act-rules/act-rules.github.io/pull/616/files#diff-021ac076b545925359e17fa319b1afadR3 for more information.

Also, this PR has a lot of failing test cases as a consequence of adding this validator, which I have not fixed until I get a go ahead on this concept.

**Back story:**
We identify and fix a lot of test cases during implementation phase. This can potentially help identify issues in the rule authoring phase.


jkodu:
There are 342 failing examples. I do not think this PR should handle fixing all of them.

jkodu:
Closing this PR. Have to investigate a better plugin for html validation. This one does not allow to ignore certain rules.

PR #619, opened 2019-06-21
Fix: Rule  9eb3f6 - fix passed test case 2 

audreymaniez:
Change the src of the image on the passed example 2 so the image is rendered when playing the example preview in separate html page.

Rule in question: https://act-rules.github.io/rules/9eb3f6

Closes issue: NA


Brynanders:
@audreymaniez which PR is this for? I couldn't find 9eb3f

audreymaniez:
Yes I ommit the 6 at the end : 9eb3f6

That is the rule : filename-is-accessible-name-9eb3f6.md

jkodu:
@Brynanders @audreymaniez 

Seems to be competing PR to this one - https://github.com/act-rules/act-rules.github.io/pull/619

So closing this.

Issue #609, opened 2019-06-14
Buttons have an accessible name Passed example 4


Aniort:
in Passed Example 4 of rule documentation "buttons have an accessible name", the code of the example displayed is : 
`<span role="button" aria-label="My button"></button>`
it should be : 
`<span role="button" aria-label="My button"></span>`
as in the test case html page https://act-rules.github.io/testcases/97a4e1/36c92faf3ffc34dc7a5820bdefdf5d534fb60e49.html

jkodu:
PR in place, closing issue.

PR #630, opened 2019-06-26
fix: HTML snippet in test case example for rule "Button has name" (97a4e1)

jkodu:
Closes issue:
- https://github.com/act-rules/act-rules.github.io/issues/609


danistr:
This is also being fixes in PR #631

jkodu:
@danistr 
This specifically addresses this issue where as #631 solves for the larger good by authoring test for catching errors like this as and when it happens.

They should not conflict each other.

Issue #553, opened 2019-05-15
Additional validations for rules


jkodu:
Given, that we have some tests (https://github.com/act-rules/act-rules.github.io/pull/551) to validate the various sections of the rules (format & content), below are some enhancements to the tests that can be added.

- [x] Ensure `authors` listed in `frontmatter` exist in contributors list in `package.json`.
- [x] Check order of headings in `rules`.
- [x] Check if all used `glossary` items, exist with in `pages/glossary` directory.

jkodu:
@annethyme 
- Feel free to expand on the above list, for additional validations that can be done to the rules.

Initial PR is here - https://github.com/act-rules/act-rules.github.io/pull/551

jkodu:
Multiple tests in place, here is another https://github.com/act-rules/act-rules.github.io/pull/634. Closing.

PR #604, opened 2019-06-12
chore: Generate and display implementation metrics

jkodu:
> Note: will merge this PR after one approval, given most of them are code changes.

This PR does the below:
- Parses & generates metrics from all implementations submitted by various contributor.
- Shows implementation metrics at various parts of the site.
- Refactors meta data generation, outside of gatsby createNode hooks
	- generation of testcases specified in rules.
	- generation of glossary usages in rules.
	- generation of implementation metrics.

Because of the large nature of the PR. I have deployed this branch to `netlify` to see how implementations are depicted on the site.
- https://eloquent-lumiere-13f4f9.netlify.com/pages/implementations/overview/
- https://eloquent-lumiere-13f4f9.netlify.com/rules/
- https://eloquent-lumiere-13f4f9.netlify.com/rules/73f2c2#implementation-metrics

> Note: I will remove this netlify hook after this branch is merged.

PR #634, opened 2019-06-27
fix: test if definition for  all referenced glossary terms exist

jkodu:
Added a test to check if content/ definition actually exists for a used glossary term in rules. In other words to check if a glossary referred actually exists as a file in `/pages/glossary/*.md`

@annethyme @Jym77 - could you help fix up the missing glossary items?

Failed tests report:
```
    glossary missing for #implicit-role, usages below:
    ┌─────────┬───────────────────────────────────────────┬────────────────┐
    │ (index) │                   name                    │      slug      │
    ├─────────┼───────────────────────────────────────────┼────────────────┤
    │    0    │ 'Role has required states and properties' │ 'rules/4e8ab6' │
    └─────────┴───────────────────────────────────────────┴────────────────┘

    glossary missing for #non-empty, usages below:
    ┌─────────┬───────────────────────────────────┬────────────────┐
    │ (index) │               name                │      slug      │
    ├─────────┼───────────────────────────────────┼────────────────┤
    │    0    │ 'Validity of HTML Lang attribute' │ 'rules/bf051a' │
    └─────────┴───────────────────────────────────┴────────────────┘
```

Closes issue: 
- Partially closes https://github.com/act-rules/act-rules.github.io/issues/553


test/rule-glossary.test.js
```diff
@@ -0,0 +1,20 @@
+const globby = require('globby')
+const glossaryUsages = require('./../_data/glossary-usages.json')
```
Jym77: The tests on Circle are failing on that line, not finding the file. Is this supposed to be a new file that you forgot to commit?

jkodu: @Jym77 - darn this PR cannot be merged until the implementation PR gets merged.

The `glossary-usages.json` is dynamically generated on site build & in this branch it gets generated as a part of gatsby hook, but in the PR https://github.com/act-rules/act-rules.github.io/pull/604 this has been moved to a npm script, that runs outside the scope of gatsby build.

I do not wish to commit this file.

Ignore the CI failure. 

I have listed all the missing glossary definitions above. If we work on fixing up the actual missing definitions, that will help.

Jym77: I agree that auto-generated files should not be committed.

But merging a PR that breaks the CI tests kinda destroy the main purpose of CI tests. Especially, if a further PR is breaking the tests for real, we won’t see it immediately since we'll be in the habit that "oh yeah, tests are broken, as usual"….
So in my opinion, either this PR should use the old way of finding the file (and #604 should adapt this script once its merged), or this PR should wait until #604 is merged before being merged.

Also, this PR is going to break tests (the whole point of it being to add a test that fails on missing glossary items to find these missing items). I’m not in favour of merging without either fixing the test (filling the glossary) or ignoring them (commenting them out, or equivalent, until fix is done). We should not get used to see red lights in the checks and consider that as normal…

jkodu: Clarification @Jym77. I never said we are going to merge this PR with failing tests.

I just don't want to repeat the work done on PR here - https://github.com/act-rules/act-rules.github.io/pull/604/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R155, which is why it is best if we merge this PR after merging the other one and resolving conflicts on this.

We never did "oh yeah, tests are broken, as usual", on this project, so please don't assume that.

The intent of this PR is to find the missing glossary terms and fix them & not go around in circles for currently failing test, where we have a roadmap to fix them. Thanks.

WilcoFiers:
To fix the media ones: @jkodu 

- #non-streaming -> #non-streaming-media-element
- #audio -> #audio-output
- #caption -> https://www.w3.org/TR/WCAG21/#dfn-captions

test/rule-glossary.test.js
jkodu: As a temporary measure for now have added the generated `json` file. See - https://github.com/act-rules/act-rules.github.io/pull/634/files#diff-a084b794bc0759e7a6b77810e01874f2R75 and https://github.com/act-rules/act-rules.github.io/pull/634/files#diff-f5cfbc832d01fe2cf6d74079a9346113

Jym77:
Given that (i) our definition of "semantic role" already mention implicit ones; and (ii) that in the very same rule, "explicit semantic role" is already pointing to "semantic role":
I went for the quick fix, pointing "implicit semantic role" to "semantic role".

_rules/role-has-required-states-and-properties-4e8ab6.md
```diff
@@ -18,7 +18,7 @@ authors:
 
 ## Applicability
 
-Any HTML or SVG element that has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#implicit-role) that is identical to the explicit semantic role.
+Any HTML or SVG element that has an [explicit semantic role](#semantic-role), except if the element has an [implicit semantic role](#semantic-role) that is identical to the explicit semantic role.
```
WilcoFiers: This definitely isn't the way to solve it. I think we need a definition for this.

_rules/html-lang-valid-bf051a.md
```diff
@@ -19,11 +19,11 @@ authors:
 
 ## Applicability
 
-The root element of the page, if it is an `html` element with a `lang` and/or `xml:lang` attribute that is not empty ("").
+The root element of the page, if it is an `html` element with a `lang` and/or `xml:lang` attribute that is not [empty](#empty).
 
 ## Expectation
 
-The `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is [non-empty](#non-empty).
+The `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is not [empty](#empty).
```
WilcoFiers: `empty` isn't a term in the glossary either. https://act-rules.github.io/glossary/

jkodu:
Note: All definition references are fixed, waiting for #604 before this is merged.

_rules/html-lang-valid-bf051a.md
Jym77: I’ve added empty to the glossary inside this branch: 
https://github.com/act-rules/act-rules.github.io/pull/634/commits/d35e16878f7022dac6d2261d4de6227b3f9aa4e2

_rules/role-has-required-states-and-properties-4e8ab6.md
Jym77: Then, do we also need a definition for "explicit semantic role" which is used in the very same rule and simply links to "semantic role"?

The definition of semantic role has mention of both implicit and explicit roles, with links to further explanations. I assumed that if it was OK to link "explicit semantic role" to it, then it would also be OK for "implicit semantic role".

_rules/html-lang-valid-bf051a.md
Jym77: @WilcoFiers Actually, there's something deeper with our use of `empty`. I see 69 uses of `empty` or variations (`non-empty`, `not empty`, …) in our rules and only the 2 usages in that rule tried to link to the glossary, the rest being either `empty ("")` or just `empty`… And such a discrepancy doesn’t make sense :-(

And there's also some other places that refers to `not only whitespace` which certainly means that I've got the wrong definition of `empty` over there :-((

Anyway, I feel we should unify the way we use and explain `empty` in the rules: either the inline definition `empty ("")` or linking to some glossary entry. We should not mix the two ways to do it. And we should probably also give the explanation every single time we use the word (or at least once per rule).

Jym77: @WilcoFiers I've removed the incorrect glossary entry that I added previously, and I've updated this rule to use `empty ("")` like we do all over the place.

_rules/role-has-required-states-and-properties-4e8ab6.md
Jym77: @WilcoFiers I do not understand why it is fine to have `[explicit semantic role](#semantic-role)` (as was here before), but it is not fine to have `[implicit semantic role](#semantic-role)` Especially given that the glossary entry for semantic role mention both implicit and explicit roles, with links to more details.

Could you explain why one is OK and the other is not?
Should we rather add definitions for both explicit and implicit roles? (they are going to essentially be the same links that already exist in the definition of "Semantic Role")

test/rule-glossary.test.js
jkodu: Cleaned up the temporary addition, given #604 is merged.

Jym77: Thanks. Sorry for the fuss around it.

jkodu:
@WilcoFiers can you respond on this - https://github.com/act-rules/act-rules.github.io/pull/634#discussion_r300348727

_rules/role-has-required-states-and-properties-4e8ab6.md
WilcoFiers: You're right. We shouldn't have linked explicit semantic role to semantic role either. I'll open an issue for it, as to not block this PR any further.

PR #416, opened 2019-01-30
New rule: Video or audio has no auto-play audio (80f0bf, 4c31df, aaa1bf)

annethyme:
Looking for approval of concept for rule in first round.

Replaces outdated rule: https://github.com/auto-wcag/auto-wcag/blob/master/_rules/outdated/SC1-4-2-audio-control-audio.md

# How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

PR #645, opened 2019-07-03
Add files via upload

Brynanders:
Needed for #416


jkodu:
Merged via - https://github.com/act-rules/act-rules.github.io/pull/646

Issue #588, opened 2019-06-06
Dead link in the rule-template page


audreymaniez:
On the website, [in the rule template page](https://act-rules.github.io/pages/design/rule-template/) , 
the two links "empty atomic rule template" (https://act-rules.github.io/pages/design/rule-template/atomic-template-empty.html) and "rule design"  (https://act-rules.github.io/pages/design/rule-template/rule-design.html) leads to 404.

I suppose: 
-  "empty atomic rule template" should be an anchor to the template available below and so the url be `#atomic-rule-template`
- 'rule design" should lead to this page https://act-rules.github.io/pages/design/rule-design/

jkodu:
PR in place https://github.com/act-rules/act-rules.github.io/pull/649, closing issue.

PR #649, opened 2019-07-03
chore: fix broken links in rule template page

jkodu:
Fixes broken links in `rule template` page.

Closes issue: 
- https://github.com/act-rules/act-rules.github.io/issues/588


pages/design/rule-template.md
```diff
@@ -117,7 +117,7 @@ Description...
 
 Composite rules are rules that take results from different rules and through some logic come to a single result. For example: SC 1.2.3 allows video to pass with either a transcript, an audio description, or if it is a media alternative. Each of these would be atomic rules, that are used in a composite rule. The composite rule than describes that at least one of these must pass for the composite rule to pass.
 
-For more about composite rules, see the [ACT Rules Format](https://www.w3.org/TR/act-rules-format/#composed-rules). To create a composite rule, use the [empty composite rule template](./composite-template-empty.html).
+For more about composite rules, see the [ACT Rules Format](https://www.w3.org/TR/act-rules-format/#composed-rules). To create a composite rule, use the [empty composite rule template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/composite-template-empty.md).
```
Jym77: Same remark: shouldn't we use `/pages/design/composite-template-empty/`?

pages/design/rule-template.md
```diff
@@ -4,7 +4,7 @@ title: Rule Template
 
 The rule template contains a plain language description of the rule, some background information, and a list of all assumptions. The test procedure is defined by the selector, a number of steps and a description of the possible outcomes.
 
-Use the [empty atomic rule template](atomic-template-empty.html) to create new rule. When creating a new rule, first read [rule design](rule-design.html).
+Use the [empty atomic rule template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/atomic-template-empty.md) to create new rule. When creating a new rule, first read [rule design](/pages/design/rule-design/).
```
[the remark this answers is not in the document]
jkodu: Given the templates are actual markdown files with frontmatter, a rendered page will not show the frontmatter & is no way useful for someone who wants to author a new rule. It is only beneficial if they are able to copy the template and use it elsehwere. 

I believe referring to the actual raw content is a better way to go about this.

pages/design/rule-template.md
jkodu: See above comment, additionally I have updated the GitHub templates to refer to the same (one source of truth).

See - https://github.com/act-rules/act-rules.github.io/pull/648/files#diff-1a1c3dd142f76a5fad803a0c52839881R5

pages/design/rule-template.md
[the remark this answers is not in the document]
Jym77: OK. I guess I missed the difference between the template (where raw content is what we need to share) and the design file (where we want to actually render it)…

Jym77: The `pages/design/rule-design` link seems broken (at least when I try to view this file and click on this link, I get a 404) and might need to be `/pages/design/rule-design.md` instead. (but I’m not sure how much magic is happening in the back…)

jkodu: @Jym77 - this is what this will get resolved to https://act-rules.github.io/pages/design/rule-design/

All markdown content is converted to html pages & this is one such page.

Jym77: OK. I suspected some of it, but hadn’t had time to look into the code to figure all that out.

PR #667, opened 2019-07-05
Various typos

Jym77:
A few more typos I've caught while reading rules…


_rules/image-button-has-name-59796f.md
```diff
@@ -125,7 +125,7 @@ Image button with aria-labelledby that does not reference an id that exists in t
 
 #### Failed example 4
 
-Accessible name is not only whitespace.
+Accessible name is not whitespace.
```
WilcoFiers: Did you mean for this to say "Accessible name is only whitespace"?

Jym77: Yep. Good catch…

PR #678, opened 2019-07-16
[WIP] New Rule: SVG-with-non-text-content-has-name-7d6735.md

Brynanders:
Related to #419 

This rule was the result of @Jym77 and @Brynanders trying to figure out a roadmap for SVG image rules. We think that we need two or more possible rules:
#419 to target the use of explicit roles to intentionally including elements in the a11y tree)
Other rules (like this one) to target implicit or unintentionally included elements in the a11y tree.


_rules/SVG-with-non-text-content-has-name-7d6735.md
```diff
@@ -0,0 +1,91 @@
+---
+id: 7d6735
+name: SVG with non-text content has accessible name
+rule_type: atomic
+description: |
+   This rule checks that <evg> element that has content included in the accessibility tree has an accessible name.
```
annethyme: <evg> --> <svg>

WilcoFiers:
As agreed with @Brynanders, closing this due to inactivity.

PR #688, opened 2019-07-17
chore: run prettier

jkodu:
- run prettier

Closes issue(s): 
- NA

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


WilcoFiers:
@Jym77 I don't think Prettier wraps HTML attributes for short tags like `<a href="…">Foo</a>`. Only when things get longer does it start wrapping stuff. The line length is set to 80 for it, which I think is a good length given that people review code side by side.

Use of Prettier is more important than you may realise. We identify test cases by hashing the example. If the test case changes, so does the URL, and based on that we can easily see if the test data we have is still complete. Running Prettier on all the examples ensures that everything is consistent, and that we never have to invalidate a bunch of data because someone used a space instead of a tab, or some other meaningless change like that.

Especially for manual tests, ensuring the stability of our test cases is very important. It takes hours to update manual implementation data, we don't want people to have to do that whenever a space or line break is out of place.

Jym77:
> @Jym77 I don't think Prettier wraps HTML attributes for short tags like `<a href="…">Foo</a>`. Only when things get longer does it start wrapping stuff. 

You are entirely right. I should have copied one of the many examples that prettier actually did wrap, instead of trying to condensate it. Let me rephrase my initial comment:
I personally think that
```html
<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/index.html">Contact us</a>
```
is more (human) readable than
```html
<a
	href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/index.html"
	>Contact us</a
>
```
(except for the horizontal lift that github puts because it decided that discussions should only use one third of my screen space…)
I personally have a cognitive overload when mentally parsing lines that start with `>`, or the `</a` at the end of line with its closing `>` on the next line. This makes me, as a human, spend more time reading and understanding the example just because I need some extra thinking.

> The line length is set to 80 for it, which I think is a good length given that people review code side by side.

Do we write examples so they can be easily reviewed, or so they can be easily read on the website and understood by human devs and testers trying to implement the ACT-Rules? ;-)

> Use of Prettier is more important than you may realise. We identify test cases by hashing the example. If the test case changes, so does the URL, and based on that we can easily see if the test data we have is still complete. Running Prettier on all the examples ensures that everything is consistent, and that we never have to invalidate a bunch of data because someone used a space instead of a tab, or some other meaningless change like that.

I've never said that prettier is not important. Of course it is. And my question was to increase the allowed line length (which is indeed not very clear at the start of my initial comment, my bad, the end was more explicit). Maybe 120 characters (which is arguably still OK when viewing the rule in the ACT-Rules web page) is also good and will less aggressively break lines in places where it actually hampers readability. 

> Especially for manual tests, ensuring the stability of our test cases is very important. It takes hours to update manual implementation data, we don't want people to have to do that whenever a space or line break is out of place.

Ironically, this PR is rewriting a huge lot of test cases :-D
(of course, the goal is that they do not change after that)

jkodu:
@Jym77 @WilcoFiers I have re-run prettier by setting `printWidth: 120`. This caters by preventing a lot of line wraps.

I reckon this is a good start 👍

Issue #670, opened 2019-07-11
implicit / explicit semantic role definitions are missing [4e8ab6]


WilcoFiers:
In `Role has required states and properties`, the terms "explicit semantic role" and "implicit semantic role" are used. Both link to the "semantic role" definition. These are two parts of how to compute the semantic role, but they're not equivalent. Those two terms really should have their own definitions, possibly used in the `semantic role` definition.

Jym77:
Closed by #689.

PR #689, opened 2019-07-18
Define implicit/explicit role (4e8ab6, b20e66, c487ae, e086e5, 5c01ea)

Jym77:
Adding a definition of implicit and explicit semantic roles.

Closes issue(s): 
- #670 

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
@WilcoFiers I tried to add a definition of implicit/explicit semantic roles, but it feels like I just copy/pasted half of the content of the definition of semantic role in each.

I am not really happy with the result since it duplicates a lot of stuff between definition and that is not very DRY. Yet, I do not think it would be a good idea to remove that stuff from the definition of semantic role since that would cause extra indirections to get to relevant information.

What's your opinion on that? What do we think we should do?

pages/glossary/implicit-role.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Implicit Semantic Role
+key: implicit-role
+---
+
+The Implicit Semantic Role is the default [semantic role](#semantic-role) of each element that is used when no valid role is specified. Elements with no [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute), or with a [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute) containing no valid token, are mapped to their implicit role.
```
annethyme: > when no valid role is specified 

--> 

> when no valid [explicit semantic role](#explicit-role) is specified 

?

pages/glossary/semantic-role.md
```diff
@@ -7,7 +7,7 @@ A semantic role is a semantic association that indicates an object's type. This
 
 [Roles](https://www.w3.org/TR/wai-aria-1.1/#dfn-role) can be [implicit](https://www.w3.org/TR/wai-aria/#implicit_semantics) through the element type or explicit through the [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute).
```
annethyme: explicit -- > [explicit](#explicit-role)

pages/glossary/semantic-role.md
```diff
@@ -7,7 +7,7 @@ A semantic role is a semantic association that indicates an object's type. This
...
 
-The role attribute takes a list of tokens. The semantic role is the first valid role in this list. If none of the tokens are valid, the implicit role will be used instead.
+The [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute) takes a list of tokens. The semantic role is the first valid role in this list. If none of the tokens are valid, the implicit role will be used instead.
```
annethyme: implicit role --> [implicit role](#implicit-role) ?

pages/glossary/explicit-role.md
```diff
@@ -0,0 +1,20 @@
...
+
+Non-abstract roles defined in the following specifications are considered valid:
+
+- [Accessible Rich Internet Applications (WAI-ARIA) 1.1](https://www.w3.org/TR/wai-aria-1.1/)
+- [Digital Publishing WAI-ARIA Module 1.0](https://www.w3.org/TR/dpub-aria-1.0/)
+- [WAI-ARIA Graphics Module](https://www.w3.org/TR/graphics-aria-1.0/)
+
+Other roles may be added as they become available. Not all roles will be supported in all assistive technologies. Testers are encouraged to adjust which roles are allowed according to the [accessibility support base line](https://www.w3.org/TR/WCAG-EM/#step1c). For the purposes of executing test cases in all rules, it should be assumed that all roles are supported by assistive technologies so that none of the roles fail due to lack of accessibility support.
+
+#### Acccessibility Support
```
jkodu: given the heading of each glossary terms are rendered as `h2` in the website, it would make sense that these are `h3` so we do not skip headings.

Jym77: That is a `####` in the current def of semantic role (which is why I copied it that way).
But yes, I agree that we shouldn't skip heading and use only `###`.

Jym77: Actually, the problem comes from the rule pages where "Referenced glossary" is a `h2`, and each glossary item is a `h3`, thus sub-headings like this one need to be `h4`.

It is annoying that glossary terms are rendered sometimes as `h2` and sometimes as `h3`…

Jym77: @jkodu do you have an opinion on the `h2`/`h3` inconsistence of glossary items I'm explaining in this discussion?

annethyme: For the sake of this PR, could we here do whatever is done (mostly) in other glossary terms, and then split out the task of creating consistency between the Glossary page and the embedded Glossary on the Rules pages?

pages/glossary/semantic-role.md
```diff
@@ -3,22 +3,6 @@ title: Semantic Role
...
-
-Other roles may be added as they become available. Not all roles will be supported in all assistive technologies. Testers are encouraged to adjust which roles are allowed according to the [accessibility support base line](https://www.w3.org/TR/WCAG-EM/#step1c). For the purposes of executing test cases in all rules, it should be assumed that all roles are supported by assistive technologies so that none of the roles fail due to lack of accessibility support.
-
-**Note**: For HTML elements the implicit roles are documented in [ARIA in HTML](https://www.w3.org/TR/html-aria/).
-
-#### Acccessibility Support
-
-Some browsers and assistive technologies treat the tokens of the `role` attribute as case-sensitive. Unless lowercase letters are used for the value of the `role` attribute, not all user agents will be able to interpret the tokens correctly. ARIA in HTML also specifies that [authors must use lowercase letters for the `role` and `aria-*` attributes](https://www.w3.org/TR/html-aria/#case-sensitivity).
+The semantic role is of an element is its [explicit semantic role](#explicit-role) if it has any, otherwise, the [implicit semantic role] (#implicit-role) is used.
```
carlosapaduarte: The semantic role is of > The semantic role of

Remove the space between ] and ( in [implicit semantic role] (#implicit-role)

pages/glossary/explicit-role.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Explicit Semantic Role
+key: explicit-role
+---
+
+The explicit semantic role is the [semantic role](#semantic-role) of an element set by its [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute).
```
carlosapaduarte: It's not linking to correct place. Was this what you intended? 
https://html.spec.whatwg.org/multipage/infrastructure.html#attr-aria-role

pages/glossary/explicit-role.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Explicit Semantic Role
+key: explicit-role
+---
+
+The explicit semantic role is the [semantic role](#semantic-role) of an element set by its [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute).
+
+The [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute) takes a list of tokens. The explicit semantic role is the first valid role in this list. If none of the tokens are valid, the [implicit semantic role](#implicit-role) will be used instead.
```
carlosapaduarte: same for the link

pages/glossary/implicit-role.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Implicit Semantic Role
+key: implicit-role
+---
+
+The Implicit Semantic Role is the default [semantic role](#semantic-role) of each element that is used when no valid [explicit semantic role](#explicit-role) is specified. Elements with no [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute), or with a [role attribute](https://www.w3.org/TR/html/dom.html#aria-role-attribute) containing no valid token, are mapped to their implicit role.
```
carlosapaduarte: same problem with the role attribute link

carlosapaduarte: I'm not sure about the use of "default" in the definition. I would prefer to remove it from the sentence.

pages/glossary/explicit-role.md
jkodu: Lets get this merged here & will solve for heading inconsistencies in a different PR

pages/glossary/explicit-role.md
Jym77: This was indeed not the correct link. The one you suggest is just a redirection to WAI-ARIA. Digging into it, I do not find a precise definition of the `role` *attribute* (rather than the *concept* of roles), the best I have is https://www.w3.org/TR/wai-aria-1.2/#introroles which is itself mostly a redirection to https://www.w3.org/TR/role-attribute/

I guess I'll go for https://www.w3.org/TR/role-attribute/ then…

PR #692, opened 2019-07-18
Add Accessibility Support about browsers not respecting empty alt (23a2a8)

Jym77:
When encountering a ``,
* Chrome incorrectly puts it in the accessibility tree with a role of `img`
* Firefox incorrectly puts it in the accessibility tree with a role of `graphic`.

This should be mentioned for this rule…


Jym77:
Yes, that's what I mean.
For example, for Chrome, `` has a role of `img`, not `presentation`.

Jym77:
@WilcoFiers @jkodu I tried to rephrase it, is it OK now?

_rules/image-has-name-23a2a8.md
```diff
@@ -36,6 +36,8 @@ _There are currently no assumptions._
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name](#accessible-name).
 
+There are several popular browsers that do not treat images with empty `alt` attribute as having a role of `presentation` but instead add the `img` element to the accessibility tree with a role of either `img` or `graphic`.
```
jkodu: 👍 

Small editorial nitpick, in some rules when there are more than one accessibility support concerns, we should be list them, like so.
- support sentence 1
- support sentence 2

It may be worth standardising this across the board.

PR #693, opened 2019-07-18
New Rule: Links with identical names and context serve equivalent purpose (fd3a94)

annethyme:
Creating the 2.4.4 Link Purpose (In Context) (Level A) version of this published rule: 
[Links with identical accessible names serve equivalent purpose](https://act-rules.github.io/rules/b20e66) (Level AAA)

----

## How to Review And Approve


annethyme:
For the ACT Rules CG meeting agenda on August 8: 

Is this a rule that we in the CG want to move on with?
How does the definition of "programmatically determined link context" look to you? 

Please feel free to leave comments here before the meeting.

annethyme:
Agreed that we should proceed with this rule.

Test cases need to be updated, since they do not match this rule, but are copy/pasted directly out of the AAA rule.

Feel free to review everything but the test cases :)

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,298 @@
...
+  - Anne Thyme Nørregaard
+---
+
+## Applicability
+
+This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names](#accessible-name) that do not only consist of [whitespace](#whitespace), and have [matching](#matching-characters) [programmatically determined link contexts](#programmatically-determined-link-context).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name](#accessible-name) and [matching](#matching-characters) [programmatically determined link contexts](#programmatically-determined-link-context).
+
+## Expectation
```
DagfinnRomen: According to SC 2.4.4 the purpose of each link can be determined from the link text alone or from the link text together with its programmatically determined link context.

I do not think the test rule tests this. According to the test rule, links with matching names should resolve to the same or equivalent resource. If the linked resource(s) have a different purpose than what is described in the link names, this would pass the test rule, but fail SC 2.4.4.

According to the SC 2.4.4 understanding article "it is a best practice for links with the same destination to have consistent text (and this is a requirement per Success Criterion 3.2.4 for pages in a set). It is also a best practice for links with different purposes and destinations to have different link text."

In my opinion, this test rule is more relevant for SC 3.2.4 Consistent Identification. Either way, SC 3.2.4 requires the identification to be consistent, but not necessarily identical.

pages/glossary/programatically-determined-link-context.md
```diff
@@ -0,0 +1,15 @@
+--
+title: Programmatically Determined Link Context
```
DagfinnRomen: Good definition!

Jym77:
This rule proposal does not map to SC 2.4.4 See https://github.com/act-rules/act-rules.github.io/issues/968#issuecomment-539886057

carlosapaduarte:
Definition of programmatically determined link context is being done in #555

Jym77:
[Understanding 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) says:
> It is also a best practice for links with different purposes and destinations to have different link text.

Hence, this rule proposal is testing a best practice, not a WCAG requirement.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
ShadowBB: But this rule also allows for "equivalent resources" not just "same resource", so it goes beyond the best practice and is actually a good basis to fail the SC on if it doesn't link to the same or equivalent resource despite having the same accessible name and context right?

ShadowBB:
> 
> 
> This rule proposal does not map to SC 2.4.4 See [#968 (comment)](https://github.com/act-rules/act-rules.github.io/issues/968#issuecomment-539886057)

Jym77:
> > This rule proposal does not map to SC 2.4.4 See [#968 (comment)](https://github.com/act-rules/act-rules.github.io/issues/968#issuecomment-539886057)
> 
> I think it does because of the assumption and the fact that it is not just following the best practice but also allows for equivalent resources.
> 
> > [Understanding 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) says:
> > > It is also a best practice for links with different purposes and destinations to have different link text.
> > 
> > 
> > Hence, this rule proposal is testing a best practice, not a WCAG requirement.
> 
> So I think it does test for a WCAG requirement.

I have to disagree because the best practice mention "different purposes", not just "different destinations".

I read the best practice as an implication "different purposes => different names" which is logically equivalent (by contraposition) to "same names => same purposes" that the rule is testing.

ShadowBB:
> 
> 
> > > This rule proposal does not map to SC 2.4.4 See [#968 (comment)](https://github.com/act-rules/act-rules.github.io/issues/968#issuecomment-539886057)
> > 
> > 
> > I think it does because of the assumption and the fact that it is not just following the best practice but also allows for equivalent resources.
> > > [Understanding 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) says:
> > > > It is also a best practice for links with different purposes and destinations to have different link text.
> > > 
> > > 
> > > Hence, this rule proposal is testing a best practice, not a WCAG requirement.
> > 
> > 
> > So I think it does test for a WCAG requirement.
> 
> I have to disagree because the best practice mention "different purposes", not just "different destinations".
> 
> I read the best practice as an implication "different purposes => different names" which is logically equivalent (by contraposition) to "same names => same purposes" that the rule is testing.

After reading it again I come to the conclusion that actually makes more sense than my interpretation. I think I have to admit you have convinced me. :)

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,298 @@
...
+			Follow us
+		</text>
+	</a>
+</svg>
+```
+
+### Failed
+
+#### Failed Example 1
+
```
patrickhlauke: aren't all the listed failed examples exempt since those links would be confusing/ambiguous to *all* users? e.g. how would a sighted mouse user not be confused by them versus an AT user?

Jym77: That rule is Work in Progress, so anything in it might need to change ;-)
(and locally, it's waiting for #555 to be finished so that we can reuse the definition of context for the links…)

carlosapaduarte:
Rule has been updated to include the definition of programmatically determined link context. Ready for reviews.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,429 @@
...
+---
+
+## Applicability
+
+This rule applies to any set of any two or more HTML or SVG elements which
+
+- have the [semantic role][] of `link`, or a role that inherits from the `link` role; and
+- are in the same [web page (HTML)][]; and
+- are [included in an accessibility tree][included in the accessibility tree]; and
+- that have [matching][] [accessible names][accessible name] that are not empty (`""`); and
```
Jym77:
```suggestion
- have [matching][] [accessible names][accessible name] that are not empty (`""`); and
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,429 @@
...
+- that have the same [programmatically determined link context][].
```
Jym77:
```suggestion
- have the same [programmatically determined link context][].
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,429 @@
...
+
+```html
+<html lang="en">
+	<p>
+		<span
+			role="link"
+			tabindex="0"
+			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/index.html'"
+		>
+			Link text
```
Jym77: We should use a descriptive name for a Passed Example.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,429 @@
...
+
+```html
+<html lang="en">
+	<p>
+		<span
+			role="link"
+			tabindex="0"
+			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html'"
+		>
+			Link text
```
Jym77: We should use descriptive names.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,429 @@
...
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/redirect1.html">Contact us</a>
+	</p>
+</html>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+These `a` and `area` elements have no `href` attribute.
```
Jym77:
```suggestion
These `a` and `area` elements do not have a role of `link`.
```

carlosapaduarte:
@Jym77 I've added content to the context in the examples. Can you check it again?
@ShadowBB I haven't done any more changes to the rule besides adding the content to the context and small editorials. But you can quickly re-check if you like.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,457 @@
...
+			aria-hidden="true"
+			tabindex="-1"
+			>Contact Us</a
+		>) and get in touch (<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/page2.html"
+			>Contact Us</a
+		>)
+	</p>
+</html>
+```
+
```
Jym77: Note: the first link is actually included in the Accessibility Tree (with a role of `link`) in Chrome who chooses to expose focusable elements over `aria-hidden` (Edge and Firefox will prioritise `aria-hidden`). This is a part of the spec that is actually not specified but is linked to presentational role conflict resolution.

I am working on a PR to try and fix this problem for us. In the meantime, I suggest removing examples with `aria-hidden` on focusable elements (`tabindex=-1` removes it from sequential order, but it is still focusable). Because the behaviour is not specified and browsers disagree (so we can neither follow the spec nor the de facto industry standard).

See #1189 for my work in progress on that topic and links to more discussion…

carlosapaduarte: Removed it

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,438 @@
...
+- are in the same [web page (HTML)][]; and
+- are [included in an accessibility tree][included in the accessibility tree]; and
+- have [matching][] [accessible names][accessible name] that are not empty (`""`); and
+- have the same [programmatically determined link context][].
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
+
+## Expectation
+
+When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource). Resolving the links includes potential redirects, if the redirects happen instantly.
```
WilcoFiers: I think this is better. That second sentence does not add to the expectation, it just clarifies it.

```suggestion
When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource).

**Note**: Resolving the links includes potential redirects, if the redirects happen instantly.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,438 @@
...
+- This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+
+## Accessibility Support
+
+This rule assumes that assistive technologies are exposing all links on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its links, then it is possible for two links to have identical name and context but resolve to different resources without failing [Success Criterion 2.4.4 Link Purpose (In Context)][sc244] (if said links are in separate [documents][document] or [shadow trees][shadow tree])
+
+## Background
+
+- [CSS Scoping Module Level 1 (editor's draft)](https://drafts.csswg.org/css-scoping/)
```
WilcoFiers: How does this related to it? Can we link to the relevant section if this is necessary?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,438 @@
...
+- [Understanding Success Criterion 2.4.4: Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)
+- [HTML Specification - URL parsing](https://html.spec.whatwg.org/#resolving-urls)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+A set of two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context] and link to the [same resource][].
```
WilcoFiers: Can you make sure these example descriptions follow the format outlined here: https://github.com/act-rules/act-rules.github.io/pull/946/files#diff-9772ff397ce17764837b9ae6f1fd71acR142

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,438 @@
...
+### Inapplicable
+
+#### Inapplicable Example 1
+
+These `a` and `area` elements do not have a role of `link`.
+
+```html
+<html lang="en">
+	<a>Link text</a>
+	<area aria-label="Link text" />
```
WilcoFiers: That's two test cases in one. Can you split them out please?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,438 @@
...
+	</p>
+</html>
+```
+
+#### Inapplicable Example 5
+
+These links have empty [accessible names][accessible name].
+
+```html
+<a href="http://facebook.com"></a> <a href="http://twitter.com"></a>
```
WilcoFiers: These are inaccessible links. I don't think these satisfy the SC. We should make sure all inapplicable examples satisfy the SC.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+#### Inapplicable Example 6
+
+These image links have empty [accessible names][accessible name].
+
+```html
+<a href="http://facebook.com"><img src="facebook.jpg"/></a> <a href="http://twitter.com"><img src="twitter.jpg"/></a>
```
WilcoFiers: As above, I don't think these satisfy the SC.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Not needed.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Done

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Not sure I understand what you mean. What is being tested here besides elements that do not have the link role?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Links now satisfy the SC.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Removed this example, given that it was testing the same as the previous.

carlosapaduarte:
@ShadowBB and @Jym77 After your approval, following @WilcoFiers review, I've updated all example descriptions, inapplicable example 5 and removed inapplicable example 6. This should be a quick review for you again.

patrickhlauke:
Coming in a bit out of the blue and very late, admittedly...but as this only just bubbled up in my notifications: how does this take into account the 

> except where the purpose of the link would be ambiguous to users in general

clause of 2.4.4, if at all?

patrickhlauke:
i mean i can see it's mentioned in the assumptions, but then looking at the failed examples I can't help but wonder if at least some of those wouldn't actually count as falling under that exemption (like, for instances, those two links that just read "My University" and go to different locations...how are those not ambiguous to everybody?)

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,435 @@
...
+			</a>
+		</svg>
+	</p>
+</html>
+```
+
+#### Failed Example 6
+
+The two HTML `a` elements with the same [accessible name][] and [context][programmatically determined link context] resolve to the [same resource][] after redirect, but the redirect is not instant.
+
```
patrickhlauke: isn't what counts the end result? not sure i agree with this as a failure

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,435 @@
...
+			>Contact us</a
+		>)
+	</p>
+</html>
+```
+
+#### Failed Example 3
+
+The two HTML `span` elements have an [explicit role][] of link, same [accessible name][] and [context][programmatically determined link context], but lead to resources that offer different content.
+
```
patrickhlauke: this example seems to me to be "ambiguous to all users"

jkodu:
@carlosapaduarte I ignored this HTML validation test - https://github.com/htmlhint/HTMLHint/wiki/attr-lowercase, that should help with passing the case of `viewBox`.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -411,18 +416,10 @@ These `span` elements do not have a [semantic role][] of link.
...
-```
-
-#### Inapplicable Example 6
-
-These image links have empty [accessible names][accessible name].
+The two HTML `a` elements have empty [accessible names][accessible name].
 
 ```html
-<a href="http://facebook.com"><img src="facebook.jpg"/></a> <a href="http://twitter.com"><img src="twitter.jpg"/></a>
+<a href="http://facebook.com"></a> <a href="http://facebook.com"><img src="facebook.jpg"/></a>
```
WilcoFiers: Neither of these are accessible links. Can you make sure this satisfies the SC?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: But we can't tell when the redirect will happen if it does not happen instantly. That's why we have a note stating "Resolving the links includes potential redirects, if the redirects happen instantly."

Imagine a redirect that happens half an hour after the link has been followed. Almost all users will not even perceive that redirect, so they will experience different content and the SC should fail.

(of course we can tell when the redirect happens, but then the issue becomes where do we draw the line between redirects that will be perceived and those that won't)

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Agree that each link can be ambiguous to all users. Updated the text to be similar to what is used in some of the other examples.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: Modified the example so that they are both in the accessibility tree but only one has an accessible name.

carlosapaduarte:
Following @WilcoFiers and @patrickhlauke comments, I've updated a couple of examples. @Jym77 can you do another quick review?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+  - Language
+acknowledgements:
+  authors:
+    - Anne Thyme Nørregaard
```
jkodu: It might be worth moving some authors to previous authors.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+  - Language
+acknowledgments:
+  authors:
+    - Anne Thyme Nørregaard
```
WilcoFiers, line 19: I think we should put Anne as previous author. She has been out of the loop 6 months.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+
+**Note**: Resolving the links includes potential redirects, if the redirects happen instantly.
+
+## Assumptions
+
+- This rule assumes that the purpose of the links with identical [accessible names][accessible name] and [context][programmatically determined link context] would not be ambiguous to users in general, which is the exception mentioned in [Success Criterion 2.4.4 Link Purpose (In Context)][sc244]. If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links, which makes it more of a general user experience concern than an accessibility issue.
+- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of a link can only accurately describe one resource (notably, homonyms alone are not used as link names). Thus, if two or more links have the same [accessible name][] but resolve to different resources, at least one of them does not describe its purpose.
+- This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
```
WilcoFiers, line 50: Language is an input. We don't need to assume we can determine it, if we can't, input is missing and the rule can't be executed. This can be taken out I think.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+- [Understanding Success Criterion 2.4.4: Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)
+- [HTML Specification - URL parsing](https://html.spec.whatwg.org/#resolving-urls)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+These` two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context] and link to the [same resource][].
```
WilcoFiers, line 68:
```suggestion
These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context] and link to the [same resource][].
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+			>Contact us</a
+		>) and get in touch (
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/index.html">Contact us</a>)
+	</p>
+</html>
+```
+
+#### Passed Example 2
+
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], and resolve to the [same resource][] after instant redirect.
```
WilcoFiers, line 83:
```suggestion
These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], and resolve to the [same resource][] after an instant redirect.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+		>) and get in touch (<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/page2.html"
+			>Call us</a
+		>)
+	</p>
+</html>
+```
+
+#### Passed Example 7
+
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], and go to pages that use different layouts but have the same advertised key content.
```
WilcoFiers, line 160:
```suggestion
These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], and go to pages that use different layouts but have the same purpose.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -0,0 +1,431 @@
...
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 2.4.4: Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)
+- [HTML Specification - URL parsing](https://html.spec.whatwg.org/#resolving-urls)
+
+## Test Cases
```
WilcoFiers, line 62: I think you've overdone it a little on the test cases. The test cases aren't there to check if someone knows how to implement the accessible name algorithm. That's not what this rule is / should be about. I think several of these examples can be removed without.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
Jym77: The CG decision was to keep this assumption nonetheless:
https://www.w3.org/2020/01/09-act-r-minutes.html#item04

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: OK. I'm pruning the examples.

carlosapaduarte:
Final call ends April, 22nd

carlosapaduarte:
Final call ended. Merging.

PR #695, opened 2019-07-18
Add many missing links to glossary

Jym77:
Many rules had glossary terms without the links to the actual items.
Presence or absence of links was somewhat inconsistent between rules.


Jym77:
I tried to improve things a bit, notably the rule-writer overload, by using markdown internal references.

jkodu:
Also, how can we stop future pull requests from missing a reference? Can this be caught in tests? We should add tests as a part of this PR if that is the case.

jkodu:
> Action: Based on call on August 8th 2019
@jkodu and @Jym77 to come up with better tests to catch this for the future PR's

_rules/aria-hidden-focus-6cfa84.md
```diff
@@ -42,9 +42,9 @@ There are no major accessibility support issues known for this rule.
 
 ## Background
 
-By adding `aria-hidden="true"` to an element, content authors ensure that assistive technologies will ignore the element. This can be used to hide decorative parts of a web page, such as icon fonts - that are not meant to be read by assistive technologies.
```
daniel-montalvo: there I see that decorative has not been marked as a glossary term. Am I supposed to review all these files in the pull request to see if there are missing links to glossary?

Jym77: Good catch.
That is indeed a somewhat tough question since there can be a lot of items that I missed :-(

I would say that it is a safe bet that I missed some, hopefully not too many, and these will be catch at later point in time (in the same way that we constantly catch typos, *e.g.* I'm constantly submitting some editorial changes for typos/consistency/…), especially once we start adding some testing to this. 

So, my guess is that the important part of the current review is to check that my changes are not accidentally breaking the intend of the rules (due to some reformulation), and that they are correct (*e.g.* not linking "accessible name" to "included in the accessibility tree"). Checking that they are exhaustive is as  time consuming as proofreading a document to perfection and I think that 100% accuracy (whether on these links or on "no typo") is not a realistic goal to achieve with the current resources we have.

…so, I say, check what I’ve done rather than what I haven't done. The second point will slowly improve with time and we'll try and set up some test to make it easier.

Jym77: (and you seem to be commenting on the old file, not on the new one where at least this instance of decorative is correctly linking the glossary)

PR #702, opened 2019-07-24
Editorial changes

Jym77:
Mostly unifying descriptions style ("This rule checks that" + declarative form) and adding backticks in descriptions needing it.


EmmaJP:
Appreciate this already got merged. Thumbs up on the changes.

However, while it might be pedantic, it would be nice to be consistent about punctuation too. Specifically whether or not these phrases end with a full stop. I would lean toward having one, but either way think it should be consistent.

Jym77:
@EmmaJP good point. I'll add some in the next batch of editorial changes…

Issue #182, opened 2018-07-12
Add examples for definitions


jkodu:
Add test cases for definitions

@annethyme - can you elaborate on this, for further discussion.

WilcoFiers:
I found this today. It's a document that uses Gherkin to test implementations of the accessible name. That could be a way for us to start writing tests for definitions:

```
if given
  <div id="ID1">Rich's button</div>
  <input type="button" aria-labelledby="ID1" id="test">
then the accessible name of the element with id of "test" is "Rich's button"
```


https://www.w3.org/wiki/AccName_1.1_Testable_Statements

WilcoFiers:
@kasperisager any thoughts on using Gherkin for tests of definitions?

jkodu:
Priority to add examples is based on usage.

Jym77:
Partial PR in #994. This is probably enough for a start given that the other widely used definitions are not so tricky (_e.g._ "non-streaming video").

Keeping this for now, but won't object closing it…

jkodu:
Myself and @Jym77 agreed, we will add definitions as well/ improve other definitions. So #994 now completes this issue.

Issue #480, opened 2019-04-17
More scalable approach to test cases?


kasperisager:
Test cases are currently written in the same Markdown document as the rules they're testing and are manually assigned consecutive numbers. While this works fine when dealing with a small number of test cases, it quickly becomes unwieldy as the number of test cases increases. To ensure consistent implementations of rules, I think that it is in our best interests to make it easy to write, document, and maintain a large number of test cases for rules. As such, finding a more scalable approach to writing test cases would be great.

Some initial thoughts on requirements:

- Adding a new test case to a rule must not cause any merge conflicts. As such:
  - Identifiers for test cases must be constructed without regard for existing test cases.
  - Each test case must live in its own file named by its identifier.
- The reason for why a test case passes, fails, or is inapplicable must be able to be documented alongside the test case.

WilcoFiers:
Decision: We will add a feature to collapse test cases in the rules page, but keep all test cases in a single markdown file.

PR #717, opened 2019-07-26
[RFC] Examples in glossary

Jym77:
Follow-up to #182.

# Why?
Many examples in the rules are actually variations on some glossary definition rather than illustrating the rule. For example, in [Links have an accessible name](https://act-rules.github.io/rules/c487ae), Passed Example 4, 5, 6, 7, and 8 are just variations on how to give an accessible name to a link, *i.e.* they are illustrations of the definition of accessible name, not illustration of the rule.

This is bad from a rule-writing point-of-view since we have to remember to create examples of every single variation of every glossary item used which creates a pointless overhead. (and, by the way, many rules are missing accessible names given by `aria-labelledby`)

This is bad from a pedagogical point of view since it adds tons of examples who are illustrating something else than the current topic (*i.e.* the glossary item rather than the rule), creating a cognitive overhead of "oh, yeah, this is just another way to give accessible name"

# What?
I've tried to put some examples in one of the glossary item and remove redundant examples from some of the corresponding rule to give an idea of what it could look like. This obviously would require more work, and more polished work…

# Pros
Rules become easier to write and more to the point, hence easier to read.

Glossary becomes easier to understand with examples illustrating the definition.

# Cons
Glossary becomes harder to write but this should be balanced by the ease of writing rule.
In some case, it might however, become a real mess, typically with `value` giving an accessible name to `input` but not to `button`… 

Rules would still need a bit of variation in the examples. If all rules always only use `aria-label` to give an accessible name, this is not very good. We need one rule to use one way, and another one to use another way (or variation between examples), which can be harder to maintain in a reasonably good balance between the possibilities.

Glossary will need to be tested by checker engines. Currently, we do not test that all checker engines compute the accessible name in the same way, but the many examples all over the place act as such a test: if an engine forgets about `aria-label`, it will fail on many passed examples and we will detect that. If these variations are mostly written in the glossary, we need to run tests on these. It is less clear what the outcome of these tests are (what does "passed" mean?, not much… the outcome of accessible name should be the computed accessible name, but then how to store that within the test case?)

### DIFF
--- a/_rules/button-has-name-97a4e1.md
+++ b/_rules/button-has-name-97a4e1.md
@@ -55,22 +55,6 @@ Regular button.
 <button>My button</button>
 ```
 
-#### Passed Example 2
-
-Value attribute as the accessible name.
-
-```html
-<input type="submit" value="Submit" />
-```
-
-#### Passed Example 3
-
-`aria-label` for the accessible name.
-
-```html
-<button aria-label="My button"></button>
-```
-
 #### Passed Example 4
 
 Span tag with role button and has name defined by aria-label.
@@ -131,14 +115,6 @@ Input button has an accessible name that comes from the default "reset" text.
 ```
 
 ### Failed
-
-#### Failed Example 1
-
-Value attribute does NOT give an accessible name, only for input elements.
-
-```html
-<button type="button" value="read more"></button>
-```
 
 #### Failed Example 2
 
--- a/_rules/image-has-name-23a2a8.md
+++ b/_rules/image-has-name-23a2a8.md
@@ -59,22 +59,6 @@ The HTML `img` element has an accessible name
 <img alt="W3C logo" />
 ```
 
-#### Passed Example 2
-
-The element with role of `img` has an accessible name
-
-```html
-<div role="img" aria-label="W3C logo"></div>
-```
-
-#### Passed Example 3
-
-The element has an accessible name, though the name is not always accessibility supported
-
-```html
-<img title="W3C logo" />
-```
-
 #### Passed Example 4
 
 The HTML `img` element is marked as [decorative](#decorative) through an empty `alt` attribute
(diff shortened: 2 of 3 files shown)


EmmaJP:
I've no objection to providing more detail in the glossary that uses plainer language than other references that are being linked to. Though I think a full list of examples could quickly become very cumbersome to maintain and to read through.

However, I think the individual rules are where the variety of examples should exist, so each rule can be clear about a range of scenarios that comply/pass, don't comply/fail, or don't apply.

jkodu:
I agree with @EmmaJP here. Condensing examples in a rule given there are more examples in a glossary term does not seem ok.

WilcoFiers:
Hey Jean-Eyves,
Thank you for putting this together. This is definitely an interesting idea, one that has been discussed in the past. There's a reason we didn't continue with it at the time. My biggest problem with something like this is the question of what this means for implementors.

You can go one of two ways with that. The easiest one is that an implementor is free to use test cases if they want, but that implementors aren't required to show that they have an implementation that includes some internal representation of that definition. The problem with that is, because this removes test cases from the rule, this lowered the bar for implementors, which leads to less consistency.


The other solution is that we require that implementors show they have something in their test that represents that definition, that can be executed independently on the test cases in that glossary term, and that has to produce the results expected in the glossary term. This breaks one of the (in my opinion) more important pieces of ACT, the idea that it doesn't matter how you implement a rule as long as you get consistent results.

One example is axe-core's implementation of the Button has accessible name rule. Instead of calling a single accessible name method, it checks different attributes that can be used to provide an accessible name (vale, title, aria-label, etc.). Whether or not axe-core has an accessible name computation function that you can give a node, and that returns a text shouldn't matter as long as the rule produces the same end result.


That result of running a rule is the unit we're working with. That's the only thing that we can require from an implementation. You are absolutely correct that certain rules have test cases that aren't all that relevant to everyone reading that rule. That's why we proposed [collapsing test cases](https://github.com/act-rules/act-rules.github.io/issues/480#issuecomment-495125412) on a call in may. I'm also absolutely on board with adding test cases to definitions. But we can't require that an implementor pass those test cases, and so I don't think they can be a replacement of test cases in rules.

Jym77:
@WilcoFiers 
> This breaks one of the (in my opinion) more important pieces of ACT, the idea that it doesn't matter how you implement a rule as long as you get consistent results.

This is an extremely good point that I missed. Indeed the glossary doesn't really exists for ACT and thus everything must (also) live in the rules.

> I'm also absolutely on board with adding test cases to definitions.

I think that could help understand some of the more complex definition.



I'm closing this PR and keeping the  related issue open since we might still want to add examples to glossary, just without removing them from rules.

Issue #679, opened 2019-07-16
Ensure valid HTML for test cases in rules


jkodu:
Test cases have to be valid HTML

See example - https://github.com/act-rules/act-rules.github.io/pull/412/files#diff-8705f0115ad2b64712efb7544f98842dR130

jkodu:
PR in place, closing issue.

PR #631, opened 2019-06-26
chore: add tests to check validity of examples (snippets) in rules.

jkodu:
- Added a test that checks all example code snippets on each rule against a HTML validator. 
- Believe this will help writing valid test cases, given we identify and fix a lot of test cases during implementation phase.

More context & issues:
- https://github.com/act-rules/act-rules.github.io/pull/616/files#diff-021ac076b545925359e17fa319b1afadR3.
- https://github.com/act-rules/act-rules.github.io/issues/609

> Note: There are currently 53 failed examples, which can be fixed as a part of this PR>

PR #720, opened 2019-07-31
chore: add html-validator for testing code snippets

jeeyyy:
> Note: This was attempted using `html-hint` in this PR.

I came across [html-validator](https://www.npmjs.com/package/html-validator) which uses `validator.w3.org/nu` for validating code fragments. This PR integrates a test which shows all testcases being validated using the above and which fail/ pass.

I have not fixed the failing testcases as I wanted to get more opinion on if this is the right way to go about solving for validity of testcases. Please see output https://circleci.com/gh/act-rules/act-rules.github.io/2715 for failed testcases.

If I get a thumbs up that this is the step in the right direction. I am happy to fix the failing testcase code snippets.

Closes issue(s): 
- https://github.com/act-rules/act-rules.github.io/issues/679

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/__tests__/testcases.js
+++ b/_rules/__tests__/testcases.js
@@ -1,0 +1,45 @@
+const codeBlocks = require('gfm-code-blocks')
+const validator = require('html-validator')
+const describeRule = require('../../test-utils/describe-rule')
+
+describeRule('testcases', ruleData => {
+  const { frontmatter, body } = ruleData
+  const { id } = frontmatter
+  
+  /**
+   * get test case code snippets of the rule
+   */
+  const codeSnippets = codeBlocks(body).map(({ code }) => code)
+
+  /**
+   * Check if testcases are valid
+   * - using `html-validator`
+   * - check fragment `isFragment = true`
+   */
+  test.each(codeSnippets)('is valid testcase: `%p`', async (snippet) => {
+    const options = {
+      data: snippet,
+      isFragment: true,
+      format: 'json'
+    }
+    const result = await validator(options)
+
+    expect(result).toBeObject()
+    expect(result).toContainKey('messages')
+
+    const { messages } = result
+    const errorMessages = messages.filter(({ type }) => type === `error`)
+
+    if (errorMessages.length) {
+      const output = errorMessages.map(({ message }) => {
+        return {
+          "Rule Id": id,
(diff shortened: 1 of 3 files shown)


Jym77:
I think this looks nice, and the error messages of htmlvalidator are pretty good too.

👍

EmmaJP:
I definitely think we should be providing valid HTML code in test cases. Otherwise the permutations become too many to reasonably cover.

jkodu:
Cheers for the +1's. I will go ahead with the fixes.

carlosapaduarte (approved):

Jym77:
One issue I see with validating everything it that some examples need to become way bigger to be valid HTML, typically the ones that are supposed to be full documents (for rules whose applicability is the full document).

Typically, I want to have an example for using `main` landmarks, like this:
```html
<html>
  <div role="main" aria-label="Translation by C. H. Brewitt-Taylor">
    The empire, long divided, must unite; long united, must divide. Thus it has ever been.
  </div>
  <div role="main" aria-label="Translation by Yu Sumei">
    Unity succeeds division and division follows unity. One is bound to be replaced by the other after a long span of time.
  </div>
</html>
```

But if I want this to be validated by [HTML validator](https://validator.w3.org/nu), it needs to be:
```html
<!DOCTYPE html>
<html lang="en">
  <head><title>Comparing translations of the Romance of the Three Kingdoms</title></head>
  <div role="main" aria-label="Translation by C. H. Brewitt-Taylor">
    The empire, long divided, must unite; long united, must divide. Thus it has ever been.
  </div>
  <div role="main" aria-label="Translation by Yu Sumei">
    Unity succeeds division and division follows unity. One is bound to be replaced by the other after a long span of time.
  </div>
</html>
```

which adds a lot of "useless" stuff not related to the example. This both takes up space to read (and time to write) and dilute the purpose of the example (less focused, harder to find out the parts actually relevant for the rule).

(this may be better be validating snippets, typically the `DOCTYPE` is probably not needed)

But then, I say we can always try and evaluate :-D

jkodu:
Closing this PR, as discussed with @WilcoFiers that `html-hint` is a better solution.
So will be opening a similar PR like this - https://github.com/act-rules/act-rules.github.io/pull/631/files

Issue #625, opened 2019-06-26
Composite rule (c3232f) with another composite rule as input


Jym77:
Rule https://act-rules.github.io/rules/c3232f "video only has an accessible alternative" is listing rule https://act-rules.github.io/rules/eac66b "Video has audio alternative" as input rule.

However, both are composite rules and the ACT-R format requires that only atomic rules can be input of composite rules.

My guess is that the input rule should rather be "Video **only** has audio alternative" (which does not exist yet).

annethyme:
@ShadowBB and @WilcoFiers, you were the original authors for this rule. Could you maybe help provide a bit of clarity on this?
Or is the composite-composite rule on purpose? It is currently not in lign with the ACT Rules Format on this, but there has already been discussions about allowing composite-composite rules.

ShadowBB:
This issue is indeed an issue.

According to the media plan (https://github.com/act-rules/act-rules.github.io/issues/180) this should indeed have input from a rule called "Video only has audio alternative".

In fact, in the original pull request this was the case:
https://github.com/act-rules/act-rules.github.io/pull/276/files

I am unsure how that got changed later on but we should reverse that change.

We should revert his change and link to the correct rule (that does already exist):
https://github.com/act-rules/act-rules.github.io/blob/develop/_rules/video-only-audio-alternative-d7ba54.md

Jym77:
@jkodu Ain't this one closed by your recent PRs?

jkodu:
Yup. true. Can close this @Jym77

PR #722, opened 2019-08-01
chore: test if composite rules only refer to atomic rules as `input_rules`

jkodu:
- added test to verify if `composite` rule type only refers to `atomic` rules as `input_rules`.
- the actual fix for changing the rule id is here - https://github.com/act-rules/act-rules.github.io/pull/721

Closes issue(s): 
- https://github.com/act-rules/act-rules.github.io/issues/625

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/test-utils/describe-rule.js
+++ b/test-utils/describe-rule.js
@@ -1,19 +1,42 @@
 const getRulesMarkdownData = require('../utils/get-rules-markdown-data')
 
 /**
+ * Get ids of rules that are of given type
+ * @param {Array<Object>} rules list of rules
+ * @param {String} ruleType type of rule
+ */
+const getRuleIdsOfRuleType = (rules, ruleType) => {
+  return rules.reduce((out, ruleData) => {
+    const { frontmatter } = ruleData
+    const { rule_type, id } = frontmatter
+    if (rule_type === ruleType) {
+      out.push(id)
+    }
+    return out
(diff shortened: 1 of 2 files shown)


test-utils/describe-rule.js
```diff
@@ -1,19 +1,42 @@
 const getRulesMarkdownData = require('../utils/get-rules-markdown-data')
 
+/**
+ * Get ids of rules that are of given type
+ * @param {Array<Object>} rules list of rules
+ * @param {String} ruleType type of rule
+ */
+const getRuleIdsOfRuleType = (rules, ruleType) => {
```
Jym77: I'd rather split the two steps (filtering and getting Ids) in order to have simpler code, easier to understand and maintain.

Something like:
```javascript
return rules
  .filter(rule => rule.frontmatter.rule_type === ruleType)
  .map(rule => rule.frontmatter.id)
```

jkodu: 2 iterations vs 1 iteration ;). Given we are only iterating and adding to an array think it is easy to follow. Can change it to a for.. loop instead of reduce if you prefer.

Jym77: Given that this is not exactly time-critical, I'd go for readability over performance without hesitation.

PR #745, opened 2019-08-09
Remove `aria-label` from inapplicable test case as it otherwise passes (6a7281)

kasperisager:


jkodu:
@kasperisager once there are required number of approvals, this needs to go through Final Call (as advised by @annethyme). Please do the needy.

kasperisager:
The final call for this PR will end on September 16.

kasperisager:
@jkodu Any way to trigger Circle CI on your end? I've tried poking it with an empty commit, but it didn't seem to make a difference 😕

Issue #700, opened 2019-07-22
Citing the AAMs


Jym77:
There are a few places in rules and glossary where we reference the [Core AAM](https://www.w3.org/TR/core-aam-1.2/), [HTML AAM](https://www.w3.org/TR/html-aam-1.0/),  [SVG AAM](https://www.w3.org/TR/svg-aam-1.0/), or [ARIA in HTML](https://www.w3.org/TR/html-aria/).

All of these document are WIP (which is fine).

But they also all state:
>  It is inappropriate to cite this document as other than work in progress.

I understand that we should explicitly write things like "look into the [Core AAM](https://www.w3.org/TR/core-aam-1.2/) (work in progress)", which we totally don't do.

Should we change the way we are citing these documents?
@nitedog I guess you're the expert on that topic…

Jym77:
Closed by #716

PR #768, opened 2019-08-13
Handling shadow trees and iframe for "Links with identical accessible names..." (b20e66)

annethyme:
Add "within the same document" to Applicability to ensure that running the rule on a set of pages (instead of a single page) will not result in false positives.

Published rule: https://act-rules.github.io/rules/b20e66
----

## How to Review And Approve


Jym77:
Have fun reviewing all the shadow tree examples 🤣 
Note that the "complexity progression" of these examples is jumping between passed/failed/inapplicable examples, which does not make anything easier…

Jym77:
Testing with NVDA on Chrome, when asked to list links, it list all of them without making any mention of them being in another document (iframe) or not. That is
```html
<a href="foo.html">Foobar</a>
<iframe title="bar" srcdoc=’<a href="bar.html">Foobar</a>’</iframe>
```
has (for NVDA) two links that are both named "Foobar" and nothing from the list of links show that they are in different context (document). Thus, they appear as links with the same name and should lead to the same place.

I haven't checked how it is handled for iframe and how the interaction goes for getting into some context. I haven't checked with other AT…

Jym77:
I've added an Accessibility Support about AT exposing all links together, which is the behaviour I've seen on Chrome+NVDA.

Jym77:
Fixed (hopefully) the applicability to correctly include all links that are within iframe (recursively).

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -24,21 +25,26 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names][accessible name] that do not only consist of [whitespace](#whitespace).
+This rule applies to any set of any two or more HTML or SVG elements which
 
-**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name][].
+- are inside the [flat trees][flat tree] (work in progress) of one or more [documents][document] of the same [set of browsing context related documents][]; and
```
WilcoFiers: What's this "work in progress" thing doing here? If this PR is WIP, please put that in the name, or make it a draft PR.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -24,21 +25,26 @@ htmlHintIgnore:
...
+- are [included in the accessibility tree][] of their respective [document][]; and
+- have the [semantic role][] of `link`, or a role that inherits from the `link` role; and
```
WilcoFiers: I suggest making this the first item of the list. It's the most important one.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -152,19 +158,183 @@ A set of two SVG `<a>` elements have the same [accessible name][] and link to th
...
+		<title>Links in the shadow</title>
+	</head>
+
+	<body>
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html">Contact us</a>
+		all the time.
+
+		<div id="host">
+			<span
+				><a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html"
```
WilcoFiers: Please run Prettier.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -152,19 +158,183 @@ A set of two SVG `<a>` elements have the same [accessible name][] and link to th
...
+			const host = document.getElementById('host')
+			const shadowRoot = host.attachShadow({ mode: 'open' })
+
+			shadowRoot.innerHTML = '<slot></slot>'
+		</script>
+	</body>
+</html>
+```
+
+#### Passed Example 13
```
WilcoFiers: The purpose of these examples is to show what the rule is about, not to check whether or not shadow DOM is correctly implemented. You can potentially add that to any rule. That's just creates a lot of duplicate tests. It's also kind of an anti-pattern to test more than what you are actually trying to show. The only thing this rule needs to verify is that shadow DOM was considered. Not that the shadow DOM implementation is correct. Please remove passed examples 12 - 14.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -152,19 +158,183 @@ A set of two SVG `<a>` elements have the same [accessible name][] and link to th
...
+All three links have the same [accessible name][]. The second link ("from the light") is only part of the [light tree][]. When the [shadow tree][] is attached to `host` and flattened, this link is overwritten and therefore not part of the [flat tree][] (work in progress). Hence, only the first and third link are considered by this rule and they both point to the [same resource][].
+
+```html
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>Links in the shadow</title>
+	</head>
+
+	<body>
```
WilcoFiers: Please only use the content of the body. The doctype, html element, and the head stuff is not relevant to this example. The same goes for most of the new examples below.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -222,26 +392,220 @@ A set of two SVG `<a>` elements have the same [accessible name][] but link to di
...
+			const shadowRoot = host.attachShadow({ mode: 'open' })
+
+			shadowRoot.innerHTML =
+				'<span><a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html">Contact us</a> from the shadow.</span>'
+		</script>
+	</body>
+</html>
+```
+
+#### Failed Example 8
```
WilcoFiers: See my comment on passed example 12. This rule isn't about correct use of shadow DOM, so it's test shouldn't be either. Failed examples 8 through 11 should be removed.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -282,18 +646,98 @@ Links created via scripting, but without the semantic role of link:
...
 #### Inapplicable Example 6
 
-Image links do not have [accessible names][accessible name]:
+These image links do not have [accessible names][accessible name].
 
 ```html
 <a href="http://facebook.com"><img src="facebook.jpg"/></a> <a href="http://twitter.com"><img src="twitter.jpg"/></a>
 ```
 
+#### Inapplicable Example 7
```
WilcoFiers: As above, this rule isn't about shadow DOM. inapplicable example 7 and 8 should be removed.

pages/glossary/browsing-context-related-documents.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Set of browsing context related documents
+key: browsing-context-related-documents
+---
+
+Two [documents](https://dom.spec.whatwg.org/#concept-document) are *browsing context related* if they are [fully active](https://html.spec.whatwg.org/#fully-active); and their respective [browsing context](https://html.spec.whatwg.org/#browsing-context) have a common [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context).
```
WilcoFiers: Please use W3C documentation. The WHATWG spec is a living standard. These links, and the contents of them can change.

pages/glossary/browsing-context-related-documents.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Set of browsing context related documents
```
WilcoFiers: I think we can make this a lot easier to understand if we say something like "on the same web page". I suggest you expand this to not just be about documents, but about nodes, which either have the same owned document, or are owned by two document with the same top-level browser context. That'll make the definition easier to use.

pages/glossary/browsing-context-related-documents.md
kasperisager: @WilcoFiers The HTML and DOM specifications are no longer published by the W3C, please see https://www.w3.org/2019/04/WHATWG-W3C-MOU.html.

Jym77: @WilcoFiers @kasperisager here is what I'm understanding of the situation (and your respective positions):
* W3C stopped *writing* HTML/DOM specs, WHATWG starts writing them as Living Standard.
* W3C will nonetheless *publish* these as REC sometimes.
* HTML 5.2 is the latest *stable* standard (W3C REC) and does have a stable URL at https://www.w3.org/TR/html52/ while the editor draft of the "next" version at https://www.w3.org/TR/html/ redirects to WHATWG Living Standard.
* It is likely that at some point in the future, the WHATWG LS gets frozen into HTML 5.3 and that would get its own (non-redirect) URL at https://www.w3.org/TR/html53/ (that's how I understand the fact that W3C will *publish* as REC the result of WHATWG work). WHATWG (and https://www.w3.org/TR/html/) will still hold the newer Living Standard.

---
When it comes to us, the question is what we should link. @WilcoFiers says that we should always link a frozen version (*i.e.* 5.2) because the LS can change without notice and breaks our rules. @kasperisager says that we should link the latest version (*i.e.* WHATWG) because (I guess) if browsers follow it and it differs from 5.2, then our rules would be completely outdated.

I do not have a strong opinion on that choice. 

Recently (since I joined), we've been moving links from https://www.w3.org/TR/html/ to WHATWG. It is worth noting that links to https://www.w3.org/TR/html/ were anyway "living" since that was not a frozen version number.

Whatever the choice we make (5.2 or WHATWG), the important thing (for me) is that we are coherent among all our rules. Currently the choice is more toward WHATWG (20 occurrences vs 4 for html52), but this is largely due to the move that was recently made (notably by @annethyme , @jkodu and me). So it is in no way a clear decision.

---
I think we (likely meaning @WilcoFiers and @kasperisager since you seem to have a strong opinion about it) agree on which document we should cite and stick to it (and update the other refs). This ripples through all rules and several current PR (eg, #853) and so should be settled soonish. Probably a good agenda item for ACT-R CG.

As I said, I don't care which way we go for citing (5.2 or WHATWG), but I want to know which way we go so I can do it correctly…

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
Jym77: The CSS Scoping Module Level 1 (as well as a few other specs) are only WiP and require user to "Don’t cite this document other than as work in progress."

See #700 and #716 for more on that, this has also been discussed on an ACT-R call in July (the 25th, I guess).

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
Jym77: I do not fully understand…

I mean, in July I made the point that many rules have examples that are unrelated to the rule but only duplicate tests with a lot of variation on how to got an accessible name. See #717 for the PR.

Back then, you dismissed the PR in https://github.com/act-rules/act-rules.github.io/pull/717#issuecomment-516377867 by stating, among other thing, that we do not care how or whether implementors have an implementation of "accessible name" and thus should keep all the examples in the rules to check that the implementation of that rule is using a correct way of getting it.

I do not understand how this argument does apply to accessible name but not to the shadow DOM. The current rules have very few shadow DOM examples and thus do not check if implementation correctly handle them, including all the subtleties of slots of stuff. 

Why would it be OK to only verify that shadow DOM was considered, without verifying that it is correct; while it was KO to only verify that accessible name was considered without verifying that it is correct (_e.g._ having only a `aria-label` example and no `aria-labelledby` example)?

I agree that shadow trees could potentially be added to many rules (not all rules actually care about shadow trees). In the exact same way that 4 or 5 different possibilities of giving an accessible name can be added to many rules. I agree that it clutters the rules with useless examples that do not illustrate the rule at all, this was exactly my point in #717.

pages/glossary/browsing-context-related-documents.md
Jym77: The problem being that "web page" is less well defined that "browsing context", as you pointed in #869 

I agree that nodes will be better for that def.

pages/glossary/browsing-context-related-documents.md
Jym77: Similarly to #853, I will keep the current citations until #879 is resolved and we have a global PR fixing everything.

pages/glossary/browsing-context-related-documents.md
Jym77: I tried to use node inside the definition. I am not completely sure this is better since browsing context is ultimately a document concept more than a node concept, so this creates an extra indirection (node => document => browsing context).

Maybe it could be done by still having "browsing context related **documents**" and use it to define "browsing context related **nodes**" (as all nodes inside browsing context related documents). What do you think?

pages/glossary/browsing-context-related-nodes.md
```diff
@@ -0,0 +1,17 @@
...
+
+A _set of browsing context related nodes_ is the set of all [nodes](https://dom.spec.whatwg.org/#concept-node) that are browsing context related with a [node](https://dom.spec.whatwg.org/#concept-node) whose [shadow including root]'s [browsing context](https://html.spec.whatwg.org/#browsing-context) is a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context).
+
+A _set of browsing context related documents_ is the set of all [documents](https://dom.spec.whatwg.org/#concept-document) that are browsing context related with a [document](https://dom.spec.whatwg.org/#concept-document) whose [browsing context](https://html.spec.whatwg.org/#browsing-context) is a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context).
+
+**Note**: By construction, a set of browsing context related nodes is always maximal for inclusion.
```
carlosapaduarte: Why do we need this note?

Jym77: Not strictly needed, but I felt it was a useful clarification. A set of BCR nodes could also be defined as "a set of nodes that are pairwise BCR" (it's an equivalence anyway) in which case it would not necessarily be maximal for inclusion.

In a more math-y text, I would totally define "a set of foo" as the non-maximal for inclusion one, and use (possibly with definition) "a maximal set of foo" for the one that has to be maximal for inclusion. It did hurt a bit my math-self to use "a set of BCR nodes" for what should rather be "a maximal set of BCR nodes". Since we probably won't need the former, it was simpler not to use it. But then, I had to add the clarifying note to point out that these sets are maximal even if the name does not suggest so.

pages/glossary/browsing-context-related-nodes.md
```diff
@@ -0,0 +1,17 @@
...
+title: Set of browsing context related nodes
+key: browsing-context-related-nodes
+---
+
+Two HTML [nodes](https://dom.spec.whatwg.org/#concept-node) are _browsing context related_ if:
+
+- their [shadow including roots](https://dom.spec.whatwg.org/#concept-shadow-including-root) are [fully active](https://html.spec.whatwg.org/#fully-active); and
+- the [browsing contexts](https://html.spec.whatwg.org/#browsing-context) of their [shadow including roots](https://dom.spec.whatwg.org/#concept-shadow-including-root) either have a common [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context), or one is [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context) of the other, or are the same.
+
+A _set of browsing context related nodes_ is the set of all [nodes](https://dom.spec.whatwg.org/#concept-node) that are browsing context related with a [node](https://dom.spec.whatwg.org/#concept-node) whose [shadow including root]'s [browsing context](https://html.spec.whatwg.org/#browsing-context) is a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context).
```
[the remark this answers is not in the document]
Jym77: I prefer 'a' since I see the definition as "a set of BCR nodes among all the nodes on the internet" in which case there is no unicity of top-level browsing context. If it is "a set of BCR nodes among all nodes in that browsing context tree" (where top-level browsing context is unique and we can use 'the'), then it's just the set of all nodes and does not really need to be that complex.

It is also more consistent with the use in the specs, _e.g._ the definition of [fully active](https://html.spec.whatwg.org/multipage/browsers.html#fully-active) uses "…its browsing context is a top-level browsing context" and https://html.spec.whatwg.org/multipage/browsers.html#security-nav is also using 'a' rather than 'the'.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -24,21 +25,26 @@ htmlHintIgnore:
...
 
 ## Accessibility Support
 
-There are no major accessibility support issues known for this rule.
+This rule assume that assistive technologies are exposing all links on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its links, then it is possible for two links to have identical name but resolve to different resources without breaking success criterion [2.4.9 Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html) (if said links are in separate [documents][document] or [shadow trees][shadow tree])
 
 ## Background
```
WilcoFiers: Because you're using shadow tree, which is in a draft spec, can you add a link to the draft spec in the background?

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -24,21 +25,26 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names][accessible name] that do not only consist of [whitespace](#whitespace).
+This rule applies to any set of any two or more HTML or SVG elements which
 
-**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name][].
+- have the [semantic role][] of `link`, or a role that inherits from the `link` role; and
+- are in the same [set of browsing context related nodes][]; and
```
WilcoFiers: Let's change this definition title to "on same web page"

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -24,21 +25,26 @@ htmlHintIgnore:
...
+- are [included in the accessibility tree][] of their respective [document][]; and
```
WilcoFiers: I don't think you need to include this "of their respective document" part. At least, I don't see what it adds here.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -24,21 +25,26 @@ htmlHintIgnore:
...
 
 ## Accessibility Support
 
-There are no major accessibility support issues known for this rule.
+This rule assume that assistive technologies are exposing all links on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its links, then it is possible for two links to have identical name but resolve to different resources without breaking success criterion [2.4.9 Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html) (if said links are in separate [documents][document] or [shadow trees][shadow tree])
```
WilcoFiers: Nice add. Great job!

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
```diff
@@ -222,26 +305,87 @@ A set of two SVG `<a>` elements have the same [accessible name][] but link to di
...
 
-Links resolves to same resource after redirect, but the redirect is not instant:
+Both links resolve to [same resource][] after redirect, but the redirect is not instant.
 
 ```html
 <a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/index.html">Contact us</a>
 <a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/redirect1.html">Contact us</a>
 ```
 
+#### Failed Example 7
```
WilcoFiers: This is testing whether or not an implementor supports shadow DOM, not whether or not the rule is properly implemented. I think this should be removed. Failed example 8 covers this scenario well enough I think.

pages/glossary/browsing-context-related-nodes.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Set of browsing context related nodes
```
WilcoFiers: Rename this to "on the same web page"

pages/glossary/browsing-context-related-nodes.md
```diff
@@ -0,0 +1,15 @@
...
+key: browsing-context-related-nodes
+---
+
+Two HTML [nodes](https://dom.spec.whatwg.org/#concept-node) are _browsing context related_ if:
```
WilcoFiers: I think this can be simplified by saying that they need to have the same [top-level browser context](https://www.w3.org/TR/html50/browsers.html#top-level-browsing-context). I also wouldn't worry about whether or not something is activated. Nodes don't show up in the accessibility tree if they're not, and we're always looking for that condition.

_rules/links-with-identical-names-serve-equivalent-purpose-b20e66.md
Jym77: "**the accessibility tree**" is not correct since there is one per document, hence I felt the need for precision.
Replacing with "**an** accessibility tree" (which is less precise but probably good enough).

pages/glossary/browsing-context-related-nodes.md
Jym77: I think we need to worry about full activation.
Each document has an accessibility tree, even if it's not fully activated. So, nodes can share the top level browsing context, and be in the accessibility tree of an inactive document.

Jym77:
@carlosapaduarte @ShadowBB can you (re)review the newest version of the def and its usages?

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -224,7 +225,7 @@ all the time.
 
 #### Passed Example 13
 
-The [browsing context][] of the `iframe` has the [browsing context][] of the main [document][] as an [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context). Hence, its [nodes][node] are part of the same [set of browsing context related nodes][] as the [nodes][node] of the main [document][]. Therefore, both links are considered and, since they refer to the same document, the rule passes.
+The [browsing context][] of the `iframe` has the [browsing context][] of the main [document][] as an [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context). Hence, They share the same [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context) (namely, the [browsing context][] of the main [document][]) and are part of the same [web page][]. Therefore, both links are considered and, since they refer to the same document, the rule passes.
```
carlosapaduarte: They > they

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -360,9 +337,9 @@ all the time.
 </script>
 ```
 
-#### Failed Example 9
+#### Failed Example 8
 
-The [browsing context][] of the `iframe` has the [browsing context][] of the main [document][] as an [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context). Hence, its [nodes][node] are part of the same [set of browsing context related nodes][] as the [nodes][node] of the main [document][]. Therefore, both links are considered and the rule fails.
+The [browsing context][] of the `iframe` has the [browsing context][] of the main [document][] as an [ancestor browsing context](https://html.spec.whatwg.org/#ancestor-browsing-context). Hence, They share the same [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context) (namely, the [browsing context][] of the main [document][]) and are part of the same [web page][]. Therefore, both links are considered and the rule fails.
```
carlosapaduarte: They > they

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -24,24 +25,30 @@ htmlHintIgnore:
...
+- are in [documents][document] of the same [web page][]; and
+- are [included in an accessibility tree][included in the accessibility tree]; and
+- that have [matching][] [accessible names][accessible name] that do not only consist of [whitespace](#whitespace).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][].
 
 ## Expectation
 
-When followed, the links in each set of target elements resolve to the [same resource](#same-resource) or to [equivalent resources](#equivalent-resource). Resolving the links includes potential redirects, if the redirects happen instantly.
+When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource). Resolving the links includes potential redirects, if the redirects happen instantly.
```
DagfinnRomen: According to SC 2.4.9 the purpose of each link can be determined from the link text alone. I do not think the test rule tests this.

According to the test rule, links with matching names should resolve to the same or equivalent resource. In my opinion, this is best practice, but not a requirement.

Pasted from the SC 2.4.9 understanding article: "Best practice is that links with the same destination would have the same descriptions, but links with different purposes and destinations would have different descriptions (see also Success Criterion 3.2.4 which calls for consistency in identifying components that have the same functionality)." https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html 

Also, if the linked resource(s) with matching accessible names have a different purpose than what is described in the link name, this would pass the test rule, but fail the SC.

In my opinion, this test rule is more relevant for SC 3.2.4 Consistent Identification. Either way, SC 3.2.4 requires the identification to be consistent, but not necessarily identical.

I have also left a similar comment, related to SC 2.4.4 in #693

Jym77: …Hmmmm…

---
I agree with the fact that it is more a best practice than a failure for [2.4.9](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-link.html). However, I am not sure this is a failure for [3.2.4](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-consistent-functionality.html) either because of this:
> While it is desirable and best practice always to be consistent within a single web page, 3.2.4 only addresses consistency within a set of web pages where something is repeated on more than one page in the set.

---
Also, the problem is present in [the current rule](https://act-rules.github.io/rules/b20e66) and not really related to this PR. It certainly need to be fixed but this might require more discussion and a separate PR to keep things focused.

---
My current approach would be to keep this PR as is (not changing the SC), and open an issue for the SC mismatch that can be discussed further and fixed later. Would that be OK for you @DagfinnRomen ?

DagfinnRomen: That would be ok, @Jym77 . Could you open an issue for this?

Jym77: Will do.

kasperisager: This is definitely something for a separate pull request, yes. My opinion on the matter is that the rule is really making an assumption that has not yet been written out. That assumption is that when two links have the same accessible name, they describe the same purpose, and the resources they reference should therefore be equivalent as they otherwise do not serve the same purpose. 

Put another way, the word "coffee" is not really describing the purpose of "a cup of coffee" if I also use the word to refer to "a cup of tea". In reverse, I should only use the word "coffee", and not also "joe", "java", or "mocha", when I want to refer to "a cup of coffee"; this seems to be the best practice described in the understanding document.

Jym77: > Pasted from the SC 2.4.9 understanding article: "Best practice is that links with the same destination would have the same descriptions, but links with different purposes and destinations would have different descriptions (see also Success Criterion 3.2.4 which calls for consistency in identifying components that have the same functionality)." https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html

Hmmmm…
This seems to go the other way around. (and that is likely what @kasperisager is also saying…)

The best practice is that if the links go to the same place, then they should have the same name. The rule is that if the links have the same name, then they should lead to the same place.

So, to keep on with @kasperisager examples:
```html
<a href="#cup-of-coffee">coffee</a>
<a href="#cup-of-coffee">java</a>
```
is failing the best practice (both links have same destination but different names), but not the SC. And it is inapplicable for the rule (there is no set of two or more links with the same name).

On the other hand,
```html
<a href="#cup-of-coffee">coffee</a>
<a href="#cup-of-tea">coffee</a>
```
is not failing the best practice (the links do not have the same destination), but is failing the rule (they have same name but different destination), and likely failing the SC (the link text alone is not enough to get the purpose of the link given that the same text cannot accurately describes two different purposes).

---
Therefore, I am now thinking that the accessibility requirement mapping is actually correct. What do you think @DagfinnRomen ?

DagfinnRomen: I am with you so far, @Jym77 and @kasperisager . But what about the following?
```
<a href="#cup-of-coffee">tea</a>
<a href="#cup-of-coffee">tea</a>
```
This passes both the best practice and the rule (the links are identical and have identical destinations). However, it fails the SC, as the link texts do not describe the purpose.

Jym77: Yes.

As always, the rule is only a partial check and only has a fail-on-fail relationship with the SC. The rule can pass, but the SC still fail, hence the "All passed outcomes: further testing is needed" requirement mapping.

Locally, (for the "in context" version), it would fail #555 (I do not think we have a similar rule without context given that 2.4.9 is AAA, but it would probably be super easy to have by just taking #555 and removing context…)

Jym77: @DagfinnRomen I made a separate PR #959 to improve one inapplicable example by having the "same link, different names" case.

Resolving this conversation.

DagfinnRomen: I am not sure if the test rule has a fail-on-fail relationship. The failed test cases are all based on the assumption that the same text cannot accurately describe two different purposes.

This may be considered best practice, but it is not a part of the requirement.

For example
```
<a href="#cup-of-coffee">coffee</a>
<a href="#cup-of-joe">coffee</a>
```
This fails both best practice and the rule, but it does not fail the SC.

The requirement states that each link describes the purpose of the link. In my opinion, the links are to be assessed individually. It is not necessary to compare links to verify SC 2.4.9.

I support the suggestion of adapting #555 to a new rule for SC 2.4.9, though.

Jym77: While the links are not **identical**, they are likely [**equivalent**](https://act-rules.github.io/glossary/#equivalent-resource) and thus the rule passes.

On the other hand, if `#cup-of-joe` is about the personal cup of a guy named Joe (and `#cup-of-coffee` is about coffee), then the links are not equivalents and the rule fails, but so does the SC (although for a different reason, namely that `"coffee"` is not an accurate description of the cup belonging to Joe).

So, this should be good 🤞 

---
I do agree, however, that the SC is about individual links rather than comparing links.
I do not know why this rule was here in the first place.
We may actually want to kill it once we have #555 (and its "no context" version) since if each individual link is descriptive, then links with same name must lead to equivalent resources.

This may require further discussion (and if we consider that the rule has been superseded and we plan to kill it, there's no need for this PR 🙄 )

Jym77: @DagfinnRomen, check also #962.

Jym77: As a conclusion to #968, an assumption is added to this rule in #980.
This should fix this concern.

pages/glossary/web-page.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Web page
+key: web-page
```
jkodu: Same comment as I have left in other one about webpage definition.
I would prefer if we do not add a definition & later update all rules that need webpage definition based on the conclusion arrived on whether we need this.

Jym77:
@jkodu I updated the definition of web page according to this morning ACT-R call and @EmmaJP suggestion in https://github.com/act-rules/act-rules.github.io/issues/869#issuecomment-545808574

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -24,25 +25,31 @@ htmlHintIgnore:
...
+- are in [documents][document] of the same [web page (HTML)][]; and
+- are [included in an accessibility tree][included in the accessibility tree]; and
+- that have [matching][] [accessible names][accessible name] that are not empty (`""`).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][].
 
 ## Expectation
 
-When followed, the links in each set of target elements resolve to the [same resource](#same-resource) or to [equivalent resources](#equivalent-resource). Resolving the links includes potential redirects, if the redirects happen instantly.
+When followed, the links in the set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource). Resolving the links includes potential redirects, if the redirects happen instantly.
```
WilcoFiers: each, not the

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -24,25 +25,31 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names][accessible name] that do not only consist of [whitespace](#whitespace).
+This rule applies to any set of any two or more HTML or SVG elements which
 
-**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name][].
+- have the [semantic role][] of `link`, or a role that inherits from the `link` role; and
+- are in [documents][document] of the same [web page][]; and
```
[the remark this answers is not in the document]
ShadowBB: I think it is linked to passed example 13. But I also don't understand it.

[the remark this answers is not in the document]
ShadowBB: I think I understand now. Without this clause the links can be any two links on the internet. This makes sure they are on the same web page. But the "are in documents" is really confusing and not necessary.

It would be enough to simply say "are on the same web page (HTML)"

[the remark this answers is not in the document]
Jym77: > I think I understand now. Without this clause the links can be any two links on the internet. This makes sure they are on the same web page. But the "are in documents" is really confusing and not necessary.
> 
> It would be enough to simply say "are on the same web page (HTML)"

The definition of "web page (HTML)" is a set of documents. Thus, links are technically not on the same web page, but are merely part of documents which appear on the same web page.

At the same time, I do agree that this is maybe over-technicalling the rule and that abusively writing "elements on the same web page (HTML)" instead of the correct "elements in documents that are part of the same web page (HTML)" is probably OK. I can also add a note on the definition to make that clear.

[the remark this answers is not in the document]
Jym77: I simplified the formulation here and added a note in the definition to justify this simplification. Please check that this is OK.

Jym77:
Final Call ends on November 19.

mraccess77:
I have a question about inclusion of iframes.    Some tools may insert things into the DOM or a copy of the DOM during the process of testing.  iFrames may be in different domains and this may not be possible with cross domain scripting policies.  Also a href in one iframe and in the document may appear the same but resolve do different locations "../contactus.html" and "../contactus.html" may resolve to different relative places and may require more complex ways to figure out the true location of the url to know if it's the same or different.  Also it's my experience that screen readers list of links may only work within an iFrame but not provide links from the whole page.   Because of this I wonder if we should really be including iFrames due to issues with actual implementations outside of a browser extension.  That is testing engines can run in many different environments and in some like a script added to a page they may not have access to these items.

Jym77:
@mraccess77 thanks for feedback. I'll try to answer the various concerns, re-ordering them a bit…

>  Also it's my experience that screen readers list of links may only work within an iFrame but not provide links from the whole page.

My experience with NVDA was that the "list links" button is listing all links in the page, included in nested `iframe`, without any difference. The Accessibility Support part of the rule is also addressing that: if AT is handling differently links in and out of `iframe`, things are a bit different.

> Also a href in one iframe and in the document may appear the same but resolve do different locations "../contactus.html" and "../contactus.html" may resolve to different relative places and may require more complex ways to figure out the true location of the url to know if it's the same or different.

The rule is not checking that the links resolve to identical URLs (=characters strings) but to equivalent resources. This is, most likely, not going to be a fully automated rule (because deciding if `../contact.html` and `../contactus.html` are equivalent is hard to automate). Therefore, I do no see this as too much of a problem. The (human) evaluator will in any case need to look at the linked resource and decide whether they are equivalent or note.

It is however a very interesting point that same URL does not necessarily mean same resource when `iframe` are involved. I will add a note about that.

> Some tools may insert things into the DOM or a copy of the DOM during the process of testing. iFrames may be in different domains and this may not be possible with cross domain scripting policies. Because of this I wonder if we should really be including iFrames due to issues with actual implementations outside of a browser extension. That is testing engines can run in many different environments and in some like a script added to a page they may not have access to these items.

I do not think this is really a concern for the rule. It might mean that some tools cannot automate this rule. But that is no different from the situation with many manual or semi-automated rules. Consider a rule like "audio has accurate transcript". We could similarly argue that some tools do not have speech recognition and thus cannot implement such a rule. This is, however, not a concern for the rule.

It might, however, be a concern for the ACT rules community. If the rule does not get enough implementations it will never be considered for publication (https://act-rules.github.io/pages/implementations/overview/#status-reports) and that essentially means that work on the rule has been wasted.

In that case, I'm not too worried about this because the "equivalent resources" part make this largely a manual rule anyway.

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -154,19 +161,96 @@ A set of two SVG `<a>` elements have the same [accessible name][] and link to th
...
+<script>
+	const host = document.getElementById('host')
+	const shadowRoot = host.attachShadow({ mode: 'open' })
+
+	shadowRoot.innerHTML =
+		'<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html">Contact us</a> from the shadow.'
+</script>
+```
+
+#### Passed Example 12
```
WilcoFiers: IMO one shadow DOM example is enough. We're not testing shadow DOM implementations, we're testing links.

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -280,4 +401,62 @@ Links created via scripting, but without the semantic role of link:
...
+
+These image links have empty [accessible names][accessible name].
+
+```html
+<a href="http://facebook.com"><img src="facebook.jpg"/></a> <a href="http://twitter.com"><img src="twitter.jpg"/></a>
+```
+
+#### Inapplicable Example 7
+
+Only the first link ("all the time") is present in the [flat tree][]. Hence, there is no set of two or more links to apply the rule. The [shadow tree][] does contain a [slot][], but because its [named][name], the second link ("from the light") is _not_ [slotted][slot] into it.
```
WilcoFiers: This is more a test for shadow DOM than it is for links. IMO this should be removed.

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -25,25 +26,31 @@ htmlHintIgnore:
...
-This rule applies to any set of any two or more HTML or SVG elements that have the [semantic role](#semantic-role) of `link`, or a role that inherits from the `link` role, are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names][accessible name] that do not only consist of [whitespace](#whitespace).
+This rule applies to any set of any two or more HTML or SVG elements which
 
-**Note:** The test target for this rule is the full set of link elements that share the same [matching](#matching-characters) [accessible name][].
+- have the [semantic role][] of `link`, or a role that inherits from the `link` role; and
+- are in the same [web page (HTML)][]; and
+- are [included in an accessibility tree][included in the accessibility tree]; and
+- that have [matching][] [accessible names][accessible name] that are not empty (`""`).
+
+**Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][].
```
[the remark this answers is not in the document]
Jym77: Not sure, but it's already here. Removing it is a completely different concern and should be done in another PR.
(similar notes exist in other rules and might be removed at the same time)

Jym77:

PR #769, opened 2019-08-13
[WIP] Add "shadow including root" to Applicability of "Iframes with identical accessible names..." (4b1c6c)

annethyme:
Updated the rule to take into account shadow trees as well.

----

## How to Review And Approve

### DIFF
--- a/_rules/iframe-unique-name-4b1c6c.md
+++ b/_rules/iframe-unique-name-4b1c6c.md
@@ -20,7 +20,7 @@ authors:
 
 ## Applicability
 
-This rule applies to any set of any two or more `iframe` elements in the same [document tree](https://www.w3.org/TR/dom41/#document-trees) that are [included in the accessibility tree](#included-in-the-accessibility-tree), and that have [matching](#matching-characters) [accessible names](#accessible-name) that do not only consist of [whitespace](#whitespace).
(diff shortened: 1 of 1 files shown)


Jym77:
@WilcoFiers from what I understand, you can have an `iframe` in the light tree and one in the shadow tree that are not in the same document per se but once you plug the shadow tree into the light tree and flatten all that, they will be rendered together and thus need different id.

I thing the current proposal (same shadow including root) is both a bit too complex formulation and not as precise as we want (the descendant (in the light tree) of a shadow host might disappear from the flat tree if it's not slotted). @kasperisager agrees that "within the same flat tree" is OK and I think it's slightly easier to understand. I will rewrite in that direction and include examples…

(same goes for "link with identical name" #768)

kasperisager:
My burning question, whose answers will perhaps also help resolve #835, is why we even require that `<iframe>` elements with the same accessible name embed the same or equivalent resources? 4.1.2 requires that the names of user interface components can be programmatically determined, but doesn't require that the names uniquely identify each component; I'm guessing that this is something we're however all reading into the success criterion. 

From a user perspective, it does make sense: If I, for example, announce my intent to interact with a user interface component named "foo", I would hope that every user interface component named "foo" would serve an equivalent purpose as to make my intent unambiguous. Is this really what we're testing then? If so, then my next question is: How does AT deal with shadow and document boundaries? 

Building on the previous example, would AT look for a user interface component named "foo" across shadow and document boundaries when instructed to interact with this component? Would it only cross shadow boundaries within the current document context, and provide some other way of ascending into or descending out of a given document context? Regardless of the answer, it should be documented in the rule and used as the basis of whatever decision we make. Without an answer, we have no way of deciding on whether or not to cross shadow and/or document boundaries when looking for duplicate names.

Jym77:
Reimplemented by #982.

Issue #759, opened 2019-08-13
Put the changelogs on a separate page


WilcoFiers:
The ACT Rules Format does not require the changelog to be included in the rule itself. The table is pretty big, and it would help the readability of the page if the changelog wasn't part of the rules page. I propose we put a link to the changelog under "Useful links" instead.

jkodu:
PR in place. Closing issue.

Issue #734, opened 2019-08-08
"See file at given version" in Changelog seems to be broken for updates before end of April


annethyme:
On rule https://act-rules.github.io/rules/2779a5, the older items in the Changelog has broken links for "See file at given version"

> Apr 29, 2019 | chore: rename files and update associations (#489) **<-- WORKS**
> 
> Apr 16, 2019 | chore: add unique id to all rules (#478) **<-- Gives a 404**

If it's because we don't have anything to link to, could we just remove the links?

jkodu:
Good catch, this has happened due to file name changes between 2 commits. 
But I will research if this can be fixed easily.

jkodu:
Action for @jkodu - remove column on changelog for see file at version.

jkodu:
This PR - https://github.com/act-rules/act-rules.github.io/pull/787 removes the column which allows to see file at given commit.

PR #787, opened 2019-08-15
fix: move changelog of each rule to its own page

jkodu:
- introduce a `..url/rules/(ruleId)/changelog` page for each rule, so that changelog can be kept separately out of the rule description page.
- also removed respective sections from the rule page.

Closes issue(s): 
- https://github.com/act-rules/act-rules.github.io/issues/759
- https://github.com/act-rules/act-rules.github.io/issues/734

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


src/templates/rule.js
```diff
@@ -111,6 +108,11 @@ export default ({ data }) => {
 								Propose a change to the rule
 							</a>
 						</li>
+						<li>
```
Jym77: I think we should put the Changelog link close to the Issues link since I feel they are closely related.

Issue #642, opened 2019-07-01
Some space characters missing from glossary/whitespace.md


dd8:
There are some zero-width Unicode characters that don't have the Unicode White_Space property, but should be treated equivalently to whitespace when detecting empty text alternatives and labels:

U+200B ZERO WIDTH SPACE
U+200C ZERO WIDTH NON-JOINER
U+200D ZERO WIDTH JOINER
U+2060 word joiner
U+FEFF zero width no-break space

The Unicode chars with the White_Space property are:

0009..000D    ; White_Space # Cc   [5] <control-0009>..<control-000D>
0020          ; White_Space # Zs       SPACE
0085          ; White_Space # Cc       <control-0085>
00A0          ; White_Space # Zs       NO-BREAK SPACE
1680          ; White_Space # Zs       OGHAM SPACE MARK
2000..200A    ; White_Space # Zs  [11] EN QUAD..HAIR SPACE
2028          ; White_Space # Zl       LINE SEPARATOR
2029          ; White_Space # Zp       PARAGRAPH SEPARATOR
202F          ; White_Space # Zs       NARROW NO-BREAK SPACE
205F          ; White_Space # Zs       MEDIUM MATHEMATICAL SPACE
3000          ; White_Space # Zs       IDEOGRAPHIC SPACE

Jym77:
I have mixed feelings about this…

As far as I can tell, the definition of whitespace is mostly used in cases like "the accessible name is not only whitespaces". And it of course makes sense to say that `<img alt=" ">` is not helping anybody…

---

On the one hand, I do agree that a "ZERO WIDTH SPACE" is kind of a space and having `<img alt="[ZERO WIDTH SPACE]">` is not helping at all.

---

On the other hand, the "non only whitespaces" check is just a very basic test that there is actually an accessible name and not just a lazy developer/designer adding `alt=" "` all over the place. But this is an extremely basic test. `<img alt="aaaaaa">` is not only whitespaces and is not really helping either.

We are clearly not checking whether the accessible name makes sense or not. This does require manual testing anyway. "Not only whitespace" is the basic minimum check we can do, but is extremely barebone. Manual testing (needed to catch `<img alt="aaaaa">`) will catch `<img alt=[ZERO WIDTH SPACE]>`, so I don't feel a strong urge of adding the case.

---

[...]

PR #791, opened 2019-08-16
fix: rule "Autocomplete valid" should be inapplicable to whitespace only attribute value

jkodu:
Noticed this when running some validations/ implementations. After discussing with @WilcoFiers decided that the `autocomplete-valid` rule should not be applicable to attribute value that consists of only `whitespace`.

This PR makes the changes to:
- the applicability
- and updates the respective test cases.

Closes issue(s): 
- NA

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/autocomplete-valid-73f2c2.md
```diff
@@ -19,7 +19,7 @@ authors:
 
 ## Applicability
 
-The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` attribute that is not empty (""), except if one of the following is true:
+The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` attribute that is not empty ("") or that is not only [whitespace](#whitespace), except if one of the following is true:
```
Jym77: This should be "not empty or only whitespace", or (better) "neither empty nor only whitespace".

Currently, the alternative is "(not empty) OR (not whitespace)" and `""` being (not whitespace) pass this applicability.

*i.e.* you wrote `applicability = (!(autocomplete=="")) || (!(autocomplete is only whitespace))` instead of `applicability = !( (autocomplete=="") || (autocomplete is only whitespace) )` 
(at least, that's how I understand the sentence)

jkodu:
@Jym77 cheers for the eagle eyes catch here - https://github.com/act-rules/act-rules.github.io/pull/791#pullrequestreview-277768610. 🍺

_rules/autocomplete-valid-73f2c2.md
```diff
@@ -279,3 +271,11 @@ Non-widget element that does not participate in sequential focus navigation.
 ```html
 <input type="button" role="none" tabindex="-2" autocomplete="username" />
 ```
+
+#### Inapplicable example 11
```
Jym77: Actually, need uppercase E at *E*xample, as you probably have figured out with the tests :-D

jkodu:
Final call ends on Sep 9, 2019.

dd8:
I think there's problem with this due to the `Whitespace` definition used. If you run this example through the W3 validator, the first autocomplete is valid but the others aren't:

```
<input id="one" autocomplete=" " />
<input id="two" autocomplete=" " /> <!-- non-break space pasted in --> 
<input id="three" autocomplete="&nbsp;" /> <!-- non-break space entity --> 
<input id="four" autocomplete=" " /> <!-- Em Space U+2003 pasted in -->
<input id="five" autocomplete="&#x2003;" /> <!-- Em Space U+2003 entity -->
```

That's happening because `autocomplete` is a set of ASCII-space separated tokens
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute
but the ACT-R `whitespace` definition includes non-ASCII whitespace.
https://act-rules.github.io/glossary/#whitespace

I think ACT-R needs at least 2 whitespace definitions per this comment:
https://github.com/act-rules/act-rules.github.io/issues/642#issuecomment-522909629

CSS has three different whitespace definitions:
- syntactic whitespace
- document whitespace
- other whitespace 
https://drafts.csswg.org/css-text-3/#white-space

Similar issues happen if the whitespace definition is used on IDREFs like `for` or `aria-labelledby` because the only restriction on `id` values is they "must not contain any ASCII whitespace." 
https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute

For example the following code is valid and the label association works:
```
   <label for="&nbsp;">Test</label>
   <input id="&nbsp;">
```

_rules/autocomplete-valid-73f2c2.md
```diff
@@ -19,7 +19,7 @@ authors:
 
 ## Applicability
 
-The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` attribute that is not empty (""), except if one of the following is true:
+The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` attribute that is neither empty ("") nor only [whitespace](#whitespace), except if one of the following is true:
```
WilcoFiers: Editorial: I suggest we put this "is neither empty ("") nor only [whitespace](#whitespace)" bit in the list of exceptions, rather than have 1 inline exception, and 4 listed ones.

Jym77: I think @dd8 has a very valid point and that will require changes…
https://github.com/act-rules/act-rules.github.io/pull/791#issuecomment-524865504

In short, `autocomplete="&nbsp &nbsp"` is "only whitespace" per our definition of whitespace, but composed of two (invalid) autocomplete values (each being non-breaking space) as per autocomplete specs (break tokens on ASCII spaces (TAB/LF/FF/CR/SPACE)).  Thus, this should be a Failure (autocomplete is not valid) rather than a Inapplicable.

jkodu:
@Jym77 and @dd8, having spoken to @WilcoFiers  about the ambiguity in whitespace definition as currently being discussed - https://github.com/act-rules/act-rules.github.io/issues/642, we both believe that this PR should not be affected because of this definition.

Once a consensus is arrived on the whitespace definition is arrived, perhaps we may have to update a number of rules.

With that in mind, I am dismissing the changes requested. Cheers.

Jym77:
> @Jym77 and @dd8, having spoken to @WilcoFiers about the ambiguity in whitespace definition as currently being discussed - #642, we both believe that this PR should not be affected because of this definition.
> 
> Once a consensus is arrived on the whitespace definition is arrived, perhaps we may have to update a number of rules.
> 
> With that in mind, I am dismissing the changes requested. Cheers.

I disagree.

1. there is no need of consensus for ASCII whitespace, or space separated tokens that are needed for that rule. It's in the specs, we don't have anything to discuss.
2. What is the point of this PR? Why do we want `autocomplete="   "` to be inapplicable rather than fail? (I think it should fail since `autocomplete` needs at least one token as far as I understand the specs) It is worth making `autocomplete="&nbsp; "` inapplicable rather than fail?
3. Given that this is already an update on a published rule, I don't really see why we should make an incorrect update and more work for later rather than making the correct fix which is not that difficult given that we can point to the specs for the definition of ASCII spaces.

Jym77:
And on a more global note on this PR: I think that `autocomplete=""` should be a fail rather than inapplicable, since `autocomplete` attribute may not be empty as far as I understand https://html.spec.whatwg.org/#autofilling-form-controls:-the-autocomplete-attribute
(it may be omitted (when wearing the autofill expectation mantle) but it may not be empty)

Which make this whole PR goes in the wrong direction. I think the rule should apply to any `autocomplete` attribute (empty or not) and match the token (as Expectation 2 is doing) which would immediately fail both `autocomplete=""`, `autocomplete=" "` and `autocomplete="&nbsp;"` (all three being invalid `autocomplete` attribute as far as I understand the specs).

Also you should totally link https://www.w3.org/WAI/WCAG21/Techniques/html/H98 in the Background section.

jkodu:
@WilcoFiers thoughts on above - https://github.com/act-rules/act-rules.github.io/pull/791#issuecomment-526064612

Jym77:
More thoughts… (after discussing with @kasperisager).
 We may want to consider `autocomplete=""` (or even `autocomplete=" "`) as OK-ish even though it's not OK as per HTML specs. But it might be something auto-added by framework and where the user has little control. The idea being that `autocomplete=""` has the intend of "nothing" (that might have been overlooked by the framework) while `autocomplete="foo"` is a blatant error. That probably was the reason why the initial rule was on "non-empty" values.

In that case, I guess that a good applicability would be in the line of "any `autocomplete` attribute whose value is a non-empty list of space separated tokens". That would put both `autocomplete=""` and `autocomplete=" "` as inapplicable (no token) but keep `autocomplete="&nbsp;"` as failed (one token which is not valid).

WilcoFiers:
@Jym77 I've updated the PR, could you please have a look?

_rules/autocomplete-valid-73f2c2.md
```diff
@@ -278,4 +278,4 @@ Autocomplete attribute contains only [whitespace](#whitespace).
```
Jym77: Description of this example need to be updated to autocomplete contains no tokens.

jkodu:
Final call for this PR ended yesterday. Merging this PR.

Issue #784, opened 2019-08-15
Create tests for verifying hyperlinks used in rules


jkodu:
Create tests to check links used in rules:
- if they are broken links
- if link text is URL as against a valid name
- if link refers to only of the below outdated resources
   - WCAG20
   - Please add links that should not be referred.

Jym77:
Other outdated resource: https://www.w3.org/TR/dom41/ (instead of https://dom.spec.whatwg.org). There's currently a redirection, but we may still rather use the correct URL…

jkodu:
Have a PR in place.

PR #808, opened 2019-08-20
test: hyperlink in rules does not point to outdated content

jkodu:
- hyperlink referred in rules does not point to a blacklist items of outdated content

> Note: there is a lot of failing links.

Closes issue(s): 
- https://github.com/act-rules/act-rules.github.io/issues/784

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/__tests__/hyperlinks.js
```diff
@@ -0,0 +1,19 @@
+const describeRule = require('../../test-utils/describe-rule')
+const markdownLinkExtractor = require('markdown-link-extractor')
+const isUrl = require('is-url')
+
+const blacklistOfLinks = ['w3.org/TR/dom41', 'WCAG20']
```
jkodu: TODO: 

Add more links to blacklist
- https://www.w3.org/TR/wai-aria-1.0/, these should point to 1.1

jkodu: Ensure this test covers all markdown pages

WilcoFiers: I'm not sure of blacklisting "wcag20" outright. I'd blacklist the following instead:

://www.w3.org/TR/WCAG20/
://www.w3.org/TR/UNDERSTANDING-WCAG20/
://www.w3.org/TR/WCAG20-TECHS/
://www.w3.org/TR/wai-aria-1.0/

jkodu:
@Jym77 @annethyme  there are some hyperlinks that are failing. See - https://circleci.com/gh/act-rules/act-rules.github.io/3441?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Could either of you help find the right links?

annethyme:
@jkodu, I won't be able to make it before going on leave.

jkodu:
I need help finding the new urls for below:

| Old Url | New Url |
|---|---|
| https://www.w3.org/TR/html/editing.html#sec-sequential-focus-navigation | |
| https://www.w3.org/TR/html/editing.html#the-tabindex-attribute | |
| https://www.w3.org/TR/html/infrastructure.html#valid-integer | |
| https://www.w3.org/TR/html/document-metadata.html#the-title-element | |
| https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type | |
| https://www.w3.org/TR/html/sec-forms.html#element-attrdef-input-type | |
| https://www.w3.org/TR/html/document-metadata.html#statedef-http-equiv-refresh | |

jkodu:
@Jym77 helped with above. Thanks. Up for review now.

_rules/__tests__/hyperlinks.js
```diff
@@ -0,0 +1,26 @@
+const describeRule = require('../../test-utils/describe-rule')
+const markdownLinkExtractor = require('markdown-link-extractor')
+const isUrl = require('is-url')
+
+const blacklistOfLinks = [
```
Jym77: I think it would be nice to add some comment as to what is the correct URL, it's not a one-to-one relation, but something like:
```javascript
'://www.w3.org/TR/WCAG20/', // Use WCAG 2.1
…
'://www.w3.org/TR/html/' // use spec.whatwg.org
```
would be nice (and easy to do). 

Maybe even go all the way and have the comment/message be printed out in the test in case of error, so that people looking at test will immediately get the hint as to what to do to fix it (but that's more work…)

PR #809, opened 2019-08-20
fix: duplicate heading for testcase in rule "HTML has lang attribute" (b5c3f8)

jkodu:
When validating some test cases noticed this duplicate heading.

Closes issue(s): 
- NA

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
@WilcoFiers 
> Can we create a test that checks that the numbers increment correctly?

I think it's done as part of #810 
(well, not really, it's only checking that there is no duplicate number, but doesn't check if we skip some number…)

PR #822, opened 2019-08-21
Rule Update: Lang rules applicability and test cases

### LINKED ISSUE #546: Passing test cases need to meet the accessibility requirement

jkodu:
We are updating the test cases and applicability, because we want the test cases to be consistent with the accessibility requirement that they map towards.

Closes issue(s): 
- partially fixes #546

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


jkodu:
@WilcoFiers can you double check if I missed something based on our discussion?

_rules/lang-valid-de46e4.md
```diff
@@ -19,11 +19,11 @@ authors:
 
 ## Applicability
 
-This rules applies to each HTML or SVG element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) (work in progress) of the `body` element and has a `lang` or `xml:lang` attribute that is not empty ("").
+This rules applies to each HTML or SVG element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) (work in progress) of the `body` element and has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
 
 ## Expectation
 
-The `lang` and `xml:lang` attributes of the test target have a [valid language subtag](#valid-language-subtag) if the attribute is not empty ("").
+The `lang` and `xml:lang` attributes of the test target have a [valid language subtag](#valid-language-subtag) if the attribute is neither empty ("") nor only [whitespace](#whitespace).
```
Jym77: Given that the Applicability already rules out empty/whitespace attribute, why is it needed again in the Expectation?

_rules/html-has-lang-b5c3f8.md
```diff
@@ -86,46 +86,6 @@ The `xml:lang` attribute specified has a non-empty value. The rule expects a non
 <html xml:lang="en" lang=""></html>
 ```
 
-#### Passed Example 6
```
Jym77: I would be in favour of adding a note explicitly saying that the rule checks for presence, not validity of the value.
(cutting down the number of useless examples *is* good :-D)

_rules/html-lang-valid-bf051a.md
```diff
@@ -160,20 +160,4 @@ The rule applies to `html` element and hence usage of `xml:lang` attribute in `s
 
 ```svg
 <svg xmlns="http://www.w3.org/2000/svg" xml:lang="fr"></svg>
-```
-
-#### Inapplicable Example 3
```
Jym77: Why removing these inapplicable examples?

Why is this rule not updated with "not whitespace" in the applicability?

jkodu:
@kasperisager @DagfinnRomen can you please go through the recent changes and give this a tick or na.

jkodu:
Not: This needs to go through Final call.

_rules/html-lang-valid-bf051a.md
```diff
@@ -160,20 +168,4 @@ The rule applies to `html` element and hence usage of `xml:lang` attribute in `s
...
```
Jym77: Why have inapplicable examples 3 and 4 been removed?
(and you should even add examples with whitespace attributes)

WilcoFiers: @Jym77 these test cases don't satisfy the accessibility requirement. In order for the mapping to work, we can only have passed / inapplicable test cases that also satisfy the success criterion.

Jym77: OK. I think I see the point.

I'm still a bit uneasy with that because it means that the test cases don't match the accessibility requirement **mapping** of the rule. Namely, in that case, we say that if the rule passes or is inapplicable, then further testing is needed, but we have no example that show that further testing is actually needed.

jkodu: @Jym77 your concern about whether further testing is needed may be something that needs a bigger discussion & may apply to a larger subset of rules. Can you open an issue to enable discussion on this?

_rules/lang-valid-de46e4.md
```diff
@@ -165,7 +153,7 @@ The `lang` attribute value is not empty ("") and is not a valid primary language
 </html>
 ```
 
-#### Failed Example 6
+#### Failed Example 5
 
 The `lang` attribute value is not empty ("") and is not a valid primary language subtag.
```
carlosapaduarte: This is the same as Failed Example 4. Shouldn't one of these be for `xml:lang`?

jkodu:
Final call ends on 23rd September.

jkodu:
Final call period ends. Merging PR.

PR #836, opened 2019-08-26
fix: rule 80af7b - failing tests for rule - No keyboard trap standard navigation

audreymaniez:
Change the failed test case 2. This test case is the same as in the atomic rule  "No keyboard trap standard navigation" where it was fixed.

See the related commit : https://github.com/act-rules/act-rules.github.io/commit/af55b85eda8a8aa26a7965aaed5f23e6a267d43a#diff-ebbcb950ef3b644c6747d9b3c601e329

This PR is just applying the same correction on the composite rule.


jkodu:
This is a minor change & hence merging this without 3 approvals.

PR #831, opened 2019-08-23
aria-labelledby example for image has accessible name (23a2a8)

Jym77:
* Add an example where accessible name is given by `aria-labelledby`.
* Add final period at the end of example descriptions.
* Unify a bit example descriptions.

(Ouch github diff tool is mixing things up… The new example is Passed Example 3, the rest has been pushed down but github still tries to compare old Passed Example 4 with new Passed Example 4 instead of new Passed Example 5…)

This will require a new Final Call as new example is added.

Closes issue(s): 
- #830 

----

### **When creating PR:**
- Make sure you requesting to **pull a branch** (right side) to the `develop` branch (left side).

### **After creating PR:**
- Add yourself (and co-authors) as "Assignees" for PR.
- Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- Close the issue that the PR resolves (and make sure the issue is referenced in the top of this comment)
- Optionally request feedback from anyone in particular by assigning them as "Reviewers".

## How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

PR #849, opened 2019-08-28
Revert "aria-labelledby example for image has accessible name (23a2a8) (#831)

Jym77:
I messed up and merged too early :-(


jkodu:
Closing this. See - https://github.com/act-rules/act-rules.github.io/pull/831#issuecomment-525726740

Issue #250, opened 2018-09-11
Naming conventions for rules


annethyme:
Skimming over our list of published rules on https://auto-wcag.github.io/auto-wcag/pages/rules.html, there are big differences in the naming. Singular or plural, with or without dashes, wordy or to the bone, with or without verbs. 
Now that we have tried different things out for a while, I suggest we discuss and:
- decide on a best practice
- write down the best practice
- align the existing rules, PRs and issues to the agreement we make

WilcoFiers:
Some thoughts:

- Front load rule names so that appropriate element names / roles go first: "button has accessible name" instead of "accessible name for button"
- Use singular tense: "button has accessible name" instead of "buttons have an accessible name"
- Describe what passes, not what fails: "button has accessible name" instead of "button without accessible name"
- ...?

WilcoFiers:
Moving this off the September agenda. We need to get open PRs resolved. We'll pick it up in October.

annethyme:
How about these?

- "button has accessible name" or "button has *an* accessible name"?
- Capitalization? "*b*utton has accessible name" or "*B*utton has accessible name"
- "iFrame has name" or "iframe has name" or "Iframe has name"

WilcoFiers:
| Old name                                   | New name      |
|--------------------------------------------|---------------|
| Old rule name                              | New rule name |
| ARIA Attribute Is Valid                    | ARIA attributes are valid |
| ARIA State Or Property Allowed             | ARIA attributes are allowed for role |
| ARIA State Or Property Has Valid Value     | ARIA attribute values are valid |
| Audio-Only As A Media Alternative For Text | audio elements are a media alternative |
| Autocomplete Valid                         | Autocomplete attributes are valid |
| Buttons Have An Accessible Name            | buttons have an accessible name |
| Form Field Has Accessible Name             | form fields have an accessible name |
| Form Field Label Is Descriptive            | form field labels are descriptive |
| HTML Page Has A Title                      | html page has a title |
| HTML Has Lang Attribute                    | html element has a lang attribute |
| HTML Lang And Xml:lang Match               | html element has matching lang and xml:lang attribute |
| Heading Is Descriptive                     | headings are descriptive |
| Id Attribute Is Unique                     | id attributes are unique |
| Image Has Accessible Name                  | images have an accessible name |
| Links Have An Accessible Name              | links have an accessible name |
| Meta-Refresh No Delay                      | meta element does not have a delayed redirect |
[...]

PR #851, opened 2019-08-28
chore: Improve consistency around lang rules

WilcoFiers:
This is based on feedback from the ACT Task Force. This basically aligns the language use between these 5 rules. There is obviously more that can be improved, in other rules as well as in these. But I ask reviewers to look only at these changes.

This is also related to consistency in rule titles, #250 The titles used in this are not the same as proposed in that issue. We can revisit titles as part of 250.

**Important**: I do not believe this change requires a final call, although 3 reviews are still necessary for this to go in.

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/html-lang-valid-bf051a.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: bf051a
-name: Validity of HTML Lang attribute
+name: HTML page languages are valid 
 rule_type: atomic
 description: |
-  This rule checks that the `lang` or `xml:lang` attribute has a valid language subtag.
+  The rule checks that all HTML pages with a `lang` or `xml:lang` attribute on the root element, have a valid primary language subtag.
```
Jym77: I believe the descriptions usually start with "*this* rule" and not "the rule".

_rules/html-xml-lang-match-5b7ae0.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: 5b7ae0
-name: HTML lang and xml:lang match
+name: HTML page lang and xml:lang attributes match
 rule_type: atomic
 description: |
-  The rule checks that for the `html` element, there is no mismatch between the primary language in non-empty `lang` and `xml:lang` attributes, if both are used.
+  The rule checks that all HTML pages with both a `lang` and `xml:lang` attributes on the root element, have the same primary language subtag.
```
Jym77: "the rule" => "*This* rule"

_rules/lang-valid-de46e4.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: de46e4
-name: Valid body lang attribute
+name: Elements with lang attribute in body are valid 
 rule_type: atomic
-description: |
-  This rule checks that `lang` or `xml:lang` attributes on elements within the `body` of a web page have a valid language subtag.
+description: |	
+  The rule checks that all elements in the body of HTML pages, with a `lang` or `xml:lang` attribute, have a valid primary language subtag.
```
Jym77: *This* rule

_rules/image-has-name-23a2a8.md
```diff
@@ -53,55 +53,64 @@ There is a known combination of a popular browser and assistive technology that
 
 #### Passed Example 1
 
-The HTML `img` element has an accessible name
+The HTML `img` element has an accessible name given by the `alt` attribute.
```
Jym77: Arf. Me messing up with the image-has-name early merge is now causing it to appear in that PR where it shouldn't :-(

You need to revert that file to its previous version (the one that is now back on `develop`), I guess.
Sorry ☹️

_rules/html-has-title-2779a5.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: 2779a5
-name: HTML Page has a title
+name: HTML pages have a title
 rule_type: atomic
 description: |
-  This rule checks that the HTML page has a title.
+  This rule checks all HTML pages have a title.
```
carlosapaduarte: checks all > checks that all

_rules/html-has-lang-b5c3f8.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: b5c3f8
-name: HTML has lang attribute
+name: HTML pages have lang attribute
```
carlosapaduarte: Why 'pages' instead of 'page'?
The applicability if for the root element of the web page. Also, in 'HTML page languages are valid' you use the singular (but in 'HTML pages have a title' you use the plural - so it's not coherent)
I do prefer the singular form instead of using plural by default, because that leaves us the plural for any instance where we have rules that apply to sets

_rules/html-has-title-2779a5.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: 2779a5
-name: HTML Page has a title
+name: HTML pages have a title
```
carlosapaduarte: Same comment re: pages vs page

_rules/html-xml-lang-match-5b7ae0.md
```diff
@@ -19,11 +19,13 @@ authors:
 
 ## Applicability
 
-The root element of the page, if it is an `html` element with both `lang` and `xml:lang` attributes that has a [valid language subtag](#valid-language-subtag).
+The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element with both `lang` and `xml:lang` attributes that has a [valid language subtag](#valid-language-subtag).
```
carlosapaduarte: that has a [valid language subtag](#valid-language-subtag) > that have [valid language subtags](#valid-language-subtag)

_rules/lang-valid-de46e4.md
```diff
@@ -23,7 +23,7 @@ This rules applies to each HTML or SVG element that is a [descendant](https://do
 
 ## Expectation
 
-The `lang` and `xml:lang` attributes of the test target have a [valid language subtag](#valid-language-subtag) if the attribute is not empty ("").
+For each test target, the `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is not empty ("").
```
carlosapaduarte: `lang` and `xml:lang` > `lang` or `xml:lang`
(to match the applicability)

_rules/html-has-lang-b5c3f8.md
WilcoFiers: The group consensus on page titles (see #250) is that plural is preferable. "HTML page languages" is also plural, so I think that is consistent.

The other thing to consider is collections of pages. If you are testing WCAG on a collection of pages, all of them have to have a lang attribute. This is for example true for EPUB, which is a single EPUB page, that is a bundle of HTML web pages.

_rules/html-has-title-2779a5.md
WilcoFiers: Using plural seems to have a majority preference in the group (it wasn't mine), so I'm sticking with that until I hear otherwise.

_rules/lang-valid-de46e4.md
WilcoFiers: Interesting point. I think it can go either way. That's not related to this change though. Please open a separate issue.

_rules/html-has-lang-b5c3f8.md
carlosapaduarte: I was only able to find support for single tense in #250 (that's also what I prefer). Perhaps there was consensus on a call. No need to drag this.

WilcoFiers: Right you are... :/ I'm not sure where my head was at yesterday. Sorry about that!

WilcoFiers:
@carlosapaduarte I've updated the titles. Can you give it another look?

_rules/html-has-lang-b5c3f8.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: b5c3f8
-name: HTML has lang attribute
+name: HTML page has lang attribute
 rule_type: atomic
 description: |
-  This rule checks that the `html` element has a non-empty `lang` or `xml:lang` attribute.
+  This rule checks that an HTML pages has a non-empty `lang` or `xml:lang` attribute.
```
carlosapaduarte: pages > page

_rules/html-title-is-descriptive-c4a8a4.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: c4a8a4
-name: Document title is descriptive
+name: HTML page title is descriptive
 rule_type: atomic
 description: |
-  This rule checks that the document title describes the topic or purpose of the document.
+  This rule checks that first title in an HTML page describes the topic or purpose of that page.
```
carlosapaduarte: that first > that the first

_rules/html-lang-valid-bf051a.md
```diff
@@ -19,11 +19,13 @@ authors:
...
 
-The root element of the page, if it is an `html` element with a `lang` and/or `xml:lang` attribute that is not empty ("").
+The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element with a `lang` and/or `xml:lang` attribute that is not empty ("").
+
+**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
 
 ## Expectation
 
-The `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is not empty ("").
+For each test target, the `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is not empty ("").
```
Jym77: This expectation should get the same wording as the one you did in https://github.com/act-rules/act-rules.github.io/blob/0d3d29695abbffa7c09eda289ae72f72aaf2ad8a/_rules/lang-valid-de46e4.md (**unless** the attribute is empty).
And does have the same issue about the non-existing attribute.

_rules/html-has-lang-b5c3f8.md
```diff
@@ -19,11 +19,13 @@ authors:
 
 ## Applicability
 
-The root element of the [page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element.
+The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element.
```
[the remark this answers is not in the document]
WilcoFiers: I don't have a problem with that, but I do think this is out of scope for this PR. I'll open a separate issue for it. https://github.com/act-rules/act-rules.github.io/issues/869

_rules/lang-valid-de46e4.md
```diff
@@ -23,7 +23,7 @@ This rules applies to each HTML or SVG element that is a [descendant](https://do
 
 ## Expectation
 
-The `lang` and `xml:lang` attributes of the test target have a [valid language subtag](#valid-language-subtag) if the attribute is not empty ("").
+For each test target, the `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) unless the attribute is empty ("").
```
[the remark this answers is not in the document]
WilcoFiers: I disagree. The statement is still true if the target only has a `lang` attribute, but I agree this can probably be worded better. Given that this is the original wording of the rule I'm using, I think this is out of scope for this change. Feel free to open a separate issue for this.

PR #868, opened 2019-09-03
Definition of context

Jym77:
3 rules currently in progress need a definition of context. By putting it here, we can factor the work and hopefully avoid a merge deadlock later one…

Once in a reasonably correct state, the defs will be shipped with whichever rule seems closest to be done.

Closes issue(s): 
- Needed for #447, #555, and #693.

----

## How to Review And Approve


pages/glossary/context.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Context
+key: context
+---
+
+The context of some piece of information is any other information that helps a user understand it.
```
Brynanders: The context of some piece of information is any other additional information that helps a user understand the original piece of information.

haha its like a tongue twister :)

carlosapaduarte: Is this better or worse? 
The context of a piece of information is any other information that helps a user understand it by clarifying its meaning.

pages/glossary/context.md
```diff
@@ -0,0 +1,12 @@
...
+
+Both the [visual context](#visual-context) and the [programmatically determined context](#programmatically-determined-context) are part of the context.
```
carlosapaduarte: Suggestion:
are part of the context > contribute to the context

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Programmatically determined context
+key: programmatically-determined-context
+---
+
+The programmatically determined context of some piece of information (_e.g_ an element) is all the additional information that has a [programmatically determinable](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) [relationship](https://www.w3.org/TR/WCAG21/#dfn-relationships) to it.
```
carlosapaduarte: of some piece > of a piece
e.g > e.g.

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,8 @@
...
+
+It includes, but is not restricted to, `aria-label` attribute, element that are referred by `aria-labelledby` or `aria-describedby` attributes, ancestor `label` elements, `label` elements referring the current element with a `for` attribute, ancestor elements, table header associated with the cell containing the information, text inside the same paragraph (`p` element), table cell, list item, …
```
carlosapaduarte: element that are > element that is

pages/glossary/visual-context.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Visual context
+key: visual-context
+---
+
+The visual context of some piece of information (_e.g._ an element) is everything that is visually or logically located near it. Only [visible](#visible) information may be part of the visual context.
+
+Visual context may include, and is not limited to, headings, text in the same sentence or paragraph, …
```
carlosapaduarte: and > but

pages/glossary/context.md
```diff
@@ -0,0 +1,12 @@
...
+
+**Note**: The maximal "distance" (both visual distance on the device, and tree distance within the DOM tree) between some information and its context depends on the precise situation, and on what kind of information is looked at.
```
ShadowBB: The note mentions a maximum "distance" but the actual definition doesn't. As written "context" can be a book I read about the subject 4 years ago.

pages/glossary/visual-context.md
ShadowBB: I do believe that paragraphs and headings have a programmatically determinable relationship. Is there an overlap between programmatically determined context and visual context? If so I would mention that in a note and I would add headings to the examples of both.

pages/glossary/visual-context.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Visual context
+key: visual-context
+---
+
+The visual context of some piece of information (_e.g._ an element) is everything that is visually or logically located near it. Only [visible](#visible) information may be part of the visual context.
```
ShadowBB: I am inclined to ask for a more precise definition of "near". I think this is an incredibly difficult definition to write and applaud your ambition.

pages/glossary/context.md
Jym77: The "distance" is used by the "located near it" part of the definition (near being the distance).
I agree that it is only present in the visual context def, and not in the prog. det. one 😞 
I also agree that it is indeed a bit blurry…

pages/glossary/visual-context.md
Jym77: Correct. I do not know why I missed headings and paragraph in that list 😞 
(and yes, there is an overlap between visual and prog. det. context and I see no problem with that)

pages/glossary/visual-context.md
Jym77: That is indeed one of the main problem with that definition and the reason of the "distance" note. The "near" that need to be used depends a lot on why the user needs the definition in their current use case (the "context" in which the definition is used, but I'd like to avoid using this term here to keep things not too complicated).

If this definition is used to differentiate otherwise similar links in a table, then "near" will be very near, typically only the cell in which the link resides. If used to check whether such a link is descriptive, then "near" can be further away and include the table headers or caption, and even the headings above it. Possibly far away from it…

Jym77:
@WilcoFiers 
> @Jym77 I don't think merging definitions before merging the rules they are used in is a good idea. It's impossible to judge the correctness of a definition, without knowing how it will be used. The rules in which they are used are still in draft, and they can change. I realise you are doing this to speed things up, but I think by doing this, reviewing becomes guess work. This is essentially asking the reviewers to guess whether or not the a draft that doesn't exist yet will work with the definition that is proposed today.

* The 3 rules that are currently in need of this definition are linked from the PR description, so it is fairly easy to check them. They are all in a fairly advanced state, even though they are likely going to change.
* The definition may evolve together with the rules that use it. Exactly in the same way that a definition that is re-used in a new rule 1 year after being written may evolve if it is not perfect for that rule. Obviously, we want to limit that, but in the same way that we want to limit definitions changes.

> In my opinion, these definitions need to be included in one of the rules, that you get through a CFC, and once you do, use those definitions in the other two rules.

* Putting the definition together with a single rule barely improve the "guess work" that you complain about. Sure if it's with rule A, you know a bit about its usage, but you're not more informed about its usage in rule B, and rule C. You might even not have the links at all since they won't be in the PR description.
* Putting the definition together with a single rule also creates an annoying bottleneck. If it's with rule A, then rule B and C cannot be merged before rule A which can very well be delayed for completely unrelated reason. I don't think any of us can accurately predict which of these rules will be merged first.
* And of course, copying the definition in all rules is a guarantee for drift and later merge nightmare.

pages/glossary/context.md
Jym77: I added a note in "context" to rule out the book that you have read 4 years ago.

pages/glossary/visual-context.md
Jym77: I added some note in "visual context" to give a better (but still blurry) definition of "near".

Jym77:
I think one way might be to copy the defs into the first rule which is ready. In the meantime we can keep discussion over the defs here.
Obviously, some changes might be needed once the defs are put in rule context, but hopefully these will be fairly minor.

pages/glossary/context.md
```diff
@@ -0,0 +1,16 @@
...
+
+**Note**: When used to differentiate between two pieces of information, context should not include anything that is "shared" by these two pieces of information: _e.g._, headings that are above only one of these pieces of information may be used as context, but headings that are above both should not.
+
+**Note**: Some elements, _e.g._ headings, may be part of both the [visual context](#visual-context) and the [programmatically determined context](#programmatically-determined-context). In many cases, a document which has proper semantic formatting while have all the [visual context](#visual-context) of its elements included in their [programmatically determined context](#programmatically-determined-context) (on the other hand, [programmatically determined context](#programmatically-determined-context) can often include information that is not [visible](#visible), _e.g._ a hidden element referred by `aria-describedby` to provide some alternative text description).
```
DagfinnRomen: The last sentence is incomplete. Something seems missing around "while". Or perhaps replace "while have" with "has".

pages/glossary/context.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Context
+key: context
+---
+
+The context of a piece of information is any other information that helps a user understand it by clarifying its meaning.
+
+Both the [visual context](#visual-context) and the [programmatically determined context](#programmatically-determined-context) contribute to the context.
+
+**Note**: To be considered as context, content must be available similarly to all users with similar abilities. Thus, for example, content from another [document] is not part of the context as there is no guarantee that all users have access to it.
```
DagfinnRomen: Is is possible to simply say that to be considered context, the content must be on the same web page?

pages/glossary/programmatically-determined-context.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Programmatically determined context
+key: programmatically-determined-context
+---
+
+The programmatically determined context of a piece of information (_e.g._ an element) is all the additional information that has a [programmatically determinable](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) [relationship](https://www.w3.org/TR/WCAG21/#dfn-relationships) to it.
+
+It includes, but is not restricted to, `aria-label` attribute, element that is referred by `aria-labelledby` or `aria-describedby` attributes, ancestor `label` elements, `label` elements referring the current element with a `for` attribute, ancestor elements, table header associated with the cell containing the information, headings, text inside the same paragraph (`p` element), table cell, list item, …
```
DagfinnRomen: You could also add "sentence" to the list.

From the WCAG definition of programmatically determined link context:
In HTML, information that is programmatically determinable from a link in English includes text that is in the same paragraph, list, or table cell as the link or in a table header cell that is associated with the table cell that contains the link. 

Note: "Since screen readers interpret punctuation, they can also provide the context from the current sentence, when the focus is on a link in that sentence."

https://www.w3.org/TR/WCAG21/#dfn-programmatically-determined-link-context

pages/glossary/visual-context.md
```diff
@@ -0,0 +1,14 @@
...
```
DagfinnRomen: Defnining "near" is difficult. Could it mean all information on the web page? Or all information in the viewport? Again this carries with it a slew of other issues...

pages/glossary/programmatically-determined-context.md
DagfinnRomen: And what about the title-attribute?

pages/glossary/visual-context.md
```diff
@@ -0,0 +1,14 @@
...
+
+**Note**: As a rule of thumb, visual context should be close enough to be displayed on the device at the same time as the piece of information it clarifies. Because device sizes vary wildly and content can further be zoomed and moved around, this is however not a strong requirement.
+
+**Note**: Elements that are visually located further from the piece of information may still be considered as visual context if they both have a visual emphasise and are logically close to the piece of information (_e.g._ near it in the [node tree](https://dom.spec.whatwg.org/#concept-node-tree)). For example, a heading can be considered as visual context for its full section of text even though it might be located far away from the end of that content, and might even be off screen due to scrolling. But it has a strong logical connection with the content which indicates that it is giving context for all of it.
+
+**Note**: Visual context that is located before (in reading order) the piece of information it clarifies is often more useful than visual context located after. Indeed, it is easier for users to use context that they have already read than context that is yet to be read.
```
DagfinnRomen: I like these three notes. They explain that context is (for the lack of a better term) context sensitive :)

pages/glossary/context.md
Jym77: I'm not sure it's that clear… 

If I'm looking at a website about movies, it's likely that the word "theatre" means "movie theatre" while on a WW2 history site, it would mean "Theatre of war (Europe/Pacific)". In that case, the topic of the website would be context. Or maybe we consider that every single page of the site is about that topic and has enough information to make the difference? Will it always be that clear?

(that note was added after Brian's remark that "a book [user] read 5 years ago should no be context", hence the formulation).

pages/glossary/visual-context.md
Jym77: Yes, "near" is annoying… The problem with viewport or similar is that "on screen" info is that this can change by zooming or scrolling and I don't want, say, a header to disappear from context just because it's been scrolled 1 line out of screen…

pages/glossary/programmatically-determined-context.md
carlosapaduarte: I've updated the definition of programmatically determined context in #555

pages/glossary/visual-context.md
Jym77: There are a bunch of notes clarifying the meaning of "near", both in that definition and in the global definition of "context". Do you think we need more?

DagfinnRomen: I agree with you. I do not have anything to add.

pages/glossary/programmatically-determined-context.md
DagfinnRomen: Editorial: Remove the ... (ellipsis).

Any thoughts on including the title attribute?

pages/glossary/visual-context.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Visual context
+key: visual-context
+---
+
+The visual context of some piece of information (_e.g._ an element) is everything that is visually or logically located near it. Only [visible](#visible) information may be part of the visual context.
+
+It may include, but is not limited to, headings, text in the same sentence or paragraph, …
```
DagfinnRomen: Editorial: Remove the ... (ellipsis).

pages/glossary/programmatically-determined-context.md
Jym77: I chose to not change it here given that @carlosapaduarte onboarded it with #555 and I didn't want to duplicate changes.

Jym77:
Since we seem to be reaching an agreement, or as close as possible to one without surrounding rules, I'm closing this PR and onboarding the definitions with #905.

PR #873, opened 2019-09-06
New rule: No keyboard shortcuts use only letter, number, sign or punctuation keys

joao-vicente:
Set of manual rules for SC 2.1.4 Character Key Shortcuts

Reviews required

----

### **When creating PR:**
- Make sure you requesting to **pull a branch** (right side) to the `develop` branch (left side).

### **After creating PR:**
- Add yourself (and co-authors) as "Assignees" for PR.
- Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- Close the issue that the PR resolves (and make sure the issue is referenced in the top of this comment)
- Optionally request feedback from anyone in particular by assigning them as "Reviewers".

## How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

PR #421, opened 2019-02-12
New Rule: 2.4.1 Bypass Blocks (cf77f2)

annethyme:
Rule for 2.4.1 Bypass Blocks.

Closes issue: https://github.com/auto-wcag/auto-wcag/issues/404

# How to Review And Approve
- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.

PR #880, opened 2019-09-09
New rule: Device motion based functionality can be disabled

joao-vicente:
Set of manual rules for SC 2.5.4: Motion Actuation

Reviewers required

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/device-motion-bf0de1.md
```diff
@@ -0,0 +1,327 @@
...
+
+- Device orientation events may include sensitive data. Implementations must fire events only on [secure browsing contexts](https://www.w3.org/TR/secure-contexts/)
+
+## Background
+
+- [Understanding Success Criterion 2.5.4: Motion Actuation](https://www.w3.org/WAI/WCAG21/Understanding/motion-actuation.html)
+- [G213: Provide conventional controls and an application setting for motion activated input](https://www.w3.org/WAI/WCAG21/Techniques/general/G213.html)
+- [DeviceOrientation Event Specification](https://www.w3.org/TR/orientation-event/)
+- [Detecting device orientation](https://developer.mozilla.org/en-US/docs/Web/API/Detecting_device_orientation)
+- [DeviceMotionEvent](https://developer.mozilla.org/en-US/docs/Web/API/DeviceMotionEvent)
```
jkodu: We do not refer to MDN documentation usually. Update this with something from w3 spec.

_rules/device-motion-bf0de1.md
```diff
@@ -0,0 +1,327 @@
...
+```html
+<div>document content</div>
+```
+
+#### Passed Example 2
+
+[HTML document](https://dom.spec.whatwg.org/#concept-document) has [functionality](https://www.w3.org/WAI/WCAG21/Understanding/motion-actuation.html#dfn-functionality) that can be operated by device motion or user motion and can also be operated by [user interface components](https://www.w3.org/WAI/WCAG21/Understanding/motion-actuation.html#dfn-user-interface-component) and can be disabled
+
+```html
+<html>
```
jkodu: you are going to get a lot of test failures because the composed HTML is not complete. Please refer to other rules in the repository where we ignore some checks done by `_rules/__tests__/testcases.js`

Please ensure you run npm test locally and ignore what needs to be ignored or fix accordingly.

Eg: see - https://github.com/act-rules/act-rules.github.io/blob/develop/_rules/html-has-title-2779a5.md#L21

_rules/motion-disabled-de951c.md
```diff
@@ -0,0 +1,188 @@
...
+  - Accessibility tree
+
+authors:
+  - João Vicente
+  - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any [HTML document](https://dom.spec.whatwg.org/#concept-document).
```
Jym77: Remark: you can use "short style" refs in markdown, _i.e._ use `[link name][]` in the text and have, at the end of the file, a bunch of lines with `[link name]: link destination 'link description'` That will make the files easier to write and to read (because you only need to write the full URL once).

For example:
```markdown
The rule applies to any [HTML document][].

[HTML document]: https://dom.spec.whatwg.org/#concept-document 'Definition of Document'
```

(and if you need to use another display name for the link than the "link name" (typically for plural), you can use, _e.g._ `[HTML documents][HTML document]`)

_rules/motion-replaced-5d9784.md
```diff
@@ -0,0 +1,193 @@
...
+      function load() {
+        //slider behavior
+        var slider = document.getElementById("motionSlider");
+        var output = document.getElementById("demo");
+        output.innerHTML = slider.value;
+        slider.oninput = function () {
+          output.innerHTML = this.value;
+        }
+
+        document.getElementById('increase').addEventListener('click', function() {
```
Jym77: I think I'd rather go for a named Javascript function (out of `load`) and a `onClick` on the button. Makes it easier to understand what happen since it is immediate that clicking on the button does something.

_rules/motion-disabled-de951c.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Passed Example 1
+
+[HTML document](https://dom.spec.whatwg.org/#concept-document) does not have [functionality](https://www.w3.org/WAI/WCAG21/Understanding/motion-actuation.html#dfn-functionality) that can be operated by device motion or user motion
+
+```html
+<div>document content</div>
+```
+
+#### Passed Example 2
```
Jym77: I understand that this example is failing the SC (because the function cannot be operated by interface control, _i.e._ it fails the other rule), which seems to be something that we normally avoid.

_rules/motion-replaced-5d9784.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Passed Example 1
+
+[HTML document](https://dom.spec.whatwg.org/#concept-document) does not have [functionality](https://www.w3.org/WAI/WCAG21/Understanding/motion-actuation.html#dfn-functionality) that can be operated by device motion or user motion
+
+```html
+<div>document content</div>
+```
+
+#### Passed Example 2
```
Jym77: I understand that this example is failing the SC (because the function cannot be disabled, _i.e._ it fails the other rule), which seems to be something that we normally avoid.

_rules/device-motion-bf0de1.md
```diff
@@ -0,0 +1,327 @@
...
+## Expectation
+
+For each test target, the outcome of all of the following rules is passed:
+
+- [Motion can be replaced by user interface components](https://act-rules.github.io/rules/5d9784)
+- [Motion can be disabled](https://act-rules.github.io/rules/de951c)
+
+## Assumptions
+
+- This test assumes that motion is not [essential](https://www.w3.org/WAI/WCAG21/Understanding/motion-actuation.html#dfn-essential) for the function
```
audreymaniez: and that _The motion is used to operate functionality through an accessibility supported interface;_  ? no?

_rules/motion-disabled-de951c.md
```diff
@@ -0,0 +1,188 @@
+---
+id: de951c
+name: Motion can be disabled
```
audreymaniez: Suggesting rephrasing:

> name: **Motion actuation** can be disabled

_rules/motion-actuation-bf0de1.md
```diff
@@ -0,0 +1,729 @@
+---
+id: bf0de1
+name: Motion actuation can be disabled and be operated by user interface components
+rule_type: atomic
+
+description: |
+  This rule checks that functionality that can be operated by device motion or user motion can also be operated by user interface components and responding to the motion can be disabled to prevent accidental actuation
```
joecron: Paragraph missing "." at the end.

jkodu:
@joao-vicente can you please dismiss reviews once you have caught up with the changes requested. Thanks.

_rules/motion-actuation-bf0de1.md
```diff
@@ -0,0 +1,737 @@
...
+        document.getElementById('decrease').addEventListener('click', function() {
+          slider.value--;
+          output.innerHTML = slider.value;
+        });
+
+        //slider motion detection
+        function handleOrientation(event) {
+
+          var z = document.getElementById("disable");
+          if (z.checked == false) {
```
Jym77: Please use `===` (triple =) for all your tests.

Jym77:
```js
if (foo === false) {
  no()
  } else {
  yes()
}
```
should be
```js
if (foo) {
  yes()
  } else {
  no()
}
```

_rules/motion-actuation-bf0de1.md
```diff
@@ -0,0 +1,737 @@
...
+          }
+        }
+
+        if (evCache.length == 2) {
+          var curDiff = Math.abs(evCache[0].clientX - evCache[1].clientX);
+
+          if (prevDiff > 0) {
+            if (curDiff > prevDiff) {
+              var disable = document.getElementById("disable");
+              if (disable.checked == false) {
```
Jym77: Should just be `if (!disable.checked)`

_rules/motion-actuation-bf0de1.md
```diff
@@ -0,0 +1,737 @@
...
+            output.innerHTML = slider.value;
+          } else if (x < -20) {
+            slider.value--;
+            output.innerHTML = slider.value;
+          }
+        }
+
+        window.addEventListener('deviceorientation', handleOrientation);
+      }
+    </script>
```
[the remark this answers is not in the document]
joao-vicente: I understand what you're saying, but most examples have subtle variations of the same code, and it's not reasonable to move everything to a separate asset test. Since most examples have their differences in the script code, i'll leave the code as it is. Nevertheless i did optimize most scripts code.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,324 @@
...
+      <p aria-live="polite">Slider Value: <span id="output">50</span></p>
+    </div>
+    <div>
+      <input type="checkbox" id="disableMotion">
+      <label for="disableMotion"></label>
+    </div>
+  </body>
+</html>
+```
+
```
Jym77: We need an example where the control for disabling is not a UI component. (that's the last part of the expectation which is currently not tested)

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+    - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any [HTML document][] that has [functionality][] that can be operated by [device motion][].
+
+## Expectation
+
+For the test target, [device motion][] can also be operated by [user interface components][] and they should be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77:
```suggestion
For the test target, [device motion][] can also be operated by [user interface components][] and they must be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Here also, bullet list might be easier to read.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Assumptions
+
+- This test assumes that the motion to operate [functionality][] is not used through an [accessibility supported][] interface
```
Jym77:
```suggestion
- This test assumes that the motion to operate [functionality][] is not used through an [accessibility supported][] interface.
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+- This test assumes that motion is not [essential][] for the [functionality][]
```
Jym77:
```suggestion
- This test assumes that motion is not [essential][] for the [functionality][].
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Accessibility Support
+
+- Device [orientation events][] may include sensitive data. Implementations must fire events only on [secure browsing contexts][]
```
Jym77:
```suggestion
- Device [orientation events][] may include sensitive data. Implementations must fire events only on [secure browsing contexts][].
```
Same remark about this not being an accessibility concern.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        function decreaseSlider() {
+          slider.value--;
+          output.innerHTML = slider.value;
+        }
+
+        function handleOrientation(event) {
+          if (event.gamma > 20) {
+            slider.value++;
```
Jym77:
```suggestion
            increaseSlider();
```
Even if that repeat the `output.innerHTML` change, this make the intend clearer and the fact that both clicking and moving do the same thing immediately obvious.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+          } else if (event.gamma < -20) {
+            slider.value--;
```
Jym77:
```suggestion
            decreaseSlider();
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,355 @@
...
+
+#### Passed Example 1
+
+The [HTML document][] has [functionality][] that can be operated by [device motion][] and can also be operated by [user interface components][] and they are [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
+
+```html
+<html>
+  <head>
+    <title>Passed Example 1</title>
+    <script>
```
Jym77: Again, this script looks like it's identical in all Failed Examples except the first. It should be possible to refactor it in a test asset. Failed Example 1 can also be using it if this `activateSlider` has a `options` object with a `allowUIControl` or similar field.

Actually, both code for this rule and for "device motion can be disabled" can be put together. `activateSlider` will take an `options` object with two fields: `canBeDisabled` and `allowUIControl`. Given that `undefined` is a falsy value in JS, testing `options.canBeDisabled` will return `false` if not set. Hence, the calls need only to be:
- `activateSlider( {canBeDisabled: false} )` => other rule, Failed Example 1.
- `activateSlider( {canBeDisabled: true} )` => other rule, all other examples.
- `activateSlider( {allowUIControl: false} )` => this rule, Failed Example 1.
- `activateSlider( {allowUIControl: true} )` => this rule, all other examples.

(I'm in favour of keeping the explicit setting of the field to `false` even if not strictly necessary because this is making obvious what the example is illustrating.)

These calls are explicit enough that there is no need to perfectly remember the JS in order to understand the examples here. So we can safely hide the JS away.

_rules/motion-actuation-bf0de1.md
```diff
@@ -0,0 +1,1459 @@
...
+
+The rule applies to any [HTML document][] that has [functionality][] that can be operated by [device motion][] or user motion.
+
+## Expectation (1)
+
+For the test target, [device motion][] and user motion [can be disabled][] and the control for disabling them must be a [user interface component][] and it should be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
+
+## Expectation (2)
+
+For the test target, [device motion][] and user motion can also be operated by [user interface components][] and they should be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77: Expectations for Composite rules can only look at the outcomes of each individual atomic rule. See, for example, https://act-rules.github.io/rules/80af7b#expectation

---
Additionally, in that case, I believe that the Expectation will be "the outcome of each of these rules is either Pass or Inapplicable". This is supported by the fact that each individual atomic rule is mapping to the SC (failing a single atomic rule is failing the SC).

This means that the composite rule **is not needed at all**! We already know that failing a single atomic rule is failing the SC, the composite rule is currently not giving any additional knowledge and is just useless.

_rules/user-motion-disabled-ed12b1.md
```diff
@@ -0,0 +1,574 @@
...
+    - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any [HTML document][] that has [functionality][] that can be operated by user motion.
+
+## Expectation
+
+For the test target, user motion [can be disabled][] and the control for disabling it must be a [user interface component][] and it should be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77:
```suggestion
For the test target, user motion [can be disabled][] and the control for disabling it must be a [user interface component][] and it must be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
+bullet list.

_rules/user-motion-disabled-ed12b1.md
```diff
@@ -0,0 +1,574 @@
...
+
+## Assumptions
+
+- This test assumes that the motion to operate [functionality][] is not used through an [accessibility supported][] interface
```
Jym77:
```suggestion
- This test assumes that the motion to operate [functionality][] is not used through an [accessibility supported][] interface.
```

_rules/user-motion-disabled-ed12b1.md
```diff
@@ -0,0 +1,574 @@
...
+- This test assumes that motion is not [essential][] for the [functionality][]
```
Jym77:
```suggestion
- This test assumes that motion is not [essential][] for the [functionality][].
```

_rules/user-motion-disabled-ed12b1.md
```diff
@@ -0,0 +1,574 @@
...
+
+## Accessibility Support
+
+- Device [orientation events][] may include sensitive data. Implementations must fire events only on [secure browsing contexts][]
```
Jym77:
```suggestion
- Device [orientation events][] may include sensitive data. Implementations must fire events only on [secure browsing contexts][].
```
+not accessibility concern.

_rules/user-motion-disabled-ed12b1.md
```diff
@@ -0,0 +1,574 @@
...
+        vertical-align: middle;
+        display: table-cell;
+        height: 50vh;
+        width: 100vw;
+      }
+      #target {
+        background: white;
+        border: 1px solid black;
+      }
+    </style>
```
Jym77: Styling is shared by all Passed/Failed examples (except the `:first-child` part in Failed Example 1) => factor it away in a test asset. (plus, if possible keep the `:first-child` part in Failed Example 1, or remove it everywhere else, to use the same styling).

_rules/user-motion-disabled-ed12b1.md
```diff
@@ -0,0 +1,574 @@
...
+    <script>
```
Jym77: Code seems to be shared by all Passed/Failed Example, except Failed Example 1 (canot be disable). Again, this can be refactored in a test asset and `init` can take an `options` parameter to handle the `canBeDisabled` part. Maybe rename it `activateColorPinch` or something similar to make the intend more obvious.

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+    - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any [HTML document][] that has [functionality][] that can be operated by user motion.
+
+## Expectation
+
+For the test target, user motion can also be operated by [user interface components][] and they should be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77:
```suggestion
For the test target, user motion can also be operated by [user interface components][] and they must be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
+bullet list

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+
+## Assumptions
+
+- This test assumes that the motion to operate [functionality][] is not used through an [accessibility supported][] interface
```
Jym77:
```suggestion
- This test assumes that the motion to operate [functionality][] is not used through an [accessibility supported][] interface.
```

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+- This test assumes that motion is not [essential][] for the [functionality][]
```
Jym77:
```suggestion
- This test assumes that motion is not [essential][] for the [functionality][].
```

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+
+## Accessibility Support
+
+- Device [orientation events][] may include sensitive data. Implementations must fire events only on [secure browsing contexts][]
```
Jym77:
```suggestion
- Device [orientation events][] may include sensitive data. Implementations must fire events only on [secure browsing contexts][].
+not accessibility concern.
```

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+
+#### Passed Example 1
+
+The [HTML document][] has [functionality][] that can be operated by user motion and can also be operated by [user interface components][] and they are [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
+
+```html
+<html>
+  <head>
+    <title>Passed Example 1</title>
+    <style>
```
Jym77: Styling is shared between examples, and also with "user motion can be disabled" => refactor in test asset.

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+        display: table-cell;
+        height: 50vh;
+        width: 100vw;
+      }
+      #target {
+        background: white;
+        border: 1px solid black;
+      }
+    </style>
+    <script>
```
Jym77: Code seems to be shared between examples, and also with "user motion can be disabled" => refactor in test asset, use `options` object for the small variability that is needed.

_rules/user-motion-user-interface-9d42d6.md
```diff
@@ -0,0 +1,592 @@
...
+  <body onload='init();' style='touch-action:none'>
+    <div id='target'>Touch and Hold with 2 pointers, then pinch in or out horizontally.<br/>
+        The background color will change to pink if the pinch is opening (Zoom In) 
+        or changes to lightblue if the pinch is closing (Zoom out).
+    </div>
+    <div aria-hidden="true">
+      <button onclick="changeBackgroundColor('lightblue');">Change background color to lightblue</button>
+      <button onclick="changeBackgroundColor('pink');">Change background color to pink</button>
+      <button onclick="changeBackgroundColor('white');">Reset color</button>
+    </div>
```
Jym77: What would happen if only one of the buttons were not accessible (or not visible, or empty name)? Would that fail the rule (I think it should)? It's probably worth updating the examples so that different parts of the UI are "out" in each one. Expectation can also be made more explicit about that ("**each** UI component controlling this must … blah blah").

Same goes for "device motion has UI control" rule.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,324 @@
...
+    - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any [HTML document][] that has [functionality][] that can be operated by [device motion][].
+
+## Expectation
+
+For the test target, [device motion][] [can be disabled][] and the control for disabling it must be a [user interface component][] and it should be [visible][] and [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77: Also, @WilcoFiers is pointing out that "ui component" it ambiguous…
https://github.com/act-rules/act-rules.github.io/pull/421#discussion_r390893467

carlosapaduarte:
Defining "user motion" in an objective and unambiguous way seems out of reach at the moment.
We've worked a way to define "device motion" in terms of the events it triggers (making it objective and unambiguous).
Therefore, we're changing this PR from one composite rule (with 4 atomic rules) that attempted to fully cover SC 2.5.4, to 2 atomic rules that cover partly cover SC 2.5.4 (considering only the device motion part of the SC).
Currently, the PR includes the revised applicability and expectation for the 2 atomic rules.
I'll work next on updating the examples and refactoring the tests.

carlosapaduarte:
Currently using the definition of "Instrument to achieve an objective" proposed in #960

test-assets/7677a9/slider.js
```diff
@@ -0,0 +1,44 @@
...
+    if (event.gamma > 20) {
+        increaseSlider();
+    } else if (event.gamma < -20) {
+        decreaseSlider();
+    }
+}
+
+function handleOrientationCanBeDisabled(event) {
+    const disableMotion = document.getElementById('disableMotion');
+    const gamma = !disableMotion.checked ? event.gamma : 0;
```
Jym77, line 21: You can invert test outcomes to remove the negation
```suggestion
    const gamma = disableMotion.checked ? 0 : event.gamma;
```

test-assets/7677a9/slider.js
```diff
@@ -0,0 +1,44 @@
...
+
+function handleMotion(event) {
+    if (event.rotationRate.gamma > 5) {
+        increaseSlider();
+    } else if (event.rotationRate.gamma < -5) {
+        decreaseSlider();
+    }
+}
+
+function handleMotionCanBeDisabled(event) {
```
Jym77, line 38: Test on `disableMotion.checked` is missing.

test-assets/7677a9/slider.js
```diff
@@ -0,0 +1,44 @@
+function increaseSlider() {
+    document.getElementById('motionSlider').value++;
+    document.getElementById('output').innerHTML = document.getElementById('motionSlider').value;
+}
+
+function decreaseSlider() {
+    document.getElementById('motionSlider').value--;
+    document.getElementById('output').innerHTML = document.getElementById('motionSlider').value;
+}
```
Jym77, line 9: I'd factor that in a `changeSlider(value)` function. `increaseSlider()` and `decreaseSlider()` can still be here but would just call `changeSlider` with ±1.
Not really necessarily. I tend to factor in and abstract a lot 😁

test-assets/7677a9/slider.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+function handleOrientationCanBeDisabled(event) {
+    const disableMotion = document.getElementById('disableMotion');
+    const gamma = !disableMotion.checked ? event.gamma : 0;
+
+    if (event.gamma > 20) {
+        increaseSlider();
+    } else if (event.gamma < -20) {
```
Jym77, line 25: Tests should be done on `gamma` otherwise it can't be disabled.
```suggestion
    if (gamma > 20) {
        increaseSlider();
    } else if (gamma < -20) {
```

test-assets/7677a9/slider.js
```diff
@@ -0,0 +1,44 @@
...
+    document.getElementById('output').innerHTML = document.getElementById('motionSlider').value;
+}
+
+function handleOrientation(event) {
+    if (event.gamma > 20) {
+        increaseSlider();
+    } else if (event.gamma < -20) {
+        decreaseSlider();
+    }
+}
```
Jym77, line 17: I'd also factor in something that take `gamma` (rather than `event` as parameter) and have the other function only set gamma to a correct value (either extracting it from `even` or looking at `disableMotion.checked` on top of it).
Again, not really necessarily.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,304 @@
...
+    - Carlos Duarte
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has one or more registered [device orientation events][device orientation] or [device motion events][device motion].
```
Jym77, line 29: I like listing the events.

Is "registered" clear enough, or does that need further definition? I am not sure…
Maybe something like "has one or more EventTarget whose event listener list contains …"
https://dom.spec.whatwg.org/#eventtarget
Using DOM specs would also make the applicability easier to automate, which is always helpful for testers…

...hmmm wondering if such a formulation could be used to target the EventTarget (or even the event listener) rather than the full page (and thus give more precise result as we could pinpoint which element is not correct).

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available to disable the event.
```
Jym77, line 33: Add a note that the same instrument can be used for multiple events.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Expectation 2
+
+The [instrument][] is [visible][].
+
+## Expectation 3
+
+The [instrument][] is [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77, line 41: These two assumptions can certainly be merged together, or event with the first one.
I have no strong feeling between keeping them separate or not. Our current practice tend to be to have as few expectations as needed, but that is not necessarily the best way to do…

_rules/device-motion-disabled-c249d5.md
Jym77, line 29: (I don't know much about motion events)
~Why restricting the event listener on the Window object and not on the elements in the DOM?~
(these events can only be fired on the window object, so this is OK)

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,304 @@
...
+		</div>
+		<div aria-hidden="true">
+			<input type="checkbox" id="disableMotion" />
+			<label for="disableMotion">Disable Motion Actuation</label>
+		</div>
+	</body>
+</html>
+```
+
+#### Failed Example 3
```
Jym77, line 211: I'd move the "not visible" example before the "not accessible" one in order to mimic the order of the conditions in the expectations.

_rules/device-motion-disabled-c249d5.md
Jym77, line 33: Is there a need for the instrument to **not** be activated through device motion…
Currently it seems that a slider on `devicemotion`, disabled on `deviceorientation`, would be good 👿 
But then the disabling on `deviceorientation` would itself need a motion disabling thingie…
Plus, instrument is currently defined as an HTML element, so an event firing on the window object probably doesn't qualify…

🤔 this should be good…

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,314 @@
...
+  - 'attr-lowercase'    
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has one or more registered [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available supporting the same outcome of the event.
```
Jym77, line 33:
```suggestion
For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available supporting the same outcomes of the event.
```
The event can have multiple outcomes (e.g., -1, 0, or +1 on the value of the slider).

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: Thanks for the suggestion. I also has some reserves about using registered. 
Given that we know the eventTarget to be the Window object, I'll focus instead on the event listener list.

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: wow, that was twisted 😁

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: In the previous version of the rule, it was just one expectation. But I always feel that it is more readable like this. If someone really wants this to be just one expectation, I can change it back.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -173,12 +175,12 @@ This [HTML document][] that can be operated through the device's orientation to
...
 #### Failed Example 2
 
-This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has a control to disable the functionality but it is not [included in the accessibility tree][].
+This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has a control to disable the functionality but it is not [visible][].
 
 ```html
 <html>
 	<head>
-		<title>Failed Example 2</title>
+		<title>Failed Example 3</title>
```
Jym77, line 183:
```suggestion
		<title>Failed Example 2</title>
```
Titles need to be updated to their new place.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -210,12 +212,12 @@ This [HTML document][] that can be operated through the device's orientation to
...
 #### Failed Example 3
 
-This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has a control to disable the functionality but it is not [visible][].
+This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has a control to disable the functionality but it is not [included in the accessibility tree][].
 
 ```html
 <html>
 	<head>
-		<title>Failed Example 3</title>
+		<title>Failed Example 2</title>
```
Jym77, line 220:
```suggestion
		<title>Failed Example 3</title>
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -177,12 +179,12 @@ This [HTML document][] that can be operated through the device's orientation to
...
 #### Failed Example 2
 
-This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has the same functionality available through `button` elements but these are not [included in the accessibility tree][].
+This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has the same functionality available through `button` elements but these are not [visible][].
 
 ```html
 <html>
 	<head>
-		<title>Failed Example 2</title>
+		<title>Failed Example 3</title>
```
Jym77, line 187:
```suggestion
		<title>Failed Example 2</title>
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -216,12 +220,12 @@ This [HTML document][] that can be operated through the device's orientation to
...
 #### Failed Example 3
 
-This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has the same functionality available through `button` elements but these are not [visible][].
+This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has the same functionality available through `button` elements but these are not [included in the accessibility tree][].
 
 ```html
 <html>
 	<head>
-		<title>Failed Example 3</title>
+		<title>Failed Example 2</title>
```
Jym77, line 228:
```suggestion
		<title>Failed Example 3</title>
```

_rules/device-motion-disabled-c249d5.md
Jym77: Thinking about it, I do like having separate expectations as it's easier to read and allow to pinpoint the problem more precisely (better for reporting error and indicating fixes).

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,318 @@
+---
+id: 7677a9
+name: Device motion can be operated by user interface
```
WilcoFiers, line 3: I don't think this title works... You mean "device motion functionality can also be controlled from the user interface"? The current phrasing sounds like you want the UI to be able to move the device.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,308 @@
+---
+id: c249d5
+name: Device motion can be disabled
```
WilcoFiers, line 3:
```suggestion
name: Device motion functionality can be disabled
```

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,308 @@
...
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available to disable the event.
+
+**Note:** The same [instrument][] can be used to disable more than one event.
+
+## Expectation 2
```
WilcoFiers, line 37: Why does it need to be visible or in the accessibility tree? Why does it need to be on the same page even? It's kind of a similar problem as I'm pointing out here:
https://github.com/act-rules/act-rules.github.io/pull/960/files#r412915505

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,308 @@
...
+
+The [instrument][] is [visible][].
+
+## Expectation 3
+
+The [instrument][] is [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
+
+## Assumptions
+
+- The motion to operate the device is not used through an [accessibility supported][] interface.
```
WilcoFiers, line 47: I don't really understand what this means. Can you make this clearer?

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,308 @@
...
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available to disable the event.
```
WilcoFiers, line 33: What does it mean to "disable the event'? That seems ambiguous to me. Is it to have the event cancelled, remove the event handler, prevent some effect from occurring in reaction to the event?

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,308 @@
...
+    - Carlos Duarte
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
```
WilcoFiers, line 29: You're making a rather substantial assumption here that just because there's an event handler, something is actually going to happen in response to the event.

carlosapaduarte: I think that is a fair assumption to make (otherwise, why is the event handler there?)
But I agree that we need to add an assumption that what will happen in response to the event is a functionality of the in the web page and not other unrelated result.

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: OK. There is nothing in this SC that explicitly requires the instrument to be visible or in the accessibility tree. It just needs to exist.
Regarding the problem of (not) being on the same page, I'll address it in the same way I did in #960

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: It's one of the exceptions in the success criterion. I'll make that clear.

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: Rephrased the expectation (without using the term disable and focusing on the outcome of the event) and added a note.

carlosapaduarte:
Moving one of the atomic rules to its own PR #1276

_rules/device-motion-disabled-c249d5.md
```diff
@@ -28,24 +28,19 @@ htmlHintIgnore:
...
 **Note:** The same [instrument][] can be used to disable more than one event.
 
-## Expectation 2
-
-The [instrument][] is [visible][].
-
-## Expectation 3
-
-The [instrument][] is [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
+**Note:** Preventing the outcome of the event can be done in multiple ways (e.g. removing the event listener; handling the event in a different manner) but the way in which it is done is not relevant for this rule.
```
Jym77, line 37:
```suggestion
**Note:** Preventing the outcome of the event can be done in multiple ways (e.g. removing the event listener; handling the event in a different manner; …) but the way in which it is done is not relevant for this rule.
```
Make it explicit that this is an open list.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -28,24 +28,19 @@ htmlHintIgnore:
...
 
 ## Assumptions
 
-- The motion to operate the device is not used through an [accessibility supported][] interface.
-- The motion is not [essential][] for the functionality it triggers.
+- The motion to operate the device is not used through an [accessibility supported][] interface, which is listed as a valid exception to SC 2.5.4.
+- The motion is not [essential][] for the functionality it triggers, which is listed as a valid exception to SC 2.5.4.
+- The event listeners listening to device motion events trigger a funcionality in the web page. If they do not trigger any such functionality failing this rule might not be a failure of the success criterion.
```
Jym77, line 43:
```suggestion
- The event listeners listening to device motion events trigger a functionality in the web page. If they do not trigger any such functionality failing this rule might not be a failure of the success criterion.
```

_rules/device-motion-disabled-c249d5.md
```diff
@@ -28,24 +28,19 @@ htmlHintIgnore:
 
 The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
 
-## Expectation 1
+## Expectation
 
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available to disable the event.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] to prevent the outcome of the event is available in the same [web page][], or in another [web page][] that is [linked][hyperlink] from the [web page][] of the test target.
```
Jym77, line 33: Actually, the SC only says that the functionality "can be disabled". It doesn't say anything about instrument (UI component). (while the "alternate operation" part does refer to a UI component).

I think we are making an assumption that "can be disabled" means "by a UI component".
A strict reading of the SC, providing a phone number to a support team that will server-side disable the motion detection passes the SC… (but not this rule).

And if the rule assumes that "can be disabled" means "by a UI component", then I think that expectations 2 and 3 should stay (as UI component "can be perceived as …" which is pretty difficult for something with `display: none`).

carlosapaduarte: Agree. Will add an assumption and reinstate expectations 2 and 3.

carlosapaduarte:
Applied similar changes to what was done in #960 and #1276

_rules/device-motion-disabled-c249d5.md
```diff
@@ -30,19 +30,20 @@ The rule applies to an [HTML document][] with an associated [Window object][] th
 
 ## Expectation 1
 
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] to prevent the outcome of the event is available in the same [web page][], or in another [web page][] that is [linked][hyperlink] from the [web page][] of the test target.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] to prevent the results of the event.
```
Jym77, line 33: I'm not completely sure about "prevent". See also https://github.com/act-rules/act-rules.github.io/pull/873#discussion_r423593131

Replacing the effect of device motion by an other effect would "prevent the results" of the event but still be pretty bad (as there is now a new device motion event). So for example, a slider where any motion increase it, and a button toggles to any motion decrease the slider, the button would "prevent the results" of the event and thus pass the rule.

I think what we went is that after activating the instrument, "nothing" happens when the event is fired. But that need a bit of definition of "nothing" (change in the DOM or visible content, playing media, redirecting, …) Note that the definition doesn't need to be too strict: if the rule pass when "disabling" still allow sound to be played, we're in one of the "further testing needed" case. Bad (especially if too broad), but not too bad.

carlosapaduarte: @Jym77 I don't think the problem here is prevent (definition of prevent is ""to keep from happening or existing"). The problem is what we are preventing. I agree with you when you say that if we replace the result of one event with another result, then we are preventing the original result from happening. I guess what we want in this instance is to prevent any results of the event from happening (i.e. disable the event). I'll rephrase it accordingly.

_rules/device-motion-disabled-c249d5.md
```diff
@@ -30,11 +30,11 @@ The rule applies to an [HTML document][] with an associated [Window object][] th
 
 ## Expectation 1
 
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] to prevent the results of the event.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] to prevent any results of the event.
```
Jym77, line 33: Coming from a strong computability background, I'm always worried when we go into the effects of programs as essentially everything is undecidable, which make testing complicated… I'll try to flesh out concerns in a more concrete way…

* What if the instrument result in adding a 2h wait between the event trigger and the effect? This should fail the rule, but this is hardly testable. Maybe we need an arbitrary timeout on the test here also (as you did in another rule).

* What if (after activating the instrument), the effect of the event is to increase an internal counter which is never shown in any way to anybody? That would fail the arguably rule (depending what we call "result of the event") but not the SC.

* What if that counter can be seen but somewhere else? For example, that counter is stored on a server-side database and admins can look at it (but users can't). This is a bit more clearly a "result of the event", but still I'm not sure it fails the SC 

* Even more obvious, the event could both change a slider and update a counter, which is shown on a totally different page (e.g., profile page says "you've moved your phone X times"). And the instrument removes the update to slider on same page, but not the counter update. This wouldn't "prevent any result" but I'm not sure it fails the SC.

* From the previous points, if device motion update a counter that is only shown on another page, does this fail the SC? It might be very hard to test if we consider this as a result of the event (because it can't be tested in isolation). Maybe we need to restrict the rule to events that have an "effect" (whatever we mean by this) on the current page?

* Ultimately, even just `if (eventHasBeenDisabled) { return }` is some sort of "result of the event". Arguably, from a pure computability point of view, something is happening and we're consuming CPU time. Can be pushed to the absurd with `if (eventHasBeenDisabled) { doHugeComputation();  return }` which could grind the browser to almost a halt for 1 minute but ultimately have no other effect. Is stopping the browser a "result of the event"? This shows the ambiguity of these terms if the first case has no result but the second has (where is the limit?)


---
I think we need:
* a timeout on "nothing happens", otherwise this is undecidable.
* a more precise definition of the kind of things that can happen. This can totally start small as being too restrictive will mean that we sometimes pass the rule when the SC fails, which is OK.
I say something like "change in the DOM", or "change in the rendered pixels" is a good start for effect we want to monitor. It would mean we'll pass (or be inapplicable), say, a page where motion plays a sound, but that doesn't break the fail-on-fail relationship, so it only mean that the rule can be improved…

carlosapaduarte: @Jym77 I think the timeout is reasonable and I'll add it.

However, for the second issue I think you are missing one of the rule's assumptions that states: 
> The event listeners listening to device motion events trigger a functionality in the web page. If they do not trigger any such functionality failing this rule might not be a failure of the success criterion.

Do you agree that this addresses your concerns?

_rules/device-motion-disabled-c249d5.md
```diff
@@ -30,12 +30,14 @@ The rule applies to an [HTML document][] with an associated [Window object][] th
 
 ## Expectation 1
 
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] to prevent any results of the event.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] to prevent any results of the event within a 10 minute time span.
```
Jym77, line 33: There might need to precise to what "within 10 minutes" applies.
* Activating the instrument "disable" the event forever, where "disable" means "no effect during 10 minutes". (real case)
* Activating the instrument "disable" the event for the next 10 minutes. (this interpretation is plausible with the text)
* Activating the instrument can take up to 10 minutes to react, after that the event is "disabled" forever. (this interpretation is a bit far fetched)

_rules/device-motion-disabled-c249d5.md
Jym77: That should be good, indeed.

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: @Jym77 I've updated the expectation. Can you please review again?

_rules/device-motion-disabled-c249d5.md
```diff
@@ -30,19 +30,22 @@ The rule applies to an [HTML document][] with an associated [Window object][] th
...
+**Note:** The 10 minute time span is an arbitrary limit which is not included in WCAG. Results that happen after this period will not fail this rule but may nonetheless fail [Success Criterion 2.5.4: Motion Actuation][sc 2.5.4]. The accessibility problem tends to be less severe for longer time periods, and without a time limit, testing this rule consistently would be impractical.
 
 ## Expectation 2
 
-The [instrument][] is [visible][].
+For each [instrument][] in the set of [instruments][instrument] from Expectation 1, one of the following is true:
 
-## Expectation 3
+- the [instrument][] is in the same [web page][] of the test target; or
+- instructions for the user to locate the [instrument][] are in the same [web page][] of the test target.
```
WilcoFiers, line 46: I don't think WCAG requires that there are instructions on every page.

carlosapaduarte: @WilcoFiers Indeed. I was just waiting for the definition of 'clearly labeled location' from #960 
Updating this expectation to use that definition.

_rules/device-motion-disabled-c249d5.md
WilcoFiers: I disagree, see my comment here: https://github.com/act-rules/act-rules.github.io/pull/1276/files#r434474714

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,205 @@
+---
+id: c249d5
+name: Device motion based functionality can be disabled
```
WilcoFiers, line 3: This isn't testing that functionality can be disabled; you're not checking that the event handler is used to operate "functionalty".

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,205 @@
...
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] to prevent any results of the event within a 10 minute time span of the [event firing][].
```
WilcoFiers, line 33: I think "any result of the event" is problematic to test. You can maybe say that it is presented to the user as the result of the event?

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: Reworked the expectation to take into account the possibility of the event not producing any results.

_rules/device-motion-disabled-c249d5.md
carlosapaduarte: Updated name and description

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target one of the following is true:
+
+- **no result:** The registered event does not present to the user results within a 10 minute time span of the [event firing][event firing]; or
```
WilcoFiers, line 35: See my comment here: https://github.com/act-rules/act-rules.github.io/pull/1276#discussion_r436615334

carlosapaduarte:
@WilcoFiers @Jym77 Updated the rule using the definition of "changes in content" proposed in https://github.com/act-rules/act-rules.github.io/pull/1276/commits/ddf45e171f692f13c2293c73868f886ad6402e13
Also updated title and description

carlosapaduarte:
> Looks good. Probably need to ripple change of vocabulary to the examples descriptions.

@Jym77 Updated examples descriptions and added another passed example for the no changes condition

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,226 @@
...
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target one of the following is true:
+
+- **no changes:** The registered event does not cause [changes to the content][changes in content] of the [web page][] within a 10 minute time span of the [event firing][event firing]; or
+- **disabled:** There is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] of the registered event or can be found in a [clearly labeled location][] from that [web page][], to prevent [changes to the content][changes in content] of the [web page][] resulting from the event within a 10 minute time span of the [event firing][].
```
WilcoFiers, line 36: Minor editorial suggestion:

```suggestion
- **disabled:** There is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] of the registered event or can be found in a [clearly labeled location][] from that [web page][]. The set of [instruments][] can be used to prevent [changes to the content][changes in content] of the [web page][] resulting from the event within a 10 minute time span of the [event firing][].
```

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,226 @@
...
+## Expectation
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target one of the following is true:
+
+- **no changes:** The registered event does not cause [changes to the content][changes in content] of the [web page][] within a 10 minute time span of the [event firing][event firing]; or
+- **disabled:** There is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] of the registered event or can be found in a [clearly labeled location][] from that [web page][]. The set of [instruments][] can be used to prevent [changes to the content][changes in content] of the [web page][] resulting from the event within a 10 minute time span of the [event firing][].
+
+**Note:** The 10 minute time span is an arbitrary limit which is not included in WCAG. Results that happen after this period will not fail this rule but may nonetheless fail [Success Criterion 2.5.4: Motion Actuation][sc 2.5.4]. The accessibility problem tends to be less severe for longer time periods, and without a time limit, testing this rule consistently would be impractical.
+
+## Assumptions
```
WilcoFiers, line 40: Link to the SC, and don't abbreviate "SC", use "success criterion"

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target one of the following is true:
+
+- **no changes:** The registered event does not cause [changes to the content][changes in content] of the [web page][] within a 10 minute time span of the [event firing][event firing]; or
```
adilsofficial, line 35:
```suggestion
- **no changes:** The registered event does not cause [changes to the content][changes in content] of the [web page][] within a 10 minutes time span of the [event firing][event firing]; or
```

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,226 @@
...
+- **disabled:** There is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] of the registered event or can be found in a [clearly labeled location][] from that [web page][]. The set of [instruments][instrument] can be used to prevent [changes to the content][changes in content] of the [web page][] resulting from the event within a 10 minute time span of the [event firing][].
```
adilsofficial, line 36:
```suggestion
- **disabled:** There is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] of the registered event or can be found in a [clearly labeled location][] from that [web page][]. The set of [instruments][instrument] can be used to prevent [changes to the content][changes in content] of the [web page][] resulting from the event within a 10 minutes time span of the [event firing][].
```

_rules/device-motion-disabled-c249d5.md
```diff
@@ -0,0 +1,226 @@
...
+
+**Note:** The 10 minute time span is an arbitrary limit which is not included in WCAG. Results that happen after this period will not fail this rule but may nonetheless fail [Success Criterion 2.5.4: Motion Actuation][sc 2.5.4]. The accessibility problem tends to be less severe for longer time periods, and without a time limit, testing this rule consistently would be impractical.
```
adilsofficial, line 38:
```suggestion
**Note:** The 10 minutes time span is an arbitrary limit which is not included in WCAG. Results that happen after this period will not fail this rule but may nonetheless fail [Success Criterion 2.5.4: Motion Actuation][sc 2.5.4]. The accessibility problem tends to be less severe for longer time periods, and without a time limit, testing this rule consistently would be impractical.
```

carlosapaduarte:
Final call ends July 14

carlosapaduarte:
Final call ended. Merging.

PR #904, opened 2019-09-17
Rename "Form field label is descriptive" (cc0f0a / e0ca31)

Jym77:
Renaming a rule file to prepare for bigger changes without loosing history.
This is part of solving the history problem of #447, as discussed with @WilcoFiers on Monday 16th of September.

New name is suboptimal for current rule, but this needs to be merged before substantial changes are made to rule in order to avoid loosing history.

Will not require a Final Call.

Closes issue(s):

- Partially replace #447
- Blocks #905 

----

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
@WilcoFiers, If it doesn't change the rule id, then [the test break](https://circleci.com/gh/act-rules/act-rules.github.io/4227?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) because the rule id does not match the filename (and merge is blocked, and if we force-merge, deployment is blocked).

If the id in the filename is not changed as part of this PR, then the next PR will have to do it, renaming the file and raising again the same issue we're trying to avoid with this PR.

PR #934, opened 2019-09-24
Remove `xml:lang` from language rules (de46e4, b5c3f8, bf051a, 5b7ae0)

### LINKED ISSUE #668: Validity of HTML Lang attribute (bf051a) - incorrect info about xml:lang
https://act-rules.github.io/rules/bf051a#accessibility-support

The Accessibility Support section says:

> While HTML5 specification indicates that xml:lang attribute takes priority over lang attribute, certain assistive technologies prioritise lang over xml:lang instead.

The HTML 5.2 and the HTML living standards actually say:

>Authors must not use the lang attribute in the XML namespace on HTML elements in HTML documents. 
>The attribute in no namespace with no prefix and with the literal localname "xml:lang" has no effect on language processing.

https://www.w3.org/TR/html52/dom.html#the-lang-and-xmllang-attributes
https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes

kasperisager:
This pull request removes any mention of the `xml:lang` attribute from the following rules, per my suggestion in https://github.com/act-rules/act-rules.github.io/issues/668#issuecomment-527434344:

- https://act-rules.github.io/rules/b5c3f8
- https://act-rules.github.io/rules/bf051a
- https://act-rules.github.io/rules/de46e4

The rule https://act-rules.github.io/rules/de46e4 has also been restricted to HTML elements as the behaviour of `lang` and `xml:lang` attributes is different for SVG elements and really warrant a separate rule.

The following assumption has also been added to the rules listed:

> This rule assumes that [documents](https://dom.spec.whatwg.org/#concept-document) are served with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`. If a document is served using a different content type, this can affect the interpretion of the `lang` attribute.

The reason for this is that for documents served with a content type of e.g. `application/xhtml+xml`, the `lang` attribute will be ignored entirely. As discussed with @dd8, our gut feeling was that the far majority of documents are served with a content type of `text/html`, so we agreed to make this an assumption rather than part of the applicability. Some data to back this up would be nice though.

~~This pull request also updates the applicabilities of the language rules so that they apply to all `<html>` document elements, ensuring that nested browsing contexts are also required to specify their language as this is not inherited from their parent browsing context.~~


_rules/html-has-lang-b5c3f8.md
```diff
@@ -25,24 +25,21 @@ The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s)
...
 
-Each test target has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
-
-**Note**: HTML5 recommends using `lang` instead of `xml:lang`. This is not known to impact accessibility, which is why use of both is permitted by this rule.
+Each test target has a `lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
 
 ## Assumptions
 
-_There are currently no assumptions_
+This rule assumes that [documents](https://dom.spec.whatwg.org/#concept-document) are served with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`. If a document is served using a different content type, this can affect the interpretion of the `lang` attribute.
```
dd8: This assumption good - Common Crawl data shows that Content-Types that cause pages to be interpreted as XHTML are very rare (approx 0.06% of pages)
https://github.com/act-rules/act-rules.github.io/issues/668#issuecomment-534480359

_rules/html-has-lang-b5c3f8.md
```diff
@@ -25,24 +25,21 @@ The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s)
 
 ## Expectation
 
-Each test target has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
-
-**Note**: HTML5 recommends using `lang` instead of `xml:lang`. This is not known to impact accessibility, which is why use of both is permitted by this rule.
+Each test target has a `lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
```
dd8: Shouldn't this be ASCII whitespace (https://infra.spec.whatwg.org/#ascii-whitespace) or is that a different pull request?

_rules/html-xml-lang-match-5b7ae0.md
```diff
@@ -27,11 +27,9 @@ The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s)
 
 For each test target, the value of the primary language subtag (characters before the first dash) for the `lang` and `xml:lang` attributes are the same.
 
-**Note**: HTML 5 specification requires the `lang` and `xml:lang` attributes to match exactly. This is not known to impact accessibility, which is why it is permitted in this rule.
-
 ## Assumptions
 
-- Although there is no known way that an inappropriate secondary subtag in the `lang` or `xml:lang` attribute (such as "en-XYZ") can lead to accessibility issues, it is conceivable that with specific assistive technologies on some languages, there can be exceptions. In that case `lang` and `xml:lang` should have matching secondary subtags.
+- Although there is no known way that an inappropriate secondary subtag in the `lang` or `xml:lang` attribute (such as `en-XYZ`) can lead to accessibility issues, it is conceivable that with specific assistive technologies on some languages, there can be exceptions. In that case `lang` and `xml:lang` should have matching secondary subtags.
```
dd8: Secondary subtags are important for some languages - and the wrong subtag results in voicing that is hard (or impossible) to understand:
https://github.com/act-rules/act-rules.github.io/issues/668#issuecomment-534465673

_rules/html-has-lang-b5c3f8.md
kasperisager: Gah, I forgot to change it 🤦‍♂

_rules/html-has-lang-b5c3f8.md
```diff
@@ -56,88 +53,32 @@ The `lang` attribute specified has a non-empty value.
...
-
-The `xml:lang` attribute consists of only [whitespace](#whitespace).
-
-```html
-<html xml:lang=" "></html>
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
```
dd8: Might be worth adding an Inapplicable Example for MathML as well - there's no `lang` attribute in MathML but `xml:lang` is allowed
https://www.w3.org/TR/MathML3/mathml.html#appendixi_index.att

_rules/html-lang-valid-bf051a.md
```diff
@@ -56,46 +53,6 @@ The `lang` attribute specified is neither empty ("") nor only [whitespace](#whit
 <html lang="fr"></html>
 ```
 
-#### Passed Example 2
```
dd8: There's only one passed example - I think it's worth adding some more passed examples from BCP-47:

de-AT
sr-Latn-RS
es-419
en-a-aaa-b-ccc-bbb-x-xyz
i-klingon

Without these it would be easy to write code that passes the ACT-R unit tests but flags valid BCP-47 language codes as invalid.

_rules/html-has-lang-b5c3f8.md
```diff
@@ -25,24 +25,21 @@ The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s)
...
 
-Each test target has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
-
-**Note**: HTML5 recommends using `lang` instead of `xml:lang`. This is not known to impact accessibility, which is why use of both is permitted by this rule.
+Each test target has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 
 ## Assumptions
 
-_There are currently no assumptions_
+This rule assumes that [documents](https://dom.spec.whatwg.org/#concept-document) are served with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`. If a document is served using a different content type, this can affect the interpretion of the `lang` attribute.
```
daniel-montalvo:
```suggestion
[Documents](https://dom.spec.whatwg.org/#concept-document) are served with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`. If a document is served using a different content type, this can affect the interpretation of the `lang` attribute.
```

_rules/lang-valid-de46e4.md
```diff
@@ -19,15 +19,17 @@ authors:
...
+For each test target, the `lang` attribute has a [valid language subtag](#valid-language-subtag).
 
 ## Assumptions
 
-This rule assumes that the presence of a lang or xml:lang attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+This rule assumes that [documents](https://dom.spec.whatwg.org/#concept-document) are served with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`. If a document is served using a different content type, this can affect the interpretion of the `lang` attribute.
+
+This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
 
 ## Accessibility Support
```
dd8: The accessibility support should be same 'language of page'

_rules/html-lang-valid-bf051a.md
kasperisager: We're not testing BCP 47 with these test cases though and I don't think we should.

_rules/html-has-lang-b5c3f8.md
```diff
@@ -56,88 +53,32 @@ The `lang` attribute specified has a non-empty value.
...
-```html
-<html xml:lang="en" lang=""></html>
-```
-
 ### Failed
 
 #### Failed Example 1
 
-There were no `lang` or `xml:lang` attribute specified.
+There were no `lang` attribute specified.
```
dd8: Typo - should be: "There was no `lang` attribute specified."

_rules/html-has-lang-b5c3f8.md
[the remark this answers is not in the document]
kasperisager: Only failed test cases are allowed to actually fail the associated accessibility requirements, for the purpose of the implementation mapping: https://act-rules.github.io/pages/implementations/mapping/. This doesn't seem to be super explicit at the moment though; did we document this somewhere, @WilcoFiers?

_rules/html-xml-lang-match-5b7ae0.md
kasperisager: Any suggestions on how to phrase the assumption to account for this?

_rules/html-has-lang-b5c3f8.md
```diff
@@ -25,24 +25,21 @@ The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s)
```
dd8: > Note: Documents embedded into other documents, such as through iframe or object elements are not applicable because they are not web pages according to the definition in WCAG.

I think this leaves a gap in the rules because iframes are not checked by https://act-rules.github.io/rules/de46e4 (language of parts) or https://act-rules.github.io/rules/b5c3f8 (language of page). `lang` is not inherited from the parent document according to this comment https://stackoverflow.com/a/53759156

The seems right because iframe security restrictions (e.g. sandbox) are supposed to prevent the iframe documents from accessing parent document state (including lang)

_rules/html-has-lang-b5c3f8.md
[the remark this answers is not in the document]
dd8: Yep, what I was meaning is that rule code is usually written to detect failures e.g. a missing lang attribute. 

That means an inapplicable SVG example with no 'lang' attribute will find implementations that wrongly detect missing 'lang' on SVG. 

But an inapplicable SVG example with a 'lang' attribute cannot find implementations that wrongly detect missing 'lang' on SVG (because the incorrect implementation never triggers when `lang` is present).

_rules/html-has-lang-b5c3f8.md
kasperisager: That's a good point. Updating.

_rules/html-has-lang-b5c3f8.md
```diff
@@ -19,30 +19,26 @@ authors:
...
 ## Accessibility Support
 
-There are known combinations of a popular operating system with browsers and assistive technologies that do not support the `lang` and `xml:lang` attributes.
+There are known combinations of a popular operating system with browsers and assistive technologies that do not support the `lang` attribute.
 
 ## Background
 
 - [H57: Using language attributes on the html element](https://www.w3.org/WAI/WCAG21/Techniques/html/H57)
 - [BCP 47: Tags for Identifying Languages](https://www.ietf.org/rfc/bcp/bcp47.txt)
 - [MDN: `lang` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)
```
Jym77: Don't we usually avoid citing MDN?
Probably for another PR, however...

_rules/html-xml-lang-match-5b7ae0.md
dd8: Chatted to @nitedog  and @WilcoFiers about this - there's a W3 Pronunciation Task force that might have some useful input on this https://www.w3.org/WAI/APA/task-forces/pronunciation/

_rules/html-has-lang-b5c3f8.md
```diff
@@ -19,30 +19,26 @@ authors:
...
 
-Each test target has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
-
-**Note**: HTML5 recommends using `lang` instead of `xml:lang`. This is not known to impact accessibility, which is why use of both is permitted by this rule.
+Each test target has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 
 ## Assumptions
 
-_There are currently no assumptions_
+- [Documents](https://dom.spec.whatwg.org/#concept-document) are served with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`. If a document is served using a different content type, this can affect the interpretation of the `lang` attribute.
```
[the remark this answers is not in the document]
daniel-montalvo: This is an editorial change that we are introducing in #935 . This is why I am suggesting it for this rule.

_rules/html-xml-lang-match-5b7ae0.md
dd8: Here's an example that shows the problem with ignoring subtags - these two code fragments specify exactly the same languages  (because lang=zh-yue and lang=yue are aliases that identify the same language) but only one of the examples fails the rule. 

See https://tools.ietf.org/html/bcp47#section-2.2.2

```
<!-- this fails the rule because primary tags don't match -->
<html lang='yue' xml:lang='cmn' >
```

```
<!-- this passes the rule because primary tags match -->
<html lang='zh-yue' xml:lang='zh-cmn' >
```

kasperisager: That's exactly the kind of argument I needed 💯 Cheers!

kasperisager: The rule now requires both the primary and the extended language subtags to match. Is there a similar argument for requiring script and/or other subtags to match as well? Are these even used by UAs and ATs?

_rules/html-has-lang-b5c3f8.md
```diff
@@ -56,94 +52,46 @@ The `lang` attribute specified has a non-empty value.
...
-<svg xmlns="http://www.w3.org/2000/svg" lang="en"></svg>
+<svg xmlns="http://www.w3.org/2000/svg"></svg>
+```
+
+#### Inapplicable Example 2
+
+The rule does not apply to `math` element.
+
+```svg
+<math></math>
```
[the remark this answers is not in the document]
kasperisager: There's no content though and so no language is needed.

_rules/html-lang-valid-bf051a.md
```diff
@@ -3,7 +3,7 @@ id: bf051a
 name: HTML page language is valid
 rule_type: atomic
 description: |
-  This rule checks that the `lang` and `xml:lang` attributes of the root element of an HTML page have a valid primary language subtag.
+  This rule checks that the `lang` attribute of the root element of an HTML page have a valid primary language subtag.
```
carlosapaduarte: have > has

_rules/lang-valid-de46e4.md
```diff
@@ -3,7 +3,7 @@ id: de46e4
 name: lang attribute in body is valid
 rule_type: atomic
 description: |
-  This rule checks that the `lang` and `xml:lang` attributes of an element in the page body have a valid primary language subtag.
+  This rule checks that the `lang` attribute of an element in the page body have a valid primary language subtag.
```
carlosapaduarte: have > has

_rules/html-has-lang-b5c3f8.md
kasperisager: I think that listing MDN as background material is fine; it's not a normative reference.

_rules/html-lang-valid-bf051a.md
```diff
@@ -19,83 +19,39 @@ authors:
...
 
-The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element with a `lang` and/or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
-
-**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element with a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 
 ## Expectation
 
-For each test target, the `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is neither empty ("") nor only [whitespace](#whitespace).
+For each test target, the `lang` attribute has a [valid language subtag](#valid-language-subtag).
```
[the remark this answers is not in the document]
kasperisager: Let's continue the discussion here: https://github.com/act-rules/act-rules.github.io/pull/934#pullrequestreview-292304082

_rules/html-xml-lang-match-5b7ae0.md
```diff
@@ -23,11 +23,10 @@ This rule applies to any [document element](https://dom.spec.whatwg.org/#documen
 
 ## Expectation
 
-For each test target, the value of the primary language subtag (characters before the first dash) for the `lang` and `xml:lang` attributes are the same.
+For each test target, the values of the [primary language subtags](https://tools.ietf.org/html/bcp47#section-2.2.1) and the [extended language subtags](https://tools.ietf.org/html/bcp47#section-2.2.2), if any exist, for the `lang` and `xml:lang` attributes are the same.
 
 ## Assumptions
 
-- Although there is no known way that an inappropriate secondary subtag in the `lang` or `xml:lang` attribute (such as `en-XYZ`) can lead to accessibility issues, it is conceivable that with specific assistive technologies on some languages, there can be exceptions. In that case `lang` and `xml:lang` should have matching secondary subtags.
```
Jym77: Given how long it took us to figure out that extended subtags are important, maybe we should replace that assumption with a note on the expectation?

dd8: Based on code inspection and browser testing `xml:lang` is always ignored in Safari when calculating the language exposed through the Accessibility Tree (although `xml:lang` is used when calculating the CSS `:lang()` selector).

That means `lang` and `xml:lang` mis-match has no effect on accessibility in Safari.

The same looks to be true for Firefox and Chromium based on reading their code, but I still need to confirm in a browser (and also test IE11).

If this is confirmed then `lang` and `xml:lang` mis-match has no effect on accessibility in current user agents.

kasperisager: Do we have any insights into AT behaviour as well?

dd8: Here are the current results - results guessed from code still have '?' after them. Once confirmed in AT/Accessibility Inspector I've removed the  '?' 

https://github.com/act-rules/act-rules.github.io/issues/668#issuecomment-534606384

kasperisager: I'll limit this pull request to only requiring that extended language subtags also match and then we can take a look at either extending this requirement to other subtags in the future.

_rules/html-xml-lang-match-5b7ae0.md
```diff
@@ -72,15 +72,7 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
...
-`html` element with non-matching primary language subtags for `lang` and `xml:lang`.
-
-```html
-<html lang="yue" xml:lang="cmn"></html>
-```
-
-#### Failed Example 3
-
-`html` element with non-matching primary and extended language subtags for `lang` and `xml:lang`.
+`html` element with non-matching extended language subtags for `lang` and `xml:lang`. While the primary language subtag is Chinese for both the `lang` and `xml:lang` attribute, the extended language subtag is Cantonese for the `lang` attribute and Mandarin for the `xml:lang` attribute.
```
kasperisager: @Jym77 As I don't know anything at all about the relationship between the different dialects of Chinese, this is about as good an explanation I could come up with for why the extended language subtags matter. Do you have any suggestions for improvements?

_rules/html-xml-lang-match-5b7ae0.md
kasperisager: Let's continue the discussion here: https://github.com/act-rules/act-rules.github.io/pull/934/commits/dd90193a6b00afa9feec21351b7add501059ff4d#r328112033

_rules/html-xml-lang-match-5b7ae0.md
kasperisager: See: https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r328105044

_rules/html-xml-lang-match-5b7ae0.md
dd8: It's not just extended language subtags - the script also matters - you can write Chinese Mandarin in Han Traditional, Han Simplified script (which uses simpler character shapes than traditional script) or Pinyin script (uses Latin characters to phonetically spell words - very similar to Rōmaji in Japanese)

The word 'dragon' has 3 different forms in different scripts:

- Dragon in Han Simplified is 龙 (zh-cmn-Hans or cmn-Hans)
- Dragon in Han Traditional is 龍 (zh-cmn-Hant or cmn-Hant)
- Dragon in Pinyin is lóng (zh-Latn-pinyin)

See https://sites.psu.edu/symbolcodes/languages/asia/chinese/
and https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

kasperisager: Is there any effect on how AT announces these though? For example, would something like the following be announced correctly?

```html
<p lang="zh-cmn-Hant">龙</p>
```

dd8: Han Simplified and Han Traditional characters have different shapes, which means they're represented by different Unicode code points - that might be enough to pronounce them correctly. It might be more complicated than this because some Han characters are also used in Japanese (where they're called Kanji and pronounced differently).

For Pinyin/Rōmaji then script might matter to get the latin letters pronounced correctly. In Japanese Rōmaji latin letters shouldn't be pronounced exactly like English or any other European language because Japanese has a more limited set of sounds (phonemes) compared to European languages.

More investigation needed...

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -108,66 +66,34 @@ The `lang` attribute value is not a valid primary language subtag.
...
-```html
-<html lang=" "></html>
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-The rule applies to `html` element and hence usage of `lang` attribute in `svg` element is not applicable.
+The rule does not apply to `svg` element.
```
WilcoFiers: "to `svg` element" -> "to `svg` elements"

_rules/html-page-lang-b5c3f8.md
```diff
@@ -19,30 +19,25 @@ authors:
 
 ## Applicability
 
-The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element.
-
-**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
```
WilcoFiers: This rule shouldn't be applicable to document elements that are embedded into frames / objects. That's the reason for the original wording. This problem exists with the other rules too.

_rules/element-lang-valid-de46e4.md
```diff
@@ -19,15 +19,18 @@ authors:
 
 ## Applicability
 
-This rules applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of a `body` element and has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
+This rules applies to any HTML element that:
+
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
```
dd8: I think it should use [type](https://dom.spec.whatwg.org/#concept-document-type) instead of [content-type](https://dom.spec.whatwg.org/#concept-document-content-type) since type is what affects the APIs. 

This covers documents nodes without a content-type created by JavaScript APIs like createDocumentFragment or createDocument

kasperisager: Documents will always have a content type, even when created via JavaScript. For [`createDocument()`](https://dom.spec.whatwg.org/#dom-domimplementation-createdocument), the content type will either be `application/xhtml+xml`, `image/svg+xml`, or `application/xml`, whereas for [`createHTMLDocument()`](https://dom.spec.whatwg.org/#dom-domimplementation-createhtmldocument) the content type will always be `text/html`. For [`createDocumentFragment()`], its node document will be the context object on which the API was invoked so the created `DocumentFragment` will effectively inherit the content type of its content object.

_rules/html-page-lang-b5c3f8.md
kasperisager: This change was entirely intentional, see https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r327576703.

kasperisager: To illustrate the issue, consider this example:

```html
<!doctype html>
<html lang=da>
  <p>Hej med dig</p>
  <iframe srcdoc="<!doctype html><p>Hej med dig</p>"></iframe>
</html>
```

VoiceOver correctly announces the outer "Hej med dig" in Danish, but the inner "Hej med dig" is announced in my default system locale, English. The following works as expected:

```html
<!doctype html>
<html lang=da>
  <p>Hej med dig</p>
  <iframe srcdoc="<!doctype html><html lang=da><p>Hej med dig</p>"></iframe>
</html>
```

The rule should therefore absolutely be applicable to nested browsing contexts as well.

_rules/html-xml-lang-match-5b7ae0.md
Jym77: There seem to exists a `zh-Latn-wadegiles` tag: 
https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry (and  search for "wadegiles" in the page). And I am sure that while this uses latin characters, this should be pronounced pretty differently than `zh-Latn-pinyin` (http://www.eastasianlib.org/ctp/RomTable/Chipinyintowade.pdf)

So script subtags (and their extension) might be important…

(it seems that `ja-Latn` only has a `hepburn` (and `heploc`) subtag, but nothing for Nihon-shiki or Kunrei romanisation (https://www.japanesewithanime.com/2017/12/romaji-systems-hepburn-nihon-kunrei.html))


The difference in pronunciation between Han characters in Chinese and kanjis in Japanese should be taken into account by the primary language subtag (_i.e._ `<p lang="zh">日</p>` and `<p lang="ja">日</p>` should be enough to have a difference in pronunciation).

_rules/html-page-lang-b5c3f8.md
WilcoFiers: I see your point. Even so, the success criterion this tests is whether or not the language of the page is set, and a web page is defined as a non-embedded resource. What you're pointing out could be considered a failure of 3.1.2, but I don't think that's a failure of 3.1.1.

dd8: I think there's a bug in WCAG here with language of page / language of parts - because it uses a different definition for 'page' than HTML. The HTML language computation does not use WCAG's definition of web page when computing language. 

There's similar problem with `longdesc` where accessing the URL referenced by longdesc in Firefox is treated as page navigation to a new page, but WCAG defines long descriptions as part of the full web page:
https://www.w3.org/TR/UNDERSTANDING-WCAG20/conformance.html#uc-conf-req2-head

To see this in practice:
- open https://www.davidmacd.com/test/longdesc.html in Firefox
- right click on the first image
 - choose the 'View Description' command  

Moving this in a different PR makes sense since this is going to be hard to resolve.

kasperisager: I don't think the definition of "web page" is the issue here; I think the definition is perfectly fine. Our individual readings of the WCAG differ though and that's really were I think the issue lies. Let me try to visualise it. First, here's hopefully something we can all agree on:

```
+---------------------+
|Web page             |
|                     |
|         +----------+|
|         |Embed     ||
|         |          ||
|         +----------+|
+---------------------+
```

The above shows a web page with an embedded resource. From the perspective of WCAG, the embedded resource forms a subset of the web page, but is not a web page in its own right; only the superset is. My hope is that you're nodding in agreement to this statement.

So, let's pretend for a second that the web page does not embed a resource and we've set its language to `da`:

```
+---------------------+
|Web page             |
|lang: da             |
|                     |
|                     |
|                     |
|                     |
+---------------------+
```

Assuming that the web page is indeed authored in Danish, this passes [success criterion 3.1.1](https://www.w3.org/TR/WCAG/#language-of-page) as the _default_ human language of the web page can be programmatically determined. 

Keeping the above in mind, consider this: What if, for whatever reason, the default human language only applied to _half_ the web page and we needed to do _something_ additional to also have it apply to the other half? I hope we can agree that without doing that additional _something_, the web page does not pass success criterion 3.1.1.

Now, let's take a long, hard look at this:

```
+---------------------+
|Web page             |
|lang: da             |
|         +----------+|
|         |Embed     ||
|         |lang: ?   ||
|         +----------+|
+---------------------+
```

This is really just a graphical depiction of the code I posted in https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r328210245. Here, the default human language can be determined for everything but the embedded resource. For the sake of argument, let's pretend that the embedded resource contains half the content of the entire web page. This is _precisely_ the situation I described previously; the default human language can only be determined for _half_ the web page. Hopefully, we've all already agreed that this does _not_ pass success criterion 3.1.1.

WilcoFiers: @dd8; I disagree this is a "bug". The choice for the definition of web page is very much intentional. We can discuss whether or not they should have chose a different term, such as "non-embedded web page", but that's beside the point that this SC was designed to only apply to the top-level document.

@kasperisager; I see where you're coming from, but I don't share your interpretation. The way I see it is this: If a Danish page has an English ad embedded in a div, that's a failure of 3.1.2, correct? Now take that ad and put it into an iframe with `html[lang=da]`. Still a failure of 3.1.2, and not of 3.1.1, right? So if that's the case, why then, if we were to remove that incorrect lang=da and not replace it, would it suddenly become a failure of 3.1.1? That is still the same piece of content, where the language of a section of the page can't be programatically determined. 

Regardless of what we conclude on, this change shouldn't be in a PR about xml:lang. We should not have one PR making multiple changes to the rule. Can you please undo it, and either open an issue, or create a different PR.

kasperisager: > If a Danish page has an English ad embedded in a div, that's a failure of 3.1.2, correct?

Yes, since the _default language_ of that ad would be Danish.

> Now take that ad and put it into an iframe with html[lang=da]. Still a failure of 3.1.2, and not of 3.1.1, right?

Yes, because the _default language_ is _still_ Danish.

> So if that's the case, why then, if we were to remove that incorrect lang=da and not replace it, would it suddenly become a failure of 3.1.1?

Because now the _default language_ of the ad can no longer be determined programmatically and the web page therefore fails 3.1.1.

Regardless, I'll put this change into a separate PR and adjust the applicability in this PR.

kasperisager: @DagfinnRomen and @ShadowBB I'd love your take on this as well.

_rules/html-has-lang-b5c3f8.md
kasperisager: Will move this update to a separate PR.

_rules/html-xml-lang-match-5b7ae0.md
kasperisager: Resolving this, let's move the discussion to a separate issue or pull request.

_rules/html-page-lang-b5c3f8.md
```diff
@@ -19,30 +19,25 @@ authors:
 
 ## Applicability
 
-The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element.
-
-**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
```
WilcoFiers: I think this note needs to be put back.

_rules/html-page-lang-b5c3f8.md
```diff
@@ -19,30 +19,25 @@ authors:
...
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html` and whose [browsing context](https://html.spec.whatwg.org/#browsing-context) is a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context).
```
WilcoFiers: Can we simplify the language here? You've made this technically precise, but it has made this much harder to read. Perhaps:

> The rule applies to any `html` element that is in a [document] where the [content type] is `text/html`, and not embedded in an iframe or object (e.g. in the [top-level browser context]).

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -19,83 +19,41 @@ authors:
 
 ## Applicability
 
-The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element with a `lang` and/or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that:
```
WilcoFiers: As before, this has gotten quite technically dense, making it difficult to read. Saying things like that the content type is a property of node document adds unnecessary complexity. The content-type definition already describes explains this. Similarly with top-level browser context. That can only mean one thing, we don't need to specify that it's a type of browser context.

_rules/html-page-lang-b5c3f8.md
kasperisager: > The rule applies to any html element that is in a [document] where the [content type] is text/html, and not embedded in an iframe or object (e.g. in the [top-level browser context]).

The above is saying something entirely different though, and would mean that the following code has two applicable `<html>` elements:

```html
<!doctype html>
<html>
  <body>
    <script>
      document.body.appendChild(document.createElement("html"));
    </script>
  </body>
</html>
```

Only one of those `<html>` elements is a _document element_; there's a reason for the technically precise language used. I should be able to avoid the "stuttering" use of "browsing context" though, if that helps.

_rules/html-page-lang-valid-bf051a.md
kasperisager: See https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r329934298.

_rules/html-page-lang-b5c3f8.md
kasperisager: I'll add a clarifying note 👍

_rules/element-lang-valid-de46e4.md
```diff
@@ -180,7 +124,7 @@ The rule applies to elements with the `body` of a webpage. `html` elements are i
 
 #### Inapplicable Example 2
 
-An empty value for `lang` attribute is ignored as the rule only applies to `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
+An empty value for `lang` attribute is ignored as the rule only applies to `lang` or `xml:lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
```
carlosapaduarte: Remove the mention to `xml:lang` from the description

_rules/element-lang-valid-de46e4.md
```diff
@@ -19,15 +19,19 @@ authors:
...
 
 ## Expectation
 
-For each test target, the `lang` and `xml:lang` have a [valid language subtag](#valid-language-subtag) if the attribute is neither empty ("") nor only [whitespace](#whitespace).
+For each test target, the `lang` attribute has a [valid language subtag](#valid-language-subtag).
 
 ## Assumptions
 
-This rule assumes that the presence of a lang or xml:lang attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+- The presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
```
carlosapaduarte: I was under the impression that we don't do bullets when there is just a single item in the list (but I might be wrong)
Also, don't we usually start the sentence with "This rule assumes that the presence ..."

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -19,83 +19,41 @@ authors:
...
 
-_There are currently no assumptions_
+- The presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
 
 ## Accessibility Support
 
-There are known combinations of a popular operating system with browsers and assistive technologies that do not support the `lang` and `xml:lang` attributes.
-
-While HTML5 specification indicates that `xml:lang` attribute takes priority over `lang` attribute, certain assistive technologies prioritize `lang` over `xml:lang` instead.
+There are known combinations of a popular operating system with browsers and assistive technologies that do not support the `lang` attribute.
```
Brynanders: I'm not sure this still rings true for just the `lang` attribute. I mean we experienced inconsistencies when testing with the `lang` and `xml:lang` attributes but I am not aware of a lack of support for the `lang` attribute.

_rules/element-lang-valid-de46e4.md
```diff
@@ -154,18 +110,6 @@ The `lang` attribute value is not empty ("") and is not a valid primary language
...
-<html>
-	<body>
-		<article xml:lang="123"></article>
-	</body>
-</html>
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
```
carlosapaduarte: In the description change "with" to "within"

_rules/html-page-lang-b5c3f8.md
```diff
@@ -19,30 +19,30 @@ authors:
...
 
-The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element.
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that:
 
-**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
+- is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
 
-## Expectation
+**Node:** `html` elements within `iframe` and `object` elements are not applicable as `iframe` and `object` elements create [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context).
```
carlosapaduarte: Node > Note

_rules/html-page-lang-b5c3f8.md
```diff
@@ -19,30 +19,30 @@ authors:
...
-Each test target has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
+## Expectation
 
-**Note**: HTML5 recommends using `lang` instead of `xml:lang`. This is not known to impact accessibility, which is why use of both is permitted by this rule.
+Each test target has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 
 ## Assumptions
 
-_There are currently no assumptions_
+- The presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
```
carlosapaduarte: Same comments regarding assumptions

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -19,83 +19,44 @@ authors:
...
 
 ## Expectation
 
-For each test target, the `lang` and `xml:lang` attributes have a [valid language subtag](#valid-language-subtag) if the attribute is neither empty ("") nor only [whitespace](#whitespace).
+For each test target, the `lang` attribute has a [valid language subtag](#valid-language-subtag).
 
 ## Assumptions
 
-_There are currently no assumptions_
+- The presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
```
carlosapaduarte: Same comments regarding assumptions

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -19,83 +19,44 @@ authors:
...
 
-The root element of the [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), if it is an `html` element with a `lang` and/or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that:
 
-**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
+- is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
+- has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
+
+**Node:** `html` elements within `iframe` and `object` elements are not applicable as `iframe` and `object` elements create [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context).
```
carlosapaduarte: Node > Note

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -50,52 +53,36 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
...
 
-`html` element with non-matching value for `lang` and `xml:lang`.
+`html` element with non-matching extended language subtags for `lang` and `xml:lang`. While the primary language subtag is Chinese for both the `lang` and `xml:lang` attribute, the extended language subtag is Cantonese for the `lang` attribute and Mandarin for the `xml:lang` attribute.
 
 ```html
-<html lang="fr" xml:lang="en"></html>
+<html lang="zh-yue" xml:lang="zh-cmn"></html>
 ```
 
 ### Inapplicable
```
carlosapaduarte: We're missing inapplicable examples with empty or whitespace only `lang` attributes

You might also want to add another example with a `<math>` element

_rules/html-page-lang-valid-bf051a.md
kasperisager: I tend to agree. Any thoughts on this, @jkodu, as one of the authors of this rule?

kasperisager:
Note to self: Put note in the rule about nested browsing contexts perhaps also needing default languages.

_rules/html-page-lang-b5c3f8.md
DagfinnRomen: @kasperisager Are we moving this discussion to a separate PR?

According to WCAG, a web page is a non-embedded resource (...you know the rest, I won't paste the definition here) including all embedded images and media. The way I see it, a web page with an embedded iframe, would include the iframe as a part of the page. However, the content in this iframe would itself not count as a separate webpage.

I also agree with Kasper in https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r329500292

ShadowBB: @kasperisager I agree with Wilco his interpretation. I would state that indeed the default language of the ad would be Danish. But 3.1.1 doesn’t requires that the default language of the ad has to be programmatically determinable. It only requires that the default language of the page can be programmatically determined.

kasperisager: > It only requires that the default language of the page can be programmatically determined.

@ShadowBB That's exactly my point in https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r329500292; the default language of the page _cannot_ be programmatically determined. Part of it can, yes, but not the part contained within the ad. It's really no different than talking about conformance of a page; if all non-embedded content on a page conforms to the WCAG, but some or all embedded content does not, the page as a whole does not conform to the WCAG.

In general, I'm having a really hard time understanding how one can regard a glass as full if only the top half is filled, but the bottom half isn't. Disregarding the physical impossibility of that analogy, I don't see it as any different.

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -19,19 +19,22 @@ authors:
...
+- has an `xml:lang` attribute.
 
-## Expectation
+**Node:** `html` elements within `iframe` and `object` elements are not applicable as `iframe` and `object` elements create [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context).
 
-For each test target, the value of the primary language subtag (characters before the first dash) for the `lang` and `xml:lang` attributes are the same.
+## Expectation
 
-**Note**: HTML 5 specification requires the `lang` and `xml:lang` attributes to match exactly. This is not known to impact accessibility, which is why it is permitted in this rule.
+For each test target, the values of the [primary language subtags](https://tools.ietf.org/html/bcp47#section-2.2.1) and the [extended language subtags](https://tools.ietf.org/html/bcp47#section-2.2.2), if any exist, for the `lang` and `xml:lang` attributes are the same.
```
WilcoFiers: I don't think this PR should add the extended language subtags here. I'm still not convinced that should be required at all, but even if it is, that should be in a different PR. Here's why i disagree this is required:

An element with `lang="nl-NL" xml:lang="nl-BE"` does not create accessibility problems. While I agree with you that there are scenarios in which having a mismatch between the extended subtags can be a problem, it definitely isn't always a problem. So we shouldn't always fail it. Passing the rule does not mean the SC is satisfied. So even in scenarios where the primary subtag matches and the extended subtags don't, and this does cause an accessibility issue, this rule is still correct.

_rules/html-page-lang-valid-bf051a.md
kasperisager: I will remove it.

kasperisager:
The final call for this pull request will end on October 18.

_rules/element-lang-valid-de46e4.md
```diff
@@ -19,15 +19,19 @@ authors:
 
 ## Applicability
 
-This rule applies to each HTML or SVG element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the `body` element and has a `lang` or `xml:lang` attribute that is neither empty ("") nor only [whitespace](#whitespace).
+This rules applies to any HTML element that:
+
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
+- is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of a `body` element; and
```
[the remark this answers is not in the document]
dd8: For this to fail the element has to have human language content. The rule currently fails this code where the lang attribute doesn't conform to BCP47, but it doesn't cause an accessibility problem:

```
<html>
<body>
 <script lang="jscript">
  window.close();
 </script>
</body>
</html>
```

There are other elements without any text content or text attributes that fall into the same category: `<br>` `<col>` `<hr>` `<wbr>` and probably others

dd8:
I've filed an issue on html-aam on the consistency issues surrounding the language calculation, including iframes. 
https://github.com/w3c/html-aam/issues/258

_rules/element-lang-valid-de46e4.md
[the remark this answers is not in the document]
kasperisager: This will be dealt with in a separate pull request per https://github.com/act-rules/act-rules.github.io/pull/934#discussion_r329500292.

_rules/html-page-lang-b5c3f8.md
```diff
@@ -19,30 +19,30 @@ authors:
...
+This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that:
 
-**Note**: Documents embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
+- is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
 
-## Expectation
+**Note:** `html` elements within `iframe` and `object` elements are not applicable as `iframe` and `object` elements create [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context). However, as these elements are meant to provide a layer of isolation, the declared language of their [parent browsing context](https://html.spec.whatwg.org/#parent-browsing-context) will likely not be inherited, making it possible for empty `lang` attributes in [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context) to also cause accessibility issues.
```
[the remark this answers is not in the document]
kasperisager: As we discussed in person, I'll leave this as a note as it's really an explanation for why the rule is limited to top-level browsing contexts more than something related to accessibility support.

kasperisager:
@dd8 @Jym77 I've hopefully addressed your feedback by now; let me know if this is not the case!

PR #942, opened 2019-09-25
Fix test cases for orientation lock rule (b33eff)

kasperisager:


kasperisager:
I've requested reviews from the people who reviewed #408. In the interest of getting correct test results for this rule, and seeing how it was merged only 16 hours ago, would anybody mind if we just push these changes through?

PR #943, opened 2019-09-25
Updating rule name(s) and filename(s)

EmmaJP:
Updating rule name(s) and filename(s) ... after discussions on what naming conventions should be.

Closes issue(s):
- #250 

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/audio-text-alternative-e7aa44.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: e7aa44
-name: Audio only has a text alternative
+name: `audio` element content has text alternative
 rule_type: composite
 description: |
-  This rule checks if audio only elements have a text alternative available.
+  This rule checks if `audio` elements have a text alternative available.
```
audreymaniez: We should use singular, not plural. And in the previous rule,the "element" is not specified  "This rule checks `audio` is a media alternative for text on the page."

Proposal

>  This rule checks `audio` has a text alternative available

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: 4e8ab6
-name: Role has required states and properties
+name: Element with role has required states and properties
```
audreymaniez: Backsticks and 'attribute' word no ? 

> Element with *`role` attribute* ?

carlosapaduarte: explicit should also be in the name

EmmaJP: In this case I believe the role is implicit rather than an attribute

EmmaJP: Agreed, though wondering if explicit or implicit

_rules/audio-text-alternative-e7aa44.md
EmmaJP: I don't recall changing the description yesterday. It shouldn't have been changed.

_rules/role-required-states-and-properties-4e8ab6.md
EmmaJP: Apparently not ... this rule checks elements with the role attribute.

EmmaJP: Explicit ... ie. elements with a role attribute. eg. `<div role="main">

_rules/form-control-label-descriptive-cc0f0a.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: cc0f0a
-name: Form field label is descriptive
+name: Form control label is descriptive
```
EmmaJP: In this situation label is generic and not specifically the `label` element

PR #955, opened 2019-09-30
[WIP] chore: migrate to using "act-rules-implementation-mapper" for generating implementation metrics

jkodu:
> Note: It was discussed during the F2F, that we split the repository for rules, website, and implementation mapper logic.

This PR takes out the implementation mapping logic into its own module.

- use [act-rules-implementation-mapper](https://github.com/act-rules/act-rules-implementation-mapper)
- remove implementation mapper code from rules repo.

Closes issue(s):

- NA


jkodu:
This PR creates a lot of noise & better to solve this as a part of the website only repo that will separate rules from website and implementation code.

PR #979, opened 2019-10-18
Rule: Table header cells have assigned data cells (d0f69e)

jkodu:
I made a mistake. I merged and reverted the rule - 
"Rule: Table header cells have corresponding data cells (d0f69e)"

See - https://github.com/act-rules/act-rules.github.io/pull/417#issuecomment-543657301.

Unfortunately GitHub does not allow me to re-open a merged PR, so I am creating a new one. As a consequence, we have lost the history of conversations from https://github.com/act-rules/act-rules.github.io/pull/417.

Apologies. Reviews welcome for this rule.


_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,201 @@
...
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 1.3.1: Information and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics)
+
+## Test Cases
```
Jym77: Examples to add:
* `headers` attribute giving a cell to an otherwise empty header (should pass)
```html
<table>
  <tr> <th id="foo"></th> <th id="bar"></th> </tr>
  <tr> <td></td> </tr>
  <tr> <td headers="bar"></td> </tr>
</table>
```

```html
<table>
  <tr> <th id="foo"></th> <th id="bar"></th> </tr>
  <tr> <td headers="foo bar"></td> </tr>
</table>
```

* `headers` attribute removing cell from a header (should fail)
```html
<table>
  <tr> <th id="foo"></th> <th id="bar"></th> </tr>
  <tr> <td></td> <td headers="foo"></td> </tr>
</table>
```

```html
<span id="foo">
<table>
  <tr> <th></th> </tr>
  <tr> <td headers="foo"></td> </tr>
</table>
```

Jym77: The [algorithm for assigning headers cells](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics) is somewhat complex, but Step 3 is very clear about one thing: if there is a `headers` attribute, its content, and only it (even if it result in am empty list of headers), is taken into consideration; if there is no `headers` attribute, the cell position, and only it, it take into consideration.

As per this algorithm, the last two examples above have header cells with no data cell assigned to it.

So, if this rule consider that "is the header for cells" is the same thing as "as some cells assigned by this algorithm", then the last two examples above fail. Otherwise, we need to define what we mean by "is the header for cells".

jkodu:
@carlosapaduarte & @DagfinnRomen, you both had approved the rule - https://github.com/act-rules/act-rules.github.io/pull/417#issuecomment-543657301, so I very kindly request you to have a read of this one.

Jym77:
I'm not seeing any changes. (the only commit since my latest review is a merge commit)

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,201 @@
...
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), and is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the following [semantic roles](#semantic-role):
+
+- [cell](https://www.w3.org/TR/wai-aria-1.1/#cell)
+- [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell)
+- [rowheader](https://www.w3.org/TR/html-aria/#index-aria-rowheader)
```
DagfinnRomen: Use these links?

- rowheader: https://www.w3.org/TR/wai-aria-1.1/#rowheader
- columnheader: https://www.w3.org/TR/wai-aria-1.1/#columnheader

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,201 @@
...
+authors:
+  - Jey Nandakumar
+  - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), and is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the following [semantic roles](#semantic-role):
+
+- [cell](https://www.w3.org/TR/wai-aria-1.1/#cell)
```
DagfinnRomen: I am not sure if I fully understand the applicability. How does the rule handle data cells that have no descendants?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,201 @@
...
+- [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell)
+- [rowheader](https://www.w3.org/TR/html-aria/#index-aria-rowheader)
+- [columnheader](https://www.w3.org/TR/html-aria/#index-aria-columnheader)
+
+## Expectation
+
+The target element is either the row or column header for cells that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the same [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid) element, as that of the target element.
+
+**Note:** Assigning headers cells to data cell is performed as per algorithm - [Forming relationships between data cells and header cells](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics).
```
kasperisager: This is not just informative; it normatively defines the algorithm to be used for forming header/data cell relationships. I'd suggest making it part of a definition of "row header" and "column header".

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Assumptions
+
+_There are currently no assumptions._
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
```
kasperisager: The `headers` attribute has extremely wonky support across different OS / UA combinations. I know @2biazdk did a bit of testing, so maybe he has something to add?

For context, this rule checks if all header cells have associated data cells. If the `headers` attribute is used for associating data cells with header cells, it will likely not work reliably.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: @DagfinnRomen we are referring to descendants of any element with a role of `table` or `grid` which are the cells. Does that help?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: @kasperisager I did some testing on AT's, based on which I will add some support notes.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
DagfinnRomen: I am trying to wrap my head around it, and I may be going out on a tangent here. Following the DOM standard, an object A is called a descendant of an object B, if either A is a child of B or A is a child of an object C that is a descendant of B.

With the applicability as it is now, the rule applies to all cells in the table/grid. That is both rowheaders, coumnheaders and data cells. This is also mentioned in the listed semantic roles, that includes both cells and gridcells.

What will the outcome be when testing a "regular" data cell that has no descendants? It would not make sense that these cells even be tested as header cells.

Would it be better to not include cells and gridcells in the applicability?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,238 @@
...
+			<td>13:00</td>
+			<td>18:00</td>
+		</tr>
+	</tbody>
+</table>
+```
+
+#### Passed Example 5
+
+Each column header has corresponding cells. Usage of `headers` attribute does not change relationship between column headers and cells.
```
carlosapaduarte: Did you really meant "does not change relationship"? It should be "changes the relationship"

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
carlosapaduarte: It's also unclear to me why you are applying this to elements with cell and gridcell semantic roles. Since you're targeting headers, aren't rowheader and columnheader enough?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+</table>
+```
+
+#### Passed Example 4
+
+Each row and column header element has assigned cells, within the same element having a [semantic role](#semantic-role) of `grid`.
+
+```html
+<table role="grid">
+	<thead role="rowgroup">
```
carlosapaduarte: Why include explicit roles for the elements that have the same implicit role? In what way does that improve the example?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+		<th id="Column2"></th>
+	</tr>
+	<tr>
+		<td></td>
+		<td headers="col1"></td>
+	</tr>
+</table>
+```
+
+### Inapplicable
```
carlosapaduarte: Add an example of a table without headers

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,234 @@
...
+  - DOM Tree
+acknowledgements:
+  authors:
+    - Jey Nandakumar
+    - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an element having a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid), that is [visible](#visible) as well as [included in the accessibility tree](#included-in-the-accessibility-tree), where the element has any of the following [semantic roles](#semantic-role):
```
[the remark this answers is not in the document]
jkodu: I worked with Jym on this and came up with this. Please feel free to suggest a better version.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,246 @@
...
+
+#### Passed Example 5
+
+Each column header is assigned to a cell. Usage of `headers` attribute changes the relationship between column headers and cells.
+
+```html
+<table>
+	<tr>
+		<th id="col1"></th>
+		<th id="col2"></th>
```
DagfinnRomen: Shouldn't there be some text in these table header cells?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,246 @@
...
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The column header element has an assigned cell, within the same `table` element.
```
Jym77:
```suggestion
The column header element has an [assigned](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics) cell, within the same `table` element.
```
Same goes for most examples descriptions.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,246 @@
...
+
+- [rowheader](https://www.w3.org/TR/wai-aria-1.1/#rowheader)
+- [columnheader](https://www.w3.org/TR/wai-aria-1.1/#columnheader)
+
+## Expectation
+
+Each target element is [assigned](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics) to at least one element with a [semantic role](#semantic-role) of [cell](https://www.w3.org/TR/wai-aria-1.1/#cell) or [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell), that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the same [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid) element, as that of the target element.
```
Jym77: I'm still not sure that the "descendant of the same table" part is not correct in case of nested tables (where cells in the inner table are descendant of both the inner and outer `table` element).
I think it's OK because the assignation algorithm can only assign to cell "in the same table".

Maybe it's worth changing this to "…that is a cell in the same table as the target element". Or adding a note to make that point explicit.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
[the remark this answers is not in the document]
carlosapaduarte: Here is my suggestion:

The rule applies to any HTML element that has the [semantic role](#semantic-role) of [rowheader](https://www.w3.org/TR/wai-aria-1.1/#rowheader) or [columnheader](https://www.w3.org/TR/wai-aria-1.1/#columnheader) and is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an element that:

- is [visible](#visible); and
- is [included in the accessibility tree](#included-in-the-accessibility-tree); and
- has a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid)

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: I think it is OK as is, because the assignation algorithm as you point out can only point to cell in the same table.

Jym77: Actually, if we rely on the assignation algorithm for pointing into the same table, this can be completely shortened to "each target element is assigned to at least one cell or gridcell" (i.e., no need for the "that is in the same table" part). + a note pointing out that this will necessarily be in the same table.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+  - DOM Tree
+acknowledgements:
+  authors:
+    - Jey Nandakumar
+    - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that has the [semantic role](#semantic-role) of [rowheader](https://www.w3.org/TR/wai-aria-1.1/#rowheader) or [columnheader](https://www.w3.org/TR/wai-aria-1.1/#columnheader) and is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an element that:
```
WilcoFiers: By looking at descendent, you're making an assumption that the header is part of the same table. That's not necessarily the case. I think it's better to keep the relationship explicit. The header has to be a child of a row, and the row has to be a child of either a table|grid or a group that is a child of a table|grid. I know that's a little harder to document, but that's far more accurate, and it's likely how anyone who tests this will look at it.

You're probably going to want to add some definitions for this to avoid having to write this all out into the applicability.

WilcoFiers: Why not check if the header is visible / included in the accessibility tree?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+
+- has a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid); and
+- is [visible](#visible); and
+- is [included in the accessibility tree](#included-in-the-accessibility-tree)
```
WilcoFiers: I think this rule should ignore tables with out cells, and tables with only empty cells. There is nothing wrong with an unpopulated data table that I can see.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+	<tr>
+		<td></td>
+	</tr>
+	<tr>
+		<td headers="col2"></td>
+	</tr>
+</table>
+```
+
+### Failed
```
WilcoFiers: These failures are pretty limited. I think it's worth adding a grid example, and one that's pure ARIA.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+</table>
+```
+
+#### Inapplicable Example 2
+
+The rule does not apply to table element that is not [visible](#visible) in page.
+
+```html
+<html>
+	<style>
```
WilcoFiers: Either put this style in the head, or strip out the html and body elements. Preferably the second, the browser doesn't need them.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+	</body>
+</html>
+```
+
+#### Inapplicable Example 3
+
+The rule does not apply to table element that has no [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) with [semantic roles](#semantic-role) of either [rowheader](https://www.w3.org/TR/wai-aria-1.1/#rowheader) or [columnheader](https://www.w3.org/TR/wai-aria-1.1/#columnheader).
+
+```html
+<table role="presentation">
```
WilcoFiers: Seems unnecessary. Just a simpler version of inapplicable 1.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+
+Each target element is [assigned](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics) to at least one element with a [semantic role](#semantic-role) of [cell](https://www.w3.org/TR/wai-aria-1.1/#cell) or [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell), that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the same [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid) element, as that of the target element.
+
+## Assumptions
+
+_There are currently no assumptions._
+
+## Accessibility Support
+
+- `headers` attribute may not be consistently announced by assistive technologies.
```
WilcoFiers: Table markup in general isn't well supported.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+
+- has a [semantic role](#semantic-role) of either [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid); and
+- is [visible](#visible); and
+- is [included in the accessibility tree](#included-in-the-accessibility-tree)
+
+## Expectation
+
+Each target element is [assigned](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics) to at least one element with a [semantic role](#semantic-role) of [cell](https://www.w3.org/TR/wai-aria-1.1/#cell) or [gridcell](https://www.w3.org/TR/wai-aria-1.1/#gridcell), that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of the same [table](https://www.w3.org/TR/wai-aria-1.1/#table) or [grid](https://www.w3.org/TR/wai-aria-1.1/#grid) element, as that of the target element.
```
WilcoFiers: Can't say "table or gridcell element". This could be misunderstood. There's a reason we're using this verbose "element with the semantic role of XXX" all over the place.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
WilcoFiers: Totally agreed. Although keep in mind that the header assignment algorithm does not cross shadow boundaries.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
WilcoFiers: I think this should ignore header cells that have an empty accessible name. The header assignment algorithm explicitly says to ignore empty cells, so if you follow that logic, empty table headers are never assigned to a cell.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Assumptions
+
+_There are currently no assumptions._
```
WilcoFiers: There is. You're making a rather substantial assumption that all table headers have a relationship with some other cell in the table, and that stuff that's marked up as a table header actually is supposed to be a table header.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: Agreed.
```html
<table id="ok">
<td>
  <table id="ko" style="display: hidden">
    <th>bla</th>
  </table>
</td>
</table>
```
The `th` is a descendant of `ok` which is matching the conditions of the applicability. But the `th` doesn't need to be checked because it's actually a header in `ko` which doesn't match the conditions.

Another way to express it could be "whose closest table/grid ancestor is …", I think.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: I think I agree and I'm afraid that goes a bit deeper.
Essentially, you're saying that
```html
<table>
 <tr><th>Foo</th></tr>
</table>
```
Is not failing 1.3.1 because there is no "structure conveyed through presentation" that needs to be programmatically determinable.
I agree with that.

Going further, if I look at
```html
<table>
 <tr><th>Foo</th><th>Bar</th></tr>
 <tr><td>Hello</td></tr>
</table>
```
Then the second `th` (Bar) has no assigned data cell, but there is no "structure conveyed through presentation" for it either, so I don't think that this example is failing 1.3.1, but it is failing the rule.

Looking a bit into [H43](https://www.w3.org/WAI/WCAG21/Techniques/html/H43), which is cited in the Background, there's nothing in it that says header cells must have assigned data cells.

Which is annoying because that's more or less challenging the very basics of this rule 😢 (actually, that's good with the Assumption suggested by @WilcoFiers).

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: Totally agree on the first part. I'd precise it saying that we assume that header cells have a "**visual** relationship" or "relationship **conveyed through presentation**", or something like that, to pinpoint it.

I agree also with the second part. As far as I can see, `th` and `td` are rendered in the same way (notably, https://html.spec.whatwg.org/multipage/rendering.html#tables-2 (which is **not normative**) doesn't make a difference), so what gives a cell (`th` or `td`) some relationship "conveyed through presentation" is the fact that it's at the start of a column or row, ie that it is supposed to be a table header…

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: Not really, but not perfect either 😉 
Inapplicable Example 1 has a `th` and is inapplicable because of `role="presentation"`.
This one does not have any `th` and is inapplicable because of this. The `role="presentation"` should be removed from that example.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,297 @@
...
+---
+
+## Applicability
+
+The rule applies to any HTML element that has the [semantic role][] of [rowheader][] or [columnheader][], that is [visible][], [included in the accessibility tree][], whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]; which: 1. has [descendants][descendant], that are non empty (""), 2. is [visible][], 3. as well as [included in the accessibility tree][].
+
+## Expectation
+
+Each target element is [assigned][] to at least one element with a [semantic role][] of [cell][] or [gridcell][].
+**Note:** The assigned cell is a [descendant][] in the [flat tree][] of the same [table][] or [grid][] element.
```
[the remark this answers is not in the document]
Jym77: The assignation algorithm only looks into the same table. So as soon as we talked about an "assigned cell", it has to be in the same table. The note is not an extra condition (~"is assigned to a cell AND that cell is within the same table"~) but a reminder ("by the way, assigned implies within the same table so we're not allowing weird cases with cross-table references").

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,304 @@
...
+## Applicability
+
+The rule applies to any HTML element that 
+- has the [semantic role][] of [rowheader][] or [columnheader][] and;
+- is [visible][] and;
+- is [included in the accessibility tree][] and;
+- whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]; 
+1. has [descendants][descendant], that are non empty ("") and; 
+2. is [visible][] and; 
+3. is [included in the accessibility tree][].
```
ShadowBB: Hmmmm. It seems that my suggestion was accepted. However it seems not to be a good sublist. Does anyone know how to create correct sublists in this format?

Apart from this I am approving of this rule now!

Jym77:
```
- item
  1. sub item (starts with 2 spaces)
```
should do the trick.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Background
+
+- [Understanding Success Criterion 1.3.1: Information and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells][assigned]
+
+## Test Cases
+
+### Passed
```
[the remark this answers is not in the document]
jkodu: This might be overkill in terms of examples

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+---
+
+## Applicability
+
+The rule applies to any HTML element that
+
+- has the [semantic role][] of [rowheader][] or [columnheader][] and;
+- is [visible][] and;
+- is [included in the accessibility tree][] and;
+- whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][];
```
Jym77:
```suggestion
- whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]:
```

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+  1. has [descendants][descendant], that are non empty ("") and;
```
Jym77: How would it be possible that the `table` ancestor has only empty descendant?
Can you build an example where this is the only part of the Applicability that is not matched?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+- [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [Forming relationships between data cells and header cells][assigned]
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The column header element has an [assigned][] cell, within the same `table` element.
```
Jym77: We kinda don't need the "within the same `table` element" part because it is implied by "assigned", and adding the precision makes it look like this is an extra condition rather than a consequence.
(same with other example descriptions)

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+	<tr>
+		<td></td>
+		<td headers="col1"></td>
+	</tr>
+</table>
+```
+
+#### Failed Example 3
+
+The column header ('Column B') does not have an assigned cell within the same `table` element.
```
Jym77: This example does not have any column header corresponding to the definition "Column B".
If you don't want to use my suggestion (https://github.com/act-rules/act-rules.github.io/pull/979/files#r363636645) and prefer dismissing it, you still need to somehow have the description actually matching the example.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+<table>
+	<tr>
+		<td>12:00</td>
+	</tr>
+</table>
+```
+
+#### Inapplicable Example 4
+
+There are no [descendants][descendant] within the table.
```
Jym77:
```suggestion
There are no element with a [semantic role][] of header.
```
This is actually Inapplicable for the same reason as Inapplicable example 3.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+<table>
+	<tr>
+		<th role="cell"></th>
+	</tr>
+	<tr>
+		<td></td>
+	</tr>
+</table>
+```
+
```
Jym77: Need an  Inapplicable example where the only header are not included in the accessibility tree.

Jym77: Need an Inapplicable example that fails only the part "has descendant that are non empty" and none of the other Applicability points.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77:
```suggestion
  1. has [descendants][descendant] that are non empty ("") and;
```

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: I strongly suggest re-ordering the Inapplicable examples to match the order of the conditions in the applicability:
* no header (3, 4, 6)
* no visible header (5)
* no header in the accessibility tree (missing)
* header with no `table` ancestor (missing)
* header of a table with no non-empty descendants (missing)
* header of a non-visible table (2)
* header of a table not in the accessibility tree (1)

That will make is easier to follow both in parallel and check verify that Inapplicable examples cover all cases.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+- is [included in the accessibility tree][] and;
+- whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][];
+  1. has [descendants][descendant], that are non empty ("") and;
+  2. is [visible][] and;
+  3. is [included in the accessibility tree][].
+
+## Expectation
+
+Each target element is [assigned][] to at least one element with a [semantic role][] of [cell][] or [gridcell][].
+**Note:** The assigned cell is a [descendant][] in the [flat tree][] of the same [table][] or [grid][] element.
```
Jym77:
```suggestion

**Note:** The assigned cell is a [descendant][] in the [flat tree][] of the same [table][] or [grid][] element.
```
(without the empty line, markdown doesn't even break line)

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Assumptions
+
+This rule assumes that table header cells have a relationship conveyed through presentation with other cells within the same table.
+**Note:** This assumption helps exclude edge cases like: - a table definition where there is only one header cell, or - a table definition where there are multiple headers and no other cells
```
Jym77:
```suggestion

**Note:** This assumption helps exclude edge cases like:
- a table definition where there is only one header cell, or 
- a table definition where there are multiple headers and no other cells
```

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
Jym77: Actually, it's not entirely clear that the header has an ancestor with a `table` or `grid` role. So this may be referring to 0 or 1 element and needs clarification.
( 
- has a `table`/`grid` ancestor; and
- whose closest …)

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -236,80 +238,125 @@ The only element with a header [role][semantic role] is part of a table which is
...
-		<td></td>
+		<td>Cell A</td>
 	</tr>
 </table>
 ```
 
 #### Inapplicable Example 6
 
-There are no elements with a [role][semantic role] of header within the table.
+The only element with a [role][semantic role] of header does not have a closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]
```
Jym77: Because of this, the applicability needs to be more precise.
"The rule applies to a `header` whose closest `table` ancestor…" cannot really be evaluated on this example.
It should be "The rule applies to a `header` with at least one `table` ancestor and whose closest `table` ancestor…"

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -236,80 +238,125 @@ The only element with a header [role][semantic role] is part of a table which is
...
+<div>
+	<tr>
+		<th>Column A</th>
+	</tr>
+</div>
+```
+
+#### Inapplicable Example 7
+
+The only element with header [role][semantic role] is part of a table which has [descendants][descendant] that are empty ("").
```
Jym77: This is inapplicable because the `header` is not visible (so, same reason as inapplicable example 4).
We need an inapplicable example where the only reason for which it's inapplicable is that the table has no non-empty descendants.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,371 @@
...
+---
+
+## Applicability
+
+The rule applies to any HTML element that
+
+- has the [semantic role][] of [rowheader][] or [columnheader][] and;
+- is [visible][] and;
+- is [included in the accessibility tree][] and;
+- whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]:
```
Jym77:
```suggestion
- has at least one ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]; and
- whose closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]:
```
Otherwise, this cannot be evaluated on Inapplicable example 6.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,356 @@
...
+    - Jey Nandakumar
+    - Audrey Maniez
+---
+
+## Applicability
+
+The rule applies to any HTML element that
+
+- has the [semantic role][] of [rowheader][] or [columnheader][] and;
+- is [visible][] and;
```
kensgists: I'm wondering why visibility is important in this proposed rule. 

In "Inapplicable 7," below, we have a passing example which is marked inapplicable because of this "visibility" clause. If the table were not passing but also not visible (due to positioning), would it still be inapplicable? 

A screen reader would still be able to navigate the non-visible table and, if the table were non-passing, it would encounter what is the stated purpose of this rule: to ensure that all headers have assigned data cells.

It seems that the crucial aspect is inclusion in the accessibility tree, not visibility. A non-visible table can still be navigated by a screen reader as though it were visible, so long as it is in the accessibility tree.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,356 @@
...
+	<div role="row">
+		<div role="gridcell">1A</div>
+		<div role="gridcell">1B</div>
+	</div>
+	<div role="row">
+		<div role="gridcell">2A</div>
+	</div>
+</div>
+```
+
```
kensgists: Col B header cell *does* have an assigned cell in the same table element, just not in the third row. 

Is that case captured in the rule description or applicability sections?

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,356 @@
...
+	<tr>
+		<td>12:00</td>
+	</tr>
+</table>
+```
+
+#### Inapplicable Example 2
+
+There are no element with a [semantic role][] of header.
+
```
kensgists: Agreement issue: Either "There is no element..." or "There are no elements..." will work.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,356 @@
...
+<table>
+	<tr>
+		<th style="display: none;">Column A</th>
+	</tr>
+	<tr>
+		<td>Cell A</td>
+	</tr>
+</table>
+```
+
```
kensgists: display:none not only makes the element non-visible, it also removes it from the accessibility tree, I believe.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,356 @@
...
+	<tr>
+		<th>Column A</th>
+	</tr>
+</div>
+```
+
+#### Inapplicable Example 7
+
+The only element with header [role][semantic role] is part of a table which is not [visible][].
+
```
kensgists: Missing article: "The only element with a header role..."

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
```diff
@@ -0,0 +1,356 @@
...
+				<th>Time</th>
+			</tr>
+			<tr>
+				<td>24:00</td>
+			</tr>
+		</table>
+	</body>
+</html>
+```
+
```
kensgists: This comment relates to my comment on the rule Assumptions about visibility: Though this table is not visible on the page, it is still in the accessibility tree and can be navigated with AT.

_rules/table-header-cells-refer-to-data-cells-d0f69e.md
jkodu: @kensgists Good point. I have updated accordingly. The visible criteria is not essential on a table as you pointed out that screen readers will pick this up either how. 

But from our tests, we found that visible on the applicable element, in this case the element with a role of row or column header is essential, as for examples a table marked up as below, 
does not associate headings as expected.

```
<table>
  <tr style="display:none;">
    <th>Projects</th>
    <th>Progress</th>
  </tr>
  <tr>
    <td>My Project</td>
    <td>15%</td>
  </tr>
</table>
```

**IE + JAWS**: Does not work (Reads: "My Project" "15%" as plain text)
**Chrome + JAWS**: Does not work (Reads: "My Project" "15%" as plain text)
**Firefox + NVDA**: does not work. NVDA picks up a table with one row and two
columns: Column 1 "My Projects", Column 2 "15%".
**Safari + VO**: Does not work

Issue #968, opened 2019-10-08
"link/iframe with identical name serve equivalent purpose" (b20e66/4b1c6c) have incorrect accessibility requirement mapping


Jym77:
Following up on https://github.com/act-rules/act-rules.github.io/issues/962#issuecomment-539010202
```html
Joe is drinking a <a href=#cup-of-coffee>cup of joe</a>) in the 
<a href=#cup-of-a-guy-named-joe>cup of Joe</a>.
```
is an example of links with identical name that serve different purposes (thus failing rule [Links with identical accessible names have equivalent purpose](https://act-rules.github.io/rules/b20e66)) but the links are descriptive enough to pass [SC 2.4.9](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only). 
Hence, the accessibility requirement mapping for this rule is wrong.

---
In the same line,
```
<iframe title="cup of joe" srcdoc="A cup of joe is slang for a cup of coffee" />

<iframe title="cup of joe" srcdoc="There's a guy named Joe who owns a cup. It's called the cup of Joe"/>
```
is an example of `iframe` with identical name that serve different purposes (thus failing rule [`iframe` elements with identical accessible names have equivalent purpose](https://act-rules.github.io/rules/4b1c6c)) but the names are descriptive enough to pass [SC 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value).
Hence, the accessibility requirement mapping for this rule is also wrong.

---
The rules might still map to another SC, or just be kept as a best practice…

Jym77:
Thinking more about it, we get plenty of examples with all homonyms. "square" is an accurate description of both the geometric drawing and the arithmetic operation. Many people have identical names, and essentially the need for "disambiguation" pages on Wikipedia show that the same name can accurately describes two different things (the disambiguation pages themselves use context to make the purpose clear).

Note that context might not be enough, _i.e._ we might have links/iframe with same name and context, serving different purposes and still being descriptive enough. For example:
```html
The sailor <a href=https://en.wikipedia.org/wiki/John_Smith_(Medal_of_Honor,_born_1826)>
John Smith</a> received the medal of honor during the American Civil War.

The sailor <a href="https://en.wikipedia.org/wiki/John_Smith_(Medal_of_Honor,_born_1831)>
John Smith</a> received the medal of honor during the American Civil War.
```
So #693 has the same problem.

Jym77:
Put assumption.

[...]

PR #980, opened 2019-10-21
Add assumption of unicity of description (4b1c6c, b20e66)

Jym77:
Add an assumption that a given name can only describe one resource.

Closes issue(s):

- #968 

Need for Final Call:
This will not require a Final Call (changes to assumptions)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
+++ b/_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
@@ -32,7 +32,7 @@ The `iframe` elements in each set of target elements embed the the [same resourc
 
 ## Assumptions
 
-- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of an `iframe` can only accurately describe one resource (notably, homonyms alone are not used as `iframe` names). Thus, if two or more `iframe` elements have the same [accessible name][] but embed different resources, at least one of them does not describe its purpose and this is a violation of [Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/quickref/#name-role-value).
+- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of an `iframe` can only accurately describe one resource (notably, homonyms alone are not used as `iframe` names). Thus, if two or more `iframe` elements have the same [accessible name][] but embed different resources, at least one of them does not describe its purpose.
 
 ## Accessibility Support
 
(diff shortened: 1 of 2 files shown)


_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
```diff
@@ -32,7 +32,7 @@ The `iframe` elements in each set of target elements embed the the [same resourc
 
 ## Assumptions
 
-_There are currently no assumptions_
+- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of an `iframe` can only accurately describe one resource (notably, homonyms alone are not used as `iframe` names). Thus, if two or more `iframe` elements have the same [accessible name][] but embed different resources, at least one of them does not describe its purpose and this is a violation of [Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/quickref/#name-role-value).
```
WilcoFiers: I think you can leave this part off:

> and this is a violation of [Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/quickref/#name-role-value).

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -34,7 +34,8 @@ When followed, the links in each set of target elements resolve to the [same res
...
+- This rule assumes that the purpose of the links with identical [accessible names][accessible name] would not be ambiguous to users in general when seen in context on the web page, which is the exception mentioned in success criterion [2.4.9 Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html). If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links out of context, e.g. on a list of links in a screen reader, which makes it more of a general user experience concern than an accessibility issue.
+- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of a link can only accurately describe one resource (notably, homonyms alone are not used as link names). Thus, if two or more links have the same [accessible name][] but resolve to different resources, at least one of them does not describe its purpose and this is a violation of [Success Criterion 2.4.9: Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/quickref/#link-purpose-link-only).
```
WilcoFiers: As above.

Jym77:
Note: one case where the assumption is actually wrong is when the the test subject is not a single page but several pages.

For example, both https://en.wikipedia.org/wiki/St_Sepulchre-without-Newgate and https://en.wikipedia.org/wiki/Abel_Moysey have a link whose (visible) name is just "John Smith" but refer to different persons and thus different pages. If the test subject is "Wikipedia website", then the new assumption is wrong.

Issue #976, opened 2019-10-15
Can accessible name be non-existent? Only whitespace?


Jym77:
This stems from the discussions in https://github.com/act-rules/act-rules.github.io/pull/400#discussion_r321176440 and https://github.com/act-rules/act-rules.github.io/pull/400#discussion_r321201851

This is really several different questions but the way they interact with each other (and the ACT Rules) makes it worth tackling both at the same time…

Useful link: [Algorithm for computing accessible name](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te).

Useful test case (using `role="link"` to allow name from content):
```html
<div role="link" id="A">A</div>
<div role="link" id="empty"></div>
<div role="link" id="1-space"> </div>
<div role="link" id="2-spaces">  </div>
<div role="link" id="newline">

</div>
<div role="link" id="br"><br /></div>
<div role="link" id="nbsp">&nbsp;</div>
<div role="link" id="A-nbsp-B">A &nbsp; B</div>
<div role="link" id="flat-string">A           B
  C
</div>
<div role="link" id="trim">  hello  </div>

<div aria-label="A" id="aria-A"></div>
<div aria-label="" id="aria-empty"></div>
<div aria-label=" " id="aria-1-space"></div>
<div aria-label="  " id="aria-2-spaces"></div>
<div aria-label="
" id="aria-newline"></div>
<div aria-label="&nbsp;" id="aria-nbsp"></div>
<div aria-label="A &nbsp; B" id="aria-A-nbsp-B"></div>
<div aria-label="A          B
    C" id="aria-flat-string"></div>
<div aria-label="  hello  " id="aria-trim"></div>
```

# Can accessible names be non-existent?
According to the algorithm, no. [Step 1](https://www.w3.org/TR/accname-1.1/#step1) set the `total accumulated text` to the empty string (`""`); [Steps 2](https://www.w3.org/TR/accname-1.1/#step2) only adds to it, never removes from or deletes it; and the end of the algorithm (no step number number) states:
> After all steps are completed, the total accumulated text is used as the accessible name or accessible description of the element that initiated the computation.

Therefore, as per the algorithm, the accessible name can be an empty string but always exists. Any formulation in the line of "something has an accessible name" is always true or should be specified as "something has a non-empty accessible name".

When testing with the code above and elements with id `empty` and `aria-empty`:
* Chrome gives them a name of `""` (empty string).
[...]

PR #981, opened 2019-10-21
Use empty accessible name

Jym77:
* Add notes and accessibility support on definition of accessible name to mention trimming and flattening and incoherent browsers behaviour.
* Replace "has an accessible name that is not only whitespace" by "that is not empty (`""`)" where relevant.

Closes issue(s): #976 (mostly)

Need for Final Call:
This will require a 1 week Final Call (small change that should not have any implication on test cases).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/image-accessible-name-23a2a8.md
+++ b/_rules/image-accessible-name-23a2a8.md
@@ -149,7 +149,7 @@ The `img` element inside a `div` positioned off screen has an empty [accessible 
 
 #### Failed Example 4
 
-The HTML `img` element has an emtpy [accessible name][].
+The HTML `img` element has an empty [accessible name][].
 
 ```html
 <img alt=" " />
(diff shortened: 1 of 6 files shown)


_rules/image-accessible-name-23a2a8.md
```diff
@@ -125,34 +125,34 @@ The HTML `img` element has an [accessible name][] that does not only consist of
 
 #### Failed Example 1
 
-The HTML `img` element is not marked as [decorative][] and does not have an [accessible name][].
+The HTML `img` element is not marked as [decorative][] and has an empty [accessible name][].
```
Brynanders: The `img` is missing an attribute used to provide an accessible name rather than the accessible name being empty.

Jym77: As per accessible name computation, and the new note added to the definition of accessible name in that PR, this results in an empty accessible name.

Brynanders: Got it thanks.

Jym77:
Final Call ends on October 29.

Jym77:
Final Call has ended. So I'm merging this 😄

Issue #835, opened 2019-08-23
iframe rule should not be limited to the same tree (4b1c6c)


WilcoFiers:
In the rule "Iframes with identical accessible names serve equivalent purpose", applicability says the following:

> This rule applies to any set of any two or more iframe elements in the same document tree

The way this is formulated, causes iframes that are embedded each in their own shadow tree to be exempt from the rule. This seems odd, as there's no difference in how AT users will end up interacting with them. Even iframes embedded into the page through other iframes should not be exempt in my opinion.

kasperisager:
Duplicate of #769

Jym77:
@WilcoFiers Are you saying that AT do not respect iframe boundaries?
In relation to #768 I've benn trying the following document:
```html
<!DOCTYPE html>
<html>
<head><title>I have been framed!</title></head>

<body>
	<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html">Contact us</a> all the time.

	<iframe srcdoc='<span><a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html">Contact us</a> from the iframe.</span>'></iframe>
</body>
</html>
```

With Chrome, the top-level accessibility tree only shows the first link ("all the time") and an `iframe` with no child and I need to manually go into the second document to see the second link ("from the iframe"). So, it seems that Chrome accessibility tree does not list all the links together but respects the iframe boundary.
This needs to be tested a bit more thoroughly.

Jym77:
@WilcoFiers @kasperisager 
So, I'e been running some test with NVDA on Chrome. NVDA does not respect iframe boundaries and is dropping all the content upon seeing the page.

Which mean that, for #768 we likely want links to have different names *even if they are in an iframe*; and, for #769 (and this issue), we want iframes' ids to be unique *even if they are nested inside another iframe*.

Now, the correct applicability (for both rules) is (maybe?) links/iframe that are in the same flat tree. That should take care of iframe nesting. On the other hand, that seems weird that a dev has to take care of whatever their framework put in the components shadow tree on which they have no control…

Jym77:
Closed by #982.

PR #973, opened 2019-10-10
Form field label (cc0f0a) update rule

Jym77:
Update the form field labels rule. Hopefully fixing issues of previous versions…

Closes issues:
- #423 
- #953 

Reimplements PRs:
- #447 
- #904 & #905 

Need for Final Call:
This will require a 2 weeks Final Call

---

### **When creating PR:**

- Make sure you requesting to **pull a branch** (right side) to the `develop` branch (left side).

### **After creating PR:**

- Add yourself (and co-authors) as "Assignees" for PR.
- Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- Close the issue that the PR resolves (and make sure the issue is referenced in the top of this comment)
- Optionally request feedback from anyone in particular by assigning them as "Reviewers".

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Final Call period. In case of disagreement, the longer period wins.

PR #982, opened 2019-10-22
Handle shadow tree and nested iframe for "iframe with identical name serve identical purpose" (4b1c6c)

Jym77:
Change applicability and add example to allow descending through nested browsing context and shadow trees.

Closes issue(s):
- #835 

Reimplements #769 to circumvent file renaming problems.

Need for Final Call:
This will require a 2 weeks Final Call.

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/web-page.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Web page
+key: web-page
```
jkodu: As far as I remember, we are still discussing around webpage definition, so we may not need this.

pages/glossary/web-page-html.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Web page (HTML)
+key: web-page-html
+---
+
+Placeholder for definition which is part of this PR:
```
jkodu: Sorry can we not have placeholder definitions. Perhaps the definition can be a separate PR to expedite definitions getting merged earlier, to allow for them to be referred in rules.

Jym77: @jkodu: Having separate PR for definition has been ruled out by @WilcoFiers: https://github.com/act-rules/act-rules.github.io/pull/868#pullrequestreview-283508565
(at the same time, placeholder has also been ruled out by @WilcoFiers: https://github.com/act-rules/act-rules.github.io/pull/973#discussion_r338995324 and given that I'm not a huge fan of duplicating the def (high risk of drift), I'm not sure what to do…)

My idea with placeholder is to get feedback on the rest and not put the PR into Final Call without onboarding the actual definition into it. Thus, there will at the very least be the 2 weeks Final Call period to catch errors in it.

jkodu:
@Jym77 - give me a nudge once the other PR is merged, so I can approve this. Thanks.

_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
```diff
@@ -37,12 +42,13 @@ The `iframe` elements in each set of target elements embed the the [same resourc
 
 ## Accessibility Support
 
-_There are no major accessibility support issues known for this rule._
+This rule assumes that assistive technologies are exposing all `iframe` elements on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its content (notably nested `iframe`), then it is possible for two `iframe` to have identical name but embed different resources without breaking [Success Criterion 4.1.2: Name, Role, Value][sc412] (if said `iframe` are in separate [documents][document] or [shadow trees][shadow tree])
```
WilcoFiers: Editorial: You can't "break" a success criterion. An SC can be either satisfied or not satisfied.

_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
```diff
@@ -130,6 +136,59 @@ Two `iframe` elements within the same [document tree][] have the same [accessibl
 <iframe title="advertising" src="/test-assets/iframe-unique-name-4b1c6c/advertising-two.html"> </iframe>
 ```
 
+#### Passed Example 9
+
+All three `iframe` elements have the same [accessible name][]. The second `iframe` (with `id` `"light"` ) is only part of the [light tree][]. When the [shadow tree][] is attached to `host` and flattened, this `iframe` is overwritten and therefore not part of the [flat tree][]. Hence, only the first and third `iframe` are considered by this rule and they both point to the [same resource][].
```
WilcoFiers: I think you've gone a little over board on the shadow DOM testing. These tests are intended to check if this iframe rule was implemented correctly, not if shadow DOM was correctly implemented.

Jym77: Yes, possibly 🙄 
I've been putting here essentially the same examples as in #768. Which might still be too much… 

=> can we move the discussion concerning which examples to keep to #768 (which is in Final Call). I'll reflect the changes here.

WilcoFiers: That's fine, but I'm still leaving this on changes requested. This is testing the wrong thing.

Jym77:
> @Jym77 - give me a nudge once the other PR is merged, so I can approve this. Thanks.

@jkodu Nudging you right now.

Jym77:
Final Call ending on December 3rd.

PR #986, opened 2019-10-23
Remove acc name tests from 2.4.6 Form control label descriptive (cc0f0a)

WilcoFiers:
I had a long conversation with @ShadowBB a few weeks back about what exactly a label is, and we concluded that the accessible name is not a label. This is based on the following text:

> Further, in the case of labels, this Success Criterion (2.4.6) does not take into consideration whether or not alternative methods of providing an accessible name for form controls and inputs has been used - this aspect is covered separately by 4.1.2: Name, Role and Value. It is possible for controls and inputs to have an appropriate accessible name (e.g. using aria-label="...") and therefore pass Success Criterion 4.1.2, but to still fail this Success Criterion (if the label is not sufficiently clear or descriptive).
> Source: https://www.w3.org/WAI/WCAG21/Understanding/headings-and-labels.html

There is a similar argument to be made for 2.5.3 label in name. If the label can be the name, that success criterion doesn't make much sense.

The reason to remove this, specifically is Passed example 5. An input field without a visible label **does not** satisfy SC 2.4.6.

Need for Final Call:
This will require a 1 week Final Call << small changes affecting a small number of test cases, if in doubt do not use this


_rules/form-control-label-descriptive-cc0f0a.md
```diff
@@ -92,17 +102,6 @@ Label is [visible][], but not included in accessibility tree
 <input aria-labelledby="label_fname" type="text" name="fname" />
 ```
 
-#### Passed Example 5
```
WilcoFiers: This is the main reason to change this rule. This test case does not pass SC 2.4.6.

Jym77:
Can we try to finish #973 instead?
Given that it also fix the problem and has been waiting for reviews for almost 2 weeks…

_rules/form-control-label-descriptive-cc0f0a.md
```diff
@@ -139,22 +138,11 @@ Label is [visible][], but not included in accessibility tree, and does not descr
...
-</p>
-<input aria-labelledby="label_fname" type="text" name="fname" />
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-`Label` that is neither [visible][] to users, nor [included in the accessibility tree][].
+`Label` that is neither [visible][] to users.
```
carlosapaduarte: that is neither > that is not

_rules/form-control-label-descriptive-cc0f0a.md
```diff
@@ -165,7 +153,7 @@ Label is included in accessibility tree, but not [visible][], and does not descr
 
 #### Inapplicable Example 2
 
-Programmatically associated `p` element that is neither [visible][] nor [included in the accessibility tree][].
+Programmatically associated `p` element that is neither [visible][].
```
carlosapaduarte: that is neither > that is not

WilcoFiers:
@Jym77 Yes there is, the content of that label element is the WCAG label. There is no requirement in WCAG that the label be part of the accessible name.

_rules/form-control-label-descriptive-cc0f0a.md
ShadowBB: How would this fail SC 2.4.6? The label is descriptive. I think this fail SC 3.3.2 (depending on your definition of "provided".

Jym77:
> @Jym77 Yes there is, the content of that label element is the WCAG label. There is no requirement in WCAG that the label be part of the accessible name.

I have to disagree with this. As per the note on [WCAG definition of label](https://www.w3.org/TR/WCAG21/#dfn-labels):
> A label is presented to all users

I understand this as labels needing to be presented both to user of AT and people not using AT. I agree that an accessible name alone is not a WCAG label (it is not presented to people not using AT). But I think that a WCAG label **also** need to be presented to users of AT.

This interpretation (WCAG label has to be presented to users of AT) was confirmed by @DagfinnRomen. Here's relevant excerpt from mail:

---
Only a visual identification, hence not a “WCAG label” (not presented to all users), hence 2.4.6 passes/is inapplicable (1.3.1 or 3.3.2 may fail):
```html
<span aria-hidden=true>First name:</span>
<input type="text" name="first name"/>
```
•	DR: The span is only visual. There is no WCAG label and this case is inapplicable for 2.4.6.

---
Replacing the `span` with a `label` does not really change this case nor the argument.

PR #989, opened 2019-10-23
page-title-descriptive: remove inaccessible title example (c4a8a4)

WilcoFiers:
Assistive technologies tend to ignore punctuations, which is why a title like ";)" does not work well. We shouldn't have that as a passing example. Nothing in the rule applicability / expectation mentions anything about punctuations, so there's no reason for the example.

Need for Final Call:


EmmaJP:
What about failed example 3? It appears to be a duplicate of failed example 1 (page title present, with content that does not describe page content), but someone may think the punctuation is the reason for it being a failed example.

PR #990, opened 2019-10-23
page-has-title, passed example must satisfy the SC

WilcoFiers:
All passed examples need to satisfy the success criterion it is designed for. This passed example 6, with only punctuations in it, does not.

Need for Final Call:


WilcoFiers:
@kengdoj because the example still passes the rule. [HTML page title is descriptive](https://act-rules.github.io/rules/c4a8a4) tests for that type of issue.

EmmaJP:
Would an emoji in a title not meet the success criterion? For example, `<title>;-)</title>` could be considered meaningful and is not only whitespace.

WilcoFiers:
@EmmaJP No, I don't think so. Things like that are non-text content according to WCAG. It needs a text alternative along side that emoji.

PR #991, opened 2019-10-23
fix: update repo url to `https` instead of `git`

jkodu:
This PR updates the repository url to use `https` as against `git`.
The repo url is used by the ACT Rules website to resolve some URL's dynamically when the site is built & these need to be https.

Closes issue(s):
- https://github.com/act-rules/act-rules-web/issues/44


jkodu:
This is not a rule related change & will not require 3 approvals.

PR #994, opened 2019-10-24
Examples in glossary

### LINKED ISSUE #182: Add examples for definitions

Jym77:
Add some examples in three of the most used (and maybe most subtle) definitions.

These examples are intended to be non-normative, not testable. They are here to illustrate some common pitfalls about these definitions and help users of ACT rules understand the definitions.

Closes issue(s):
- Partially closes #182 

Need for Final Call:
This will not require a Final Call (adding non-normative text)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/accessible-name.md
```diff
@@ -8,3 +8,53 @@ The programmatically determined name of a user interface element that is [includ
...
+<a href="https://act-rules.github.io/">ACT rules</a>
+```
+
+This `span` element has an empty accessible name (`""`) because `span` does not allow [name from content](https://www.w3.org/TR/wai-aria/#namefromcontent).
+
+```html
+<span>ACT rules</span>
+```
+
+**Note**: when the same element can have an accessible name from several sources, the order of precedence is: `aria-labelledby`, `aria-label`, own attributes, `label` element, name from content. The examples here are listed in the same order.
```
kasperisager:
```suggestion
**Note**: When the same element can have an accessible name from several sources, the order of precedence is: `aria-labelledby`, `aria-label`, own attributes, `label` element, name from content. The examples here are listed in the same order.
```

pages/glossary/accessible-name.md
```diff
@@ -8,3 +8,53 @@ The programmatically determined name of a user interface element that is [includ
...
+
+**Note**: For more examples of accessible name computation, including many tricky cases, check the [Accessible Name Testable Statements](https://www.w3.org/wiki/AccName_1.1_Testable_Statements)
```
kasperisager:
```suggestion
**Note**: For more examples of accessible name computation, including many tricky cases, check the [Accessible Name Testable Statements](https://www.w3.org/wiki/AccName_1.1_Testable_Statements).
```

pages/glossary/included-in-the-accessibility-tree.md
```diff
@@ -8,3 +8,42 @@ Elements included in the accessibility tree of platform specific accessibility A
...
+<span aria-hidden="true">ACT rules</span>
+```
+
+This `span` element is positioned off screen, hence is not [visible](#visible), but is nonetheless included in the accessibility tree.
+
+```html
+<span style="position: absolute; top:-9999em">ACT rules</span>
+```
+
+Although the `span` element with an `id` of "label" is not itself included in the accessibility tree, it still provides an [accessible name](#accessible-name) to the other `span`, via the `aria-labelledby` attribute. Thus, it is still indirectly exposed to users of assistive technologies.
```
carlosapaduarte: This might be confusing to some readers. The element is not in the accessibility tree but is indirectly exposed to users of AT (thus it seems it is in accessibility tree). Perhaps we can leave the last sentence out?

pages/glossary/accessible-name.md
```diff
@@ -8,3 +8,53 @@ The programmatically determined name of a user interface element that is [includ
...
 For native markup languages, such as HTML and SVG, additional information on how to calculate the accessible name can be found in [HTML Accessibility API Mappings 1.0, Accessible Name and Description Computation (working draft)](https://www.w3.org/TR/html-aam/#accessible-name-and-description-computation) and [SVG Accessibility API Mappings, Name and Description (working draft)](https://www.w3.org/TR/svg-aam/#mapping_additional).
+
+#### Examples
+
+The `span` element with an `id` of `"target"` has an accessible name of "ACT rules" given by the `aria-labelledby` attribute and both associated elements. The fact that the element with an `id` of `"label-1"` is hidden to all users does not prevent it from giving an accessible name to other elements.
+
+```html
+<span id="label-1" style="display: none">ACT</span>
+<span id="label-2">rules</span>
+<span id="target" aria-labelledby="label-1 label-2"></span>
```
EmmaJP: I think it would be helpful to use realistic examples, and possibly not on span elements (https://www.w3.org/TR/using-aria/#practical-support-aria-label-aria-labelledby-and-aria-describedby). 

For example, a common use of aria-labelledby may be to combine a group label with an element's label and establish that association. There is a nice example at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute.

pages/glossary/accessible-name.md
```diff
@@ -8,3 +8,53 @@ The programmatically determined name of a user interface element that is [includ
...
+```
+
+This `span` element has an accessible name of "ACT rules" given by its `aria-label` attribute.
+
+```html
+<span aria-label="ACT rules"></span>
```
EmmaJP: As before, possibly better to use an element other than span and a realistic example. Perhaps a button with a more expansive area-label, eg:

`<button aria-label="Share ACT Rules">Share</button>`

pages/glossary/included-in-the-accessibility-tree.md
ShadowBB: Or changed into something explicit like "Thus, it is still indirectly exposed to users of assistive technologies without being included in the accessibility tree"

Jym77: I'm more in line with @ShadowBB on that one. The goal of that sentence is to say that "just because you put a `aria-hidden` attribute, it does not mean you can do whatever you want without breaking accessibility".

Typically,
```html
<img src="https://picsum.photos/100" aria-hidden=true id="img" />
<button aria-labelledby="img">Foo</button>
```
Although the image is not in the accessibility tree, not giving it a name is still creating accessibility issues (namely, not giving a name to the button).

Jym77: @carlosapaduarte @ShadowBB I tried to improve the description of the example to better communicate the intend.

PR #995, opened 2019-10-24
fix: `autoplay` attribute usage in `no-autoplay-audio` rule (aaa1bf / 4c31df / 80f0bf)

jkodu:
When implementing this rule in `axe`, I noticed perhaps we should refer to `autoplay` as `auto-play`.

Also noticed that the URL's to atomic rules from composite rules were not correct.

Closes issue(s):
- NA


_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -76,15 +76,15 @@ The `<video>` element audio output does not last longer than 3 seconds.
 
 #### Failed Example 1
 
-The `<audio>` auto-plays for more than 3 seconds.
+The `<audio>` `autoplay's` for more than 3 seconds.
```
Jym77: I think that "auto-play" was used as a verb, rather than as an attribute name. This is particularly visible on that case where the conjugation (final 's') make the whole syntax look weird.

I'd rather go for "automatically play" if you want to disambiguate "auto-play" from the `autoplay` attribute.

jkodu: In most places where just `auto-play` is used, it is rather confusing, hence the change. 

In the case of a plural, I am happy to change it to automatically play. I will await a few more people to have an opinion, before changing this.

Jym77: Yes, I agree that "auto-play" was a bit confusing. I think most places it was used as a verb and should be fixed as "automatically play" rather than "`autoplay`". Best w%ould be to have @Brynanders opinion on that since he wrote the rules…

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: aaa1bf
-name: auto-play audio does not exceed 3 seconds
+name: '`autoplay` audio does not exceed 3 seconds'
```
DagfinnRomen: If autoplay is used as an adjective (?) or a verb, there is no reason for the backticks. In these cases, autoplay does not refer to the autoplay attribute.

To avoid confusion, consider rewriting where autoplay does not refer to the attribute.

Suggestion in this case: Autoplay audio does not exceed 3 seconds -> Audio that plays automatically does not exceed 3 seconds.

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -1,9 +1,9 @@
...
 rule_type: atomic
 description: |
-  auto-play audio content does not output audio for more than 3 seconds
+  '`autoplay` audio content does not output audio for more than 3 seconds'
```
DagfinnRomen: Same as my previous comment.

Suggestion: Autoplay audio content -> Audio content that plays automatically

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
DagfinnRomen: I am in favour of using "automatically play" or "play automatically" in stead of autoplay, to avoid confusion.

In this case, "autoplay's" is incorrect and should be "autoplays" without the apostrophe.

Refering to my previous comment, the backticks should be removed where autoplay is used as a verb and does not refer to the attribute.

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -76,15 +76,15 @@ The `<video>` element audio output does not last longer than 3 seconds.
...
+The `<audio>` `autoplay's` for more than 3 seconds.
 
 ```html
 <audio src="../test-assets/moon-audio/moon-speech.mp3" autoplay="true" controls></audio>
 ```
 
 #### Failed Example 2
 
-The `<video>` element audio auto-plays for more than 3 seconds.
+The `<video>` element audio `autoplay` for more than 3 seconds.
```
DagfinnRomen: Remove backticks or consider rewriting.

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -55,7 +55,7 @@ There are no major accessibility support issues known for this rule.
 
 #### Passed Example 1
 
-The `<audio>` element does not auto-play for more than 3 seconds.
+The `<audio>` element does not `autoplay` for more than 3 seconds.
```
DagfinnRomen: Remove backticks, or autoplay -> play automatically.

_rules/auto-play-audio-has-control-mechanism-4c31df.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: 4c31df
-name: auto-play audio has a control mechanism
+name: '`autoplay` audio has a control mechanism'
```
DagfinnRomen: If autoplay is used as an adjective (?) or a verb, there is no reason for the backticks. In these cases, autoplay does not refer to the autoplay attribute.

To avoid confusion, consider rewriting where autoplay does not refer to the attribute.

Suggestion in this case: Autoplay audio -> Audio that plays automatically

This also goes for the other uses of the word autoplay in this rule.

_rules/no-auto-play-audio-80f0bf.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: 80f0bf
-name: Video or audio has no auto-play audio
+name: 'Video or audio has no `autoplay` audio'
```
DagfinnRomen: If autoplay is used as an adjective (?) or a verb, there is no reason for the backticks. In these cases, autoplay does not refer to the autoplay attribute.

To avoid confusion, consider rewriting where autoplay does not refer to the attribute.

This also goes for the other uses of the word autoplay in this rule.

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
Brynanders: @jkodu @Jym77 @DagfinnRomen I agree with "automatically play(s)" or "play(s) automatically."

PR #1020, opened 2019-11-14
Ignore .idea

Jym77:
Ignoring a folder which is auto-created by my IDE…

Need for Final Call:
This can be merged with 1 approval (editorial change to git config)


jkodu:
@Jym77 you may want to add workspace settings on webstorm to auto ignore this

Issue #1009, opened 2019-11-06
"Language of parts" and failure when ancestors/document itself lack language definition


patrickhlauke:
Looking over [Element within body has valid lang attribute](https://act-rules.github.io/rules/de46e4), it seems to only apply to elements with an actual `lang` attribute present.

Just to gauge initial thoughts: assuming a page lacks a `lang` attribute altogether (i.e. fails 3.1.1 Language of Page), elements in the body without a `lang` attribute also fail 3.1.2 automatically (since if the human language of the overall page was not set, THEIR human language also cannot be programmatically determined). So, I'd argue, in the below, the `<p>` would fail as well:

```
<html>
  <body>
    <p></p>
  </body>
</html>
```

Is it worth me making an attempt at defining this as a new rule?

patrickhlauke:
In essence, you'd want to go through every node in the `<body>`, and for each determine if it, or any of its ancestors all the way up to `<html>` have some defined and valid human language definition (wondering if there's even a property on the node itself that can provide this, without the need to climb down the whole DOM tree)

Jym77:
Yes, this rule only check that **if** you use a `lang` attribute, it is valid. The rule does not check whether a `lang` attribute` is needed or not. Like many (all?) rules, it is only a partial check for its SC: you can pass the rule and still fail the SC.

I agree that the `p` in your example fails the SC. It is inapplicable for the rule (there is no `lang` attribute).

You are, of course, welcome to try and write a new rule to check whether a `lang` attribute is needed in the first place 😄 Be aware that the review process can be very lengthy and time consuming (for the author also). But if the rule looks good you might also find people willing to help for the final polishing…

An issue I see with the current crude formulation is that if a page has no `lang` attribute at all, you're going to flag every single element as a failure. For page with thousands of elements, that is going to be a lot of failures. This might not be a real issue for tools, but tools users might be drowned in thousands of errors when it's just the page needing a `lang` attribute (and worse case, somebody would believe that every single element needs a `lang` attribute).

A better middle ground might be to target specific elements where it make sense to have a `lang` attribute (typically, landmarks). <= this is a personal opinion only.

patrickhlauke:
> For page with thousands of elements, that is going to be a lot of failures.

[...]

PR #1027, opened 2019-11-21
New Rule: Element language is programmatically determinable

patrickhlauke:
This rule is primarily aimed at catching situations where 3.1.1 Language of Page fails, which also leads to automatically failing 3.1.2 (since if the language for the overall page cannot be programmatically
determined, neither can the language for any text nodes in the page itself unless there's been some explicit language definition further down in the DOM tree)

Closes issue(s):

- #1009 

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


patrickhlauke:
(note that I made this contribution from my own fork, as I don't think I can just clone and add a new branch directly here? for that reason, I'm also unable to add a label/assign myself)

patrickhlauke:
I'll also add that this is my first stab at doing an ACT rule, so I expect that I made some fundamental logic/form errors...

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
...
+
+This rules applies to any HTML element that:
+
+* contains a [text node](https://dom.spec.whatwg.org/#text)
+* is a descendant in the flat tree of a `body` element
+* is included in the accessibility tree
+
+## Expectation (1)
+
+For each test target, the test target itself or any of its ancestor elements (up to and including the root element of the document) has a `lang` attribute with a valid language subtag.
```
Jym77: "ancestor" should probably be specified "in the flat tree".

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
...
+<html lang="en">
+	<body>
+		<p>Content</p>
+	</body>
+</html>
+```
+
+#### Passed Example 2
+
+Language of the `p` element is inherited by the `lang` attribute on the `div` parent element. Note that this example fails 3.1.1 Language of Page (Level: A) as the `html` root element lacks a `lang` or `xml:lang` attribute.
```
Jym77: Maybe mention that both the `p` and the `div` pass the rule.

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
...
+		</div>
+	</body>
+</html>
+```
+
+### Failed
+
+#### Failed Example 1
+
+Language of the `p` element cannot be programmatically determined as the element itself and none of its ancestors, including the `html` root element, has a defined `lang` or `xml:lang` attribute. Note that this example also fails 3.1.1 Language of Page (Level: A) as the `html` root element lacks a `lang` or `xml:lang` attribute.
```
Jym77: I think we should remove all references to `xml:lang` to follow what we did we the other language rules.

_rules/element-lang-programmatically-determinable-7ed469.md
Jym77: The failure of 3.1.1 is not really relevant for this rule and is already mentioned in the background. It could probably be removed from the examples.

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
+---
+id: 7ed469
+name: Element language is programmatically determinable
+rule_type: atomic
+description: |
+  This rule checks that the language of an element in the page body can be programmatically determined.
+accessibility_requirements: # Remove whatever is not applicable
```
Jym77: I think the rule is also mapping to 3.1.1: if an element has no ancestor with a valid `lang`, then the document element has no valid `lang`.

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
...
+		<div lang="en">
+			<div>
+				<p>Content</p>
+			</div>
+		</div>
+	</body>
+</html>
+```
+
+### Failed
```
Jym77: Add a Failed Example where the `lang` attribute is invalid. This is one of the failure conditions in the expectation and thus needs to be tested.

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
...
+
+```html
+<html>
+	<body>
+		<p aria-hidden="true">Content</p>
+	</body>
+</html>
+```
+
+#### Inapplicable Example 2
```
Jym77: This doesn't add much to the previous Inapplicable Example and can probably be removed.

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,142 @@
...
+
+```html
+<html>
+	<body>
+		<p>Content</p>
+	</body>
+</html>
+```
+
+### Inapplicable
```
Jym77: Add an Inapplicable example where the `body` element has no descendant. This is an applicability condition that needs to be tested.

patrickhlauke:
Just on this point

> 5. There can also be a failed example where one element is OK but another one is not. This illustrates that the SC fails if there is at least one failed outcome.

doesn't the rule get applied to every single element separately, in isolation? Don't quite understand therefore why there should be an example with two separate elements, as they'd be tested individually?

Jym77:
Yes, the rule is applied to every element.
```html
<body>
  <p lang="en" id="ok">Hello</p>
  <p id="ko">World</p>
```
the `ok` element is passing the rule. The `ko` element is failing the rule. As a result of a single failure of the rule, the SC also fails. Having a Failed Example like that remind of that "single fail is enough" relationship to the SC.
Not a very strong point for adding the example, I admit 😄 And I won't push it further if you or others think it's not necessarily.

_rules/element-lang-programmatically-determinable-7ed469.md
patrickhlauke: wondering if the `div` gets tested directly as it doesn't contain a text node (or if the applicability needs to be changed to make it clear that the text node needs to be the immediate child of the target)

_rules/element-lang-programmatically-determinable-7ed469.md
patrickhlauke: makes sense, this was already suggested by a colleague - i only included it here because there's still one existing rule that does mention `xml:lang` https://act-rules.github.io/rules/5b7ae0

_rules/element-lang-programmatically-determinable-7ed469.md
patrickhlauke: assuming the various ways in which something is hidden/omitted from the accessibility tree gets checked properly, sure

_rules/element-lang-programmatically-determinable-7ed469.md
Jym77: I'm currently understanding the applicability "contains a text node" as "has a text node as a descendant".
Could be worth indeed to update it to "has a text node as a child" to make it clear (and also to greatly reduce the number of targets…)

_rules/element-lang-programmatically-determinable-7ed469.md
Jym77: Yes. It's the rule that specifically test `xml:lang` 😄 
Other language rules were also using `xml:lang` but that has been cleaned recently (not that you can guess, of course…) Notably because it seems that `xml:lang` is almost not used in practice.
Additionally, your rule specifically requires a `lang` attribute (not an `xml:lang` one), so there no reason to mention `xml:lang`.

_rules/element-lang-programmatically-determinable-7ed469.md
Jym77: Yes. But this rule is not about checking all the possible ways to remove something from the accessibility tree. The examples should illustrate different aspects of the rule, but not the various ways to achieve these aspects.

We could similarly argue that the examples are currently not showing a case where the `lang` attribute is set on the 10th ancestor, or something similar. But we probably agree that we are not going to illustrate all the ways an element can have an ancestor…

Jym77:
Also, I forgot to mention that you need to add yourself as "contributor" in [`package.json`](https://github.com/act-rules/act-rules.github.io/blob/develop/package.json). You can either do it as part of this PR or open a separate one. Given the usual delay we have for finishing rule PRs, it might be worth making a quick separate PR for this instead…

_rules/element-lang-programmatically-determinable-7ed469.md
patrickhlauke: i understood `accessibility_requirements` as the list of (in this case) WCAG SCs that this is used to test conformance to. as this rule is scoped to descendants of `body`, it cannot be used to test 3.1.1 language of the page itself, as that needs to (in HTML) be defined for the `html` element?

patrickhlauke:
> 1. many links are missing. Notably, "included in the accessibility tree", "flat tree", all the Success Criteria, … Please add them all over the place. You can use "footnote style" reference of markdown for links that are used several time.
> 2. Add the Understanding document to Background. Also add CSS scoping with the WiP mention given that you reference it (check what is done in other rules).
> 3. Given that you work on the flat tree (which makes sense), you need to add example with shadow tree. Given the nature of the rule, probably something where the shadow tree + flattening is "moving" a lang attribute. Maybe something with a slot where the slotted element has no lang ancestor in the light tree, but shadow tree manipulation put it under an element with a lang (or the other way around).
> 4. Maybe having an example with an iframe to show that the rule stops at document boundaries.
> 5. There can also be a failed example where one element is OK but another one is not. This illustrates that the SC fails if there is at least one failed outcome.

Done 1, 2 and 5. Have to admit that for 3 I have no idea what that even means, and 4 not quite sure how that would be shown/expressed in a code sample. Any hints for those two would be appreciated.

Jym77:
> > 3. Given that you work on the flat tree (which makes sense), you need to add example with shadow tree. Given the nature of the rule, probably something where the shadow tree + flattening is "moving" a lang attribute. Maybe something with a slot where the slotted element has no lang ancestor in the light tree, but shadow tree manipulation put it under an element with a lang (or the other way around).
> > 4. Maybe having an example with an iframe to show that the rule stops at document boundaries.
> 
> Done 1, 2 and 5. Have to admit that for 3 I have no idea what that even means, and 4 not quite sure how that would be shown/expressed in a code sample. Any hints for those two would be appreciated.

From the top of my head…
```html
<html>
<body>
  <div lang="en">Hello
  <iframe srcdoc="<body><div>World</div></body>" />
  </div>
</body>
</html>
```
The `div` inside the `iframe` has no `lang` because in your expectation the "ancestor" is restricted in the same document. This the rule fails on it.
It's been discussed whether `lang` should leak inside `iframe` or not (it is a leak of information to an external resource). I'm not sure what the actual conclusion is 😕 but it's worth illustrating how the rule work.

For the shadow tree, … hmmm…
```html
<html>
<body>
      <div id="host">
        <span>Hello world</span>
      </div>
      
      <script>
        const host = document.getElementById('host')
        const shadowRoot = host.attachShadow({ mode: 'open' })
      
        shadowRoot.innerHTML =
          '<div lang="en"><slot></slot></div>'
      </script>
  </body><html>
```
In the flat tree, the `span` is slotted inside the shadow tree and thus has the `div lang="en"` as parent. So this passes the rule even if it looks like the `span` has no language…
(the substitution can be made a bit more explicit by naming the slot if you want)

If you use instead `shadowRoot.innerHTML='<div></div>'`, then there is no slot and the `span` disappears from the flat tree (and is not rendered), hence the rule is inapplicable. But it is less stricking (a simple `host.innerHTML='<div></div>'` would have the same effect).

_rules/element-lang-programmatically-determinable-7ed469.md
Jym77: I agree that the rule is not designed to test SC 3.1.1, but it still does test a bit of it indirectly…
As always, we focus on the fail-on-fail relationship. A failure of this rule indicates a failure of SC 3.1.1 (in addition to 3.1.2). A Pass on this rule does not indicate a Pass of 3.1.1, but neither of 3.1.2… That's how it is for most rules.

Then, it is true also that a failure of this rule implies a failure of [HTML page has lang attribute](https://act-rules.github.io/rules/b5c3f8), and that is why it fails 3.1.1. That's also OK. This rule checks more.

patrickhlauke: wondering now if another failure purely for this rule (which i've not added here, but can if it's deemed an actual failure) would be something like
```
<html lang="en">
  <body>
    <div lang="foo">
      <p>Content</p>
    </div>
  </body>
</html>
```

where the `p` fails because its ancestor `div` has an invalid language attribute, *but* the page itself does pass 3.1.1...and so a failure of this rule does not indicate a failure of 3.1.1

Jym77: The rule currently does not fail that because "any of its [the `p`] ancestor" has a valid `lang` tag.
I agree that this is a failure of 3.1.2 but not of 3.1.1

I see 2 possibilities:
=> Change the rule to "its closest ancestor with a `lang` attribute" or something similar. This makes the rule more tricky. In this case, it would indeed not map to 3.1.1
=> Leave the rule as is. This keeps the rule simpler and more focused. This case is already handled by [Element within `body` has valid `lang` attribute](https://act-rules.github.io/rules/de46e4), so there is no real need to enforce it in this rule. The main "problem" with this solution is that Passed Examples must pass the SC, so we cannot use this as a Passed Example to illustrate the limitation of the rule. This is not really a big deal…

Given that this case is already handle by anther rule, I'd rather keep this rule simple.

_rules/element-lang-programmatically-determinable-7ed469.md
```diff
@@ -0,0 +1,180 @@
...
+<html>
+	<body>
+		<div lang="en">
+			<div>
+				<p>Content</p>
+			</div>
+		</div>
+	</body>
+</html>
+```
```
Jym77: These two examples are actually pretty much the same (the `lang` attribute is on an ancestor which is not the `html` element). You can remove one of them.

OTOH, I think it would be good to add an example to illustrate the "**child** text node" part of the rule.
```html
<body>
<div>
<p lang="en">Hello</p>
</div>
</body>
```
The `p` is good. The `div` is inapplicable. So this passes the rule.

patrickhlauke: > Change the rule to "its closest ancestor with a lang attribute" or something similar

I think I'm actually tending towards this one, as the closest ancestor is indeed what plays the major part in the essence of this rule, as it relates more closely in how the language of the specific element itself is calculated/programmatically determined.

And mainly, because I do think there's value in making sure that it's clear that there can be failures of this rule that fail 3.1.2 but not 3.1.1.

patrickhlauke: > These two examples are actually pretty much the same

ah, good spot. wanted to make the difference between parent and ancestor clear, but probably no need to do two examples that are so close to each other. will go with ancestor one, as parent is just a special case of ancestor

Jym77: Note that you can also go even simpler: 
check the existence of a `lang` attribute, not its validity. Then you can go for any ancestor without problem, the rule is much simpler and keep a single concern (it does not need to know what are valid/invalid language tags). Given that we already have rules that check validity of language tags, this is not going to be a real problem.

The point here being that we don't need the rule to be completely covering a single SC. It is better (imho) to have simple rules checking a simple thing and use a combination of rules to cover a SC (rules are already complicated beast, keeping them simple helps writers, readers and testers; simpler rules are also better for separation of concern).

This is a personal opinion and mostly showing what can be done. Of course, go the direction you want 😄 given that you are the one doing the job in the first place…

patrickhlauke: i'll mull it over. appreciate the thorough discussion on this, particularly as i dove in on the deep end here :)

WilcoFiers:
I'm going to close this as this PR has not been updated in 8 months. If someone wants to pick this up again, we can reopen the PR or create a new one.

PR #1029, opened 2019-11-21
Change "not whitespace" to "not empty" button-acc-name

WilcoFiers:
This rule was overlooked in an update in: https://github.com/act-rules/act-rules.github.io/pull/981/files

Need for Final Call: 1 week

### DIFF
--- a/_rules/button-accessible-name-97a4e1.md
+++ b/_rules/button-accessible-name-97a4e1.md
@@ -143,7 +143,7 @@ Value attribute does NOT give an [accessible name][], only for input elements.
 
 #### Failed Example 2
 
-Span tag with role button with no name.
+Span tag with role button with an empty [accessible name][].
 
 ```html
 <span role="button"></span>
@@ -151,7 +151,7 @@ Span tag with role button with no name.
 
 #### Failed Example 3
 
-Off screen element without an [accessible name][].
+Off screen element has an empty [accessible name][].
 
 ```html
 <html>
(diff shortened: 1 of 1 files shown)


_rules/button-accessible-name-97a4e1.md
```diff
@@ -25,7 +25,7 @@ The rule applies to elements that are [included in the accessibility tree][] wit
 
 ## Expectation
 
-Each target element has an [accessible name][] that is not only [whitespace][].
+Each target element has an [accessible name][] that is not not empty (`""`).
```
carlosapaduarte:
```suggestion
Each target element has an [accessible name][] that is not empty (`""`).
```

_rules/button-accessible-name-97a4e1.md
```diff
@@ -143,15 +143,15 @@ Value attribute does NOT give an [accessible name][], only for input elements.
 
 #### Failed Example 2
 
-Span tag with role button with no name.
+Span tag with role button with an empty [accessible name][].
```
EmmaJP: change second 'with' to 'has'

_rules/button-accessible-name-97a4e1.md
```diff
@@ -170,7 +170,7 @@ Off screen element without an [accessible name][].
 
 #### Failed Example 4
 
-Button has an [accessible name][] that is only [whitespace][].
+Button has an [accessible name][] that is empty (`""`).
```
EmmaJP: Button has an empty accessible name.

For consistency with other examples.

PR #1033, opened 2019-11-21
Improve image-button-name examples

WilcoFiers:
- Better descriptions for the examples
- Ensure all images exist used in the examples
- Give proper accessible names for each image
- Remove passed example that does not satisfy the SCs

Need for Final Call: 1 week


_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,127 +55,103 @@ There is a known combination of a popular browser and assistive technology that
 
 #### Passed Example 1
 
-Image button element with [accessible name][] through `alt` attribute
+The image button has an with [accessible name][] through the `alt` attribute.
```
carlosapaduarte:
```suggestion
The image button has an [accessible name][] through the `alt` attribute.
```

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,127 +55,103 @@ There is a known combination of a popular browser and assistive technology that
...
 
 ```html
-<input type="image" name="submit" src="button.gif" alt="Submit" />
+<input type="image" src="/test-assets/shared/search-icon.png" alt="Search" />
 ```
 
 #### Passed Example 2
 
-Image button element with [accessible name][] through `aria-label`
+The image button has an with [accessible name][] through the `aria-label` attribute.
```
carlosapaduarte:
```suggestion
The image button has an [accessible name][] through the `aria-label` attribute.
```

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,127 +55,103 @@ There is a known combination of a popular browser and assistive technology that
...
 
 ```html
-<input type="image" name="submit" src="button.gif" aria-label="Submit" />
+<input type="image" src="/test-assets/shared/search-icon.png" aria-label="Search" />
 ```
 
 #### Passed Example 3
 
-Image button element with [accessible name][] through `title` attribute
+The image button has an with [accessible name][] through the `title` attribute.
```
carlosapaduarte:
```suggestion
The image button has an [accessible name][] through the `title` attribute.
```

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,127 +55,103 @@ There is a known combination of a popular browser and assistive technology that
...
 
 ```html
-<input type="image" name="submit" src="button.gif" title="Submit" />
+<input type="image" src="/test-assets/shared/search-icon.png" title="Search" />
 ```
 
 #### Passed Example 4
 
-Image button element with [accessible name][] through `aria-labelledby`
+The image button has an with [accessible name][] through the `aria-labelledby` attribute.
```
carlosapaduarte:
```suggestion
The image button has an [accessible name][] through the `aria-labelledby` attribute.
```

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,127 +55,103 @@ There is a known combination of a popular browser and assistive technology that
...
 
 ```html
-<img alt="W3C logo" />
+<img src="/test-assets/shared/w3c-logo.png" alt="W3C logo" />
 ```
 
 #### Inapplicable Example 5
 
-Image button is not included in the accessibility tree
+The image button is ignored by assistive technologies because it is not [included in the accessibility tree][]. These are not required to have an accessible name. If at some future state of the page the element gets [included in the accessibility tree][], an [accessible name][] will be necessary.
```
carlosapaduarte: Missing the [included in the accessibility tree] definition (twice)

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,128 +55,105 @@ There is a known combination of a popular browser and assistive technology that
...
+The `button` element is not an image button. [Success Criterion 1.1.1 Non-text Content](https://www.w3.org/TR/WCAG21/#non-text-content) can not fail text buttons. Only non-text content is applicable.
 
 ```html
 <button>My button</button>
 ```
 
 #### Inapplicable Example 2
 
-HTML `input` element with type with a `type` attribute in the `Button` state is not an image button
+The `input` element with type with a `type` attribute in the `Button` state is not an image button. [Success Criterion 1.1.1 Non-text Content](https://www.w3.org/TR/WCAG21/#non-text-content) can not fail text buttons. Only non-text content is applicable.
```
daniel-montalvo:
```suggestion
The `input` element with a `Button` value in the `type` attribute is not an image button. [Success Criterion 1.1.1 Non-text Content](https://www.w3.org/TR/WCAG21/#non-text-content) can not fail text buttons. Only non-text content is applicable.
```

Just suggesting, since I guess this is the message we want to communicate here. `Button` is a value for the `type` attribute. I am not sure what "state" tries to convey there.

Jym77: @daniel-montalvo: the `type` attribute (and a few other) can have any string as value. These values are converted to "states" which are only from a limited (finite) set.

Notably, both the "Button" and "button" **values** will be converted into the "Button" **state** (ie, case insensitivity).

So, it is indeed the **state** of the attribute we want to look at…
https://html.spec.whatwg.org/multipage/input.html#attr-input-type

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,128 +55,105 @@ There is a known combination of a popular browser and assistive technology that
...
 ### Failed
 
 #### Failed Example 1
 
-Image button element with empty [accessible name][].
+The image button element has an empty [accessible name][]. The `name` attribute can not be used to provide an [accessible name][].
 
 ```html
-<input type="image" name="submit" src="button.gif" />
+<input type="image" name="search" src="/test-assets/shared/search-icon.png" />
```
cblouch: While the input can't get its accessible name from the name attribute, fallback for input type="image" is the literal string "Submit Query", so this still has an accessible name and should pass.
See step 4 of:
https://w3c.github.io/html-aam/#input-type-image

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,128 +55,105 @@ There is a known combination of a popular browser and assistive technology that
...
 ```
 
 #### Failed Example 2
 
-Image button element with empty [accessible name][].
+The image button has an empty `alt` attribute, and no other attributes that can give it an [accessible name][].
 
 ```html
-<input type="image" name="submit" src="button.gif" alt="" />
+<input type="image" src="/test-assets/shared/search-icon.png" alt="" />
```
cblouch: As in fail example 1, the usual methods for finding an accessible name for this element fail so it should get an accessible name of "Submit Query"

_rules/image-button-accessible-name-59796f.md
```diff
@@ -55,128 +55,105 @@ There is a known combination of a popular browser and assistive technology that
...
-```
-
-#### Failed Example 4
-
-[accessible name][] is empty.
+The image button has an `aria-labelledby` attribute, but the referenced element does not exist. This gives the button an empty [accessible name][].
 
 ```html
-<input type="image" name="submit" src="button.gif" alt=" " />
+<input type="image" src="/test-assets/shared/search-icon.png" aria-labelledby="non-existing" />
```
cblouch: As in fail example 1, the usual methods for finding an accessible name for this element fail so it should get an accessible name of "Submit Query"

In general, because the API Mappings for input type=image have step #4, is there every a scenario where it will not have an accessible name?

_rules/image-button-accessible-name-59796f.md
cblouch: To make things a bit more complicated, Step 4 actually has:
"…user agent **may**…"
which makes it more squishy. In testing the above html I found:
Safari OSX VoiceOver – “Submit Button”
Chrome OSX VoiceOver – “Submit Button”
Chrome Win10 NVDA – “Button Submit”
Chrome Win10 JAWS – “Submit Button”
Edge Win10 NVDA – “Button Blank”
Edge Win10 JAWS – “Button Blank”
FF Win10 NVDA – “Button Submit Query”
FF Win10 JAWS – “Submit Query Button”
So it would appear only Edge is not taking the "may" approach. This is problematic. Whether an image input lack of an accessible name is somewhat inconsistent and flagging the markup appears to be a false positive in several browser/AT stacks.

WilcoFiers: Thank you for providing this test data. Would you mind opening a separate issue or pull request for this? The purpose of this PR is to improve the test case descriptions. I don't think your comment relates to this.

_rules/image-button-accessible-name-59796f.md
WilcoFiers: This will still fail, because of the empty alt the default accessible name isn't used.

_rules/image-button-accessible-name-59796f.md
WilcoFiers: You're right. This will use the fallback. This example would only fail if the label actually exists, but has no text in it.

_rules/image-button-accessible-name-59796f.md
WilcoFiers: I'm going to merge this. I'll open a separate issue.

PR #1034, opened 2019-11-21
Add Patrick H. Lauke to contributors

patrickhlauke:
See https://github.com/act-rules/act-rules.github.io/pull/1027#issuecomment-557059516


jkodu:
This does not require 3 approvals.

PR #1036, opened 2019-11-21
Editorial changes to html-page-lang-valid

WilcoFiers:
- Changed the order of bullets in the applicability
- Shortened the note to make it simpler
- Added understanding link to the background

Need for Final Call: nope


_rules/html-page-lang-valid-bf051a.md
```diff
@@ -22,11 +22,11 @@ acknowledgements:
 
 This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that:
 
+- has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 - is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
 - has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
```
daniel-montalvo:
```suggestion
- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
```

This is the last bulletted point. I think we should remove the "and" here and use the period.

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -22,11 +22,11 @@ acknowledgements:
 
 This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element that:
 
+- has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
```
Jym77:
```suggestion
- has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace); and
```
And this one is not the last bulletted point anymore 😄

_rules/html-page-lang-valid-bf051a.md
Jym77: Agreed.

PR #1037, opened 2019-11-21
chore: auto generate technique ids to be ignored for spelling tests

jkodu:
- This PR arbitrarily generates matching technique ids from https://www.w3.org/WAI/WCAG21/Techniques, that are then ignored by the spelling checker. 
- This removes the overhead to keep updating ignore words.

Closes issue(s):
- NA


__tests__/spelling.js
```diff
@@ -9,10 +9,21 @@ const reporter = require('vfile-reporter')
 
 const describeRule = require('../test-utils/describe-rule')
 
-const ignore = yaml.safeLoad(fs.readFileSync('./__tests__/spelling-ignore.yml', 'utf8'))
+const ignoreWords = yaml.safeLoad(fs.readFileSync('./__tests__/spelling-ignore.yml', 'utf8'))
+// https://www.w3.org/WAI/WCAG21/Techniques
+const ignoreTechniques = [`ARIA`, `C`, `F`, `G`, `H`].reduce((out, techniquePrefix) => {
+	let i = 1
+	while (i < 500) {
```
Jym77: Why not a classical for loop? 😄

__tests__/spelling.js
```diff
@@ -9,10 +9,21 @@ const reporter = require('vfile-reporter')
...
+		// Arbitarily chosen number
```
Jym77:
```suggestion
		// Arbitrarily chosen number
```
Arbitrarily indeed, but should be harmless so I see no need to fine tune that number now…

__tests__/spelling.js
```diff
@@ -9,10 +9,21 @@ const reporter = require('vfile-reporter')
 
 const describeRule = require('../test-utils/describe-rule')
 
-const ignore = yaml.safeLoad(fs.readFileSync('./__tests__/spelling-ignore.yml', 'utf8'))
+const ignoreWords = yaml.safeLoad(fs.readFileSync('./__tests__/spelling-ignore.yml', 'utf8'))
+// https://www.w3.org/WAI/WCAG21/Techniques
+const ignoreTechniques = [`ARIA`, `C`, `F`, `G`, `H`].reduce((out, techniquePrefix) => {
```
Jym77: Some prefixes are missing (FLASH, SCR, …) but I agree that they shouldn't concern us immediately and can be added if needed…

Issue #1021, opened 2019-11-14
page-title-descriptive, failed example 3 is a duplicate of example 1


WilcoFiers:

jkodu:
PR in place. Closing issue.

PR #1038, opened 2019-11-21
fix: remove identical failed example for rule `html-page-title-descriptive-c4a8a4`

jkodu:
See title

Closes issue(s):
- https://github.com/act-rules/act-rules.github.io/issues/1021

### DIFF
--- a/_rules/html-page-title-descriptive-c4a8a4.md
+++ b/_rules/html-page-title-descriptive-c4a8a4.md
@@ -143,23 +143,6 @@ Even though a correct `<title>` element is put in the `<head>` of the document, 
 </html>
 ```
 
-#### Failed Example 3
-
-`<title>` element with content present in document, but it is not descriptive of the content.
-
-```html
-<html>
-	<head>
-		<title>;)</title>
-	</head>
-	<body>
-		<p>
-			Clementines will be ready to harvest from late October through February.
-		</p>
-	</body>
-</html>
-```
-
 ### Inapplicable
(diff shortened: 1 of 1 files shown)


jkodu:
Final call Ends **December 3rd 2019**

mraccess77:
It seems like allowing two titles and title in body puts the rules at odds with WCAG 2.   This rule now seems like an HTML validation rule and not a accessibility related rule and then all test writers have to write a second rule to deal with 2 titles and title in body which are violations of WCAG.

patrickhlauke:
xref https://github.com/w3c/wcag/issues/980

The rule only cares about whether there's at least one `<title>` at all. It does not express any sentiment about situations where there is more than one `<title>`, nor whether it is in the `<head>` or elsewhere, as UAs also don't care and seem to - per spec - just use the first one they encounter. As such, I don't think it's at odd with WCAG, nor is it a validation rule...it just has examples that show how even in odd cases that are valid HTML 5.2, the rule needs to look at only the first one as that's what UAs do.

WilcoFiers:
@mraccess77 Two titles is not a violation of WCAG according to the people who have reviewed this rule, at least, not when the first title describes the page. We can definitely revisit that conversation if you'd like. Can you open a separate issue for it though, since this is not related to this pull request?

mraccess77:
If the group agrees this approach is acceptable then I'm fine with it -- I have no need to open a new issue.

PR #1040, opened 2019-11-25
test: ensure assets referred under `test-assets` directory exists

jkodu:
- Added test to catch missing assets referred under the `test-assets` directory.
- Fix tests by adding missing test assets or wrongly referred assets

Closes issue(s):
- https://github.com/act-rules/act-rules-web/issues/9


_rules/__tests__/test-assets.js
```diff
@@ -0,0 +1,101 @@
...
+	const codeBlocks = gfmCodeBlocks(ruleData.body)
+	const usedAssetPaths = getUniqueTestAssetsUsedInCodeBlocks(codeBlocks)
+
+	if (usedAssetPaths && usedAssetPaths.length) {
+		test.each(usedAssetPaths)('%s', assetPath => {
+			const message = `Asset at path -> ${assetPath} does not exist`
+			expect(allTestAssetPaths.includes(assetPath), message).toBe(true)
+		})
+	}
+})
```
Jym77: I think we should also do the reverse test (check that files in the `test-assets` directory are used somewhere) in order to clean up old test assets that are not used anymore (if any).

Given that you already have `allTestAssetPaths` and `usedAssedPaths`, that should be fairly easy.

jkodu:
Given this is a `test`, there is no need for multiple approvals on the PR.

Issue #879, opened 2019-09-09
Citing HTML spec: v5.2 or WHATWG?


Jym77:
There are essentially 2 versions of the HTML specs: The "stable" HTML5.2 published by W3C at https://www.w3.org/TR/html52/ and the "living standard" published by WHATWG at https://html.spec.whatwg.org/ (and redirected to by W3C from https://www.w3.org/TR/html/).

Previously, the living standard was W3C editor's draft published only at https://www.w3.org/TR/html/ but the task of writing it has been delegated to WHATWG on May 28, 2019 by this MOU: https://www.w3.org/2019/04/WHATWG-W3C-MOU.html

Many of our links were to https://www.w3.org/TR/html/ and have been updated to https://html.spec.whatwg.org/ some links are nonetheless to the stable v5.2.

---
We should decide which version to cite and keep consistency between all rules (consistency is relatively easy to enforce by tests banishing the wrong URL).

* The main drawback of citing WHATWG is that it is a living standard. Thus, it can theoretically change any day and either our link could get broken or our rules could get broken before they refer to a definition that has change and changes the meaning of the rules. Broken links are a mild annoyance, but having our rules broken by external changes is pretty bad.
* The main drawback of citing HTML5.2 is that it may drift quite a bit from the living standard and its implementations in browsers. Thus, our rules could link to a _de facto_ outdated definition and not reflect reality. We have a lot of discussion about accessible names and such about how such browser is computing it (no matter what the spec we cite says), it would be pretty bad to start adding the same kind of discussion about HTML (if browsers implement the living standard rather than v5.2).

---
Anyway, I guess this needs some discussion and decision and unification…

Jym77:
See https://github.com/act-rules/act-rules.github.io/pull/768#discussion_r322190389 (and previous comments) for a bit of discussion on that topic.

Jym77:
Note: links to DOM specs are also concerned.

kasperisager:
To be frank, I don't see _any_ reason why we should continue to reference the now superseded W3C HTML and DOM specifications. All future W3C specifications will also reference the WHATWG HTML and DOM specifications per the MoU:

> For these other specifications published at WHATWG that are not parts of the WHATWG HTML and DOM specifications, the W3C will: [...] 3. prefer that normative references from W3C specifications reference the WHATWG versions.

https://www.w3.org/2019/04/WHATWG-W3C-MOU.html#transition

jkodu:
[...]

PR #1043, opened 2019-11-26
add link to HTML 5.2 to the background section

WilcoFiers:
As requested by AG.

Need for Final Call: nope


_rules/html-page-title-2779a5.md
```diff
@@ -56,6 +56,7 @@ _There are no major accessibility support issues known for this rule._
 - [Understanding Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled)
 - [G88: Providing descriptive titles for Web pages](https://www.w3.org/WAI/WCAG21/Techniques/general/G88)
 - [H25: Providing a title using the title element](https://www.w3.org/WAI/WCAG21/Techniques/html/H25)
+- [HTML 5.2: the `title` element](https://www.w3.org/TR/html52/document-metadata.html#the-title-element)
```
Jym77: Other concepts in this rule are linked to WHATWG spec (descendant, document). Should this be WHATWG instead?
(we haven't really solved #879).

WilcoFiers: I use the same link to the title element that is in several other places in this document. Yes we should shift over to WHATWG spec, but that's not what this PR is for.

Jym77: OK.

PR #1067, opened 2019-12-05
refactor: helper fn's under `test-utils` used by `jest`

jkodu:
This PR is more of a clean-up.

Closes issue(s):

- NA


pages/glossary/__tests__/glossary-exists.js
```diff
@@ -1,27 +0,0 @@
-const globby = require('globby')
```
jkodu: > Note: There is an existing issue asking to author tests to ensure validity of referenced glossary items. Removing this redundant file until we get to that issue.

PR #1073, opened 2019-12-10
Examples descriptions lang rules (5b7ae0 / bf051a / b5c3f8)

### LINKED ISSUE #834: Unifying example descriptions
Examples description varies wildly from rule to rule, and also inside a rule…

* Ending the sentence is sometimes done with a full period (`.`), sometimes with a colon (`:`) and sometimes with nothing at all. We should always do it the same way.
(I'm roughly counting 438 periods, 19 colons and 103 nothing, numbers are not 100% accurate, seems to indicate that period is the preferred way)
* Use of "this" + declarative form (notably in failed examples).
"this element has no foo" vs "element without foo"
* fully qualifying entities
"the attribute `foo` is blah blah" vs "`foo` is blah blah"
* naming elements/attributes
"this `a` element" vs "this element"
"a set of two links that blah blah" vs "links blah blah"
* others… I'm sure there are more…
* descriptions have to be precise: e.g "`title` element" instead of "`title`"
* front-load the description with the applicable "thing". "This X element has a Y" instead of "Y is available on element X"

I guess it's a quite huge task but we should at least:
* agree on some general rules on how to write these descriptions
* probably document these guidelines in a "how to contribute" page (same as we should document the guidelines for rule names of #250) 
* Keep that in mind when writing/reviewing/updating rules so that the overall quality increases.

I think that having inconsistent wording for descriptions, especially when it's often barely even a real English sentence, is really bad for the overall quality of the rules.
(often, the "description" of examples look more like a short bullet list of all cases that need to be written, intended for a co-writer rather than for further readers…)

## preferred format

This [test target] [meets / does not meet condition] [because optional reasons].

Examples of [test target]:
- `button` element
- `div` element with [semantic role](#semantic-role) of `button`
- `aria-required` property
- `id` attribute
- page

Examples of [meets / does not meet condition]:
- has an accessible name
- does not have an accessible name
- contains a `title` element with content
- have the same accessible name and embed the same resource

Example of [because optional reasons]:
- given by the `alt` attribute
- because it is not [visible](#visible)
- even though the referenced element is not yet in the DOM

Jym77:
Improving examples descriptions in the lang rules, according to #834. 

Plus adding some missing cases for the lang/xml:lang rule.

Handling of XTML examples is done in https://github.com/act-rules/act-rules-web/pull/86

Closes issue(s):
Partially closes #834.

Need for Final Call:
This will require a 1 week Final Call (adding a couple of examples).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/html-page-lang-b5c3f8.md
+++ b/_rules/html-page-lang-b5c3f8.md
@@ -29,7 +29,7 @@ This rule applies to any [document element](https://dom.spec.whatwg.org/#documen
 
 ## Expectation
 
-Each test target has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
+Each test target has a `lang` attribute that is neither empty (`""`) nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 
 ## Assumptions
 
@@ -51,7 +51,7 @@ _There are no major accessibility support issues known for this rule._
 
 #### Passed Example 1
 
-The `lang` attribute specified has a non-empty value.
+This `html` element has a `lang` attribute with a non-empty (`""`) value.
 
 ```html
 <html lang="en"></html>
@@ -61,7 +61,7 @@ The `lang` attribute specified has a non-empty value.
 
 #### Failed Example 1
 
-There is no `lang` attribute specified.
+This `html` element does not have a `lang` attribute.
 
 ```html
 <html></html>
@@ -69,7 +69,7 @@ There is no `lang` attribute specified.
 
 #### Failed Example 2
 
-The `lang` attribute specified is empty ("").
+This `html` element has a `lang` attribute with an empty (`""`) value.
 
 ```html
 <html lang=""></html>
@@ -77,7 +77,7 @@ The `lang` attribute specified is empty ("").
 
 #### Failed Example 3
 
-The `lang` attribute consists of only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
+This `html` element has a `lang` attribute whose value is only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
 
 ```html
 <html lang=" "></html>
@@ -85,7 +85,7 @@ The `lang` attribute consists of only [ASCII whitespace](https://infra.spec.what
 
 #### Failed Example 4
 
-There is no `lang` attribute specified, only a `xml:lang` attribute.
+This `html` element has no `lang` attribute, only a `xml:lang` attribute.
 
 ```html
 <html xml:lang="en"></html>
--- a/_rules/html-page-lang-valid-bf051a.md
+++ b/_rules/html-page-lang-valid-bf051a.md
@@ -30,7 +30,7 @@ This rule applies to any [document element](https://dom.spec.whatwg.org/#documen
 
 ## Expectation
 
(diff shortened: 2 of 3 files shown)


Jym77:
> Under Assumptions, have the first instance of 'WCAG' say 'Web Content Accessibility Guidelines (WCAG)'

I agree, and it should also be a link.
That is however unrelated with this PR and I'd rather keep it for another.

_rules/html-page-lang-b5c3f8.md
```diff
@@ -61,31 +61,31 @@ The `lang` attribute specified has a non-empty value.
 
 #### Failed Example 1
 
-There is no `lang` attribute specified.
+This `html` element does not have a `lang` attribute.
```
WilcoFiers: Instead of saying what it doesn't have, should we say what it should have? For example:

```suggestion
This `html` element should have a `lang` attribute.
```

_rules/html-page-lang-b5c3f8.md
```diff
@@ -61,31 +61,31 @@ The `lang` attribute specified has a non-empty value.
...
 
 ```html
 <html></html>
 ```
 
 #### Failed Example 2
 
-The `lang` attribute specified is empty ("").
+This `html` element has a `lang` attribute with an empty (`""`) value.
```
WilcoFiers:
```suggestion
This `html` element should not have a `lang` attribute with an empty (`""`) value.
```

_rules/html-page-lang-b5c3f8.md
```diff
@@ -61,31 +61,31 @@ The `lang` attribute specified has a non-empty value.
...
 
 ```html
 <html lang=""></html>
 ```
 
 #### Failed Example 3
 
-The `lang` attribute consists of only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
+This `html` element has a `lang` attribute whose value is only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
```
WilcoFiers:
```suggestion
This `html` element should not have a `lang` attribute whose value is only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
```

_rules/html-page-lang-b5c3f8.md
```diff
@@ -61,31 +61,31 @@ The `lang` attribute specified has a non-empty value.
...
 
 ```html
 <html lang=" "></html>
 ```
 
 #### Failed Example 4
 
-There is no `lang` attribute specified, only a `xml:lang` attribute.
+This `html` element has no `lang` attribute, only a `xml:lang` attribute.
```
WilcoFiers:
```suggestion
This `html` element should have a `lang` attribute that matches the `xml:lang` attribute.
```

_rules/html-page-lang-b5c3f8.md
Jym77: That wouldn't match the "[meets / does not meets condition]" pattern that we suggest in #834, I think.

_rules/html-page-lang-b5c3f8.md
EmmaJP: I think this is a good example of why "should" is maybe not as clear. Is this failing because the attribute is empty or because it should have an empty attribute.

It also could lead to double negative confusion, and Jey indicated it could be an issue for unit test stuff.

Jym77:
Closing due to the added examples in one of the rule.

PR #1081, opened 2019-12-13
Examples descriptions lang rules (5b7ae0 / bf051a / b5c3f8)

Jym77:
Rewrite examples descriptions in lang rules, according to #834.

Closes issue(s):
Partially closing #834.

Need for Final Call:
This will not require a Final Call (editorial changes only).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/html-page-lang-valid-bf051a.md
```diff
@@ -53,7 +53,7 @@ _There are no major accessibility support issues known for this rule._
 
 #### Passed Example 1
 
-The `lang` attribute specified is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace) and has a valid primary language subtag.
+This `html` element has a `lang` attribute whose value is a [valid][valid language subtag] [primary language subtag][].
```
jkodu: Missed a `[]` after valid language subtag?

Jym77: No. The ref is `[valid][valid language subtag]` which is correct.
And the (new) test agrees with me 😄

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -63,31 +63,31 @@ The `lang` attribute specified is neither empty ("") nor only [ASCII whitespace]
...
+This `html` element has a `lang` attribute whose value is not a [valid language subtag][].
 
 ```html
 <html lang="xyz"></html>
 ```
 
 #### Failed Example 2
 
-The `lang` attribute value has a valid primary language subtag, but a syntactically invalid region subtag.
+This `html` element has a `lang` attribute whose value is not a [valid language subtag][]. Even though the [primary language subtag][] is valid, the [region subtag] is not
```
EmmaJP: Missing a full stop at the end of the second sentence.

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -63,31 +63,31 @@ The `lang` attribute specified is neither empty ("") nor only [ASCII whitespace]
...
 
 ```html
 <html lang="en-US-GB"></html>
 ```
 
 #### Failed Example 3
```
EmmaJP: This failed example appears to be almost identical to failed example 1

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -63,31 +63,31 @@ The `lang` attribute specified is neither empty ("") nor only [ASCII whitespace]
...
 
-The `lang` attribute value is not a valid primary language subtag.
+This `html` element has a `lang` attribute whose value is not a [valid language subtag][].
 
 ```html
 <html lang="123"></html>
 ```
 
 #### Failed Example 4
```
EmmaJP: Does this really need 3 examples that are not valid subtags?

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -57,23 +57,23 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
...
+This `html` element has identical [primary language subtags][] for its `lang` and `xml:lang` attributes.
 
 ```html
 <html lang="en" xml:lang="en"></html>
 ```
 
 #### Passed Example 2
 
-`html` element with matching primary and extended language subtags for `lang` and `xml:lang`.
+This `html` element has identical [primary language subtags][] for its `lang` and `xml:lang` attributes. The [extended language subtags][] also match.
```
EmmaJP: Is there a reason for not keeping this as a single sentence?

This `html` element has identical [primary language subtags][] and [extended language subtags][] for its `lang` and `xml:lang` attributes.

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -83,7 +83,7 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
 
 #### Failed Example 1
 
-`html` element with non-matching primary language subtags for `lang` and `xml:lang`.
+This `html` element has different [primary language subtags][] for its `lang` and `xml:lang` attributes.
 
 ```html
 <html lang="fr" xml:lang="en"></html>
```
EmmaJP: Should there be a second failing example where the primary subtags do not match but the extended subtags do not? 

Appreciate the PR is about the language being used in the rule, rather than the rule itself.

_rules/html-page-lang-valid-bf051a.md
Jym77: Pro-tip: For small changes like this, you can use the "suggestion" button on github to directly make the suggested change as part of your review. That's usually faster both for the reviewer (adding the full stop is faster than writing that it is needed) and for the rule writer (github offers the possibility to accept the change in one click rather than dig through the file manually).

_rules/html-page-lang-valid-bf051a.md
Jym77: I disagree.
Failed Example 1 is about garbage lang attribute.
Failed Example 2 is about "the primary tag is not enough for this rule" (some other rules only check the primary tag).

_rules/html-page-lang-valid-bf051a.md
Jym77: I agree that Failed Example 1 and 3 are the same.
Subject for some other PR…

_rules/html-page-lang-xml-lang-match-5b7ae0.md
Jym77: I felt that it was closer to the "[test target] [matches expectation]" format. Here, the expectation is about the primary tag. The second sentence has nothing to with the expectation and is additional information and remark. That's why I think it is better suited as a separate sentence, keeping the first one strictly on the expectation.

_rules/html-page-lang-xml-lang-match-5b7ae0.md
Jym77: Could be a good idea. Not sure. Totally subject for another PR in any cases 😄

Jym77:
> Would be good to add the missing full stop. Other comments may or may not require changes, possibly in a different PR.

I put these other changes in #1094 and assigned @EmmaJP on it…

_rules/html-page-lang-valid-bf051a.md
EmmaJP: And failed examples 3 and 4? The description text is identical on both old and reworded versions.

I was comparing failed examples 1 with 3 and 4, but don't think my comment was clear about that. Apologies. Is it really necessary to have 3 versions on an invalid language tag?

_rules/html-page-lang-xml-lang-match-5b7ae0.md
EmmaJP: So is the subtags matching is incidental to the success criteria and rule requirement? Wondering if "Incidentally" should be added to the second sentence, or if Passed Example 3 alone would be sufficient if the second sentence read "The extended language subtags are not required to match."

EmmaJP: Probably also mostly conversation for a different PR.

_rules/html-page-lang-valid-bf051a.md
EmmaJP: Another subject for a different PR!

_rules/html-page-lang-valid-bf051a.md
EmmaJP: Thanks ... I wasn't aware of that... and also can't see how to do that. But I'll look out for it.

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -63,31 +63,31 @@ The `lang` attribute specified is neither empty ("") nor only [ASCII whitespace]
...
 
-The `lang` attribute value is not a valid primary language subtag.
+This `html` element has a `lang` attribute whose value is not a [valid language subtag][]
```
EmmaJP:
```suggestion
This `html` element has a `lang` attribute whose value is not a [valid language subtag][].
```

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -93,15 +93,15 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
 
 #### Inapplicable Example 1
 
-`svg` element is not applicable for this rule.
+This rule does not apply to `svg` elements
```
EmmaJP:
```suggestion
This rule does not apply to `svg` elements.
```

_rules/html-page-lang-valid-bf051a.md
EmmaJP: Figured it out ... thanks, that's really helpful to know.

_rules/html-page-lang-valid-bf051a.md
Jym77: I agree that 1, 3 and 4 are very similar. Maybe not all are needed.
You've won #1097 😁

_rules/html-page-lang-xml-lang-match-5b7ae0.md
Jym77: It is incidental for the rule. It is less clear for the SC…
There has been some discussion around it (can't find the issue/PR) between @kasperisager and @dd8 if I remember correctly.
Essentially, I remember that secondary tags don't really cause accessibility issues. E.g., if a page is written in British but with a `lang="en-US"`, that's not going to be a problem. So, similarly for matching `lang`/`xml:lang`, if one of them says "en-US" and the other "en-UK", that's not going to be a real problem.

One notable case where this would cause accessibility issue is between Cantonese (`zh-yue`) and Mandarin (`zh-cmn`) which are written the same way but pronounced very differently (to the point that speakers don't understand each other). So the wrong subtag would cause screen readers to output incomprehensible speech.

Given that this was the only case we could find where the subtag really matters for accessibility, the rule only focus on primary tags and there is a note in Expectation mentioning this.

It is not clear whether WCAG require perfect identification of subtags also…

I will nonetheless update the description to make this a bit more obvious.

Issue #984, opened 2019-10-23
Test foot note links for correctness


WilcoFiers:
Things like `[semantic role][]`. We need to add tests to catch if these definitions are missing.

Jym77:
Yes, I thought I had one when I added support for the footnote links, but somehow I guess I messed up my regex or something 😞 

Hmm, looking at it, it seems I only gathered the refs and checked that the corresponding definition exists, and did not check that the internal link references an existing ref 😞 (_i.e._ I checked that if `[foo]: #foo` is used, then `foo.md` exists; but didn't check that if `[foo][]` or `[bar][foo]` exists, then `[foo]: something` also exists in the same file).

This should however be super standard markdown linting, so I kinda assume that tools for doing that already exist and we don't need to reinvent them.

Relevant code is here:
* https://github.com/act-rules/act-rules-web/blob/master/build/create-glossary-usages.js#L93-L121
* https://github.com/act-rules/act-rules-web/blob/master/utils/reg-exps.js

jkodu:
PR in place - https://github.com/act-rules/act-rules.github.io/pull/1083

Issue #829, opened 2019-08-22
Check for broken glossary-to-glossary links


Jym77:
We check if glossary links used in rules actually refer to existing glossary items. This is however not the case for glossary links used in glossary items, as shown in #783.

We should add a test for this.

jkodu:
PR in place - https://github.com/act-rules/act-rules.github.io/pull/857

jkodu:
See comment - https://github.com/act-rules/act-rules.github.io/pull/857#issuecomment-529893002

Issue #828, opened 2019-08-22
Check for missing links to glossary


Jym77:
Add test to check if some term is a glossary item without a link.
(*e.g.* if some rule uses `accessible name` instead of `[accessible name](#accessible-name)` or `[accessible name][]`)

There will likely be some false negative going through the test anyway…

We might also want to check validity of internal "footnote references", *i.e.* that if either `[accessible name][]` or `[accessible names][accessible name]` appears, then `[accessible name]: foo` must also exist.

PR #1083, opened 2019-12-16
test: identifiers and urls used as `link`, `definition`, `linkReference` in markdown exists

jkodu:
This change uses `markdownAST` to traverse markdown for nodes [`link`](https://github.com/syntax-tree/mdast#link), [`definition`](https://github.com/syntax-tree/mdast#definition) and [`linkReference`](https://github.com/syntax-tree/mdast#linkreference) in every rule and pages markdown file and verifies the integrity of each of the references.

**The main tests (jest) are:**
- https://github.com/act-rules/act-rules.github.io/commit/d9e83c8e52ff2c85e78f23e93a37df081b0d1ccc#diff-6e64e206c7d5f108603aa729409a5530
- https://github.com/act-rules/act-rules.github.io/commit/d9e83c8e52ff2c85e78f23e93a37df081b0d1ccc#diff-b825a2ad6ce3c8d730204d7201cc743a
- https://github.com/act-rules/act-rules.github.io/commit/d9e83c8e52ff2c85e78f23e93a37df081b0d1ccc#diff-23bf7fcfd9e2619cbed39361b30d2fce

**Other changes are:**
- updates to pre-existing functions and other helper/ utility functions.
- fixes to links in markdown files (https://github.com/act-rules/act-rules.github.io/commit/65c591479cf12f87f14f7b8db5de5e5a2d65e4ee)

**Closes issue(s):**
- https://github.com/act-rules/act-rules.github.io/issues/984
- https://github.com/act-rules/act-rules.github.io/issues/829
- https://github.com/act-rules/act-rules.github.io/issues/828


jkodu:
@Jym77 for the failing test - can you help me find the right reference for [Text Alternative Computation](#text-alternative) as `text-alternative` definition does not exist.

__tests__/link-to-glossary-term-valid.js
```diff
@@ -0,0 +1,63 @@
...
+
+function validateGlossaryReferences({ markdownAST }, { glossaryKeys = [] }) {
+	/**
+	 * get all links
+	 * -> eg: [Alpha](https://....) or [Beta](#semantic-role)
+	 */
+	const pageLinks = getMarkdownAstNodesOfType(markdownAST, 'link').map(({ url }) => url)
+	/**
+	 * get all definition links
+	 * -> eg: [Alpha]: https:// 'Link to something' or [Beta]: #some-glossary 'Def to some glossary'
```
Jym77:
```suggestion
	 * -> eg: [alpha]: https:// 'Link to something' or [beta]: #some-glossary 'Def to some glossary'
```

_rules/form-control-label-descriptive-cc0f0a.md
```diff
@@ -181,4 +181,5 @@ The element with `aria-labelledby` is not a form field.
 ```
 
 [semantic role]: #semantic-role 'Definition of semantic role'
+[semantic roles]: #semantic-role 'Definition of semantic role'
```
Jym77: I'd change the link reference to `[semantic roles][semantic role] rather than adding a new link definition.

__tests__/link-is-outdated.js
```diff
@@ -0,0 +1,58 @@
...
+	 * -> eg: [Alpha]: https:// 'Link to something'
+	 */
+	const definitionLinks = getMarkdownAstNodesOfType(markdownAST, 'definition').map(({ url }) => url)
+
+	/**
+	 * get all links that is a valid
+	 * -> this test does not cover glossary/ definition referencing links (see test - 'link-to-glossary-term-valid.js')
+	 */
+	const links = uniqueArray([...pageLinks, ...definitionLinks].filter(isUrl))
+	if (!links || !links.length) {
```
Jym77: `uniqueArray` is always returning an Array. so you may want to simplify these tests to
```suggestion
	if (links.length === 0) {
```

__tests__/link-is-outdated.js
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+
+	test.each(links)('%s', link => {
+		const badLink = Object.keys(badLinksAndRecommendations).find(badLink => link.includes(badLink))
+		expect(!!badLink, badLinksAndRecommendations[badLink]).toBe(false)
+	})
+}
```
Jym77: And I'm not sure exactly how `test.each` behaves on an empty array. It should pass (that's the expected behaviour of universal quantifier on empty set).
In that case, the full `if` statement of lines 50-52 can be removed.

(same thing on other tests).

Jym77:
> @Jym77 for the failing test - can you help me find the right reference for `[Text Alternative Computation](#text-alternative)` as `text-alternative` definition does not exist.

"text content" seems to not be used: 
https://act-rules.github.io/glossary/#text-content

So, just drop that def…

__tests__/link-is-outdated.js
jkodu: Jest expects each describe to have a `test` so it fails. We can bypass this behaviour by setting `passwithnotests` - https://jestjs.io/docs/en/cli#--passwithnotests, but I think that is bad.

Issue #1003, opened 2019-10-29
Objective / Unambiguous metadata in definitions?


WilcoFiers:
I've seen a few places where a subjective or sometimes even ambiguous definition is used in some of our rules. I'm wondering if, to ensure that reviewers look at this, and to prevent us from using definitions we shouldn't be using, we should add the some frontmatter info to all the definitions.

Example:

```
---
title: Focusable
key: focusable
objective: true
unambiguous: true
---

Elements that can become the target of keyboard input as described in the [HTML](https://www.w3.org/TR/html) specification of [focusable](https://html.spec.whatwg.org/#focusable-area) and [can be focused](https://html.spec.whatwg.org/#specially-focusable).
```

If nothing else, I'm hoping this can maybe remind reviewers to check that these things are correct.

Jym77:
I like that. It should even be possible (although somewhat tricky) to automatically figure out where the definition is used (applicability, expectation, example description, note, assumption, …) and test if this is allowed with the current "ambiguity level".

Could also be a single "value" instead of several booleans. It looks like there is a gradation "ambiguous < subjective < objective" as far as I understand.

WilcoFiers:
Objective and ambiguous are two distinct things. The distinction is important in rules, because in ACT, applicability has to be objective, whereas the expectation can be subjective. For example "descriptive" is a subjective term. In the expectation you can say that the alt text has to be descriptive of the image, but you can't write an applicability that says all images with a descriptive alt text.

Ambiguous also matters, sometimes we have ambiguous definitions in assumptions, to reference concepts from WCAG, things like "essential".

Jym77:
I agree that objective/subjective and (un)ambiguous are different things.

What I fail to see is how it's possible to have something which is both objective and ambiguous.

---
As far as I understand,
* "objective" means "all users will always interpret that the same way".
* "subjective" means "two different users may interpret that in different way".
* "ambiguous" means "a given user may interpret that in different ways depending on other stuff" (possibly, user's mood).

Maybe I'm wrong with these understandings…

Thus, I fail to find something that could be objective (you and me will always interpret it the same way) but ambiguous (I might interpret it in different ways, you might interpret is in different ways). Which is why I feel there is a gradation rather than two orthogonal dimensions.

This is a completely minor point and I'm fine with two booleans.

Jym77:
ACT rules format states under [Applicability](https://www.w3.org/TR/act-rules-format/#applicability):
[...]

PR #1085, opened 2019-12-17
Metadata in glossary

Jym77:
Add "unambiguous" and "objective" metadata in each definition.
Co-authored by @jkodu 

Closes issue(s):
- #1003 

Need for Final Call:
This will not require a Final Call (internal metadata with no external visibility).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/accessible-name.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Accessible Name
 key: accessible-name
+unambiguous: true
```
Jym77: Technically, "accessible name" **is** ambiguous because the algorithm doesn't explain which kind of whitespace should be trimmed.
However, we seem to always have considered it as unambiguous… So we kept it that way…
(it is used in the Applicability or Expectation of a dozen rules so far…)

pages/glossary/decorative.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Decorative
 key: decorative
+unambiguous: false
```
Jym77: This is used in the examples description of 2 rules (which is OK).

This is used in the Expectation of [Image has accessible name](https://act-rules.github.io/rules/23a2a8#expectation), but as part of the expression "marked as decorative" which is unambiguously defined there. So this is OK.

We might want to create a full definition for "marked as decorative", but this would be the topic of another PR.

jkodu:
@WilcoFiers @Jym77 

Are we supposed to be using these attributes to be shown somewhere on the website?

Jym77:
> @WilcoFiers @Jym77
> 
> Are we supposed to be using these attributes to be shown somewhere on the website?

As far as I understand, "not now".

Currently, we just add them and hope that rule authors/reviewers will think about them.
Later, we may want to add some tests.

pages/glossary/decorative.md
WilcoFiers: Can you explain why you think this is ambiguous? Keep in mind, ambiguous in our context means it can mean multiple things. I don't think that's true here. Sure something like "aesthetic" is debatable, but that word doesn't have two different meanings.

pages/glossary/section-of-content.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Section of content
 key: section-of-content
+unambiguous: true
```
WilcoFiers: I think this is ambiguous. "A distinct part or subdivision of a document" is pretty vague and can mean a lot of different things. For example, visually distinct, an element could be a "distinct part", things in a landmark, etc.

pages/glossary/accessible-name.md
EmmaJP: Presume the same ASCII Whitespace as defined elsewhere

pages/glossary/semantic-role.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Semantic Role
 key: semantic-role
+unambiguous: true
+objective: true
 ---
 
 A semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role) is a semantic association that indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
```
EmmaJP: Wondering if, to be unambiguous, this definition within ACT Rules needs to be clear that a role should be one of those defined by the WAI-ARIA spec role taxonomy (https://www.w3.org/TR/wai-aria-1.1/#role_definitions). Even for custom elements.

pages/glossary/decorative.md
Jym77: On that topic, what do you think of #1095?

pages/glossary/section-of-content.md
Jym77: Hmm… You might be right 😢 

The definition is currently used in the expectation of https://act-rules.github.io/rules/b49b2e so this need to be fixed somehow (and I'm making a heavy use of it in the Bypass block PR also…)

The fact that a `p` element might, or maybe not, be a section of content in itself doesn't really bother me. I see that as subjectivity, not ambiguity (i.e., yes, a `p` can be a section of content and its up to each of us to decide whether this specific `p` in this specific page is).

pages/glossary/semantic-role.md
Jym77: I don't think this is needed to make the def unambiguous.

Additionally, roles can be defined somewhere else (e.g., the Graphics AAM defines new roles: https://w3c.github.io/graphics-aam/#mapping_role_table).

Lastly, linking to the WAI-ARIA roles kinda ties down to the HTML technology. A semantic role could be defined for pdf, for LaTeX, or any other kind of document and these technologies could choose a different taxonomy of roles without this being intrinsically bad.

We could use add a note to point to the list, it is indeed a valuable addition to the definition.
Can you do that? 😁

pages/glossary/accessible-name.md
Jym77: The problem is that the algorithm uses "whitespace" without more precision (and different browsers interpret that differently).
And we are not in control of the spec of that algorithm…
(the issue has been raised: https://github.com/w3c/accname/issues/55)

pages/glossary/semantic-role.md
EmmaJP: I'll make it separate to this PR.

EmmaJP: https://github.com/act-rules/act-rules.github.io/issues/1129

PR #1090, opened 2020-01-02
Remove similar passed example for rule CSS Orientation Lock (b33eff)

jkodu:
This example is very similar to passed example 1, and hence removing the same.

Closes issue(s):

- NA


jkodu:
Started 1 week final call.

jkodu:
End of final call. Merging PR.

Issue #1092, opened 2020-01-09
iframe element has accessible name: add examples using role=presentation


maryjom:
From the ACT TF Survey on the 'iframe element has accessible name' rule Charu brought up the case where role=presentation could produce different test results from different test tools. Role=presentation but not removing the iframe from the tab order should result in a fail condition but role=presentation AND removing the iframe from the tab order should not require an accessible name as it would be ignored both by the AT and would not be navigable.  The rule should be updated to account for these situations. Here are the [survey results](https://www.w3.org/2002/09/wbs/93339/ValidPageLang/results) if you need to look at them and the [ACT TF meeting minutes](https://www.w3.org/2020/01/09-wcag-act-minutes.html#item05) where we discussed this comment.

WilcoFiers:
There's a PR open to resolve this.

PR #1102, opened 2020-01-15
Consider tabindex in iframe has acc name

WilcoFiers:
Consider whether or not iframes are in the sequential focus navigation when dealing with accessible names. This resolves a few comments from the ACT Task Force, including #1092.

Need for Final Call: 1 week


_rules/iframe-accessible-name-cae760.md
```diff
@@ -30,18 +31,18 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Assumptions
 
-- The rule assumes that the target `iframe` is used as a [user interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components).
+If an `iframe` is not perceived by the user as a single control, it does not qualify as a [user interface component][] under WCAG 2. In such a scenario, failing this rule would not fail [success criterion 4.1.2](https://www.w3.org/TR/WCAG21/#name-role-value). Unless the `iframe` is both removed from the accessibility tree and removed from [sequential focus navigation][], they usually are considered to be [user interface components][user interface component].
```
Jym77: I had the impression that we usually start the assumptions by "this rule assumes that".
Not a big deal, but uniformity helps.

_rules/iframe-accessible-name-cae760.md
```diff
@@ -30,18 +31,18 @@ Each target element has an [accessible name][] that is not empty (`""`).
...
 
 ## Accessibility Support
 
-_There are no major accessibility support issues known for this rule._
+Some browsers include `iframe` elements in the [sequential focus navigation][]. This ensures that `iframe` elements can always be scrolled using the keyboard. When an `iframe` is removed from the accessibility tree, this rule is still applicable for those browsers, unless the `tabindex` attribute is set to `-1`.
```
```suggestion
Some browsers include `iframe` elements in the [sequential focus navigation][]. This ensures that `iframe` elements can always be scrolled using the keyboard. When an `iframe` is removed from the accessibility tree, this rule is still applicable for those browsers, unless the `iframe` is explicitly removed from [sequential focus navigation][] (by having the `tabindex` attribute set to a negative value).
```
Because `tabindex=-2` also works…

_rules/iframe-accessible-name-cae760.md
```diff
@@ -132,10 +133,10 @@ Usage of `alt` attribute to describe content is not valid.
 
 #### Failed Example 7
 
-[Accessible name][] is empty.
+This `iframe` that is part of [sequential focus navigation][] has an empty [Accessible name][].
```
Jym77: We need a similar Passed Example (not in the accessibility tree, in sequential focus order, has an accessible name).

_rules/iframe-accessible-name-cae760.md
```diff
@@ -150,12 +151,22 @@ Does not apply to non `iframe` element.
 
 #### Inapplicable Example 2
 
-`iframe` is not [included in the accessibility tree][].
+This `iframe` is not part of [sequential focus navigation][], nor [included in the accessibility tree][] because of `display: none;`.
```
```suggestion
This `iframe` is neither part of [sequential focus navigation][], nor [included in the accessibility tree][] because of `display: none;`.
```

Jym77: As far as I understand, browsers that "force push" iframes in sequential navigation won't do it here because of the `display:none`, which is fine.
The description, however, looks a bit like the "because of `display:none` is related only to the second part (not in the accessibility tree) rather than to both.

But I don't really have any idea to improve it…

_rules/iframe-accessible-name-cae760.md
```diff
@@ -150,12 +151,22 @@ Does not apply to non `iframe` element.
...
 
 ```html
 <iframe style="display:none;" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
+#### Inapplicable Example 3
+
+This `iframe` is not part of [sequential focus navigation][] because it has `tabindex="-1"` and not [included in the accessibility tree][] because of `role="presentation"`
```
```suggestion
This `iframe` is neither part of [sequential focus navigation][] (because it has `tabindex="-1"`) nor ([included in the accessibility tree][] because of `role="presentation"`).
```

Jym77:
> The same thing applies to all rules that test for accessible names of "user interface components". I do recall discussions on this very matter previously, but I don't recall when nor where.

Should there be a generic rule like "element in tab order has accessible name"?
Since that is probably confusing anyway to have a tab-stop in something that you have no clue what it is…

_rules/iframe-accessible-name-cae760.md
WilcoFiers: I don't think that's necessary.

_rules/iframe-accessible-name-cae760.md
WilcoFiers: OK, leaving it as is, since it got 2 approvals.

_rules/iframe-accessible-name-cae760.md
WilcoFiers: I don't know what this is for, seems unnecessary.

PR #1106, opened 2020-01-20
Adjust definition of valid language subtag (de46e4, bf051a, 5b7ae0)

### LINKED ISSUE #1105: Make lang matching less strict (bf051a, de46e4)
Currently the "valid language" rules require that the language tags follow the BCP-47 syntax. @kengdoj pointed out to the ACT Task Force that this might be too strict. Looking at what browsers are doing today, CSS 2.2 and CSS 3 have a simpler algorithm defined for matching language tags. We could consider adopting this. There are some drawbacks to this, mainly that grandfather tags may be more difficult to match.

Another approach could be to look at RFC 4647, which is specifically about matching language tags. It is used in CSS 4 in the `:lang()` pseudo class. There are some parts to that (like wildcards "*") which do not work for what we are doing with rules though, so we'd have to deal with that.

kasperisager:
This pull request makes the definition of "valid language subtag" much more lenient to better align with how known user agents and assistive technologies behave. The relevant bit:

> A language tag consist of a primary language subtag from the language subtag registry, optionally followed by a hyphen (`-`) and any sequence of characters.

Who might be up for reviewing this? I've assigned @kengdoj so far as you were the one who voiced concern over the current definition. Let me know what you think of the suggested definition!

Closes #1105.


pages/glossary/valid-language-subtag.md
```diff
@@ -3,9 +3,13 @@ title: Valid Language Subtag
 key: valid-language-subtag
 ---
 
-A text string following the syntax as defined in [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt), with a [primary language subtag](https://tools.ietf.org/html/bcp47#section-2.2.1) from the [language subtag registry](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry).
+A language tag consist of a [primary language subtag][] from the [language subtag registry][], optionally followed by a hyphen (`-`) and any sequence of characters.
```
daniel-montalvo:
```suggestion
A language tag consists of a [primary language subtag][] from the [language subtag registry][], optionally followed by a hyphen (`-`) and any sequence of characters.
```

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -59,6 +63,14 @@ This `html` element has a `lang` attribute whose value is a [valid][valid langua
 <html lang="fr"></html>
 ```
 
+#### Passed Example 2
+
+This `html` element has a `lang` attribute value is a [valid language subtag][] even though the [region subtag][] is not.
```
kengdoj: ... has a 'lang' attribute that is a [valid language subtag]...

kasperisager:
The final call will end on February 14.

_rules/element-lang-valid-de46e4.md
```diff
@@ -73,33 +77,33 @@ The `lang` attribute has a value that is not empty ("") and has a valid primary
...
+#### Passed Example 3
 
-The `lang` attribute value is not a valid primary language subtag.
+The `lang` attribute value has a valid primary language subtag, but a syntactically invalid region subtag.
 
 ```html
 <html>
 	<body>
-		<article lang="dutch"></article>
+		<p lang="en-US-GB"></p>
```
dd8: This is right - BCP 47 has an algorithm for determining best matching language in 'Matching of Language Tags' at the end of BCP 47:

> In the lookup scheme, the language range is progressively truncated
   from the end until a matching language tag is located

So lookup order is `en-US-GB` to `en-US` to `en`

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -34,7 +34,11 @@ For each test target, the `lang` attribute has a [valid language subtag](#valid-
 
 ## Assumptions
 
-This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+- This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+
+- This rule assumes that user agents and assistive technologies can programmatically determine [valid language subtags](#valid-language-subtag) even if these do not conform to the [BCP 47][] syntax.
```
dd8: This is a reasonable assumption - BCP 47 has an algorithm for determining best matching language in 'Matching of Language Tags' at the end of BCP 47:

> In the lookup scheme, the language range is progressively truncated
   from the end until a matching language tag is located

Even if an implementor hasn't read this part of the spec, it's still likely they'll implement it because it's easy to implement and fairly obvious.

Jym77:
@kasperisager: need to add "bcp" to `spelling_ignore.yml` to pass test.

kasperisager:
Gah, the ignore list is very literal in its handling of casing; I added `BCP` 🤦‍♂ We should really consider moving to a personal dictionary (https://github.com/wooorm/nspell#personal-dictionary-documents) rather than an ignore list.

WilcoFiers:
Looks good, but as JY says tests are failing. Also I think this could have been a 1 week final call, since it's an update to the rule.

Issue #1060, opened 2019-12-05
meta viewport does not prevent zoom [b4f0c3]


patrickhlauke:
Wondering if for https://act-rules.github.io/rules/b4f0c3 it's worth mentioning that this only affects mobile/tablet browsers (as desktop browsers simply ignore viewport meta), and that in most modern browsers the viewport meta that prevents zooming/scaling is ignored (Safari/iOS for the last few years) or that there's a user setting to always allow zoom (Chrome, Firefox, etc on Android).

Additionally, wondering if under assumptions it should also include something about there not being any alternative ways to zoom. For instance, even IF a site tried to block viewport scaling, but provided its own custom text resizing buttons/functionality/setting, it would pass the SC's requirement. Either that, or change the "Any failed outcomes:" to "further testing", rather than an outright "not satisfied".

Jym77:
As per ACT Rules format, there must be a fail-on-fail relationship with the SC (if the rule fails, then the SC must fail; i.e. rules are always flagging actual errors, maybe not all of them).
https://www.w3.org/TR/act-rules-format/#outcome-mapping

If this is not the case, we indeed need to add an assumption to restrict the usage of the rule.

WilcoFiers:
@patrickhlauke I think you're absolutely correct. I think the bit about only certain browsers blocking zoom with meta viewport should got into accessibility supported, and the assumption obviously in the assumptions section. Would you like to put together a pull request with a fix for this?

WilcoFiers:
I've had some testing done for this, and all modern mobile browsers either ignore this, or have a setting which allows users to bypass this. I think the fix is to update the accessibility support note to make it clear this is only really a problem with older mobile browsers.

WilcoFiers:
We're updating the accessibility support of this issue. Discussed on the call whether or not this rule should be deprecated. Group decided against it as there are still use cases for it, even if those are less common.

PR #1108, opened 2020-01-21
Rule: meta viewport does not prevent zoom - update accessibility support (b4f0c3)

jkodu:
Closes issue(s):

- https://github.com/act-rules/act-rules.github.io/issues/1060


_rules/meta-viewport-b4f0c3.md
```diff
@@ -35,7 +35,7 @@ For each test target, the [`content`](https://www.w3.org/TR/html52/document-meta
 
 ## Accessibility Support
 
-_There are no major accessibility support issues known for this rule._
+Support for `meta name="viewport"` in older mobile browsers is not consistent.
```
WilcoFiers: I think this is too vague. We should be more explicit about what the implications of accessibility support are for the rule:

```suggestion
Desktop browsers ignore the viewport `meta` element, and most modern mobile browsers either ignore it by default, or have an accessibility option which will allow zooming. This rule is not relevant for desktop browsers, nor for most modern mobile browsers. Only users with older mobile browsers can experience issues tested by this rule.
```

jkodu:
According to our PR template suggestions we do not need final call for accessibility support.

PR #1114, opened 2020-01-21
Add note on difference between accessibility and DOM trees

Jym77:
There seems to have been some confusion, even amongst ourselves, as to the subtleties of "child in the accessibility tree". Adding a note to hopefully reduce further confusion.

Closes issue(s):
- N/A

Need for Final Call:
This will not require a Final Call (editorial)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/pages/glossary/owned-by.md
+++ b/pages/glossary/owned-by.md
@@ -7,6 +7,6 @@ An element A is owned by element B if element A is a child of element B in the [
 
 **Note:** This definition is different from the definition of ["owned element" in WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element). Because browsers have different accessibility trees, which element owns which other elements can vary between browsers. Until there is a standard accessibility tree, testing with multiple accessibility trees may be necessary.
 
-**Note:** Being a child in the [accessibility tree][] is very different from being a child in the DOM tree. Many DOM nodes have no corresponding node in the [accessibility tree][] (for example, because they are marked with `aria-hidden="true"`). Thus, a child in the [accessibility tree][] can correspond to a distant descendant in the DOM tree. Additionally, the use of `aria-owns` attribute can change the tree structure.
+**Note:** Being a child in the [accessibility tree][] can be different from being a child in the DOM tree. Some DOM nodes have no corresponding node in the [accessibility tree][] (for example, because they are marked with `role="presentation"`). A child in the [accessibility tree][] can correspond to a descendant in the DOM tree. Additionally, the use of `aria-owns` attribute can change the tree structure.
 
 [accessibility tree]: https://www.w3.org/TR/act-rules-aspects/#input-aspects-accessibility 'Definition of accessibility tree'
--- a/pages/glossary/owned-by.md
+++ b/pages/glossary/owned-by.md
@@ -3,6 +3,10 @@ title: Owned by
 key: owned-by
 ---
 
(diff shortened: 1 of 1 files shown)


carlosapaduarte:
I'm assuming that this change is a result of the conflict Brian identified. I don't think this note addresses the issue that is causing the conflict.

A simplified version of the code is
````html
<ul role="menubar">
  <li>
    <a href="#" role="menuitem">...</a>
  </li>
</ul>
````

The accessibility tree will be something like:
> menubar
>> listitem
>>> menuitem

The "menuitem" is not a child of the "menubar" in the accessibility tree (it has nothing to do withh the elements in the DOM). But it is a descendant. Why don't we use descendant instead of child in the defenition of owned by?

**Comment by Jym77** [comment]
> ```
> <ul role="menubar">
>   <li>
>     <a href="#" role="menuitem">...</a>
>   </li>
> </ul>
> ```

Hum… This looks quite similar to [Failed Example 2](https://act-rules.github.io/rules/ff89c9#failed-example-2).
Given that @WilcoFiers was pretty adamant in defining "owned by" in term of children, not descendants, I suggest that you to get together and figure out what to do 😄

**Comment by Jym77** [comment]
For Agenda item: there seems to be disagreement between @carlosapaduarte and @WilcoFiers whether "owned by" should target **children** or **descendants* in the accessibility tree.

File: pages/glossary/owned-by.md [hunk 1]
```diff
@@ -3,6 +3,10 @@ title: Owned by
...
 
-> **Note:** This definition is different from the definition of ["owned element" in WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element). Because browsers have different accessibility trees, which element owns which other elements can vary between browsers. Until there is a standard accessibility tree, testing with multiple accessibility trees may be necessary.
+**Note:** This definition is different from the definition of ["owned element" in WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element). Because browsers have different accessibility trees, which element owns which other elements can vary between browsers. Until there is a standard accessibility tree, testing with multiple accessibility trees may be necessary.
+
+**Note:** Being a child in the [accessibility tree][] is very different from being a child in the DOM tree. Many DOM nodes have no corresponding node in the [accessibility tree][] (for example, because they are marked with `aria-hidden="true"`). Thus, a child in the [accessibility tree][] can correspond to a distant descendant in the DOM tree. Additionally, the use of `aria-owns` attribute can change the tree structure.
```
WilcoFiers: The text is a little verbose. Trimming it down a bit:

```suggestion
**Note:** Being a child in the [accessibility tree][] can be different from being a child in the DOM tree. Some DOM nodes have no corresponding node in the [accessibility tree][] (for example, because they are marked with `role="presentation"`). A child in the [accessibility tree][] can correspond to a descendant in the DOM tree. Additionally, the use of `aria-owns` attribute can change the tree structure.
```

kasperisager:
@carlosapaduarte Alfa fails that example so I'm not quite sure about the results that @ShadowBB was seeing.

Jym77:
> Is `aria-hidden="true"` the only reason why a DOM node does not have a corresponding node in the accessibility tree? Couldn't it be just that there are DOM nodes that do not convey accessibility information and thus there is no need for the accessibility tree to expose them? If so, `aria-hidden="true"` would be one particular reason. Maybe it is worth broadening scope here.

Yes, there are a lot of way to remove an element from the accessibility tree. The single case is provided within parenthesis and starting with "for example". I hope this is enough and we don't need to list all of the possibilities…

PR #1116, opened 2020-01-22
Fix failed example for "Heading has accessible name" (ffd0e9)

### LINKED ISSUE #1110: Heading has accessible name - ffd0e9 - Failed example 1
The Failed Example 1 in rule [https://act-rules.github.io/rules/ffd0e9m](https://act-rules.github.io/rules/ffd0e9m) is wrong. As you can see in [https://www.w3.org/TR/html-aam-1.0/#section-elements-and-grouping-content-elements-not-listed-elsewhere](https://www.w3.org/TR/html-aam-1.0/#section-elements-and-grouping-content-elements-not-listed-elsewhere), if the element has an `aria-label` then you must use the algorithm defined in [https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) 2C, which states that if the `arial-label` is empty (""), then it should be ignored. So by following the algorithm the accessible name is derived by the text from the node. This example is similar to the Passed Example 1.

### LINKED ISSUE #1126: Heading has accessible name failed example 1 is not a failure
```
<h1 aria-label="">ACT rules</h1>
```

https://act-rules.github.io/rules/ffd0e9#failed-example-1

The acc-name calculation ignores an `aria-label` if it's empty or only contain whitespace, so 'ACT rules' is used as the name in this example. Safari, Firefox and Chrome all use 'ACT rules' as the accessible name for this test case.

Here's the step from the acc-name calculation

> Otherwise, if computing a name, and if the current node has an aria-label attribute whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
If traversal of the current node is due to recursion and the current node is an embedded control as defined in step 2E, ignore aria-label and skip to rule 2E.
Otherwise, return the value of aria-label.

https://www.w3.org/TR/accname-1.2/

Jym77:
Make the Failed example actually fail the rule

Closes issue(s):
- closes #1110 
- closes #1126

Need for Final Call:
This will require a 1 week Final Call (updating a single example)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/heading-accessible-name-ffd0e9.md
+++ b/_rules/heading-accessible-name-ffd0e9.md
@@ -96,10 +96,11 @@ Even though this `h1` element is not [visible][], it is still [included in the a
 
 #### Failed Example 1
 
-This `h1` element has an empty [accessible name][] given by its `aria-label` attribute.
+This `h1` element has an empty [accessible name][] given by its `aria-labelledby` attribute.
 
 ```html
-<h1 aria-label="">ACT rules</h1>
+<span id="label"></span>
+<h1 aria-labelledby="label">ACT rules</h1>
 ```
 
 #### Failed Example 2


Jym77:
Final Call ends on January 30th.

Jym77:

Issue #1129, opened 2020-01-28
Semantic role definition improvements


EmmaJP:
https://github.com/act-rules/act-rules.github.io/blob/develop/pages/glossary/semantic-role.md

It may help to be clear that an assigned role should be defined in a suitable taxonomy, such as the WAI-ARIA spec (https://www.w3.org/TR/wai-aria-1.1/#role_definitions), Graphics AAM spec (https://w3c.github.io/graphics-aam/#mapping_role_table) or similar.

See https://github.com/act-rules/act-rules.github.io/pull/1085/files/5e18343309bbe6fbcc998f311131883db60d50a9#r365270117

-------

It is also very confusing that the files relating to these definitions link back to themselves when linking to other definitions. Seeing how they are pulled into rules, it sort of makes sense, but does not always work as expected if all the referenced definitions are not included in a rule because the rule itself does not mention them. This is particularly noticeable with rules that reference semantic role, which in turn references explicit semantic role and implicit semantic role.

Jym77:
> It is also very confusing that the files relating to these definitions link back to themselves when linking to other definitions. Seeing how they are pulled into rules, it sort of makes sense, but does not always work as expected if all the referenced definitions are not included in a rule because the rule itself does not mention them. This is particularly noticeable with rules that reference semantic role, which in turn references explicit semantic role and implicit semantic role.

Inclusion of the definition in the rules should be recursive (i.e. also including definitions that are only used in an included definition, not in the rule itself). If this is not the case, can you open a specific issue for this, pointing to the precise case and assigning @jkodu ?

EmmaJP:
Does it require a separate issue?

Jym77:
Yes, notably because this is a website building problem, not an rule problem. So it's should be an issue in the https://github.com/act-rules/act-rules-web project.

WilcoFiers:
I'm not entirely sure what you mean Emma, but would you like to put together a pull request for this?

EmmaJP:
> Yes, notably because this is a website building problem, not an rule problem. So it's should be an issue in the https://github.com/act-rules/act-rules-web project.

https://github.com/act-rules/act-rules-web/issues/120

jkodu:
Closing issue, as we have a PR in place.

PR #1134, opened 2020-01-30
Improve clarity of definition to be less ambiguous

EmmaJP:
Added a change that makes it clear that a role value will come from a recognised standardised list. This is for clarity, to reduce ambiguity, with recognition that an invalid explicit role will default to an implicit role as determined by the browser.

Closes issue(s):

- https://github.com/act-rules/act-rules.github.io/issues/1129

Need for Final Call:
This will not require a Final Call as it is clarifying a definition rather than changing any meaning or rules that depend on it.

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/semantic-role.md
```diff
@@ -5,6 +5,6 @@ unambiguous: true
 objective: true
 ---
 
-A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic association that indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
+A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic association that indicates an object's type, as defined by a standard such as the [WAI-ARIA spec](https://www.w3.org/TR/wai-aria-1.1/#role_definitions), [Graphics AAM spec](https://w3c.github.io/graphics-aam/#mapping_role_table), or similar. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
```
EmmaJP:
```suggestion
A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic association that indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.

Valid semantic roles are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications](https://act-rules.github.io/glossary/#wai-aria-specifications).
```

EmmaJP: @Jym77 does that suggestion cover it ^^

Jym77: Yes, I think it's good.

EmmaJP:
Thanks @jkodu

PR #1136, opened 2020-01-30
test: fix failing decorative reference in link

jkodu:
I have fixed this in 2 other PR's, but it looks like they have a week long final call. Fixing a reference to pass a test, should not be a final call.

Closes issue(s):
- NA

Need for Final Call:
- NA


jkodu:
> Note: I will be merging this after 1 approval

jkodu:
Closing this, as @WilcoFiers mentioned that he is taking care of it.

PR #1137, opened 2020-01-31
chore: add git lfs support to track large files

jkodu:
When the website is [published](https://github.com/act-rules/act-rules-web/blob/master/.github/workflows/publish.yml#L42), there are some large files which needs to be git tracked. 

Closes issue(s):
- NA

Need for Final Call: No


jkodu:
Given this is a chore, it will be merged with < 3 approvals.

PR #1146, opened 2020-02-04
"Id attribute value is unique" (3ea0c8): Add title to iframe in example

Jym77:
Add a title to the iframe in an example, to make it better code practice.

Closes issue(s):

- partial for https://github.com/w3c/wcag-act/issues/426

Need for Final Call:
This will require a 1 week Final Call (updating one example).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
> I am approving this as this has a clear scope and rationale. However, in the explanatory paragraph for some of the test cases we have.
> 
> > ### Passed Example 2
> > 
> > * All ids are unique within the document context.
> > 
> > Passed Example 3
> > 
> > * Two of the id are the same (my-elt), but they are in different trees (the first one in the document tree and the second in the shadow tree).
> 
> It seems we are missing the plural form (ids( for some instances in this rule.
> 
> > Passed Example 4
> > Both id are the same (my-elt), but they are in different document trees because the iframe is creating a new one.
> 
> Do you all agree that these should be updated?
> 
> If so, I can take care of it in another pull request.

I will add that to #1150. Thanks.

Jym77:
Final Call ends on February 11.

EmmaJP:
As per my comments on #1150, I think `id` can be considered implicitly plural, especially when marked as a code snippet. It feels nicer to use `id` attributes, with the plural on attribute, rather than `id`s. 

act-rule usage is almost exclusively referring to `id` as an attribute of an element and not I.D. as an acronym.

Jym77:

PR #1147, opened 2020-02-04
"Id attribute value is unique" (3ea0c8): Add note about partial coverage of the SC

Jym77:
Adds a note explaining why if the rule is passing, "further testing is needed" for the SC.

Closes issue(s):

- partial for https://github.com/w3c/wcag-act/issues/426

Need for Final Call:
This will not require a Final Call (adding a non-normative note)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


EmmaJP:
Do we need to add notes like this into rules? I don't consider an atomic test rule to be everything required to pass a success criteria. It passes or fails the specifics of the test as described, the test being relevant to a particular success criteria.

If we add a note like this into this rule, similar notes would also need to be added into almost every other rule as well.

WilcoFiers:
@EmmaJP I don't think this is strictly necessary, but we were asked to explain it. We've done this in a few other places too.

PR #1153, opened 2020-02-04
"HTML page language is valid" (bf051a): Explain assumption

Jym77:
Explain the meaning of the assumption.

Closes issue(s):

- partial for https://github.com/w3c/wcag-act/issues/429

Need for Final Call:
This will not require a Final Call (increase details level of an assumption).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
> I don't have an issue with the additional explanation ... though for a rule related to 3.1.1 it would seem reasonable that "programmatically determined" implies it is information provided in the code somewhere, and not depending on complex AI or a person. In which case, is this explanation needed?

That (tries to) answer comment from the AG…

I'd see it more like a browser (or HTML 6) deciding that `language` is a better name for the attribute and should take precedence over `lang` and thus `lang="xyz" language=`en`" would pass WCAG (for this browser) but fail the rule. A bit overstretched, maybe, but it seems to always have been the meaning of this assumption as far as I can understand. (or maybe the language is sent as part of some separate meta-data file, or …)

EmmaJP:
> That (tries to) answer comment from the AG…
> 
> I'd see it more like a browser (or HTML 6) deciding that `language` is a better name for the attribute and should take precedence over `lang` and thus `lang="xyz" language=`en`" would pass WCAG (for this browser) but fail the rule. A bit overstretched, maybe, but it seems to always have been the meaning of this assumption as far as I can understand. (or maybe the language is sent as part of some separate meta-data file, or …)

If there are actual examples of browsers that do that kind of thing, they may be useful to add as examples in the explanation to give it that context.

dd8:
> > That (tries to) answer comment from the AG…
> > I'd see it more like a browser (or HTML 6) deciding that `language` is a better name for the attribute and should take precedence over `lang` and thus `lang="xyz" language=`en`" would pass WCAG (for this browser) but fail the rule. A bit overstretched, maybe, but it seems to always have been the meaning of this assumption as far as I can understand. (or maybe the language is sent as part of some separate meta-data file, or …)
> 
> If there are actual examples of browsers that do that kind of thing, they may be useful to add as examples in the explanation to give it that context.

Browsers calculate the language in various (non-interoperable) ways and don't just use the `lang` attribute. They use a mix of `lang`, `meta http-equiv=Content-Language`, HTTP Content-Language header, HTTP Accept-Language and browser UI language. They don't all use the same mix:

https://github.com/act-rules/act-rules.github.io/issues/668#issuecomment-534606384

Jym77:
@dd8 many thanks for the examples.
@EmmaJP I've added these examples to the assumption.

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -34,7 +34,7 @@ For each test target, the `lang` attribute has a [valid language subtag](#valid-
 
 ## Assumptions
 
-This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG. WCAG only requires that the language of the page can be "programmatically determined", but this could be achieved by using other means than a `lang` attribute (e.g., HTTP headers, `meta` element, …). Thus, an invalid `lang` attribute could be overridden by some other programmatic technique and not cause any accessibility issue or WCAG failure; but this rule would nonetheless fail.
```
WilcoFiers: How about:

```suggestion
There are other methods in HTML by which the language of the page is set; using http-headers or using the `meta` element. These are generally not supported by assistive technologies. This rule assumes that these two methods are insufficient to satisfying [success criterion 3.1.1 language of the page](https://www.w3.org/TR/WCAG21/#language-of-page).
```

Jym77: I like it. Rephrased a bit to be less restrictive (there may be ways we haven't think of, these are quite likely not well supported by AT…)

PR #1155, opened 2020-02-05
fix: update link in background section (5b7ae0)

jkodu:
As per feedback from AG (https://github.com/w3c/wcag-act/issues/417#issuecomment-578440138)

Closes issue(s):

- https://github.com/w3c/wcag-act/issues/417

Need for Final Call:
- No

### DIFF
--- a/_rules/html-page-lang-xml-lang-match-5b7ae0.md
+++ b/_rules/html-page-lang-xml-lang-match-5b7ae0.md
@@ -48,7 +48,7 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
 
 - [H57: Using language attributes on the html element](https://www.w3.org/WAI/WCAG21/Techniques/html/H57)
 - [BCP 47: Tags for Identifying Languages](https://www.ietf.org/rfc/bcp/bcp47.txt)
-- [The `lang` and `xml:lang` attributes](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes)
+- [The `lang` and `xml:lang` attributes](https://www.w3.org/TR/html52/dom.html#the-lang-and-xmllang-attributes)
 
 ## Test Cases
 


_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -48,7 +48,7 @@ Since most assistive technologies will consistently use `lang` over `xml:lang` w
 
 - [H57: Using language attributes on the html element](https://www.w3.org/WAI/WCAG21/Techniques/html/H57)
 - [BCP 47: Tags for Identifying Languages](https://www.ietf.org/rfc/bcp/bcp47.txt)
-- [The `lang` and `xml:lang` attributes](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes)
+- [The `lang` and `xml:lang` attributes](https://www.w3.org/TR/html52/dom.html#the-lang-and-xmllang-attributes)
```
Jym77: I don't get it. Why are we citing HTML 5.2 again? Didn't we decide to move to WHATWG?
I guess we haven't fully resolved #879.

jkodu: I am always unsure about this. We should update our test to catch this too then.

jkodu:
Thanks for referring to the history on this. @WilcoFiers - thoughts?

_rules/html-page-lang-xml-lang-match-5b7ae0.md
WilcoFiers: @jkodu I think the comment from Mary Jo is already resolved. I don't think this PR is needed.

jkodu:
Awesome, thanks Wilco.

jkodu:
Enhanced the test - https://github.com/act-rules/act-rules.github.io/pull/1158.
Closing issue.

PR #1156, opened 2020-02-05
fix: update accessibility support per feedback (cae760)

jkodu:
For more context, see https://github.com/w3c/wcag-act/issues/428#issuecomment-582321514

Closes issue(s):

- https://github.com/w3c/wcag-act/issues/428

Need for Final Call:
- 1 week


_rules/iframe-accessible-name-cae760.md
```diff
@@ -133,7 +133,7 @@ Usage of `alt` attribute to describe content is not valid.
 
 #### Failed Example 7
 
-[Accessible name][] is empty.
+[Accessible name][] consists only [whitespace][].
```
```suggestion
The [accessible name][] of this `iframe` element consists only of [whitespace][].
```

_rules/iframe-accessible-name-cae760.md
```diff
@@ -157,6 +157,15 @@ Does not apply to non `iframe` element.
 <iframe style="display:none;" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
+#### Inapplicable Example 3
+
+The `iframe` element is [marked as decorative][] through `role="presentation"`.
```
```suggestion
This `iframe` element is [marked as decorative][] through `role="presentation"`.
```

Jym77:
> Need for Final Call:
> 
> * Perhaps?

1 week. It is adding a new example.

Jym77:
Don't add the Final Call label before the rule is actually in Final Call…

_rules/iframe-accessible-name-cae760.md
```diff
@@ -166,8 +167,17 @@ This `iframe` is not part of [sequential focus navigation][] because it has `tab
 <iframe tabindex="-1" role="presentation" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
+#### Inapplicable Example 4
```
WilcoFiers: I don't think this example should be here. Strictly speaking, focusable elements, which iframes **can** be, depending on the browser, can not have a semantic role of `presentation` or `none`. I only found this out recently. I think inapp example 3 covers this sufficiently.

jkodu:
Final call ends on 17th Feb

_rules/iframe-accessible-name-cae760.md
mraccess77: The ARIA in HTML document indicates presentation can be used. https://www.w3.org/TR/html-aria/.  What I'm trying to determine if if iFrames with tabindex-1 should be inapplicable from test regardless of role presentation -- or only when they have role presentation or none.  That is -- if the iFrame only has tabindex-1 but not a role is it a fail or inapplicable?  Seems like the iFrame still acts like a container or region of sorts -- so while it might not fit under user interface component under 4.1.2 it might fit under SC 1.3.1 information and relationships and still need an accessible name.

PR #1158, opened 2020-02-05
test: add use `http://html.spec.whatwg.org` instead of `://www.w3.org/TR/html52/`

jkodu:
Closes issue(s):

- NA

Need for Final Call: No


jkodu:
Ha ha, we seem to use the `html52` in 12 places. 
Would love a hand with finding the alternative urls.

See https://github.com/act-rules/act-rules.github.io/pull/1158/checks?check_run_id=427651950 for failing tests

PR #1159, opened 2020-02-06
Unambiguous definition of "Section of content"

Jym77:
The definition of "section of content" is currently ambiguous (https://github.com/act-rules/act-rules.github.io/pull/1085#discussion_r364145387) but is used in the [expectation of Heading is descriptive](https://act-rules.github.io/rules/b49b2e#expectation), which is not permitted by the ACT rule format.

This creates an unambiguous (and objective) definition of "programmatic section of content" and uses it in the rule.

Need for Final Call:
This will require a 1 week Final Call (changing the expectation of a rule)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/heading-section-association.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Association between headings and sections
+key: heading-section-association
+unambiguous: true
+objective: true
+---
+
+Each element with a [semantic role][] of [`heading`][heading] is _associated_ with exactly one [programmatic section of content][] in the following ways:
```
kasperisager:
```suggestion
Each element with a [semantic role][] of [`heading`][heading] is _associated_ with exactly one [programmatic section of content][] in one of the following ways:
```

pages/glossary/implicit-section-of-content.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Implicit section of content
+key: implicit-section-of-content
+unambiguous: true
+objective: true
+---
+
+An _implicit section of content_ is all the nodes, in tree order, between an element with a [semantic role][] of [`heading`][heading] and the next element (in tree order) which has both a [semantic role][] of [`heading`][heading] and an [`aria-level`][aria-level] property lower of equal, except for the first element with a [semantic role][] of [`heading`][heading] inside an [explicit section of content][].
```
kasperisager:
```suggestion
An _implicit section of content_ is all the nodes, in tree order, between an element with a [semantic role][] of [`heading`][heading] and the next element (in tree order) which has both a [semantic role][] of [`heading`][heading] and an [`aria-level`][aria-level] property lower or equal to that of the previous [`heading`][heading] element, except for the first element with a [semantic role][] of [`heading`][heading] inside an [explicit section of content][].
```

pages/glossary/programmatic-section-of-content.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Programmatic section of content
+key: programmatic-section-of-content
+unambiguous: true
+objective: true
+---
+
+A _programmatic section of content_ is either an [explicit section of content][] or an [implicit section of content][].
+
+If a programmatic section of content contains a [host][], with a [shadow root][] associated to it, or a [browsing context container][]; then the content of the corresponding shadow tree or browsing context is also part of this programmatic section of content.
```
kasperisager: "contains" will need to be better qualified. I'm guessing we're looking for a descendant?

Jym77: Not really. Descendant is OK for explicit sections of content, but not for implicit ones…

_rules/heading-descriptive-b49b2e.md
```diff
@@ -29,13 +29,16 @@ This rule applies to any element with the [semantic role][] of heading that is e
...
 
 **Note:** Headings do not need to be lengthy. A word, or even a single character, may suffice.
 
 ## Assumptions
 
-This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the [semantic roles][semantic role] (including the level of headings) are used according to their semantics.
+- This rule assumes that [nested browsing contexts][nested browsing context] included in an implicit section of content do not contain any element with a role of `heading` and an `aria-level` lower or equal to the one of the `heading` defining this implicit section of content.
+- This rule assumes that the `hgroup` element is not used.
```
WilcoFiers: We should probably just deal with this in the definition, rather than assume it isn't used.

_rules/heading-descriptive-b49b2e.md
```diff
@@ -29,13 +29,16 @@ This rule applies to any element with the [semantic role][] of heading that is e
...
+Each target element describes the topic or purpose of the [programmatic section of content][] it is [associated][] with.
 
 **Note:** Headings do not need to be lengthy. A word, or even a single character, may suffice.
 
 ## Assumptions
 
-This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the [semantic roles][semantic role] (including the level of headings) are used according to their semantics.
+- This rule assumes that [nested browsing contexts][nested browsing context] included in an implicit section of content do not contain any element with a role of `heading` and an `aria-level` lower or equal to the one of the `heading` defining this implicit section of content.
```
WilcoFiers: I think the assumption you're making here is that the content of the nested browsing context can be treated as a single section.

pages/glossary/programmatic-section-of-content.md
```diff
@@ -0,0 +1,121 @@
...
+**Note:** This definition roughly follows the [algorithm for creating an outline][outline algorithm], but using different elements for defining sections. The main focus is here on the [semantic role][] rather than the element itself, and all [landmark roles][landmark] are considered. Additionally, programmatic sections of content include content from [shadow trees][] and [nested browsing contexts][] which is ignored by the [algorithm for creating an outline][outline algorithm].
+
+**Note:** When elements and roles are used according to their semantics, programmatic sections of content and [sections of content][] should match perfectly.
+
+#### Assumption
+
+- This definition assumes that the [semantic roles][semantic role] (including the level of headings) are used according to their semantics.
+- This definition assumes that the `hgroup` element is not used.
+
+#### Examples
```
WilcoFiers: I think we should put examples in a separate page.

pages/glossary/programmatic-section-of-content.md
```diff
@@ -0,0 +1,121 @@
...
+
+Programmatic sections of content may contain nested programmatic subsections of content.
+
+Programmatic sections of content may have zero or one heading [associated][] with them.
+
+**Note:** This definition roughly follows the [algorithm for creating an outline][outline algorithm], but using different elements for defining sections. The main focus is here on the [semantic role][] rather than the element itself, and all [landmark roles][landmark] are considered. Additionally, programmatic sections of content include content from [shadow trees][] and [nested browsing contexts][] which is ignored by the [algorithm for creating an outline][outline algorithm].
+
+**Note:** When elements and roles are used according to their semantics, programmatic sections of content and [sections of content][] should match perfectly.
+
+#### Assumption
```
WilcoFiers: I'm not a fan of putting assumptions into definitions. I understand this is something we want to document is part of this assumption, but this belongs in the assumption section of the rule.

pages/glossary/programmatic-section-of-content.md
```diff
@@ -0,0 +1,121 @@
...
+
+- This definition assumes that the [semantic roles][semantic role] (including the level of headings) are used according to their semantics.
+- This definition assumes that the `hgroup` element is not used.
+
+#### Examples
+
+**Note:** The examples presented here are non-normative and not testable. They serve to illustrate some common pitfalls about the definition and to help implementers of ACT rules understand it.
```
WilcoFiers: Saying the examples are non-normative gives the impression the rest of this is normative, which it isn't. I think this should be rephrased.

_rules/heading-descriptive-b49b2e.md
```diff
@@ -29,13 +29,16 @@ This rule applies to any element with the [semantic role][] of heading that is e
...
-Each target element describes the topic or purpose of its [section of the content][].
+Each target element describes the topic or purpose of the [programmatic section of content][] it is [associated][] with.
 
 **Note:** Headings do not need to be lengthy. A word, or even a single character, may suffice.
 
 ## Assumptions
 
-This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the [semantic roles][semantic role] (including the level of headings) are used according to their semantics.
```
WilcoFiers: I'm not comfortable with this assumption. I understand this is the way to make a programatically determined section work, but I'm not so sure that that's the right way to go about this.  I don't think we have any reason to assume this is true. WCAG doesn't require it, nor is there much else that would enforce it.

In my opinion, what needs to happen to fix 'section of the content" in this rule is to make that definition explicitly about how the content is presented. Define a section of the content based on visual cues, like differences in background, border, whitespace, and based on what is the next heading with the same presentation.

pages/glossary/explicit-section-of-content.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Explicit section of content
+key: explicit-section-of-content
+unambiguous: true
+objective: true
+---
+
+An _explicit section of content_ is a subtree of the [flat tree][], rooted at an element with a [semantic role][] that inherits either from the [`landmark`][landmark] or [`document`][document] roles, except if it is the only child of an element with a [semantic role][] that inherits either from the [`landmark`][landmark] or [`document`][document] roles.
```
WilcoFiers: Should this include application roles? Also, I'm not sure here, does this include role=document itself? If not, why not?

_rules/heading-descriptive-b49b2e.md
ShadowBB: While I agree that this assumption is not the way to go, I don't see how a definition based on visual cues would work either. We can never really describe the maried ways in which this could be made visually apparent.

I think the definition should be based on what is the next heading with the same presentation... and then state that it can also be defined by visual cues (with examples) but that this is inherently subjective and will differ from audit to audit.

_rules/heading-descriptive-b49b2e.md
Jym77: The problem is that is has no default role. Handling it will make the definition mixes roles and elements. It is doable, but it gets much more dirty (the "implicit section" delimiter will be `hgroup` or `role="heading"` not within an `hgroup`).

The fact that `hgroup` was deprecated in HTML 5.2 (but kept in WHATWG living standard… 🙄) also played its part in me choosing to just ignore it…

According to https://www.advancedwebranking.com/html/#content-sectioning `hgroup` represents 0.1% of the **Content sectioning** elements used (September 2019). I say we're safe enough in ignoring it.

_rules/heading-descriptive-b49b2e.md
Jym77: I am aware that WCAG considers that styling is enough to make a heading. (and that if one uses only `font-size: large` to mark headings, this is goings to be a problem for "Info & Relationship").

If we base sections of content purely in terms of visual cues, this is however going to be completely irrelevant for screenreaders who do not have access to these visual cues.

The rule is specifically targeting `role="heading"` hinting that it is not about visual cues only (maybe that needs to change also).

pages/glossary/explicit-section-of-content.md
Jym77: I've improved the formulation to make it clear that the `document` role is included.

I've currently left out the `application` role as I'm not entirely sure about it (and how it is used), and because it's not considered at all in the outline algorithm. But it's not a strong feeling and I'll add it if experts think it's needed.

_rules/heading-descriptive-b49b2e.md
WilcoFiers: I strongly disagree with you Brian. Take this example:
https://codepen.io/wilcofiers/pen/abOvWRB

It is visually clear that the bottom paragraph is not intended to be part of the h2 subsection. This kind of a "card" design pattern is very common. As for why I think we can make this rule about programatic headings, but not about programatic sections, that's because heading markup is required by WCAG. I don't think WCAG requires the same for sections, nor should it, because there's no commonly agreed upon definition of what that even is. That's why you couldn't just use an existing definition for this proposal.

Jym77: Yes, my proposal would put the bottom paragraph as part of the last `h2` section. Which is not what is perceived by sighted users.

But, wouldn't an AT/screen reader do the same (put the bottom paragraph in the last `h2`)? Because it can't detect the visual border marking the cards. I guess this fails Info&Relationship in any case…

WilcoFiers: > But, wouldn't an AT/screen reader do the same (put the bottom paragraph in the last h2 section

I don't think they would. As far as I know, AT don't try to identify implicit sections. They might communicate something for the `section` element (I have no idea TBH, but they might), but I doubt they would do something with "implicit sections". They just list the headings and what levels they have. That structure is entirely separate from the landmark / region structure. It doesn't get combined in some way like you're trying to do here. These just two different structures that AT can communicate.

The real question is; should AT users be aware that (in my example) that bottom paragraph doesn't belong to the h2? That's a 1.3.1 question. If 1.3.1 required this, I would have an issue with this assumption. But I don't think it does. And yes, that means AT users lose out on some information that is visually evident, but isn't programatic. That is also true for not using landmarks, and most WCAG auditors don't consider that a requirement of 1.3.1 either.

Jym77: Hum, I've played a bit with your CodePen and NVDA and I think I understand what you mean (I have little experience with AT usage…)

AT don't really care where Sections end. Only where they begin. AT provide a list of heading and a way to jump to a heading but doesn't really have a "read until end of section (next heading)" thingie. Similarly, AT provide a list of landmarks and a way to jump to it, but not a "read content of that landmark thingie".

The heading and landmark structures are kept separate by AT (unlike the outline algorithm) and maybe we should keep it separate also for this rule.

---
A problem with this rule is that it is very much concerned by where the section ends…
Building on your codepen:
```
<h1>Fruits</h1>
<div class="row">
  <div class="cell"><h2>Apples</h2>Apples are cool.</div>
  <div class="cell"><h2>Pears</h2>Pears are cool.</div>
</div>
Other fruits also exist.
```
The result on this rule in this example pretty much depends on where the last sentence is. If it is part of the last `h2` (as pure heading structure tell), then this fails the rule because "Pears" is not descriptive of "Pears are cool. Other fruits also exist". But if it is not part of the last `h2`, then this pass the rule…
(same question if the `div` are replaced by `section` and we only look at the heading structure to determine the sections of content…)

=> Would you say that it fails 1.3.1? (I'd say yes because the fact that the last sentence is not in any `h2` section is not communicated programmatically)
=> Would you say that this example fails 2.4.6? (It seems to fail it for AT users but not for sighted users. Obviously related to the fact that it is failing 1.3.1 so different users perceive different structure in the page. Does that mean we should split the rule? Or have an assumption that the page passe 1.3.1?)
=> What should we (and the "heading is descriptive" rule) do about this example?

---
What about having the implicit/explicit sections (more or less as defined here) and an assumption  in the line of "if both are used, they do not overlap in weird ways" rather than "elements are used according to semantics"?
Then something like
```html
<section>
  <h1>Fruits</h1>
  <section><h2>Apples</h2>Apples are cool.</section>
  <section><h2>Pears</h2>Pears are cool.</section>
Other fruits also exist.
</section>
```
would fail the assumption (because the "Pears" implicit section overlaps in weird ways with the explicit sections) and we don't care about what we output for it…
(the same example with `div` would still be poorly handled, however)

And maybe this rule should target only the implicit sections…

ShadowBB: > 
> 
> I strongly disagree with you Brian. Take this example:
> https://codepen.io/wilcofiers/pen/abOvWRB
> 
> It is visually clear that the bottom paragraph is not intended to be part of the h2 subsection. This kind of a "card" design pattern is very common. As for why I think we can make this rule about programatic headings, but not about programatic sections, that's because heading markup is required by WCAG. I don't think WCAG requires the same for sections, nor should it, because there's no commonly agreed upon definition of what that even is. That's why you couldn't just use an existing definition for this proposal.

I don't see how that is contradicting anything I said. I agree with this message.

WilcoFiers:
Oops, wrong button.

pages/glossary/visual-section-of-content.md
```diff
@@ -0,0 +1,16 @@
...
+key: visual-section-of-content
+unambiguous: true
+objective: false
+---
+
+Something in the line of "a subtree rooted at an element with distinguishing style" using https://github.com/act-rules/act-rules.github.io/blob/896b28caeb240171d3b2db626506a7c6034ce20b/pages/glossary/distinguishing-styles.md
+
++note that these only make sense for sighted users and users of AT need to rely on something else.
+
+Problem: the def linked above is about distinguishing between two elements. This def should be about the element being distinguishable from the surrounding.
```
WilcoFiers: I don't think distinguishing style works for this. What I would suggest is something like this:

```md
Section of the content

A section of the content is a group of elements, visually presented together, usually to show
they have a common purpose. A number of characteristics should be considered when deciding if
two elements are in the same section of content:

1. **boundaries**: elements separated by borders, box shadow or by having different backgrounds
are often not part of the same section of the content

2. **layout**: elements in a different column or row are often not part of the same section of
the content

3. **whitespace and negative space**: elements separated by spacing that is greater than the
default spacing between paragraphs are often not part of the same section of content

#### Subsections

Sections are often nested, creating a visual hierarchy in the page. In addition to the first
three indicators, the following presentational mechanisms are often used to indicate which
elements belong in the same subsection of the content:

1. **Headings**: If the first element in a section is presented as a heading, the section only
includes one such heading. The next element that is presented as a heading of that same style,
or of a style where the heading has a larger font size, or the same font size and a greater
font weight is the start of another section.

2. **Text styling**: Elements with a distinct text style are often a subsection of the content.
Text styles include such things as color, font, size, weight, use of shadow or outline, etc. 
Text presented as a heading is not a subsection of content.

3. **Indentation**: Elements with a greater indentation are often a subsection of the content.

4. **Separators**: Decorative non-text content is often used to indictate that a section can be 
divided into subsections. Elements in the same section, separated by this decorative non-text 
content are often not part of the same subsection. Separators often take the form of a line or 
simple graphic.
```

pages/glossary/section-of-content.md
```diff
@@ -1,15 +1,12 @@
 ---
 title: Section of content
 key: section-of-content
-unambiguous: false
+unambiguous: true
 objective: false
 ---
 
-A _section of content_ is a distinct part or subdivision of a document.
+A _section of content_ is either an [explicit section of content][], an [implicit section of content][] or a [visual section of content][].
```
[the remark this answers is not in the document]
Jym77: Obvious examples would be headings that are either not visible (`<h1 style="position: -999px">`) or only marked by their role (`<div role="heading">`).

These would be perceived as headings by AT users (e.g., listed in the list of headings, and can be jumped to with headings navigation) and thus can pass 2.4.6, but wouldn't define any kind of "visual section of content".

pages/glossary/visual-section-of-content.md
```diff
@@ -0,0 +1,41 @@
...
+- **separated** a visual section of content must be separated from surrounding content by one or more of the following visual cues:
+  1. **boundaries**: it has borders, box shadow, or different backgrounds;
+  2. **layout**: it is displayed in separate column or row;
+  3. **indentation**: it has a greater indentation than surrounding content;
+  4. **separators**: there is [decorative][] non-text content before or after it;
+  5. **whitespace separators**: the spacing before or after it is greater than the default spacing between paragraphs;
+  6. **headings**: its first [text][] content has larger font size or greater font weight the rest of the text;
+  7. **Text styling**: its [text][] has a styling different from the rest of the text (text color, font, size, weight, shadow, outline, …)
+
+**Note:** The precise splitting of a document into its visual sections of content is a subjective matter. For example, [decorative][] lines can be used to let the text "breath" without necessarily creating a new visual section of content. Each document can use any number of the visual cues to separate its sections of content.
```
carlosapaduarte, line 21:
```suggestion
**Note:** The precise splitting of a document into its visual sections of content is a subjective matter. For example, [decorative][] lines can be used to let the text "breath" without necessarily creating a new visual section of content. Each document can use any number of visual cues to separate its sections of content.
```

pages/glossary/visual-section-of-content.md
```diff
@@ -0,0 +1,41 @@
...
+
+**Note:** Headings elements (`h1`-`h6`) are usually [rendered][rendering of sections and headings] with a larger and heavier font than normal text. When this is the case, they do fulfill the **headings** case of the **separated** condition, and they define visual sections of content that match the [implicit ones][implicit section of content].
```
carlosapaduarte, line 23:
```suggestion
**Note:** Heading elements (`h1`-`h6`) are usually [rendered][rendering of sections and headings] with a larger and heavier font than normal text. When this is the case, they do fulfill the **headings** case of the **separated** condition, and they define visual sections of content that match the [implicit ones][implicit section of content].
```

pages/glossary/implicit-section-of-content.md
```diff
@@ -0,0 +1,37 @@
...
+  - 'https://www.w3.org/TR/html52/common-idioms-without-dedicated-elements.html#common-idioms-without-dedicated-elements'
+---
+
+An _implicit section of content_ is all the nodes of the [flat tree][], in tree order, between an initial element with a [semantic role][] of [`heading`][heading] and the next element (in tree order) which has both a [semantic role][] of [`heading`][heading] and an [`aria-level`][aria-level] property lower or equal to that of the initial [`heading`][heading] element.
+
+If a node inside an implicit section of content is a [browsing context container][], then the full content of the associated [nested browsing context][] is also included in this implicit section of content.
+
+The heading defining an implicit section of content is always [associated][] with it.
+
+**Note:** Heading level indicates nesting of subsections, the lower the level, the higher the rank of its section. Thus, a level 3 heading (for example, a `h3` element) defines an implicit section of content that stop at the next heading of level 1, 2, or 3; but includes subsections defined by headings of level 4, 5, and 6.
```
carlosapaduarte, line 18:
```suggestion
**Note:** Heading level indicates nesting of subsections, the lower the level, the higher the rank of its section. Thus, a level 3 heading (for example, a `h3` element) defines an implicit section of content that stops at the next heading of level 1, 2, or 3; but includes subsections defined by headings of level 4, 5, and 6.
```

pages/glossary/explicit-section-of-content.md
```diff
@@ -0,0 +1,37 @@
...
+
+Explicit sections of content may have zero or one heading [associated][] with them.
+
+**Note:** The `article` role (inheriting from `document`) is included in this definition because the `article` element is [sectioning content][] used by the [algorithm for creating an outline][outline algorithm].
+
+**Note:** The roles considered here are close but different from the [implicit roles][] of [sectioning content][] and [sectioning root][] elements. Notably, the `body` element itself does not define an explicit section of content.
+
+**Note:** When one of these elements is the single child of a similar element, they form a single explicit section of content, rather than one each, because they have the exact same content.
+
+**Note:** a subtree rooted at a node is this node and all its descendants.
```
carlosapaduarte, line 22:
```suggestion
**Note:** A subtree rooted at a node is this node and all its descendants.
```

pages/glossary/explicit-section-of-content.md
```diff
@@ -0,0 +1,37 @@
...
+
+If a descendant of the root of an explicit section of content is a [browsing context container][], then the full content of the associated [nested browsing context][] is also included in this explicit section of content.
+
+Explicit sections of content may have zero or one heading [associated][] with them.
+
+**Note:** The `article` role (inheriting from `document`) is included in this definition because the `article` element is [sectioning content][] used by the [algorithm for creating an outline][outline algorithm].
+
+**Note:** The roles considered here are close but different from the [implicit roles][] of [sectioning content][] and [sectioning root][] elements. Notably, the `body` element itself does not define an explicit section of content.
+
+**Note:** When one of these elements is the single child of a similar element, they form a single explicit section of content, rather than one each, because they have the exact same content.
```
carlosapaduarte, line 20: According to the definition, this happens not because they have the same content, but because the child node is not a section of content (because it is the single child of an element with sectioning role)

pages/glossary/heading-section-association.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Association between headings and sections
+key: heading-section-association
+unambiguous: true
+objective: false
+---
+
+Each [section of content][] is _associated_ with the first element (in tree order) with a [semantic role][] of `heading` within this section of content, if any.
+
+**Note:** [Implicit sections of content][implicit section of content] are always associated with the element defining them. [Explicit][explicit section of content][] and [visual sections of content][visual section of content] are associated with 0 or 1 heading.
```
carlosapaduarte, line 10:
```suggestion
**Note:** [Implicit sections of content][implicit section of content] are always associated with the element defining them. [Explicit][explicit section of content] and [visual sections of content][visual section of content] are associated with 0 or 1 heading.
```

pages/glossary/explicit-section-of-content.md
Jym77: Yes. This note is here to explain why a single child is ignored, I.e. "it's OK to ignore the single child of an element with sectioning role because otherwise there would be two sections with the exact same content and that wouldn't make sense".

I tried to turn the note around to make that intention clearer.

pages/glossary/visual-section-of-content.md
```diff
@@ -0,0 +1,41 @@
...
+key: visual-section-of-content
+unambiguous: true
+objective: false
+---
+
+A _visual section of content_ is a distinct part of the [visible][] content of a document which:
+
+- **uninterrupted** all the content of a visual section of content must be rendered together, without being interrupted by content from another visual section of content, except if it is fully included in the visual section of content (subsection); and
+- **large enough** a visual section of content must contain at least all the content of one [paragraph][], or at least all the content of one [embedded content][] element; and
+- **separated** a visual section of content must be separated from surrounding content by one or more of the following visual cues:
```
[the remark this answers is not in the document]
Jym77: Good point…
I'm not sure what is the determining characteristic that makes us see that as different sections. We could go further with
```html
<p>1. Foo. Lorem…</p>
<p>Ipsum…</p>
…
<p>2.  Bar. Lorem…</p>
<p>Ispum…</p>
…
```

I am not sure we can really define unambiguously what makes these sections of content…

I guess we have to live with the fact that our "visual sections of content" are not perfect. As long as this results in us having a bit too many inapplicable cases, this is OK.

It would be a problem, for example, if we want to create a rule like "every content is included in a section of content" (but then maybe such a rule wouldn't make much sense for visual sections of content anyway…)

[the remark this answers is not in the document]
carlosapaduarte: I think what makes us see different sections in there is a combination of two things:
1 - A paragraph that is a very short sentence between larger paragraphs. In this context (a book) we recognize that as a section heading unless that very short sentence is recognized as part of a dialogue (starting with - ). This is enough for us.
2 - With the first one being enough, this works to reinforce it. The short sentence starts with a number, distinguishing it from the format of the other paragraphs. If we had multiple sentences like this one after the other we would treat that as a list. Because they are separated we treat them as headings (which are part of a special list - the table of contents).

As I said before, this does not impact the rule, so I'm okay with it.

_rules/heading-descriptive-b49b2e.md
```diff
@@ -29,17 +29,19 @@ This rule applies to any element with the [semantic role][] of heading that is e
 
 ## Expectation
 
-Each target element describes the topic or purpose of its [section of the content][].
+Each target element describes the topic or purpose of any [section of content][] [associated][] with it.
```
WilcoFiers, line 32: I don't think this idea of there being three types of sections is correct. Nor that all of them need to be described by the first semantic heading inside them. Granted, WCAG is open-ended as to what the heading is actually supposed to describe, but this seems like a stretch. I've never tested this way, I know Deque doesn't test this way, Trusted Tester v5 doesn't seem to test this way... I'm curious to know if anyone does test headings like this.

This is inconsistent with how I think web accessibility is supposed to work. The purpose of semantics is to describe the structure this is conveyed through how the content is presented. The semantic structure isn't intended to be its own independent information structure. The way I understand this, the heading isn't supposed to describe the "programatic section", it is supposed to describe whatever is visually presented as a section, and so as long as the programatic section matches the visual section, the heading will also describe the programatic section. When the visuals don't match the semantics, 1.3.1 comes into play, where it's decided if the difference is significant enough to be non-conformant.

If you want more opinions on this, we can put it on the agenda for the next CG call.

pages/glossary/heading-section-association.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Association between headings and sections
+key: heading-section-association
+unambiguous: true
+objective: false
+---
+
+Each [section of content][] is _associated_ with the first element (in tree order) with a [semantic role][] of `heading` within this section of content, if any.
```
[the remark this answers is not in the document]
Jym77: Technically, it's the rule making these assumptions, not the def 😉 

1. The rule has always only targeted elements with a role of `heading` and has always had a note that because of this it is a partial check for 2.4.6. Given that this does not result in any case failing the rule when they pass they SC, I don't think we really need to change that.

2. Yes. The "heading in the middle" is really a weird thing. That comes from the outline algorithm in HTML and I do feel uncomfortable about it without really knowing what to do about it. I feel that defining "visual heading" to work for the "visual section of content" is going to be annoying and make it even more needed to split the rule in two.

PR #1133, opened 2020-01-30
Clarify acc support and assumption for element-lang-valid

WilcoFiers:
Noticed some stuff that can be clarified on this rule.

Need for Final Call: 1 week

PR #1162, opened 2020-02-07
Copy new lang assumption to all languages rules

Jym77:
There was a generic feedback in the AG that the assumption in languages rule is unclear. We add some good discussion in #1153 and figured out something which is hopefully better. This copies the new assumption to all the language rules.

I am not sure the new assumption is that good for the "element has valid lang" rule. Input are welcome on this…

Need for Final Call:
This will not require a Final Call (changing an assumption)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/element-lang-valid-de46e4.md
```diff
@@ -37,7 +37,7 @@ For each test target, the `lang` attribute has a [valid language subtag](#valid-
 
 ## Assumptions
 
-This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+The language of the page can be set by other methods than the `lang` attribute, for example using HTTP headers or the `meta` element. These methods are not supported by all assistive technologies. This rule assumes that these other methods are insufficient to satisfying [Success Criterion 3.1.2: Language of Parts](https://www.w3.org/TR/WCAG21/#language-of-parts).
```
dd8: Is this relevant to language-of-parts? The other methods are all at the page level (HTTP headers, `meta` element in header, default browser language)

_rules/html-page-lang-b5c3f8.md
```diff
@@ -38,7 +38,7 @@ Each test target has a `lang` attribute that is neither empty ("") nor only [ASC
 
 ## Assumptions
 
-This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+The language of the page can be set by other methods than the `lang` attribute, for example using HTTP headers or the `meta` element. These methods are not supported by all assistive technologies. This rule assumes that these other methods are insufficient to satisfying [Success Criterion 3.1.1: Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page).
```
dd8: Looks good

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -38,7 +38,7 @@ For each test target, the values of the [primary language subtags](https://tools
 
 ## Assumptions
 
-This rule assumes that the presence of a `lang` attribute is being used to comply to WCAG. This rule doesn't test if the attribute is needed to comply to WCAG.
+The language of the page can be set by other methods than the `lang` attribute, for example using HTTP headers or the `meta` element. These methods are not supported by all assistive technologies. This rule assumes that these other methods are insufficient to satisfying [Success Criterion 3.1.1: Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page).
```
dd8: Looks good

_rules/element-lang-valid-de46e4.md
dd8: If there's no `lang` attribute on the element, then language of page is relevant since it affects the inherited language

EmmaJP: Perhaps this version simply needs a short sentence at the start to be clear on that relevancy. I don't fully understand the inner browser workings around language determination, but maybe something like:

"If an element lang attribute does not validate, it will default to the language of the page."

Jym77: Yes. I was really not sure about the relevance of this version of the assumption for the "language of parts" rule, but decided to get the discussion going 😄

carlosapaduarte: I agree this only makes sense if prefaced by what @EmmaJP said. Otherwise, it would not be clear how the assumption relates to "Language of Parts".

dd8: If the language is invalid the behaviour is pretty random:

- something it's the inherited lang
- something it's the page lang 
- something it's the system lang

Basically invalid lang means the screen reader picks a language at random (aka undefined behaviour). Could say something like:

"If an element lang attribute does not validate, behaviour is undefined, and content is read in a random language."

WilcoFiers:
```suggestion
The language of the page can be set by other methods than the `lang` attribute; using HTTP headers or the `meta` element. These methods are not known to be supported by assistive technologies. This rule assumes that these other methods are insufficient to satisfying [Success Criterion 3.1.2: Language of Parts](https://www.w3.org/TR/WCAG21/#language-of-parts).
```

Jym77: There was already a new assumption for "language of element" worked in #1133 that was merged today. So I just kept it 😇

_rules/html-page-lang-b5c3f8.md
WilcoFiers: Still, these aren't known to be supported by **any** AT. Saying "are not supported by all assistive technologies" implies that it can be used. If we can't name a single one that supports it, we should be clear about that. Otherwise this should be an accessibility support note.

Jym77: According to tests done by @dd8 in https://github.com/act-rules/act-rules.github.io/issues/668#issuecomment-534606384 there methods are exposed to the accessibility tree by **some** browsers, so it will be available to **some** combinations of UA+AT.

PR #1167, opened 2020-02-12
Fix syntax of boolean attributes

### LINKED ISSUE #1160: 'audio or video has no audio that plays automatically' failed examples aren't failures
'audio or video has no audio that plays automatically' failed examples 1 and and 2 are supposed to autoplay:

https://act-rules.github.io/rules/80f0bf#failed-example-1
https://act-rules.github.io/rules/80f0bf#failed-example-2

https://act-rules.github.io/testcases/80f0bf/104b6e5d9dda4eb31d06dab80f7d8d89c858f151.html
https://act-rules.github.io/testcases/80f0bf/593e65fb0017189463401a5030980c3bc6d5d824.html

Neither example autoplays in Safari, Chrome, Firefox or IE11 so aren't failures.

The W3 validator reports this:

> Bad value true for attribute autoplay on element audio
> `<audio src="/test-assets/moon-audio/moon-speech.mp3" autoplay="true">`

>Bad value true for attribute autoplay on element video.
>`<video autoplay="true">`

Jym77:
HTML boolean attributes are set simply by their presence and explicit `="true"` is bad…

Closes issue(s):

- closes #1160 

Need for Final Call:
This will require a 1 week Final Call (changing most examples)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
Final Call ends on February 20.

adilsofficial:
Applicability section for these rules should be updated:
- an `autoplay` attribute that is equal to true, and
`autoplay` attribute is not explicitly required to be equal to true as already fixed in this PR. Presence of this attribute in `audio` or `<video>` tag  is enough for being `true`.
- both `paused` and `muted` attributes equal to false, and
The default value of both `paused` and `muted` attributes is `false`. If not presented it will be considered as `false`.
These sentences should be updated/fixed according to the examples fixed in this PR.

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -107,7 +107,7 @@ The `video` element automatically plays some audio for more than 3 seconds.
 The `video` element is `muted`.
```
adilsofficial: This test case appears in other rule i.e. 80f0bf but the description is slightly different here. For example description in inapplicable example 1 of ACT-R 80f0bf is 'The `<video>` element audio autoplays for longer than 3 seconds, but is `muted`.'

_rules/auto-play-audio-has-control-mechanism-4c31df.md
```diff
@@ -251,7 +251,7 @@ The `<video>` has a [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) to
 The `<video>` element is `muted`.
```
adilsofficial: This test case appears in other rule i.e. 80f0bf but the description is slightly different here. For example description in inapplicable example 1 of ACT-R 80f0bf is 'The `<video>` element audio autoplays for longer than 3 seconds, but is `muted`.'

Jym77:
> Applicability section for these rules should be updated:
> 
> * an `autoplay` attribute that is equal to true, and
>   `autoplay` attribute is not explicitly required to be equal to true as already fixed in this PR. Presence of this attribute in `audio` or `<video>` tag  is enough for being `true`.
> * both `paused` and `muted` attributes equal to false, and
>   The default value of both `paused` and `muted` attributes is `false`. If not presented it will be considered as `false`.
>   These sentences should be updated/fixed according to the examples fixed in this PR.

Very good catch. Thank you.

I've updated the applicabilities to use a correct formulation (the value of the attribute can be true in different ways: `autoplay`, `autoplay=""`, `autoplay="autoplay"`, `autoplay="AUTOplay"`, …: https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute).

Jym77:

PR #1171, opened 2020-02-14
New rule: Zoomed text node is not clipped with CSS overflow

WilcoFiers:
Early draft for a 1.4.4 resize test rule. Please give feedback on the approach I'm taking here, let me know if you think this is the right direction.

Need for Final Call: 2 weeks


pages/glossary/display-size.md
```diff
@@ -0,0 +1,17 @@
...
+key: display-size
+unambiguous: true
+objective: true
+---
+
+The display size is the width and height at which a page is rendered. The display size is equal to the [clientWidth][] and [clientHeight][] of the [document element][] of the [top-level browsing context][].
+
+**Note**: The display size is not to be confused with the "resolution" of the operating system. Often a browser will be a single window in the operating system, with a width and height different from the resolution of the operating system. Often browsers also include additional user interface components, such as a URL bar, tab bar, and a bookmarks bar. None of these are included in the display size. In full screen mode the display size might be the same as the resolution of the operating system.
+
+**Note**: The display size includes, if rendered, all scrollbars.
```
kasperisager: It however doesn't include the viewport scrollbars as it's defined in terms of `clientWidth` and `clientHeight` which exclude the size of scrollbars.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+- **text-overflow**: A [computed][] [white-space][] of `nowrap`, and a [computed][] [text-overflow][] that is not `clip`; or
+
+- **line wrapping**: A [computed][] [line-height][] equal to or greater than its
+  height of its [bounding box][], or in case of a [computed][] [overflow][] of `clip`, its [content box][].
+
+## Assumptions
+
+If any of the following assumptions is true, failing this rule may not result in a failure of [success criterion 1.4.4 Resize text](https://www.w3.org/TR/WCAG21/#resize-text):
+
+- There is no other mechanism for resizing text available on the page, that can be used to resize text to 200% without loss of information or functionality. This includes font resizing in the browser, or a javascript mechanism of resizing in the page.
```
kasperisager: > This includes font resizing in the browser [...]

At first, I was a little hesitant about this bit. On second thought, I do like what this implies though; if a user cannot zoom to 200% without loss of information, it shouldn't really matter that they might instead be able to bump the base font size to 200% and not lose information.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+- [Understanding Success Criterion 1.4.4: Resize text](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html)
+- [G42: Using a technology that has commonly-available user agents that support zoom](https://www.w3.org/WAI/WCAG21/Techniques/general/G142)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This [text node][] is fully [visible][] at a [display size][] of 640 by 512.
```
adilsofficial, line 68: Improve example description.
```suggestion
This [text node][] is fully [visible][] and not [clipped by overflow][clipped] of an [ancestor][] in the [flat tree][] at a [display size][] of 640 by 512.
```

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+	Once upon a midnight dreary, while I pondered, weak and weary,<br />
+	Over many a quaint and curious volume of forgotten lore.<br />
+	While I nodded, nearly napping, suddenly there came a tapping,<br />
+	As of some one gently rapping, rapping at my chamber door.<br />
+	“’Tis some visitor,” I muttered, “tapping at my chamber door.<br />
+	Only this and nothing more.”
+</div>
+```
+
+#### Passed Example 2
```
adilsofficial, line 81: I seems that this is a failed example even though there is a link to full poem but code it self failed the test rule. On the other hand there is not relation between the text node and link i.e. the link actually belong to the text node above.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+<div style="white-space: nowrap; overflow: hidden; text-overflow: ellipsis;">
+	Once upon a midnight dreary, while I pondered, weak and weary, Over many a quaint and curious volume of forgotten
+	lore. While I nodded, nearly napping, suddenly there came a tapping.
+</div>
+<a href="/test-assets/59br37/poem.html">Full text of the poem</a>
+```
+
+#### Passed Example 3
+
+This [text node][] is [clipped][] using `text-overflow: ellipsis` at a [display size][] of 640 by 512. A link to a full version of the poem is also provided.
```
adilsofficial, line 95: This description is duplicate of Passed example 2 and `text-overflow: ellipsis` is not presented in the  example code.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+  assets:
+    - The Raven, poem by Edgar Allan Poe
+---
+
+## Applicability
+
+The rule applies to any [text node][] for which all of the following is true when in a [display size][] of 640 by 512:
+
+- The [text node][] is [visible][]; and
+- The [text node][] has an HTML element as a [parent][] in the [flat tree][]; and
```
jkodu, line 28: Combining point 4 with this, will make it easier to digest the applicability around parent nodes.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+- The [text node][] has an [ancestor][] in the [flat tree][] with the a [computed][] [overflow-x][overflow] or [overflow-y][overflow] of `none` or `clip`; and
```
jkodu, line 29: Does the `getComputedStyle(node).overflow` return `clip` or `none`? The `text-overflow` does return `clip`, but not `overflow`.

pages/glossary/display-size.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Display size
+key: display-size
+unambiguous: true
+objective: true
+---
+
+The display size is the width and height at which a page is rendered. The display size is equal to the [innerWidth][] and [innerHeight][] of the [window][] of the [top-level browsing context][].
```
jkodu, line 8: If referring to inner width & height of the window, perhaps using the keyword, viewport size would be a better terminology.

_rules/zoom-text-no-overflow-clipping-59br37.md
kasperisager: `overflow-{x,y}` computes as follows:

> as specified, except with `visible`/`clip` computing to `auto`/`hidden` (respectively) if one of `overflow-x` or `overflow-y` is neither `visible` nor `clip`.

https://drafts.csswg.org/css-overflow-3/#propdef-overflow-x

`none` isn't a valid value for `overflow-{x,y}` so that should probably be fixed though.

pages/glossary/display-size.md
kasperisager: That's a good point. The term "display" can really mean many things; viewport, monitor, the `display` property. The term "viewport", however, has fewer meanings.

WilcoFiers: I'll rename to viewport. Fair point.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
+---
+id: 59br37
+name: Zoomed text node is not clipped with CSS overflow
+rule_type: atomic
+description: |
+  This rule checks that text nodes are not unintentionally clipped by overflow, when a page is zoomed to 200% on 1280 by 1024 display;
```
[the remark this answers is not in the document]
WilcoFiers: That's noted in the assumption. You're right it's an odd one. It's based on what was used as the default res for SC 1.4.11.

_rules/zoom-text-no-overflow-clipping-59br37.md
WilcoFiers: I think those are different cases. I'd prefer to keep them separate.

_rules/zoom-text-no-overflow-clipping-59br37.md
WilcoFiers: good catch. That obviously should have been "hidden" instead of "none.

_rules/zoom-text-no-overflow-clipping-59br37.md
WilcoFiers: I don't think that is correct. The first exception in the expectation is true here:

> - **text-overflow**: A [computed][] [white-space][] of `nowrap`, and a [computed][] [text-overflow][] that is not `clip`; or

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+```html
+<div style="overflow: hidden; height: 1.5em;">
+	Once upon a midnight dreary, while I pondered, weak and weary, Over many a quaint and curious volume of forgotten
+	lore. While I nodded, nearly napping, suddenly there came a tapping.
+</div>
+```
+
+#### Failed Example 2
+
+This [text node][] is [clipped][] because its height is relative to the viewport height, leaving insufficient space for the page to be zoomed to 200% at a [viewport size][] of 640 by 512.
```
carlosapaduarte, line 141:
```suggestion
This [text node][] is [clipped][] because its height is relative to the viewport height, leaving insufficient space for the page to be zoomed to 200% at a [viewport size][] of 1280 by 1024.
```

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,247 @@
...
+- [Understanding Success Criterion 1.4.4: Resize text](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html)
+- [G42: Using a technology that has commonly-available user agents that support zoom](https://www.w3.org/WAI/WCAG21/Techniques/general/G142)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This [text node][] is fully [visible][] at a [viewport size][] of 640 by 512.
```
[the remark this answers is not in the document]
WilcoFiers: Can you clarify what you mean? The browser's default stylesheet is included in the CSS spec. Obviously if a browser doesn't follow the CSS specs, its results will be different, but all browsers I know of use this default; at least where font-size is concerned they do.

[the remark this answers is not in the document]
kasperisager: The base font size is not included in the suggested user agent styles of the HTML specification, no: https://html.spec.whatwg.org/multipage/rendering.html. It's true that modern browsers default to a base font size of 16 pixels, but this is entirely configurable:

image

Above is a screenshot of the font settings in Chrome.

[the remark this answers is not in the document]
carlosapaduarte: As @kasperisager points out, the spec does not specify the base font size. Any user can configure it. When I set the base font size to 24 pixels, the text in the first passed example was clipped (so, the example should fail in that instance). This can be easily solved by setting the font-size in the div's style. 

Not relying on the default styles is similar to what we have in other rules (e.g. resetting the style of every link to make sure that default styles where being used for links when looking at distinguishing styles). 

I would be happy with a solution that sets the font-size in this example (this issue does not impact the other examples).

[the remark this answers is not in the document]
WilcoFiers: I don't think it's necessary, but I don't mind having font-size added to the examples, so I've done that.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,249 @@
...
+
+```html
+<svg>
+	<text x="0" y="15">I love SVG!</text>
+</svg>
+```
+
+#### Inapplicable Example 3
+
+This [text node][] has an [ancestor][] with `overflow: auto`.
```
Jym77, line 198: Comparing with Passed Example 4, this one is inapplicable due to "no ancestor with `overflow: hidden` or `clip`.
Both this one and Passed Example 4 have an ancestor with `overflow: auto`, it is not the reason why the example is inapplicable.

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -0,0 +1,249 @@
...
+---
+
+## Applicability
+
+The rule applies to any [text node][] for which all of the following is true when in a [viewport size][] of 640 by 512:
+
+- The [text node][] is [visible][]; and
+- The [text node][] has an HTML element as a [parent][] in the [flat tree][]; and
+- The [text node][] has an [ancestor][] in the [flat tree][] with a [computed][] [overflow-x][overflow] or [overflow-y][overflow] of `hidden` or `clip`; and
+- The [text node][] does not have an [ancestor][] in the [flat tree][] with an `aria-hidden` attribute set to `true`
```
Jym77, line 30: I don't understand why this condition is needed. The SC and the rule seems to be concerned with sighted users more than AT users, and they will essentially make no difference between `aria-hidden` subtrees and the rest.

Why is it "OK" to clip a text if it is `aria-hidden`?

WilcoFiers: This is to avoid false positives. I'll rephrase the second assumption to make this clearer.

_rules/zoom-text-no-overflow-clipping-59br37.md
WilcoFiers: I don't understand this comment. This is inapplicable because it has no ancestor with `overflow: hidden`, passed example 4 does.

Jym77: > This is inapplicable because it has no ancestor with `overflow: hidden`, passed example 4 does.

Yes, precisely.
The description here only tells about having "an ancestor with `overflow: auto`", suggesting that this is the reason why the example is inapplicable.
As you point out, the reason why this example is inapplicable is that "it has no ancestor with `overflow: hidden`" (or `clip` to keep the applicability wording). This is what the example description should state.
```suggestion
This [text node][] has no [ancestor][] with `overflow: hidden` or `clip`.
```

_rules/zoom-text-no-overflow-clipping-59br37.md
Jym77: I fail to find a false positive caused by `aria-hidden="true"` 😞 
Can you provide me one?

WilcoFiers: For example a site that uses ASCII art for something decorative. If it's decorative, SC 1.1.1 requires it be marked up with aria-hidden=true, so that it can be ignored by AT. Since it's decorative, it matter if it's clipped, so keyboard accessibility isn't necessary either.

PR #1176, opened 2020-02-17
Rule: Lists follow content model (a73be2)

jeeyyy:
A rule proposed for year 3. Feedback & reviews welcome.

Closes issue(s):
- NA

Need for Final Call:
- Yes (2 Weeks)


_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+
+The rule applies to every `ul` , `ol` , and `dl` HTML elements where the [semantic role][] is the same as the [implicit semantic role][].
+
+## Expectation
+
+Each target element:
+
+- is categorized to follow the [flow content model][], and;
+- that it's contents are its [descendants][] in the DOM, and;
+- has at least one node in its contents that is a [palpable content][], which does not have the `hidden` attribute specified.
```
Jym77, line 30: It can make sense to split in 3 separate expectations. It's going to make each of them smaller to swallow, and it's going to make error reporting easier (as we can pinpoint the expectation that fails, instead of having a single "fail" result).

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+## Applicability
+
+The rule applies to every `ul` , `ol` , and `dl` HTML elements where the [semantic role][] is the same as the [implicit semantic role][].
+
+## Expectation
+
+Each target element:
+
+- is categorized to follow the [flow content model][], and;
+- that it's contents are its [descendants][] in the DOM, and;
```
Jym77, line 29: "descendant in the DOM" means that shadow trees are ignored. Is it intentional?

Jym77, line 29: The way it is currently written, that item makes no grammatical sense:
> Each target element (…) that it's contents are its descendants in the DOM

Jym77, line 29: I usually understand "content" as "web content" (https://www.w3.org/TR/WCAG21/#dfn-content) but here it seems to mean "list content" or something like that.

https://html.spec.whatwg.org/multipage/dom.html#concept-html-contents define content as children in the DOM. We probably should stick to children when talking about the DOM to avoid confusion with "web content".

But then, if content=children, the item makes little sense.

_rules/lists-follow-content-model-a73be2.md
Jym77, line 30: Same remark about "content".

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+rule_type: atomic
+description: |
+  This rule checks that list elements follow content model
+accessibility_requirements:
+  wcag20:1.3.1: # Info and Relationships (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
```
Jym77, line 13: Need "CSS" for `hidden`.
May need "Acessibility tree".

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Applicability
+
+The rule applies to every `ul` , `ol` , and `dl` HTML elements where the [semantic role][] is the same as the [implicit semantic role][].
+
+## Expectation
+
+Each target element:
+
+- is categorized to follow the [flow content model][], and;
```
Jym77, line 28: I'm not sure what this means. `dl`, `ol` and `ul` elements are flow content. 
Can you come up with a Failed Example that would fail that item (and only that item)?

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+- that it's contents are its [descendants][] in the DOM, and;
+- has at least one node in its contents that is a [palpable content][], which does not have the `hidden` attribute specified.
+
```
Jym77, line 31: It's weird that the expectation does not actually refer to https://html.spec.whatwg.org/multipage/dom.html#concept-element-content-model given that this is what the rule is supposed to test.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+- [HTML Specification - Flow content](https://html.spec.whatwg.org/#flow-content)
+- [HTML Specification - Palpable content](https://html.spec.whatwg.org/#palpable-content)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `ul` element contains [palpable content][] `li` element.
```
Jym77, line 52: `li` elements are not palpable content.

The content model for `ul` element does not require palpable content, it requires `li` elements (or script-supporting ones).

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,153 @@
...
+	<li>Bread</li>
+	<span role="listitem">Milk</span>
+</ol>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `ul` element contains [palpable content][] `li` element, that is [hidden][].
```
Jym77, line 87: The element is `aria-hidden`, not `hidden`.
The expectation only talk about `hidden`, not about being in the accessibility tree (`aria-hidden`).

_rules/lists-follow-content-model-a73be2.md
Jym77, line 30: What about inclusion in the accessibility tree (failed example 1)? Visibility (`<ul><li style="position: absolute; top: -999px">Hello</li></ul>`)? Other way of hiding (`display: none`)?

Jym77, line 30: I don't understand why "palpable content" is needed. It is not part of the content model of either `dl`, `ol`, or `ul`.

_rules/lists-follow-content-model-a73be2.md
jkodu: Agree, this expectation can be removed.

_rules/lists-follow-content-model-a73be2.md
jkodu: For comment above and this [comment](), the spec only refers to usage of the hidden attribute & not other ways of hiding the content. See spec - https://html.spec.whatwg.org/#palpable-content

jkodu:
Thanks @ShadowBB and @Jym77 for the reviews. I have made some change. Please review again.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,135 @@
...
+    - Jey Nandakumar
+---
+
+## Applicability
+
+The rule applies to every `ul` , `ol` , and `dl` HTML elements where the [semantic role][] is the same as the [implicit semantic role][].
+
+## Expectation
+
+Each test target has at least one [descendant][] in the [flat tree][] that is categorized as [palpable content][], which does not have the [hidden][] attribute specified.
```
carlosapaduarte, line 28: I believe @Jym77 has already addressed what I'm going to say, but I couldn't understand how his comment was resolved... 
This expectation does not follow the content model (which is, having 0 or more `li` or `dt`/`dd`groups). 
I don't think we need to require following the content model, but shouldn't we at least require it to have one or more elements with the semantic role of list item?

On the other hand, this expectation means the following code passes, and I'm not sure that is good:
```html
<ul>
  <p>- item 1</p>
  <p>- item 2</p>
</ul>

File: _rules/lists-follow-content-model-a73be2.md [hunk 9]
```diff
@@ -0,0 +1,135 @@
...
+		<p>Milk</p>
+	</span>
+</ol>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `ul` element contains [palpable content][] `li` element, that is [hidden][].
```
carlosapaduarte, line 88:
```suggestion
This `ul` element contains [palpable content][] that is [hidden][].
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,135 @@
...
+<ul>
+	<li>
+		<p hidden>Brocolli</p>
+	</li>
+</ul>
+```
+
+#### Failed Example 2
+
+This `ol` element does not any [descendant][] of type [palpable content][].
```
carlosapaduarte, line 100:
```suggestion
This `ol` element does not have any [descendant][] of type [palpable content][].
```

carlosapaduarte (dismissed):

_rules/lists-follow-content-model-a73be2.md
Jym77: Yes.

I am not sure what we want the rule to check in the first place…
* That the HTML content model is followed? Then we need `li` and `script` in an `ul` (https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element), not palpable content. (note that the example provided by @carlosapaduarte does **not** follow content model!)
* That the ARIA role model is followed for implicit roles? Then we need the `ul` to contain some `li` (https://www.w3.org/TR/wai-aria-1.1/#list) but they might be empty (no palpable content).
* That there is no empty list? Then yes, requiring palpable content make sense. But it is neither HTML content model nor ARIA role model. And I am not sure this really fails 1.3.1 (an empty list has nothing conveyed by presentation…)

Jym77, line 28: * `hidden` attribute is a very specific way of hiding stuff. Is it intentionally restricted to this way?
* What if the `hidden` attribute (or whatever) is not on the palpable content but on some ancestor? Failed Example 1 could have `hidden` on the `li`, or on a `div` surrounding the `p` (thus the `p` would be palpable without `hidden` attribute). I don't think this would be good.

Jym77 (dismissed):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,135 @@
...
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [HTML Specification - Content model](https://html.spec.whatwg.org/#concept-element-content-model)
+- [HTML Specification - Flow content](https://html.spec.whatwg.org/#flow-content)
+- [HTML Specification - Palpable content](https://html.spec.whatwg.org/#palpable-content)
+
+## Test Cases
```
daniel-montalvo, line 44: As per my comment above, should we be including a WAI-ARIA list in the example?

_rules/lists-follow-content-model-a73be2.md
daniel-montalvo: Maybe "one or more non-empty `li`, or `dd/dt` groups"?

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,135 @@
+---
+id: a73be2
+name: List elements follow content model
+rule_type: atomic
+description: |
+  This rule checks that list elements follow content model
```
daniel-montalvo, line 6: Maybe something like


  This rule checks that lists have one or more `li`, or `dd/dt` groups.

If I'm getting this right.

daniel-montalvo (dismissed):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,135 @@
...
+  - CSS Styling
+  - DOM Tree
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+The rule applies to every `ul` , `ol` , and `dl` HTML elements where the [semantic role][] is the same as the [implicit semantic role][].
```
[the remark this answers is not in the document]
Jym77: The "content model" is something from HTML, not from ARIA. So it only cares about HTML element. Technically no matter what their role, but we might want to keep the role clause in order to flag accessibility issues.

The corresponding thing for ARIA is the "required context role (or owned elements)" for which we also have a rule.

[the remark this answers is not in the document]
daniel-montalvo: Thanks for this @Jym77 

I get that. And I guess it could be argued then that the ARIA example above would be tested using the other rules. For example [ARIA required context role](https://act-rules.github.io/rules/ff89c9), passed examples 1, 2, and 3.

I think we can do more to illustrate this difference, as @carlosapaduarte was suggesting in other comment. Maybe adding "HTML content model" to the existing phrasing, and maybe adding ARIA examples in the inapplicable section to demonstrate that in this rule we are only focusing in HTML native elements.

_rules/lists-follow-content-model-a73be2.md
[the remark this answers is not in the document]
jeeyyy: @carlosapaduarte I have updated the rule such a way that it is more specific than using arbitrary terms like palpable content etc.,

But [content model](https://html.spec.whatwg.org/multipage/dom.html#concept-element-content-model) is referred in background & I want to keep the name/ description generic. Let me know if you think this needs a bit more detailing.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,167 @@
...
+  - CSS Styling
+  - DOM Tree
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+The rule applies to every `ul` , `ol` , and `dl` [HTML elements][] for which
```
WilcoFiers, line 24: This needs to follow the conditional list style.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,167 @@
...
+
+- the [semantic role][] is the same as the [implicit semantic role][]; and
+- is [included in the accessibility tree][]; and
+- is [visible][].
```
WilcoFiers, line 28: I'm not so sure about this. Wouldn't it make more sense for this to be applicable to all lists except those that have display:none or visibility:hidden?

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Expectation 1
+
+Each test target of type [HTML `<ul>` element][] or [HTML `<ol>` element][], has at least one [HTML `<li>` element][] as a [descendant][] in the [flat tree][] which
```
WilcoFiers, line 32: Follow the established writing style please:
- Use proper conditional list formatting
- Don't use angle brackets for elements
- Don't link to HTML elements, they don't need to be disambiguated

Additionally; 
1. Why does a list need to have an `li` element? I get that HTML says so, but I don't see that in WCAG?
2. Why would it need non-empty text? Even HTML doesn't require that?
3. Why can it not be hidden? Why only the hidden attribute, and not some other way to hide it?
4. This leaves unspecified behavior for the dl element. An expectation can't be "inapplicable". It has to be true or false for all test targets. The way this is phrased doesn't allow a true / false answer for dl elements.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,167 @@
...
+
+- has non-empty [text][]; and
+- does not have the [hidden][] attribute.
+
+## Expectation 2
+
+Each test target of type [HTML `<dl>` element][] has at least one name-value group as a [descendant][] in the [flat tree][] which
```
WilcoFiers, line 39: What is a "name-value group", and why is that required?

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Applicability
+
+The rule applies to every `ul` , `ol` , and `dl` [HTML elements][] for which
+
+- the [semantic role][] is the same as the [implicit semantic role][]; and
+- is [included in the accessibility tree][]; and
+- is [visible][].
+
+## Expectation 1
```
WilcoFiers, line 30: Why doesn't this rule fail inappropriate content for list? Seems like this should fail:

```html
<ul>
  <li>Hello</li>
  <span>World</span>
</ul>
```

WilcoFiers (changes requested):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -85,7 +98,7 @@ This `ol` element contains at least one [descendant][] of type [palpable content
 
 #### Failed Example 1
 
-This `ul` element contains [palpable content][] `li` element, that is [hidden][].
+This [HTML `<ul>` element][] contains [HTML `<li>` element][] that is [hidden][].
```
daniel-montalvo, line 101:
```suggestion
This [HTML `<ul>` element][] contains an [HTML `<li>` element][] that is [hidden][].
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -97,7 +110,7 @@ This `ul` element contains [palpable content][] `li` element, that is [hidden][]
 
 #### Failed Example 2
 
-This `ol` element does not any [descendant][] of type [palpable content][].
+This [HTML `<ol>` element][] has [HTML `<li>` element][] with empty [text][] as [descendant][].
```
daniel-montalvo, line 113:
```suggestion
This [HTML `<ol>` element][] has an [HTML `<li>` element][] with empty [text][] as [descendant][].
```

daniel-montalvo (changes requested):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+- [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- [HTML Specification - Content model](https://html.spec.whatwg.org/#concept-element-content-model)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This [HTML `<ul>` element][] has a [descendant][] of type [HTML `<li>` element][].
```
Jym77, line 60: Global: we usually don't link to HTML elements, and don't use brackets around them.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+## Applicability
+
+The rule applies to every `ul`, `ol`, `dl`, and `menu` [HTML elements][] that is [included in the accessibility tree][] and has the same [semantic role][] as its [implicit semantic role][].
+
+## Expectation 1
+
+Each test target only has [child][] [node][] in the [flat tree][] for which one of the following is true:
+
+- the [child][] [element][] follows the [content model][]; or
+- the [child][] has an [explicit role][] which is the [Implicit ARIA semantics]() for an HTML element included in the [content model][] for the test target; or
```
Jym77, line 34: `dl` accept `div` in their content model, and `div` have no role. I'm not sure how this articulate with the expectation.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+    - Jey Nandakumar
+---
+
+## Applicability
+
+The rule applies to every `ul`, `ol`, `dl`, and `menu` [HTML elements][] that is [included in the accessibility tree][] and has the same [semantic role][] as its [implicit semantic role][].
+
+## Expectation 1
+
+Each test target only has [child][] [node][] in the [flat tree][] for which one of the following is true:
```
Jym77, line 31: The content model for `dl` is a group of `dt` + one or more `dd`. That means we can't consider each child of the target individually and need to look at them more globally. 
Unless we want to pass the rule even if the group is not complete (fair enough), but then this differs a bit from the content model and the first bullet in the expectation is not really correct.

Jym77, line 31: I suggest naming the bullets and referring them in examples descriptions ("this matches condition **foo**").

_rules/lists-follow-content-model-a73be2.md
Jym77, line 60: Expectation is about children, not descendants. Which one is correct?

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+</dl>
+```
+
+#### Passed Example 3
+
+This [HTML `<menu>` element][] has a [descendant][] of type [HTML `<li>` element][].
+
+```html
+<menu>
+	<li></li>
```
Jym77, line 86: I'm not sure it's a good idea to have an Passed Example with empty `li`. Sure it passes the rule, but it's still a fairly bad code.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+</menu>
+```
+
+#### Passed Example 4
+
+This [HTML `<ol>` element][] contains at least one [descendant][] of type [HTML `<li>` element][].
```
Jym77, line 92: This description is not correct. Should be in the line of "with explicit role equal to the implicit role of `li`" or something like that.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+```html
+<ol>
+	<span role="listitem">Eggs</span>
+	<span role="listitem">Milk</span>
+</ol>
+```
+
+#### Passed Example 5
+
+This [HTML `<ol>` element][] contains at least one [descendant][] of type [HTML `<li>` element][] and contains a [HTML `<template>` element][] which follows the [content model][].
```
Jym77, line 103: Should mention that `template` is allowed as script-supporting element.

_rules/lists-follow-content-model-a73be2.md
Jym77, line 60: "has a descendant" is not super good. Expectation is about every children ("only has child that …"). Descriptions should reflect that.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+```html
+<ul>
+	<li>Eggs</li>
+	<li>Milk</li>
+</ul>
+```
+
+#### Passed Example 2
+
+This [HTML `<dl>` element][] contains at least one name-value group as [descendant][].
```
Jym77, line 71: "at least one descendant" is even worse than "a descendant" since expectation is about every children.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+	<li>Milk</li>
+	<template>
+		<li>{{item}}</li>
+	</template>
+</ol>
+```
+
+#### Passed Example 6
+
+This [HTML `<dl>` element][] contains at least one name-value group as [descendant][]. That fact that it is placed in a div still follows the [content model][].
```
Jym77, line 117: Technically, the `div` is enough to follow content model. No need to have `dt` and `dd` inside.
Not sure if we want the rule to pass on `<dl><div>foo</div></dl>` (it currently does, I think).

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+
+```html
+<ol>
+	<li></li>
+</ol>
+```
+
+#### Passed Example 8
+
+This [HTML `<dl>` element][] contains at least one name-value group as [descendant][]. The dive elements both have an [explicit role][] which is the [Implicit ARIA semantics]() for an HTML element included in the [content model][].
```
Jym77, line 140:
```suggestion
This [HTML `<dl>` element][] contains at least one name-value group as [descendant][]. The `div` elements both have an [explicit role][] which is the [Implicit ARIA semantics]() for an HTML element included in the [content model][].
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+	<div role="term">Physics</div>
+	<div role="definition">8/10</div>
+</dl>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This [HTML `<ul>` element][] doesn't contain at least one child element.
```
Jym77, line 153: Technically, this is not failing Expectation 1 (universal quantification on the empty set of children is true), but does fail expectation 2 (requiring at least one child in the accessibility tree).

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+
+This [HTML `<dl>` element][] contains at least one name-value group as [descendant][]. The dive elements both have an [explicit role][] which is the [Implicit ARIA semantics]() for an HTML element included in the [content model][].
+
+```html
+<dl>
+	<div role="term">Physics</div>
+	<div role="definition">8/10</div>
+</dl>
+```
+
```
Jym77, line 148: Should there be Passed Example to illustrate the 3rd bullet in Expectation 1?

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+#### Failed Example 3
+
+This [HTML `<ul>` element][] only contains elements with an [explicit role][] that doesn't follow the [content model][]. Menuitems are not valid children of an unordered list.
+
+```html
+<ul>
+	<li role="menuitem">New file</li>
+	<li role="menuitem">Save file</li>
+</ul>
```
Jym77, line 177: Technically, this passes the rule. It satisfies Expectation 1 since each child is matching the condition of the first bullet (it's a `li`, allowed by content model inside a `ul`).

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+
+This [HTML `<dl>` element][] contains at least one name-value group as [descendant][]. That fact that it is placed in a div still follows the [content model][].
+
+```html
+<dl>
+	<div>
+		<dt>Physics</dt>
+		<dd>8/10</dd>
+	</div>
+</dl>
```
Jym77, line 125: Actually, I'm not fully sure this passes Expectation 2.
The only child is the `div` and `div` are not really included in the accessibility tree (they're just containers). But that's a bit unclear.
(see https://act-rules.github.io/rules/ff89c9#failed-example-3 where we needed to add an `aria-label` on the `div` to force its inclusion in the accessibility tree)

Jym77 (changes requested):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - Accessibility tree
+  - CSS Styling
+  - DOM Tree
+acknowledgments:
+  authors:
+		- Brian Bors
+		- Wilco Fiers
```
WilcoFiers, line 20:
```suggestion
    - Brian Bors
    - Wilco Fiers
```

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: Right you are, fixing this.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Applicability
+
+The rule applies to every `ul`, `ol`, `dl`, and `menu` [HTML elements][] that is [included in the accessibility tree][] and has the same [semantic role][] as its [implicit semantic role][].
+
+## Expectation 1
+
+Each test target only has [child][] [node][] in the [flat tree][] for which one of the following is true:
+
+- the [child][] [element][] follows the [content model][]; or
```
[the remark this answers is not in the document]
WilcoFiers: You are right, element is not needed here.

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: You're right, but that isn't allowed by the content model of `div`. I'm going to fix this by including `dl > div` in the applicability.

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: Going to use "hidden state" instead.

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: Will add one using the `hidden` attribute.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -83,60 +79,145 @@ This [HTML `<dl>` element][] contains at least one name-value group as [descenda
...
+
+```html
+<ol>
+	Coming soon!
+</ol>
+```
+
+#### Failed Example 3
+
+This [HTML `<ul>` element][] only contains elements with an [explicit role][] that doesn't follow the [content model][]. Menuitems are not valid children of an unordered list.
```
daniel-montalvo:
```suggestion
This [HTML `<ul>` element][] only contains elements with an [explicit role][] that don't follow the [content model][]. Menuitems are not valid children of an unordered list.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -83,60 +79,145 @@ This [HTML `<dl>` element][] contains at least one name-value group as [descenda
...
+	<li role="menuitem">New file</li>
+	<li role="menuitem">Save file</li>
 </ul>
 ```
 
-#### Failed Example 2
+#### Failed Example 4
 
-This [HTML `<ol>` element][] has [HTML `<li>` element][] with empty [text][] as [descendant][].
+This [HTML `<ol>` element][] only contains elements that doesn't follow the [content model][]. Name-value groups are not valid children of an ordered list.
```
daniel-montalvo:
```suggestion
This [HTML `<ol>` element][] only contains elements that don't follow the [content model][]. Name-value groups are not valid children of an ordered list.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -83,60 +79,145 @@ This [HTML `<dl>` element][] contains at least one name-value group as [descenda
...
 <ol>
-	<li></li>
+	<dt>Physics</dt>
+	<dd>8/10</dd>
 </ol>
 ```
 
+#### Failed Example 5
+
+This [HTML `<dl>` element][] only contains elements that doesn't follow the [content model][]. List items are not valid children of a definition list.
```
daniel-montalvo:
```suggestion
This [HTML `<dl>` element][] only contains elements that don't follow the [content model][]. List items are not valid children of a definition list.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -83,60 +79,145 @@ This [HTML `<dl>` element][] contains at least one name-value group as [descenda
...
+	<li role="listitem">Save file</li>
+</dl>
+```
+
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-This rule is not applicable to `menu` element.
+This rules is not applicable to `ul` elements that are not [included in the accessibility tree][].
```
daniel-montalvo:
```suggestion
This rule is not applicable to `ul` elements that are not [included in the accessibility tree][].
```

daniel-montalvo (dismissed):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+This rule applies to any [HTML elements][] with a [hidden state][] of "false", where the [semantic role][] is the same as the [implicit role][], and for which one of the following is true:
+
+- **lists**: The element is a `ul`, `ol` or `menu`; or
+- **definition lists**: The element is a `dl`, or a `div` that is a [child][] in the [flat tree][] of a `dl`.
+
+## Expectation
+
+For each [child][] in the [flat tree][] of each test target, one of the following is true:
+
+- **implicit role**: the [child][] has no [explicit role][] and follows the [content model][]; or
```
WilcoFiers, line 36:
```suggestion
- **implicit role**: the [child][] follows the [content model][] and has no [explicit role][]; or
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+
+_There are currently no assumptions._
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+All children of `ul`, `ol`, and `menu` elements must be either an element with a [semantic role][] of `listitem`, or a node that is hidden, such as a `script` or `template` element, an HTML comment, of a text node that consists of only [ascii whitespace][].
```
WilcoFiers, line 50:
```suggestion
All children of `ul`, `ol`, and `menu` elements must be either an element with a [semantic role][] of `listitem`, or a node that is hidden, such as a `script` or `template` element, an HTML comment, or a text node that consists of only [ascii whitespace][].
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+
+All children of a `dl` element must be an element with a [semantic role][] of `term` or `definition`, a `div` with no [explicit role][], or a node that is hidden. If the child has a `definition` role, it must have a previous sibling that is a `term`. If the child is a `div` element with no [explicit role][] it is treated as a `term` and `definition` group. The children of the `div` have the same requirement as the `dl` element, except that it does not allow another `div`.
```
WilcoFiers, line 52:
```suggestion
All children of a `dl` element must be an element with a [semantic role][] of `term` or `definition`, a `div` with no [explicit role][], or a node that is hidden. If the child has a `definition` role, it must have a previous sibling that has a `term` role. If the child is a `div` element with no [explicit role][] it is treated as a `term` and `definition` group. The children of the `div` have the same requirement as the `dl` element, except that it does not allow another `div`.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+```html
+<ol>
+	<span role="listitem">Eggs</span>
+	<span role="listitem">Milk</span>
+</ol>
+```
+
+#### Passed Example 6
+
+This `ul` element has `li` and `template` [children][child]. `li` is explicitly allowed by the `ul` [content model], `template` is a script-supported element, which is also allowed by the `ul` content model.
```
WilcoFiers, line 117:
```suggestion
This `ul` element has `li` and `template` [children][child]. `li` is explicitly allowed by the `ul` [content model][], `template` is a script-supported element, which is also allowed by the `ul` content model.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+```html
+<dl>
+	<div role="term">Physics</div>
+	<div role="definition">8/10</div>
+</dl>
+```
+
+#### Passed Example 9
+
+This `dl` element has `div` [child][child] which is allowed by its [content model][]. The `div` has `div` [children][child] with an [explicit role][] of `term` and `definition`. The [content model][] of `div` allows `dt` and `dd` elements which have [implicit ARIA semantics][] of `term` and `definition`.
```
WilcoFiers, line 155:
```suggestion
This `dl` element has `div` [child][child] which is allowed by its [content model][]. The `div` element has `div` [children][child] with an [explicit role][] of `term` and `definition`. The [content model][] of `div` allows `dt` and `dd` elements which have [implicit ARIA semantics][] of `term` and `definition`.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+	<div>
+		<div role="term">Chemistry</div>
+		<div role="definition">7/10</div>
+	</div>
+</dl>
+```
+
+#### Passed Example 10
+
+This `ul` element has a `p` [child][] which is not [included in the accessibility tree][].
```
WilcoFiers, line 168:
```suggestion
This `ul` element has a `p` [child][] which is not [included in the accessibility tree][] because of the `hidden` attribute.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+<ul>
+	<p hidden>Empty list</p>
+</ul>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `ul` element had a text node as a child, which will be [included in the accessibility tree][].
```
WilcoFiers, line 180:
```suggestion
This `ul` element had a text node as a child, which will be [included in the accessibility tree][]. Text nodes are not allowed by the [content model][] of `ul` elements.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+
+```html
+<ul hidden>
+	<li role="menuitem">Eggs</li>
+</ul>
+```
+
+#### Inapplicable Example 2
+
+This `ul` element is not [included in the accessibility tree][].
```
WilcoFiers, line 246:
```suggestion
This `ul` has an [explicit role][] of menu. The `ul` content model does not need to be applied.
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+All children of `ul`, `ol`, and `menu` elements must be either an element with a [semantic role][] of `listitem`, or a node that is hidden, such as a `script` or `template` element, an HTML comment, or a text node that consists of only [ascii whitespace][].
+
+All children of a `dl` element must be an element with a [semantic role][] of `term` or `definition`, a `div` with no [explicit role][], or a node that is hidden. If the child has a `definition` role, it must have a previous sibling that has a `term` role. If the child is a `div` element with no [explicit role][] it is treated as a `term` and `definition` group. The children of the `div` have the same requirement as the `dl` element, except that it does not allow another `div`.
```
Jym77, line 52: There is a bit of confusion between the content model (which is pure HTML) and the roles (which HTML ignores). This is normal given what we want the rule to do, and I agree that the rule should do that. The fact that the rule name and description is only about content model (which I also agree is good) doesn't clarify this.

I think we can start the Background section with the explanation of this difference. Something like
> Content model is pure HTML, this rule accept ARIA replication of the content model through roles. Cases that pass **explicit role** may actually fail the HTML content model. They pass the rule because the overall structure (e.g. a list with items) is good.

Plus, maybe we need an assumption that in `<ul><div>Foo</div></ul>` the `div` is styled to appear as an item in the list, otherwise there is no structure conveyed through presentation and 1.3.1 doesn't apply.
(this also mean that Failed Examples doing this need a bit of styling (`ul > div::before { content: "·" }` should be enough))

---
Anyway, the point is that these notes keep are helpful to understand what the rule checks but don't really clear the confusion. When saying that "children of `ul` must have a role of `listitem`", we talk about the rule, not the content model (which doesn't care about role). I think we should make it explicit that this "must" is about the rule.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,277 @@
...
+- **lists**: The element is a `ul`, `ol` or `menu`; or
+- **definition lists**: The element is a `dl`, or a `div` that is a [child][] in the [flat tree][] of a `dl`.
+
+## Expectation
+
+For each [child][] in the [flat tree][] of each test target, one of the following is true:
+
+- **implicit role**: the [child][] follows the [content model][] and has no [explicit role][]; or
+- **explicit role**: the [child][] has an [explicit role][] which is the [implicit ARIA semantics][] for an HTML element included in the [content model][] for the test target; or
+- **hidden**: the [child][] is not [included in the accessibility tree][].
```
Jym77, line 38: I'd prefer calling this "not exposed" or similar, since "hidden" also has visual meaning which we don't care about here.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,277 @@
...
+### Failed
+
+#### Failed Example 1
+
+This `ul` element had a text node as a child, which will be [included in the accessibility tree][]. Except if they only have [ascii whitespace][], text nodes are not allowed by the [content model][] of `ul` elements.
+
+```html
+<ol>
+	Coming soon!
+</ol>
```
Jym77, line 185: I'm actually not sure this is failing the SC. There is no structure conveyed through presentation, so I don't think 1.3.1 applies.
(this is not about having only one "item", this is about nothing makes it look like an item)

Jym77 (changes requested):

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: Don't much like the word "exposed", it's not the most pleasant word, nor is it particularly descriptive here. Would you be okay with "not included"?

Jym77: Perfectly fine.

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: Fair point. I've written something. Let me know what you think.

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: The issue with this is that this content can't be predictably styled. If someone wants to make sure all list items have the same bullet. The could style it with something like `ul:not([role]) > li, ul:not([role]) > [role^=listitem]`. That won't work for this, even though they might still have styles that apply to the `ul` element.

This SC doesn't just mean that content that look a certain way can be understood that way; the inverse is true too. Even if you mark every heading up with proper heading markup, if you mark non-headings up with heading markup, there's still an issue. Same logic applies here. This isn't a list. It shouldn't be marked up as a list otherwise it makes it impossible to properly style actual lists.

WilcoFiers: @ShadowBB agreed with me on this BTW.

daniel-montalvo (dismissed):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,279 @@
...
+  previous_authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any [HTML elements][] with a [hidden state][] of "false", where the [semantic role][] is the same as the [implicit role][], and for which one of the following is true:
+
+- **lists**: The element is a `ul`, `ol` or `menu`; or
+- **definition lists**: The element is a `dl`, or a `div` that is a [child][] in the [flat tree][] of a `dl`.
```
carlosapaduarte, line 30: I might be mistaken, but I don't think a `div` can have an implicit role (I couldn't find anywhere that states that it is has one even as a child of a `dl`), so it cannot ever be applicable.

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,279 @@
...
+```html
+<ol>
+	<span role="listitem">Eggs</span>
+	<span role="listitem">Milk</span>
+</ol>
+```
+
+#### Passed Example 6
+
+This `ul` element has `li` and `template` [children][child]. `li` is explicitly allowed by the `ul` [content model][], `template` is a script-supported element, which is also allowed by the `ul` content model.
```
carlosapaduarte, line 119:
```suggestion
This `ol` element has `li` and `template` [children][child]. `li` is explicitly allowed by the `ol` [content model][], `template` is a script-supported element, which is also allowed by the `ol` content model.
```

carlosapaduarte (dismissed):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,279 @@
...
+	<div role="listitem">
+		<p>Some text</p>
+	</div>
+</div>
+```
+
+[html elements]: https://html.spec.whatwg.org/multipage/dom.html#htmlelement
+[explicit role]: #explicit-role 'Definition of explicit role'
+[semantic role]: #semantic-role 'Definition of semantic role'
+[implicit role]: #implicit-role 'Definition of implicit role'
```
WilcoFiers, line 271:
```suggestion
```

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,279 @@
...
+  authors:
+    - Brian Bors
+    - Wilco Fiers
+  previous_authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any [HTML elements][] with a [hidden state][] of "false", where the [semantic role][] is the same as the [implicit role][], and for which one of the following is true:
```
WilcoFiers, line 27:
```suggestion
This rule applies to any [HTML elements][] with a [hidden state][] of "false", where the [semantic role][] is not changed by an [explicit role][], and for which one of the following is true:
```

_rules/lists-follow-content-model-a73be2.md
WilcoFiers: You're right. I've redefined this in terms of an explicit role. Think this addressed the issue.

carlosapaduarte (approved):

_rules/lists-follow-content-model-a73be2.md
```diff
@@ -0,0 +1,278 @@
...
+  authors:
+    - Brian Bors
+    - Wilco Fiers
+  previous_authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any [HTML elements][] with a [hidden state][] of "false", where the [semantic role][] is not changed by an [explicit role][], and for which one of the following is true:
```
Jym77, line 27: I understand "semantic role not changed by an explicit role" as `<ul role="list">` being applicable. Is this correct? If so, I think there should be an example with it.
(`<ul role="list">` also applies to "ARIA required owned element", but then maybe these two rules are going to be merged…)

_rules/lists-follow-content-model-a73be2.md
Jym77: I do not fully follow the argument.

> The issue with this is that this content can't be predictably styled. 

The SC is not about content being stylable.

> This SC doesn't just mean that content that look a certain way can be understood that way; the inverse is true too. Even if you mark every heading up with proper heading markup, if you mark non-headings up with heading markup, there's still an issue. Same logic applies here. This isn't a list. It shouldn't be marked up as a list otherwise it makes it impossible to properly style actual lists.

The SC is pretty one-sided: "Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text." so I am not sure what you refer too when saying that "the inverse is true too".

Notably, your example with heading seems to imply that you consider `<div role="heading">` as a failure of 1.3.1.

This example (`<ul>Foo</ul>`) has essentially no presentation (just some indentation, looks similar to a `<div style="text-indent: 40px">Foo</div>`). I say that since no structure is conveyed through presentation, 1.3.1 doesn't care.
I understand your argument as "presentation conveys the information that this is not a list, so accessibility tree must conveys the information that this is not a list". Please correct me if I'm misunderstanding.

If that interpretation is correct, it means for example that `<div role="list"><span role="listitem">Foo</span></div>` is failing 1.3.1 (here also, there is no presentation, so presentation conveys the information that this is not a list. Yet this is [Passed Example 1 of ARIA required owned elements](https://act-rules.github.io/rules/bc4a75#passed-example-1). Which leaves me fairly confused.

---
All this might get clarified when (if) we merge this rule with the "ARIA required owned elements"…

Jym77 (changes requested):

daniel-montalvo (approved):

WilcoFiers:
I'm going to close this PR. it is not actively worked on anymore.

Issue #1177, opened 2020-02-18
Definition of role do not take presentational conflict into account


Jym77:
Our current definition of [Semantic role](https://act-rules.github.io/glossary/#semantic-role) is (in short) "explicit if exists, otherwise implicit". This does not take into account the [Presentational Roles Conflict Resolution](https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none).

So for example, an `<img role="presentation" src="" aria-label="foo">` has an explicit role of `presentation` (good), but also has an "ACT semantic role" of `presentation` while the conflict resolution indicate that it must fallback to the implicit role (of `img`) due to the global property `aria-label`.

---
Among other, this has implication for rule [ARIA state or property is permitted](https://act-rules.github.io/rules/5c01ea).
`<h1 role="presentation" aria-level=2 tabindex=1>Hello World!</h1>` has an explicit role of `presentation` and thus an "ACT semantic role" of presentation. This would fail the rule because `aria-level` is not permitted on a role of `presentation`.
But conflict resolution mandates UA to expose this with the implicit role of `heading` because of the `tabindex` (and test indicates that this indeed happens in Chrome, element is exposed if the `tabindex` is here, hidden otherwise). The `aria-level` is permitted on a role of `heading`, so this does not fail the ARIA role model.

---
I haven't checked the other rules using semantic role. I suspect a few, but not many, would be impacted by this.

---
Proposed fix:
- keep the current definition of implicit and explicit roles as they are simple and generally useful.
- update the definition of semantic role to take the conflict resolution into account (or maybe create a new definition if that keeps things simpler).
- need to check every other rule using semantic role to see if it is impacted…

WilcoFiers:
Related:
- https://github.com/FreedomScientific/VFO-standards-support/issues/256
- https://github.com/w3c/aria/issues/1192

This seems like a bit of a minefield to get in to...

Jym77:
I'd say the way to go would be to follow standards (are they are now), with an Accessibility Support note mentioning that not all UA/AT do that.

Given that Chrome seems rather good at implementing the standard (according to the test you link) and that it is by large the most used browser (50-65% apparently), I think we have good reasons to stick with the standard.

Jym77:
@WilcoFiers What about defining the semantic role as whatever role the element has in the accessibility tree (if any)?

[...]

Issue #1120, opened 2020-01-23
Rules ready for W3C publicatoin


WilcoFiers:
Regular agenda item.

We need to look at all the rules that have more than 2 implementations that have not yet been submitted to the ACT Task Force for publication, and assign people to submit these.

To submit a rule, create a new issue in the following repo: https://github.com/w3c/wcag-act/issues

Use the following template for the proposal

**Issue name**: "{{new | updated}} rule: {{rule title}}"

**Issue body**: 
```
The ACT-Rules community group has completed work on {{rule name}}. We would like to see this rule be added to the WCAG ruleset. You can find the rule here:

**Link to the rule**: [rule name - rule id](url to the rule)
**License**: W3C Software And Document Notice And License
**Implementors**:
- {{Implementor name}}
- {{Implementor name}}
- ...
```

WilcoFiers:
@WilcoFiers : link has acc name
@Jym77 : ARIA state or property is permitted

Jym77:
@WilcoFiers should we mention (in the original message of this issue) that before submitting a rule we should go through a clean up of examples and examples descriptions? (given that so many of them have pretty low quality and we can probably spare some work to reviewers and some time in the process)

Jym77:
@WilcoFiers how do we (submitters)  make sure the list is kept up-to-date?

Should we update https://github.com/act-rules/act-rules.github.io/issues/1120#issuecomment-585812132 (probably a bad idea to all update the same message, that will generate mistakes…)? Should we just write messages in this issues so you can handle the updates to the list (might be a bit noisy, but that should be OK-ish given that it will be the only noise in that issue, also put some centralisation work on you)?

Anyway: https://github.com/w3c/wcag-act/issues/440

jkodu:
> Note: Updated above body of the main comment to move an item (w3c/wcag-act#442) into Ready for W3C listing.

WilcoFiers:
FYI: I'll update this list before every ACT-R call. No need for others to do so.

Jym77:
Do you think it would be possible to add that template to the WCAG-ACT repo? So we could just open a new issue and fill the gaps…

> Use the following template for the proposal
> 
> **Issue name**: "{{new | updated}} rule: {{rule title}}"
> 
> **Issue body**:
> 
> ```
> The ACT-Rules community group has completed work on {{rule name}}. We would like to see this rule be added to the WCAG ruleset. You can find the rule here:
> 
> **Rule title**: ...
[...]

PR #1179, opened 2020-02-18
Improve example descriptions of "ARIA state or property is permitted" (5c01ea)

Jym77:
Before submitting the rule to the ACT TF (https://github.com/act-rules/act-rules.github.io/issues/1120#issuecomment-585812132), I'm cleaning up examples descriptions according to #834.

Note that this rule is hit by #1177 and we'll need to fix that first in any case… That will be for another PR.

Need for Final Call:
This will require a 1 week Final Call (removing 2 examples).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -16,15 +16,16 @@ input_aspects:
 acknowledgements:
   authors:
     - Anne Thyme Nørregaard
+    - Jean-Yves Moyen
 ---
 
 ## Applicability
 
-Any [WAI-ARIA state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is specified on an HTML or SVG element that is [included in the accessibility tree][].
+Any [WAI-ARIA state or property][] that is specified on an HTML or SVG element that is [included in the accessibility tree][].
```
Jym77: There are currently no SVG examples. Should we add some?

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,65 @@ _There are no major accessibility support issues known for this rule._
...
-<div role="combobox" aria-controls>My combobox</div>
-```
-
-#### Passed Example 8
-
-WAI-ARIA states and properties with empty value, specified as an empty string, are also applicable to this rule
-
-```html
-<div role="combobox" aria-controls="">My combobox</div>
-```
```
Jym77: I removed these two examples because they are essentially failing the "SC" (except it's not a WCAG SC, but that's the same idea) and we normally avoid Passed Examples that fail their Accessibility Mapping.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,65 @@ _There are no major accessibility support issues known for this rule._
...
 <div role="button" aria-sort="">Sort by year</div>
 ```
 
 #### Failed Example 2
 
-`aria-sort` attribute is neither inherited, supported, nor required for `role=button` that is the implicit role for `button` element.
+The `aria-sort` [property][] is neither [inherited][], [supported][], nor [required][] for the [semantic role][] `button` which is the [implicit role][] for the `button` element.
 
 ```html
 <button aria-sort="">Sort by year</button>
```
Jym77: I'm considering removing that example as it does not add much compared to Failed Example 1, especially given that the implicit/explicit role thingie has already been examplified in Passed Examples 1 and 2.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -133,18 +118,28 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
...
+This `div` element has no [WAI-ARIA state or property][].
 
 ```html
 <div role="region">A region of content</div>
 ```
 
 #### Inapplicable Example 2
 
-`aria-sort` property is neither inherited, supported, nor required for role `button`, but the element is not [included in the accessibility tree][].
+This `div` element is not [included in the accessibility tree][], hence its [WAI-ARIA state or property][] are not checked.
```
daniel-montalvo:
```suggestion
This `div` element is not [included in the accessibility tree][], hence its [WAI-ARIA state or property][] is not checked.
```

_rules/aria-state-or-property-permitted-5c01ea.md
EmmaJP: Perhaps one for good measure. Is there an aria state or property that is more often used with SVG rather than HTML?

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,65 @@ _There are no major accessibility support issues known for this rule._
 
 #### Passed Example 1
 
-`aria-pressed` state is supported for role `button`.
+The `aria-pressed` [state][] is [supported][] for the [semantic role][] `button` which is the [explicit role][] of this `div` element.
```
EmmaJP:
```suggestion
The `aria-pressed` [state][] is [supported][] for the [semantic role][] `button`, which is the [explicit role][] of this `div` element.
```

If the additional text about the explicit role is non-defining then I think it should be preceded with a comma. Else swap "which" for "that".
https://www.grammarly.com/blog/which-vs-that/

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,65 @@ _There are no major accessibility support issues known for this rule._
...
 
 ```html
 <div role="button" aria-pressed="false">My button</div>
 ```
 
 #### Passed Example 2
 
-`aria-pressed` state is supported for `role=button` that is the [implicit role](#implicit-role) for `button` element.
+The `aria-pressed` [state][] is [supported][] for the [semantic role][] `button` which is the [implicit role][] for `button` elements.
```
EmmaJP:
```suggestion
The `aria-pressed` [state][] is [supported][] for the [semantic role][] `button`, which is the [implicit role][] for `button` elements.
```

_rules/aria-state-or-property-permitted-5c01ea.md
EmmaJP: Perhaps there are better examples for the situation, perhaps for something like a slider, where the required states and properties have defined default values. 

Though my minimal understanding is that such states and properties don't need to be added in the code at all if the default values are being used. Is that why you are removing these examples?

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,65 @@ _There are no major accessibility support issues known for this rule._
...
+The `aria-sort` [property][] is neither [inherited][], [supported][], nor [required][] for the [semantic role][] `button`.
 
 ```html
 <div role="button" aria-sort="">Sort by year</div>
 ```
 
 #### Failed Example 2
 
-`aria-sort` attribute is neither inherited, supported, nor required for `role=button` that is the implicit role for `button` element.
+The `aria-sort` [property][] is neither [inherited][], [supported][], nor [required][] for the [semantic role][] `button` which is the [implicit role][] for the `button` element.
```
EmmaJP:
```suggestion
The `aria-sort` [property][] is neither [inherited][], [supported][], nor [required][] for the [semantic role][] `button`, which is the [implicit role][] for the `button` element.
```

_rules/aria-state-or-property-permitted-5c01ea.md
EmmaJP: Perhaps remove Failed example 1 instead. And swap passed examples 1 and 2. Prioritise/encourage semantic HTML element with implicit role over non-semantic element with role defined.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,65 @@ _There are no major accessibility support issues known for this rule._
...
+The `aria-busy` [state][] is a [global][] [state][] that is [supported][] by all elements, even without any [semantic role][].
 
 ```html
 <div aria-busy="true">My busy button</div>
 ```
 
 #### Passed Example 4
 
-`aria-label` state is inherited for role `button`
+The `aria-label` [state][] is [inherited][] for the [semantic role][] `button`.
```
WilcoFiers: This is kind of odd. aria-label is global, should we say that?

_rules/aria-state-or-property-permitted-5c01ea.md
WilcoFiers: Can you do this in a different PR?

Jym77: #1183

_rules/aria-state-or-property-permitted-5c01ea.md
Jym77: It is global, which as far as role definition goes (https://www.w3.org/TR/wai-aria/#button) means that is is inherited or all roles (from the abstract top-level `roletype`).

The thing is if you look at the description of the `button` role (or any other), then the global states and properties are listed in the "inherited" box (which is true, global are supported by `roletype` and inehirted by the rest). So just saying it's global might be a bit confusing when looking at the specs and  seeing it As "inherited".

I tried to improve the description to explain that. Feel free to suggest something better.

_rules/aria-state-or-property-permitted-5c01ea.md
Jym77: It looks like the roles defined in the WAI-ARIA Graphic module only accept the global states and properties… I've added an example anyway.

WilcoFiers:
Dismissed stale reviews.

Jym77:
Final Call ends on 09 March 2020.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,83 @@ _There are no major accessibility support issues known for this rule._
...
+The `aria-busy` [state][] is a [global][] [state][] that is [supported][] by all elements, even without any [semantic role][].
 
 ```html
 <div aria-busy="true">My busy button</div>
 ```
 
 #### Passed Example 4
 
-`aria-label` state is inherited for role `button`
+The `aria-label` [state][] is [global][] and thus [inherited][] for all [semantic role][].
```
adilsofficial: Add word `state` with global:
```html
The `aria-label` [state][] is [global][] [state][] and thus [inherited][]  for all [semantic role][].
```

adilsofficial: Add word `state` with global: 
```html The `aria-label` [state][] is [global][] [state][] and thus [inherited][]  for all [semantic role][]. ```

File: _rules/aria-state-or-property-permitted-5c01ea.md [hunk 10]
```diff
@@ -49,81 +50,83 @@ _There are no major accessibility support issues known for this rule._
...
 
 ```html
 <div role="button" aria-label="OK"></div>
 ```
 
 #### Passed Example 5
 
-`aria-checked` state is required for role `aria-checkbox`
+The `aria-checked` [state][] is [required][] for the [semantic role][] `aria-checkbox`.
```
adilsofficial: change `aria-checkbox` to `checkbox`

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,83 @@ _There are no major accessibility support issues known for this rule._
...
 
 ```html
 <div role="checkbox" aria-checked="false">My checkbox</div>
 ```
 
 #### Passed Example 6
 
-`aria-controls` property is supported for role `combobox`
+The `aria-controls` [property][] is [supported][] for the [semantic role][] `combobox`.
```
adilsofficial: The `aria-controls` [property][] is [required][] for the [semantic role][] `combobox`.
See Characteristics table under combobox: https://www.w3.org/TR/wai-aria-1.1/#combobox

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,83 @@ _There are no major accessibility support issues known for this rule._
...
 
 ```html
 <div role="combobox" aria-controls="id1">My combobox</div>
 ```
 
 #### Passed Example 7
 
-WAI-ARIA states and properties with empty value are also applicable to this rule
+The `aria-controls` [property][] is [supported][] for the [semantic role][] `combobox`. [WAI-ARIA states and properties][wai-aria state or property] with empty value are still applicable to this rule.
```
adilsofficial: The `aria-controls` [property][] is [required][] for the ...

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -49,81 +50,83 @@ _There are no major accessibility support issues known for this rule._
...
 
 ```html
 <div role="combobox" aria-controls>My combobox</div>
 ```
 
 #### Passed Example 8
 
-WAI-ARIA states and properties with empty value, specified as an empty string, are also applicable to this rule
+The `aria-controls` [property][] is [supported][] for the [semantic role][] `combobox`. [WAI-ARIA states and properties][wai-aria state or property] with empty value (specified as an empty string) are still applicable to this rule.
```
adilsofficial: The aria-controls [property][] is [required][] for the ...

Jym77: > The aria-controls [property][] is [required][] for the ...

@adilsofficial note that for small changes like that, while you write the review there is a "suggestion" button that let you make the change directly. It is often faster to do for the reviewer (you) because you just need to replace one word by another rather than explain what need to be changed,. and easier/faster for writer (me) because Github gives me a one-click "accept change" button.

Jym77:

PR #1182, opened 2020-02-20
xml-lang-match: Remove inapplicable examples that fail SC

WilcoFiers:
While not required by the ACT rules format, the CG has a policy to ensure all passed / inapplicable examples satisfy the success criteria the rule tests. This is not the case for these two inapplicable examples.

Need for Final Call: 1 week


WilcoFiers:
@Jym77 I'm aware. I think that's alright.

WilcoFiers:
Problem seems to have fixed itself.

PR #1183, opened 2020-02-21
Remove Passed Examples that do not meet the accessibility mapping in "ARIA state or property is permitted" (5c01ea)

Jym77:
Remove Passed Examples that do not pass the accessibility mapping.

Closes issue(s):

- https://github.com/act-rules/act-rules.github.io/pull/1179#discussion_r382068415

Need for Final Call:
This will require a 1 week Final Call (removing examples)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -95,22 +95,6 @@ Global state that is supported by all base markup elements.
 <div role="combobox" aria-controls="id1">My combobox</div>
 ```
 
-#### Passed Example 7
-
-WAI-ARIA states and properties with empty value are also applicable to this rule
-
-```html
-<div role="combobox" aria-controls>My combobox</div>
-```
-
-#### Passed Example 8
-
-WAI-ARIA states and properties with empty value, specified as an empty string, are also applicable to this rule
-
-```html
-<div role="combobox" aria-controls="">My combobox</div>
-```
-
 ### Failed
 
 #### Failed Example 1


Jym77:
Final Call ends March 2nd.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -95,22 +95,6 @@ Global state that is supported by all base markup elements.
 <div role="combobox" aria-controls="id1">My combobox</div>
 ```
 
-#### Passed Example 7
```
WilcoFiers: I don't understand how these are any less of a failure than passed example 6? None of them have an accessible name, and they're all missing aria-expanded="false"

Jym77: Because these two examples do not "pass" the accessibility mapping to correctly using ARIA attributes. `aria-controls` value must be an ID reference list (https://www.w3.org/TR/wai-aria-1.1/#aria-controls) which is defined as "A list of one or more ID references." (https://www.w3.org/TR/wai-aria-1.1/#valuetype_idref_list).

I'm seeing this as the same thing as "passed examples should not fail their SC" for rule mapping to WCAG. But maybe I'm reading too much into an accessibility mapping to something that is not as clearly separated as a WCAG criterion…

WilcoFiers: There is no "must" or "should" in section 7.5 that says attribute values have to be correct. But if there were, passed example 6 would also fail, because ID reference has to point to an existing element:

> **ID reference**
> Reference to the ID of another element in the same document

Jym77:
Looks like I read too much into these examples (https://github.com/act-rules/act-rules.github.io/pull/1183#discussion_r383776549). Closing this.

PR #1184, opened 2020-02-22
Amendments to button and image button rules

### LINKED ISSUE #1049: One element relevant for multiple success criteria
Following a comment (https://github.com/act-rules/act-rules.github.io/pull/1028#discussion_r351308532) and meeting discussion, it would seem there are different points of view on how rules should handle multiple success criteria that apply to the same element. 

Should there be different rules for an element that needs to meet multiple success criteria?

Or should multiple rules call out pass/fail on the same element?

----

The rule that raised the comment checks if a button (anything with a semantic role of `button`) has an accessible name, but excludes `input` elements of `type="image"` because they also have to satisfy an additional success criteria.

So currently we have two rules checking that buttons have accessible names. And no rule checking that the accessible name of an image button describes its purpose (the additional requirement in SC1.1.1 that isn't part of SC4.1.2).

For these rules, I propose that "Button has accessible name" (97a4e1) should also apply to image buttons. Inapplicable example 1 should be removed and Passed and Failed examples of image buttons added. And that "Image button has accessible name" (59796f) should be reworked as "Image button has accessible name that describes its purpose".

An image button could then pass both rules, or pass the first rule but not the second giving implementers an opportunity to provide clearer error information.

----

Of course, these two rules could have been split out to reduce complexity in the first rule. In which case I propose that "Image button has accessible name" as it currently stands does not relate to SC1.1.1 and an additional rule about the accessible name describing the purpose of the image button is still required.

(incidentally, Passed examples 1 and 6 are essentially the same in "Image button has accessible name")

----

For reference: 
- https://github.com/act-rules/act-rules.github.io/blob/93a2a770ad1d8286a8a74a3c7a7baad7edd8f3d8/_rules/button-accessible-name-97a4e1.md
- https://github.com/act-rules/act-rules.github.io/blob/93a2a770ad1d8286a8a74a3c7a7baad7edd8f3d8/_rules/image-button-accessible-name-59796f.md

EmmaJP:
Following decisions in conversation raised by #1049, this PR combines rules regarding buttons and image buttons having an accessible name, and creates a new rule about whether the accessible name of an image button describes its purpose.

When reviewing, please also point out any out of dates formats or phrasing. I have tried to adjust those as I've worked.

Issue(s):

- related to and possibly closes #1049

Need for Final Call:
This will require a 2 week Final Call, as it includes a new rule.

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/button-accessible-name-97a4e1.md
```diff
@@ -21,10 +21,12 @@ acknowledgements:
 
 ## Applicability
 
-The rule applies to elements that are [included in the accessibility tree][] and have a [semantic role](#semantic-role) of `button`, except for `input` elements whose `type` attribute is in the [`Image Button` state](<https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image)>).
+The rule applies to elements that are [included in the accessibility tree][] and have a [semantic role](#semantic-role) of `button`, including `input` elements whose `type` attribute is in the [`Image Button` state](<https://html.spec.whatwg.org/#image-button-state-(type=image)>).
```
Jym77:
```suggestion
The rule applies to elements that are [included in the accessibility tree][] and have a [semantic role](#semantic-role) of `button`.
```
`button` is the implicit role for these elements, I don't think we need the precision.

_rules/button-accessible-name-97a4e1.md
```diff
@@ -21,10 +21,12 @@ acknowledgements:
...
 
 **Note:** `input` elements have a `type` attribute in the `Image button` state if it is set to any case-insensitive match of `image` (most of the time, using `<input type="image">`).
 
+**Note:** The specification of the [`type`](https://html.spec.whatwg.org/#states-of-the-type-attribute) attribute describes in detail how to map the value of the attribute to its corresponding state.
+
```
Jym77: I think both these notes can now go away.

_rules/button-accessible-name-97a4e1.md
```diff
@@ -172,26 +208,68 @@ This off screen `button` element has no [accessible name][] because it has no co
...
+```
+
+#### Failed Example 7
+
+This image button has an `aria-labelledby` attribute, but the referenced element does not exist. This gives the button an empty [accessible name][].
+
+```html
+<input type="image" src="/test-assets/shared/search-icon.svg" aria-labelledby="non-existing" />
+```
+
```
Jym77: These examples might be hitting the corner case of the HTML AAM overloading the accessible name computation:
https://www.w3.org/TR/html-aam-1.0/#input-type-image (step 4)
Can you try and run  some tests on major browsers to check what they do?

_rules/button-accessible-name-97a4e1.md
```diff
@@ -172,26 +208,68 @@ This off screen `button` element has no [accessible name][] because it has no co
...
-<button style="display: none;"></button>
+<input type="image" src="/test-assets/shared/search-icon.svg" style="display: none;" />
 ```
 
 #### Inapplicable Example 3
 
+The `button` element is tested separately from the `img` element. [Success Criterion 4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) is applied to the button, whereas the image is tested under [Success Criterion 1.1.1 Non-text Content](https://www.w3.org/TR/WCAG21/#non-text-content)
+
+```html
+<button><img src="/test-assets/shared/search-icon.svg" alt="Search" /></button>
```
Jym77: The `button` element still has a semantic role of `button` and thus is applicable for the rule.

_rules/button-accessible-name-97a4e1.md
```diff
@@ -172,26 +208,68 @@ This off screen `button` element has no [accessible name][] because it has no co
...
+```
+
+#### Inapplicable Example 3
+
+The `img` element is not a user interface component, and so is not tested for [Success Criterion 4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value).
```
Jym77:
```suggestion
There is no element with a [semantic role][] of `button`.
```

_rules/image-button-name-descriptive-f3b293.md
```diff
@@ -60,67 +57,68 @@ There is a known combination of a popular browser and assistive technology that
...
 
 ```html
-<input type="image" src="/test-assets/shared/search-icon.svg" alt="" />
+<input type="image" src="/test-assets/shared/search-icon.svg" alt="Share" />
 ```
 
 #### Failed Example 3
 
-The image button has an `aria-labelledby` attribute, but the referenced element does not exist. This gives the button an empty [accessible name][].
+The image button has an `aria-labelledby` attribute, but the referenced element does not exist. This gives the button an empty [accessible name][], which cannot be descriptive.
```
Jym77: As with the other rule. These may actually have an accessible name according to the AAM.
OTOH, this is probably easier to get failed examples here given that the default name is "Submit Query"…

_rules/image-button-name-descriptive-f3b293.md
```diff
@@ -146,23 +144,23 @@ The `input` element with type with a `type` attribute in the `Button` state is n
...
+This `button` element is separate from the `img` element.
 
 ```html
 <button><img src="/test-assets/shared/search-icon.svg" alt="Search" /></button>
 ```
 
 #### Inapplicable Example 4
 
-The `img` element is not a user interface component, and so is not tested for [Success Criterion 4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value).
+This `img` element is not button component.
```
Jym77: Description of these two examples could also be in the line of "There is no input element in the Image button state" (to match applicability more closely). Likely adding the extra explanation as second sentence.

_rules/button-accessible-name-97a4e1.md
EmmaJP: I've added the word "all" to the change. I agree, but wanted to gauge opinion.

_rules/button-accessible-name-97a4e1.md
EmmaJP: Are either of these notes required later in regard a specific example. Required for understanding, as opposed to nice-to-have.

_rules/button-accessible-name-97a4e1.md
```diff
@@ -127,6 +130,39 @@ This `input` element has an [accessible name][] because of the default accessibl
...
+```html
+<input type="image" src="/test-assets/shared/search-icon.svg" alt="Search" />
+```
+
+#### Passed Example 10
+
+This image button has an [accessible name][] through the `aria-label` attribute.
+
+```html
+<input type="image" src="/test-assets/shared/search-icon.svg" aria-label="Search" />
```
EmmaJP: Not sure this adds anything not already covered in Passed Example 3

_rules/button-accessible-name-97a4e1.md
```diff
@@ -127,6 +130,39 @@ This `input` element has an [accessible name][] because of the default accessibl
...
+```
+
+#### Passed Example 11
+
+This image button has an [accessible name][] through the `title` attribute.
+
+```html
+<input type="image" src="/test-assets/shared/search-icon.svg" title="Search" />
+```
```
EmmaJP: Is a title attribute considered accessible in this situation?

_rules/button-accessible-name-97a4e1.md
EmmaJP: This example in Chrome and Safari, defaults to a submit button programmatically. If the image doesn't load, the word 'Submit' will become the button text.

_rules/button-accessible-name-97a4e1.md
EmmaJP: Would this code ever be considered an image button? Or is it a failed example?

_rules/image-button-name-descriptive-f3b293.md
EmmaJP: I guess in this instance the default would not be descriptive.

_rules/image-button-name-descriptive-f3b293.md
EmmaJP: Could you provide some suggestions so I can better understand what you mean. Thanks.

_rules/button-accessible-name-97a4e1.md
Jym77: If they are only required for understanding an example or two, they should probably be moved to the description of these examples. With the new applicability which has nothing to do with `type` or `Image button`, they look completely out of the place and it is not clear why a rule about buttons need this (for example, why precising how to put an `input` in the `Image button` state, but not precise how to put it in the `Button`, `Reset Button` or `Submit Button` state? And why even mention `input` elements on a rule about buttons?)

_rules/button-accessible-name-97a4e1.md
Jym77: So, the description of "empty accessible name" is not correct. And these examples do not actually fail this rule (they fail the other rule given that the name is not descriptive).

_rules/button-accessible-name-97a4e1.md
Jym77: This is not an image button as per the sense "`input` element in the `Image Button` state".
But this is the rule checking that elements with role of `button` have a (non empty) accessible name. In that case, there is an element with a role of `button` (the `button` element itself), thus the rule is applicable and passes because the `button` element gets its name from the content, here the, `alt` attribute of the image (iirc).

_rules/image-button-name-descriptive-f3b293.md
Jym77: Yes, but the description should say so rather than saying that it has an empty accessible name.

_rules/image-button-name-descriptive-f3b293.md
Jym77: The rule is applicable to `input` elements in the `Image Button` state. So the reason that these snippets are inapplicable is the lack of such elements. It has nothing to do with `button` elements or components. To match the example description syntax, both these should be
"There is no `input` element in the `Image Button` state." 
An extra sentence can be added. Maybe for Inapplicable Example 3 "A `button` element containing only an `img` element is not an Image Button, this rule specifically checks for `input` elements in the `Image Button` state". And for Inapplicable Example 4 "The `img` element alone is not an Image Button."

EmmaJP:
Note - check if this has impact on image button has accessible name, which is a rule that has already been accepted by the task force and changes may need to be resubmitted.

WilcoFiers:
Closing due to inactivity.

PR #1189, opened 2020-02-26
Handling Presentational role conflict resolution

### LINKED ISSUE #1177: Definition of role do not take presentational conflict into account

Jym77:
[Presentational roles conflict resolution](https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none) states that elements with a `role="none"` must be exposed with their implicit role if they have either a global `aria-*` attribute (e.g. `aria-label`) or are focusable. This was completely ignored by our definition of semantic role.

This resolution is not well implemented in all browsers (https://github.com/act-rules/act-rules.github.io/wiki/Tests-on-Presentational-role-Conflict-resolution). A side note being the handling of `aria-hidden` on focusable elements for which there is no specs and browsers are doing different things.

---
This PR updates the definition of semantic role to take the resolution into account. The role is linked to the actual role in the accessibility tree to circumvent both the discrepancies between browsers (we do with what we have instead of taking side) and the ambiguous "interactive" term in the spec (we don't decide what is interactive, we just do with whatever the UA is giving us). The new examples are however following the specs (which is also what Chrome, the majority browser, does).

I have also removed examples with `aria-hidden` on focusable elements. Because there is no spec for what do to in that case, and browsers disagree, I preferred taking these examples out because what to do is far from clear.

I considered that "ARIA state or property is permitted" was the main rule to test the definition and added several examples there. Other rules are mostly using the role incidentally and only got one extra example (usually a Failed one as "when do we fail?" is the important question for our rules).

---
The definition of "marked as decorative" also ended up as collateral damage of this PR. There was somewhat of an internal contradiction between "having a role of `none`" and achieving that by "explicitly setting `role="none"`" (which does not work because of the conflict resolution). Where it was important, "marked as decorative" as been replaced by "role of `none`" which seems to work.

We might still need a definition of "marked as decorative" which would be in the line of "explicit role of `none`" so that something could be marked as decorative (expressing intention of the writer) without actually having a role of `none` (due to conflict resolution) and that would likely indicate some sort of failure or at least bad practice (i.e. explicitly setting `role="none" aria-label="foo"` is a bad idea as there is conflict between the intention of hiding stuff and showing it).

---
Changes ripples all over the place and in many rules. Please take a close look and also try to figure out if I've missed something… Keep the problem in mind when looking at other rules in the near future.

Closes issue(s):

- closes #1177 

Need for Final Call:
This will require a 2 weeks Final Call (changes many rules, has lot of implications).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/semantic-role.md
```diff
@@ -9,4 +9,17 @@ A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic
 
 Valid semantic roles are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications](https://act-rules.github.io/glossary/#wai-aria-specifications).
 
-The semantic role of an element is its [explicit semantic role](#explicit-role) if it has any, otherwise, the [implicit semantic role](#implicit-role) is used.
+The _semantic role_ of an element which is [included in the accessibility tree][] is the role this element has in the accessibility tree. Elements which are not [included in the accessibility tree][] have no semantic role.
```
WilcoFiers: So elements that aren't in the accessibility tree by definition have no role? That's a substantial change to one of our most used definitions. It means every time way say "has semantic role of X and is included in the accessibility tree" the second part becomes redundant. I'm hesitant about making that change. We'd have to update quite a few rules, make this "included in the accessibility tree" bit hidden away in the definition, which I don't like, and it feels a little unintuitive to me. For example, it would mean nothing can ever have a semantic role of "none". That's different from how semantic role is used by the ARIA working group.

I think we can add this in without such a substantial change to the definition.

```suggestion
The _semantic role_ of an element is its [explicit semantic role](#explicit-role) if it has any, otherwise, the [implicit semantic role](#implicit-role) is used. If the element has an [explicit semantic role[#explicit-role] of "presentation" or "none", but the element is, or would be, included in the accessibility tree when it is not [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden), the [implicit semantic role][#implicit-role] is used instead of the [explicit semantic role][#explicit-role].
```

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -111,6 +113,22 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 <div role="combobox" aria-controls="">My combobox</div>
 ```
 
+#### Passed Example 9
+
+This `aside` element has an [explicit role][] of `none`. However, the [global][] [property][] `aria-label` is specified. Thus it has a [semantic role][] of `complementary` due to [Presentational Roles Conflict Resolution][]. The `aria-expanded` [state][] is [inherited][] for the `complementary` role.
+
+```html
+<aside role="none" aria-label="About ACT rules" aria-expanded="true">ACT rules are cool!</aside>
```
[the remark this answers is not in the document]
Jym77: The specs seems pretty clear to me on that point.

And both Chrome and Firefox do expose elements like this. Edge doesn't but Edge without Chromium is kinda dead, I guess…

https://github.com/act-rules/act-rules.github.io/wiki/Tests-on-Presentational-role-Conflict-resolution

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -111,6 +113,22 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
...
+```
+
+#### Passed Example 10
+
+This `aside` element has an [explicit role][] of `none`. However, it is [focusable][] due to the `tabindex` attribute (even if it is excluded from sequential focus order). Thus it has a [semantic role][] of `complementary` due to [Presentational Roles Conflict Resolution][]. The `aria-expanded` [state][] is [supported][] for the `complementary` role.
+
+```html
+<aside role="none" tabindex="-1" aria-expanded="true">ACT rules are cool!</aside>
```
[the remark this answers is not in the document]
Jym77: I agree that the specs are not clear on what to do with `tabindex=-1`.
But on that point, Chrome, Edge and Firefox all agree to expose the element anyway (and yes, it gets very hard to remove role from `button` that are not `disabled`…)
(and, for examples, heading exposed that way are further shown in NVDA, except on top of Edge because Edge doesn't fallback to the implicit role)
https://github.com/act-rules/act-rules.github.io/wiki/Tests-on-Presentational-role-Conflict-resolution

I am not sure we should be writing for uncertain plans in ARIA 1.3…

_rules/image-accessible-name-23a2a8.md
```diff
@@ -39,7 +40,7 @@ The rule applies to HTML `img` elements or any HTML element with the [semantic r
 
 ## Expectation
 
-Each target element has an [accessible name][] that is not empty (`""`), or is [marked as decorative][].
+Each target element has an [accessible name][] that is not empty (`""`), or has a [semantic role][] of `none` or `presentation`.
```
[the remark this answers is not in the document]
Jym77: That is _literally_ just inlining the definition of "marked as decorative".

The definition of "marked as decorative" depends on the semantic role. Changing the definition of "semantic role" does also change "marked as decorative". I think that the name does not really convey the intend of the new definition.

I think that "marked as decorative" should be "explicit role of `none`/`presentation` or (for `img`) empty `alt`". Given that it was only really used here, I though it would be better to inline the definition and remove one glossary item.

I am indeed not entirely sure whether that rule should check "accessible name or role of `presentation`" or "accessible name, or explicit role of `presentation`, or empty `alt`"

The difference between the two would be for `` which would fail the "accessible name or role of `presentation`" version (with resolution as implemented by Chrome and Firefox) but pass the "accessible name, or explicit role of `presentation`, or empty `alt`" (on all browsers).

I think this image creates an accessibility issue (because it is tabbable but has no name) and thus chose the version that fails it…

I am happy to revisit if you consider that this should pass the rule instead.

pages/glossary/semantic-role.md
Jym77: Very good point.
I wanted to try an avoid going through the full case study on our definition but this seems unavoidable…

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -34,7 +35,8 @@ _There are currently no assumptions_
 
 ## Accessibility Support
 
-_There are no major accessibility support issues known for this rule._
+- There exist popular web browsers and assistive technologies which do not correctly implement [Presentational Roles Conflict Resolution][].
+- Some browsers expose elements which are [focusable][] but have an `aria-hidden="true"` attribute, while some hide them.
```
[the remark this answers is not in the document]
EmmaJP: I haven't looked through everything yet... but it seems that if this is an issue with implementation by browsers and AT, then the ACT rules and examples should be based on the standard, and anomalies listed in the Accessibility Support for awareness.

_rules/image-accessible-name-23a2a8.md
```diff
@@ -3,7 +3,7 @@ id: 23a2a8
 name: Image has accessible name
 rule_type: atomic
 description: |
-  This rule checks that each image that is not marked as decorative, has an accessible name.
+  This rule checks that each image is either marked as decorative, or has an accessible name.
```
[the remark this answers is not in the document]
EmmaJP: I think that tweak significantly changes the meaning implied. The rule is not for checking that images are marked as decorative. I understand that needs to be worked out to check that non-decorative images have an accessible name.

It could change in a different way. Perhaps:
"This rule checks that each image has an accessible name, unless it is marked as decorative."

This states the intention of the rule up front, with the known exception following.

_rules/aria-state-or-property-permitted-5c01ea.md
daniel-montalvo: The behavior of AT here is not consistent. JAWS does not report the expanded state. NVDA only reports it in Firefox.

_rules/aria-state-or-property-permitted-5c01ea.md
daniel-montalvo: The behavior of AT here is not consistent. JAWS does not report the expanded state. NVDA only reports it in Firefox.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -129,6 +147,14 @@ WAI-ARIA states and properties with empty value, specified as an empty string, a
 <button aria-sort="">Sort by year</button>
 ```
 
+#### Failed Example 3
+
+This `aside` element has an [explicit role][], and a [semantic role][], of `none`. The `aria-expanded` [state][] in neither [inherited][], [supported][] nor [required][] for this role.
```
daniel-montalvo:
```suggestion
This `aside` element has an [explicit role][], and a [semantic role][], of `none`. The `aria-expanded` [state][] is neither [inherited][], [supported][] nor [required][] for this role.
```
Even if we want to keep it, we should correct that little typo.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -141,10 +167,20 @@ No [WAI-ARIA state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_d
 
 #### Inapplicable Example 2
 
-`aria-sort` property is neither inherited, supported, nor required for role `button`, but the element is not [included in the accessibility tree][].
+`aria-sort` property is neither inherited, supported, nor required for role `button`, but the element is not [included in the accessibility tree][] because it is not rendered.
```
daniel-montalvo:
```suggestion
`aria-sort` property is neither inherited, supported, nor required for role `button`. In addition, the element is not [included in the accessibility tree][] because it is hidden to everybody using CSS.
```
I think the element is rendered, though it is not displayed because of the `display:none;` style. Anyway, I think it helps clarify the description.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -52,7 +53,9 @@ For this rule, it is assumed that the `autocomplete` attribute is not used on fo
 
 ## Accessibility Support
 
-While `autocomplete` in a promising technique for supporting personalization in HTML, support for this is fairly limited.
+- While `autocomplete` in a promising technique for supporting personalization in HTML, support for this is fairly limited.
```
daniel-montalvo:
```suggestion
- Support for the `autocomplete` technique is fairly limited.
```
Just to remain as objective as possible. I don't see how "promising" is relevant here.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -127,7 +130,7 @@ Full length autocomplete terms.
 
 #### Passed Example 8
 
-The `input` element does not have a semantic role that is a widget role, but still participates in sequential focus navigation, and has a single autocomplete term.
+This `input` element has an [explicit role][] of `none`. However, it is [focusable][] (by default). Thus it has a [semantic role][] of `textbox` due to [Presentational Roles Conflict Resolution][]. It has a single autocomplete term.
```
daniel-montalvo: This example does not have a label. Should we be adding it?

_rules/aria-state-or-property-permitted-5c01ea.md
Jym77: @daniel-montalvo my problem when writing these examples was to find something:
- with an implicit role (to trigger conflict resolution vs `role="none"`); and
- the implicit role has some specific (non-global) ARIA attributes (because the global ones are always allowed and this rule is pointless for them); and
- the element is not focusable by default (because focusable takes precedence over the presentational role conflict resolution, so if the element is focusable it will be exposed anyway and that doesn't test the other part of the conflict resolution (exposing global attribute)) (this exclude things like `button` or links…); and
- the ARIA attribute is not set to a correct value by the AAM (because if it's already set, having it explicit is a bad practice and overwriting it to a non-default value is likely a bad practice also if there is no context to explain it) (this exclude things such as `aria-level` on headings, or `aria-posinset` on list which have correct default value).

With all these constraints, I had a somewhat limited choice… Another possibility would have been `<section><h1 aria-level=2>Foo</h1></section>` but it's in itself a bit tricky why this `h1` needs its level to be changed (and requires the `section` as context), so I ended up choosing this as a decent example where the code shown is simple and not bad practice…

I'll be happy to change it if you figure out some other case that would fit within the constraints above…

_rules/autocomplete-valid-value-73f2c2.md
Jym77: I think this is a separate topic and should be part of another PR if needed. Would you like to handle it, @daniel-montalvo ?

_rules/autocomplete-valid-value-73f2c2.md
Jym77: This is topic for another PR, and much more generic than this specific rule.
See #1215.

_rules/descriptive-link-5effbb.md
```diff
@@ -294,6 +295,16 @@ The [accessible name][] (from the link's text) does not describe the purpose of
 </table>
 ```
 
+#### Failed Example 7
```
WilcoFiers: I don't think this test case is necessary.

_rules/image-accessible-name-23a2a8.md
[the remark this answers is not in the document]
WilcoFiers: I agree with the change, now that you've changed what semantic role does with elements not in the accessibility tree. Still, I think this should be in a separate PR. It's a completely unrelated problem.

pages/glossary/semantic-role.md
```diff
@@ -9,4 +9,19 @@ A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic
 
 Valid semantic roles are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications](https://act-rules.github.io/glossary/#wai-aria-specifications).
 
-The semantic role of an element is its [explicit semantic role](#explicit-role) if it has any, otherwise, the [implicit semantic role](#implicit-role) is used.
+The _semantic role_ of an element is determined by the first of these cases that applies:
+
+1. **(Conflict)** If the element has an [explicit role][] which is either `none` or `presentation`, but the element is [included in the accessibility tree][] (or would be when it is not [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden)), then its _semantic role_ is its **[implicit role][]**.
```
WilcoFiers: I think I'd prefer:

```md
1. **topic**: description
```

This is how these labelled items are done in WCAG.

pages/glossary/semantic-role.md
```diff
@@ -9,4 +9,19 @@ A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic
...
+2. **(Explicit)** If the element has an [explicit role][], then its _semantic role_ is its [explicit role][].
```
WilcoFiers: Here's a problem... Explicit / Implicit semantic role relies on the definition of semantic role. These are recursive definitions. We're going to need to solve that.

Jym77: Hmmm :thinking: 
Actually, the circular reference already exists in the current definitions…

---
My analysis of the circularity:
* there is some concept called "semantic role".
* there is some technical stuff called explicit/implicit role that are special case of the concept of semantic role.
* there is some technical stiff called "semantic role" which is define based on the explicit/implicit role and is thus also a case of the concept of semantic role.

Thus, the definition is not logically circular. This is more an unfortunate consequence of the fact that the concept and the technical thingie are named the same way…

This should be solved nonetheless.

---
In any case, given that this circularity already exists in the current definitions, and that I don’t think this PR is making things worse, I suggest to keep resolving that for another PR (keeping this one focused).

pages/glossary/semantic-role.md
```diff
@@ -9,4 +9,19 @@ A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic
 
 Valid semantic roles are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications](https://act-rules.github.io/glossary/#wai-aria-specifications).
 
-The semantic role of an element is its [explicit semantic role](#explicit-role) if it has any, otherwise, the [implicit semantic role](#implicit-role) is used.
+The _semantic role_ of an element is determined by the first of these cases that applies:
+
+1. **Conflict** If the element has an [explicit role][] which is either `none` or `presentation`, but the element is [included in the accessibility tree][] (or would be when it is not [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden)), then its _semantic role_ is its **[implicit role][]**.
```
kasperisager:
```suggestion
1. **Conflict** If the element has an [explicit role][] which is either `none` or `presentation`, but the element is [included in the accessibility tree][], or would be when it is not [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden), then its _semantic role_ is its **[implicit role][]**.
```

Seems a little too important to be put in parentheses.

pages/glossary/semantic-role.md
```diff
@@ -11,4 +11,19 @@ A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic
 
 Valid semantic roles are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications](https://act-rules.github.io/glossary/#wai-aria-specifications).
 
-The semantic role of an element is its [explicit semantic role](#explicit-role) if it has any, otherwise, the [implicit semantic role](#implicit-role) is used.
+The _semantic role_ of an element is determined by the first of these cases that applies:
+
+1. **Conflict** If the element has an [explicit role][] which is either `none` or `presentation`, but the element is [included in the accessibility tree][], or would be when it is not [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden), then its _semantic role_ is its **[implicit role][]**.
```
kasperisager: I'm not entirely sold on the objectiveness of the WCAG definition of "hidden". The way I interpret it is that an element is hidden when it doesn't generate any layout boxes, or generates only 0-dimension boxes, meaning that it isn't rendered. Thoughts?

Jym77: @kasperisager would this be better?
```suggestion
1. **Conflict** If the element has an [explicit role][] which is either `none` or `presentation`, but the element is [included in the accessibility tree][], or would be when it is [visible](#visible), then its _semantic role_ is its **[implicit role][]**.
```

kasperisager: I'm pretty sure that's not the definition we'll want to use though. Per https://w3c.github.io/aria/#tree_exclusion, things are removed from the accessibility tree when they have either `display: none` or `visibility: hidden`.

Jym77: Then 
```suggestion
1. **Conflict** If the element has an [explicit role][] which is either `none` or `presentation`, but the element is [included in the accessibility tree][], or would be when it is not [excluded from the accessibility tree](https://w3c.github.io/aria/#tree_exclusion), then its _semantic role_ is its **[implicit role][]**.
```

Jym77:
@WilcoFiers, @kasperisager, I tried to solve https://github.com/act-rules/act-rules.github.io/pull/1189#discussion_r395601964 while keeping https://github.com/act-rules/act-rules.github.io/pull/1189#discussion_r386347075
That required introducing a definition of "(explicitly) hidden" which depends on CSS, thus cascading (if I may say so) "CSS styling" as input aspect to the definition of "semantic role" and any rule using it… 

Please check if this is OK for you…

pages/glossary/hidden.md
```diff
@@ -0,0 +1,15 @@
...
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+An HTML element is _explicitly hidden_ if at least one of the following is true for itself or any of its ancestors:
+
+- has a `hidden` attribute; or
+- has a CSS property `display` set to `none`; or
+- has a CSS property `visibility` set to `hidden`.
```
kasperisager:
```suggestion
- has a [computed](https://www.w3.org/TR/css-cascade/#computed-value) CSS property `display` of `none`; or
- has a [computed](https://www.w3.org/TR/css-cascade/#computed-value) CSS property `visibility` of `hidden`.
```

pages/glossary/hidden.md
```diff
@@ -0,0 +1,15 @@
...
+title: Explicitly Hidden
+key: hidden
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+An HTML element is _explicitly hidden_ if at least one of the following is true for itself or any of its ancestors:
```
WilcoFiers:
```suggestion
An HTML element is _hidden_ if at least one of the following is true for itself or any of its [ancestors][] in the [flat tree][]:
```

pages/glossary/hidden.md
```diff
@@ -0,0 +1,15 @@
...
+
+- has a `hidden` attribute; or
+- has a [computed](https://www.w3.org/TR/css-cascade/#computed-value) CSS property `display` of `none`; or
+- has a [computed](https://www.w3.org/TR/css-cascade/#computed-value) CSS property `visibility` of `hidden`.
```
WilcoFiers:
```suggestion
- has a [computed](https://www.w3.org/TR/css-cascade/#computed-value) CSS property `visibility` of `hidden`.
- has an `aria-hidden` attribute set to `true`
```

For semantic role to work it needs aria-hidden too.

pages/glossary/hidden.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Explicitly Hidden
```
WilcoFiers, line 2: Thinking about this a little more... I don't think I like having a "hidden" definition that isn't the opposite of the "visible" definition. We did that in axe-core and it's pretty confusing. I think we should rename this to "hidden state", and say it can be either true or false.

Jym77: Which is why I called it "explicitly hidden" in the first place https://github.com/act-rules/act-rules.github.io/pull/1189#discussion_r397726136 😆 
"hidden state" is also a good name.

Jym77:
Final Call ends on April 28th.

Jym77:

PR #1191, opened 2020-02-27
Tweak the assumption of link-acc-name

WilcoFiers:
Minor improvement to the link-acc-name assumption

Need for Final Call: none

### DIFF
--- a/_rules/link-accessible-name-c487ae.md
+++ b/_rules/link-accessible-name-c487ae.md
@@ -50,7 +50,7 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Assumptions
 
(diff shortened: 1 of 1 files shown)


_rules/link-accessible-name-c487ae.md
```diff
@@ -50,7 +50,7 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Assumptions
 
-- The rule assumes that all links are [user interface components](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components) as defined by WCAG 2. When the link role is used incorrectly, this assumption may not be true.
+The rule assumes that all links are [user interface components](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components) as defined by WCAG 2. When the link role is used on elements that are not real links, failing this rule might not mean the success criteria is failed.
```
jkodu:
```suggestion
The rule assumes that all links are [user interface components](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components) as defined by WCAG 2. When the link role is used on elements that are not native links, failing this rule might not mean the success criteria is failed.
```

PR #1193, opened 2020-02-27
Remove outdated class="target" from examples (b49b2e / a1b64e / ebe86a / 80af7b)

Jym77:
Some examples in a few rules still had a `class="target"` which was apparently an abandonned attempt to auto-highlight them in rule pages. Cleaning them up.

Closes issue(s):

- N/A

Need for Final Call:
This will require a 1 week Final Call (updating several examples, even though changes are editorial only…)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


adilsofficial:
In rule 80af7b Passed and Failed example 4 and 5, and rule ebe864a Passed and Failed example 1 and 2 still have `class="target"`. Is there special reason behind keeping this attribute in these example?

Jym77:
Good catch @adilsofficial My replacement pattern was tailored for when it appear in the middle of a line, not alone on a line of itself 😞

Jym77:
> This change is going to invalidate some implementations.

Yes.
1. Do you think it is better to wait until the rule is submitted to the TF and the TF asks us to remove this, thus invalidating the implementation later and delaying the publication even more?
2. We just decided to invalidate all implementations (https://www.w3.org/2020/02/27-act-r-minutes.html#item03) anyway…

jkodu:
Need to lose also `data-rule-target` - https://github.com/act-rules/act-rules.github.io/search?q=data-rule-target&unscoped_q=data-rule-target

PR #1194, opened 2020-02-28
Use `a` instead of `div role="link"` in examples of "Visible label is part of accessible name" (2ee8b8)

Jym77:
That rule had several examples with `div role="link"` and none with actual `a` elements… Making use of semantic HTML for good examples, especially since the rule is not about al the ways role can be given.

Closes issue(s):

- N/A

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
Final Call ends on March 09, 2020.

Jym77:

PR #1200, opened 2020-03-03
fix(674b10): update example to be inapplicable

jkodu:
`role` with only empty/ whitespace should be inapplicable and not a failure.

Closes issue(s):

- NA

Note:
- Final call 1 week


_rules/role-attribute-valid-value-674b10.md
```diff
@@ -132,6 +124,14 @@ Element with null `role` attribute.
 
 #### Inapplicable Example 4
 
+Element with role attribute that contains only whitespace.
+
+```html
+<input type="text" role=" " />
```
WilcoFiers: Agreed, but you'll need to update the applicability.

jkodu:
Closing this. It is tackled in a different PR.

Issue #1149, opened 2020-02-04
Update example for rule Form control has accessible name (e086e5)


jkodu:
Update failed example 4 - https://act-rules.github.io/rules/e086e5#failed-example-4, so that we still refer to an element that exists but is empty.

jkodu:
PR in place. Closing issue.

PR #1202, opened 2020-03-03
fix(e086e5): update failed example

jkodu:
Closes issue(s):

- https://github.com/act-rules/act-rules.github.io/issues/1149

Note:
- Final call 1 week


jkodu:
Final Call ends 11th March 2020. Can be merged then.

Issue #1125, opened 2020-01-24
None of the pass/fail tests for 'Image has accessible name' have non-text content


dd8:
This test maps to 1.1.1 Non-text Content (Level: A). There are two main cases used in the test:

- `img` with no `src` - has no image so can't have non-text content
-  an empty `div role=img` also doesn't have non-text content

There are no tests for the main use case of `img` with `src` and `div role=img` containing content. At best the test cases are testing edge conditions, and in some cases are doing the wrong thing because HTML specifies `img` with no `src` should not be rendered - see  https://github.com/act-rules/act-rules.github.io/issues/1124#issuecomment-578042900

dd8:
@jkodu I've added a PR for this, but can't add a label or assignees because I don't have write access to the report. Can you add the Rule label and assign whoever is appropriate?

jkodu:
PR in place - https://github.com/act-rules/act-rules.github.io/pull/1208, closing issue.

Issue #1124, opened 2020-01-23
Implementations that don't work can pass the 'Image has accessible name' tests


dd8:
There are no `src` attributes in any of the `img` element test cases. It's possible to create an implementation that  only checks the accessible name when the the `src` attribute is missing. 

An implementation like this would pass all the `img` test cases, but not work in practice.

https://act-rules.github.io/rules/23a2a8

dd8:
The behaviour of `<img/>` is specified in the HTML spec - with not `src` and no `alt` it should be ignored:

> If the src attribute is not set and either the alt attribute is set to the empty string or the alt attribute is not set at all
The element represents nothing

https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element

In Safari it is not exposed in the accessibility tree, but in Firefox it is.

WilcoFiers:
That's a completely valid point. These were some of the earlier rules we wrote. I don't think this would make it in today. But yeah, I agree the rule should look at whether or not there's an src attribute. Would you like to put together a pull request to change this?

dd8:
> That's a completely valid point. These were some of the earlier rules we wrote. I don't think this would make it in today. But yeah, I agree the rule should look at whether or not there's an src attribute. Would you like to put together a pull request to change this?

I'll try to put a PR together - I think it should look like this:

- change the applicability to say 'The rule applies to HTML img elements with a src attribute or any HTML element with the semantic role of img that is included in the accessibility tree.'
- add a `src` attribute to all the existing `img` examples (probably linked to /test-assets/shared/w3c-logo.png)
- create a new `<img/>` Inapplicable Example 

Does that look reasonable?

jkodu:
@dd8 thanks for offering to open a PR for this.

kasperisager:
Do be aware that `<img>` elements without a `src` attribute can still be used as images. Consider this:

```html
<img aria-label="Placeholder image" style="
    width: 100px;
    height: 100px;
    background-image: url(https://picsum.photos/200);
">
```

This will be presented to AT as an image with a name of "Placeholder image".

dd8:
Good point - also need to consider `srcset` and `picture`

```
<img srcset="photo-640x.png 2x">
```

```
<picture>
  <source type="image/svg+xml" srcset="pyramid.svg">
  <source type="image/webp" srcset="pyramid.webp"> 
  <img src="pyramid.png" alt="regular pyramid built from four equilateral triangles">
</picture>
```

[...]

PR #1208, opened 2020-03-05
Image has accessible name fixes

dd8:
Closes issue(s):
https://github.com/act-rules/act-rules.github.io/issues/1124
https://github.com/act-rules/act-rules.github.io/issues/1125

Need for Final Call:
This will require a 2 weeks Final Call due to substantial changes affecting most test cases

This PR:

- restricts the applicability to elements that display an image (to make `` inapplicable because it has no visual rendering, and is not voiced)
- removes `img alt=':-)' as a passed example because it fails 1.1.1
- changes `role=img` to `role=figure` in Inapplicable Example 1 because the example description is "The element *does not* have the semantic role of `img`"
- adds `img src` or `background-image` to most examples 
- adds `` as an inapplicable example
- adds additional failed examples like ``


dd8:
I don't have write access to the repo, so I can't:

_rules/image-accessible-name-23a2a8.md
```diff
@@ -35,7 +35,7 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements or any HTML element with the [semantic role][] of `img` that is [included in the accessibility tree][].
+The rule applies to HTML `img` elements or any HTML element with the [semantic role][] of `img` that is [included in the accessibility tree][] and displays an image.
```
Jym77: I think "displays an image" need to be clarified.
For example `` arguably does not display an image, but I think this is not what you intend to rule out.

dd8: How about changing:

> and displays an image

to
> and renders an image or an omitted image indicator

based on the rendering description here https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element:attr-img-src-5

Is that description precise enough?

A couple of options that won't work:

1) Using the [represents nothing](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element:attr-img-src-5) definition from HTML. I don't think that works if there's a CSS background-image and no src.

2) Using the [valid non-empty URL potentially surrounded by spaces](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#valid-non-empty-url-potentially-surrounded-by-spaces) definition from HTML. That doesn't work for inline SVG.

WilcoFiers: How about this?
> it's [current request][]'s [state][image request state] is not [completely available][]; or
I used it here: https://github.com/act-rules/act-rules.github.io/pull/1213/files#diff-f7f0c7b0159f5cafa41a1983a6d9389aR30

_rules/image-accessible-name-23a2a8.md
```diff
@@ -139,31 +132,58 @@ The HTML `img` element has an [accessible name][] that is not empty.
 The HTML `img` element is not [marked as decorative][] and has an empty [accessible name][].
 
 ```html
-<img />
+<img src="/test-assets/shared/w3c-logo.png" />
```
adilsofficial: Althought this code is considered `img` element with an empty accessibile name but I guess we can improve this example by adding `alt` attribute. What do you think?
```suggestion
<img src="/test-assets/shared/w3c-logo.png" alt=""/>
```

daniel-montalvo: But that would break the example, as `alt=""` would make the image decorative, which we don't want for this.

dd8:
I'll take a look at the changes requested when time allows - probably in a couple of weeks. Other stuff taking priority at the moment due to Covid-19

dd8:
Getting back to this - this rule has been changed a lot since the original PR in March (including being renamed, and some of the changes from the PR picked up in earlier merge).

I think it might be clearer to open a new PR and refer back to this one (for requested changes). Anyone disagree?

Jym77:
> I think it might be clearer to open a new PR and refer back to this one (for requested changes). Anyone disagree?

Do it. It will be much easier for everybody (especially for you).

dd8:
Yep, makes sence, new PR is https://github.com/act-rules/act-rules.github.io/pull/1382

I don't have rights to assign reviewers, so can someone do that?

WilcoFiers:
Handled in #1382, closing this PR.

PR #1213, opened 2020-03-06
new rule: Image not in the accessibility tree is decorative

WilcoFiers:
This rule checks that visible img, svg and canvas elements that are ignored by assistive technologies are decorative.

Need for Final Call: 2 weeks


_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+Any `img`, `canvas` or `svg` element that is [visible][] and is not [included in the accessibility tree][], except if one of the following is true:
+
+- **unloaded img**: the element is an `img` and it's [current request][]'s [state][image request state] is not [completely available][]; or
+- **ignored svg**: the element is an `svg` with an empty (`""`) [accessible name][] and a [semantic role][] of `graphics-document`; or
+- **ignored canvas**: the element is a `canvas` with an empty (`""`) [accessible name][] and no [explicit semantic role][]; or
```
carlosapaduarte: I understand the assumption, but I can't really understand why should these be inapplicable. If they are not in the accessibility tree, shouldn't they be descriptive?

Btw, I tested an `svg` with VoiceOver on Firefox and it was not ignored. It was ignored with VoiceOver on Chrome.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+- **named from author**: the element is a [descendant][] in the [flat tree][] of an element that is [named from author][].
+
+**Note**: An example of an image ignored because of "named from author" is when it in a `button` element that uses `aria-label` for its accessible name.
```
carlosapaduarte:
```suggestion
**Note**: An example of an image ignored because of "named from author" is when the image is a descendant of a `button` element that uses `aria-label` for its accessible name.
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Expectation
+
+Each test target is [purely decorative][].
+
+**note**: It is relatively common for an informative image such as an icon to be marked up as decorative, if the text alternative is adjacent to the image. This is a [conforming alternative version][] for the image. This fails the rule, but meets [conformance requirement 1 of WCAG 2.1](https://www.w3.org/TR/WCAG21/#cc1).
```
carlosapaduarte:
```suggestion
**Note**: It is relatively common for an informative image such as an icon to be marked up as decorative, if the text alternative is adjacent to the image. This is a [conforming alternative version][] for the image. This fails the rule but meets [conformance requirement 1 of WCAG 2.1](https://www.w3.org/TR/WCAG21/#cc1).
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Assumptions
+
+- `svg` elements with a [semantic role][] of `graphics-document` and with an empty (`""`) [accessible name][] are ignored by assistive technologies tested for this rule. If some assistive technology does not ignore these elements, and that assistive technology is required for conformance, passing this rule does not ensure all decorative `svg` elements can be ignored, and the [success criterion 1.1.1 Non-text content][] may still not be satisfied. The same is true for `canvas` elements with no [semantic role][] and an empty (`""`) [accessible name][].
+
+- A web page with informative images without an [accessible name][] may conform to WCAG 2.1 Level A when the information provided by that image is available elsewhere, either on the web page itself or on a [conforming alternative version][] of the page. For example if an equivalent text is adjacent to the image, or if the text alternative is included in the [accessible name][] of a parent element.
```
carlosapaduarte: I don't think this is an assumption for this rule. It definitely is a note (and this information is partially on the expectation note already). But we're not assuming anything with this.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Accessibility Support
+
+_No accessibility support issues known._
```
carlosapaduarte: Not really sure. In my quick test, different combinations of browser + AT resulted in different outcomes for the `svg` with empty `alt`.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+- [H67: Using null alt text and no title attribute on img elements for images that AT should ignore](https://www.w3.org/WAI/WCAG21/Techniques/html/H67.html)
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This decorative `img` element is ignored by assistive technologies because of the `alt` attribute is empty (`""`).
```
carlosapaduarte:
```suggestion
This decorative `img` element is ignored by assistive technologies because the `alt` attribute is empty (`""`).
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+_No accessibility support issues known._
+
+## Background
+
+- [H67: Using null alt text and no title attribute on img elements for images that AT should ignore](https://www.w3.org/WAI/WCAG21/Techniques/html/H67.html)
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+
+## Test Cases
+
+### Passed
```
carlosapaduarte: The descriptions of the passed examples are written in a way that seems the expectation is that the test target is ignored by AT, when the real expectation is that the test target is decorative. For example, shouldn't the first passed example be something like:
"This `img` element with an empty `alt` attribute which leaves it out of the accessibility tree is decorative."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+This decorative `img` element is ignored by assistive technologies because it has an [explicit semantic role][] of `none`.
+
+```html
+Happy new year!<br />
+<img src="/test-assets/shared/fireworks.jpg" role="none" alt="ignore me" />
+```
+
+#### Passed Example 4
+
+This decorative `svg` element is ignored by assistive technologies because it has no attribute that would give it an [accessible name][].
```
carlosapaduarte: Firefox + VoiceOver does not ignore it
But my point is that the applicability is about being in the accessibility tree, not being ignored by assistive technology (that should be hadled by the Accessibility Support section)

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+```html
+Happy new year!<br />
+<svg height="200" xmlns="http://www.w3.org/2000/svg">
+	<polygon points="100,10 40,180 190,60 10,60 160,180" fill="yellow" />
+</svg>
+```
+
+#### Passed Example 5
+
+This decorative `canvas` element is ignored by assistive technologies because it has no attribute that would give it an [accessible name][].
```
carlosapaduarte: Once again, Firefox + VoiceOver does not ignore it
Same point about the applicability being about the accessibility tree

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+		position: absolute;
+		top: -9999em;
+	}
+</style>
+<img src="/test-assets/shared/w3c-logo.png" alt="" />
+```
+
+#### Inapplicable Example 4
+
+This `svg` is ignored because it is a child of a link that has provides its [accessible name][].
```
carlosapaduarte:
```suggestion
This `svg` is ignored because it is a child of a link that provides its [accessible name][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<a href="//example.org" aria-label="SVG star">
+	<svg height="200" xmlns="http://www.w3.org/2000/svg">
+		<polygon points="100,10 40,180 190,60 10,60 160,180" fill="yellow" />
+	</svg>
+</a>
+```
+
+#### Inapplicable Example 5
```
carlosapaduarte: Missing description

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Inapplicable Example 6
+
+This `canvas` is not [visible][] because it is completely transparent.
+
+```html
+<canvas width="200" height="200"></canvas>
+```
+
+#### Inapplicable Example 7
```
carlosapaduarte: Missing description

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+[visible]: #visible 'Definition of Visible'
+[included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of Included in the accessibility tree'
+[semantic role]: #semantic-role 'Definition of Semantic Role'
+[explicit semantic role]: #explicit-role 'Definition of Explicit semantic role'
+[purely decorative]: https://www.w3.org/TR/WCAG21/#dfn-pure-decoration 'WCAG definition of Pure decoration'
+[success criterion 1.1.1 non-text content]: https://www.w3.org/TR/WCAG21/#non-text-content
+[conforming alternative version]: https://www.w3.org/TR/WCAG21/#dfn-conforming-alternate-version 'WCAG definition of Conforming alternative version'
+[flat tree]: https://drafts.csswg.org/css-scoping/#flat-tree 'CSS Scoping definition of Flat tree, working draft'
+[descendant]: https://dom.spec.whatwg.org/#concept-tree-descendant 'DOM definition of Descendant, 2020/03/06'
+[named from author]: https://www.w3.org/TR/wai-aria-1.1/#namecomputation 'WAI-ARIA definition of Named from author'
```
carlosapaduarte: Just to make sure that you really wanted to use that URL instead of this one:
https://www.w3.org/TR/wai-aria/#namecalculation

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+```html
+<canvas id="w3c" width="200" height="60"></canvas>
+<script>
+	const ctx = document.querySelector('#w3c').getContext('2d')
+	ctx.font = '30px Arial'
+	ctx.fillText('ACT Rules!', 20, 40)
+</script>
+```
+
+### Inapplicable
```
carlosapaduarte: Include an example with an image inserted through css `background-image`

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+    - Wikimedia
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+Any `img`, `canvas` or `svg` element that is [visible][] and is not [included in the accessibility tree][], except if one of the following is true:
```
adilsofficial: Just a simple question, does this rule check the elements with sementic role of image?

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
adilsofficial: I agree with @carlosapaduarte and we need to rephrase this description because it's not in line with the expectation of the rule. It could may be like:
"This `img` element with an empty `alt` attribute which is not included in accessibility tree is decorative."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: It would help to say this is the **named from author** exception (given that the items are named, better use these names…)

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: Missing examples for the **ignored svg**, **ignored canvas** and **unloaded image** exceptions.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: I am not sure that the note is entirely needed given that there is an inapplicable example to illustrate this.
But it’s not harmful either…

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
adilsofficial: I agree with @carlosapaduarte and we need to rephrase the Passed examples description because it's not in line with the expectation of the rule, it seems that the decorative images are being tested whether image is ignored by AT or not. It could may be like:
"This `img` element with an empty `alt` attribute which is not included in accessibility tree is decorative."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
adilsofficial: as mentioned earlier we need to rephrase examples description in a way which is inline with expectation of this rule. for example:
"This `img` element is not included in acccessibility tree and is decorative because `alt` attribute is empty (`""`)."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+<img src="/test-assets/shared/fireworks.jpg" alt="" />
+```
+
+#### Passed Example 2
+
+This decorative `img` element is ignored by assistive technologies because `aria-hidden` is set to `true`.
+
+```html
+Happy new year!<br />
+<img src="/test-assets/shared/fireworks.jpg" aria-hidden="true" alt="decorative" />
```
adilsofficial: Even though this `img` element is ignored because of `aria-hidden`, `alt` attribute should be empty if the sourced image is decorative.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+```
+
+#### Passed Example 3
+
+This decorative `img` element is ignored by assistive technologies because it has an [explicit semantic role][] of `none`.
+
+```html
+Happy new year!<br />
+<img src="/test-assets/shared/fireworks.jpg" role="none" alt="ignore me" />
```
adilsofficial: Even though this `img` element is ignored because of `role="none"`, `alt` attribute should be empty if the sourced image is decorative.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
adilsofficial: I guess this rule is not to check whether element have accessible name or not? We should rephrase this description to meet the expectation and also for similar examples. Something like that:
"This `svg` element is not included in the accessibility tree and is a decorative because it do not contain an attribute which provide alternative text to this element."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
adilsofficial: Consider same comment from [Passed Example 4](https://github.com/act-rules/act-rules.github.io/pull/1213/files#r391504210)

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+	ctx.moveTo(100, 10)
+	ctx.lineTo(40, 180)
+	ctx.lineTo(190, 60)
+	ctx.lineTo(10, 60)
+	ctx.lineTo(160, 180)
+	ctx.fill()
+</script>
+```
+
+### Failed
```
adilsofficial: If we used word "purely decorative" here in failed example then we should used the same word in passed and inapplicable examples.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Failed Example 1
+
+This `img` element with an empty (`""`) `alt` is not [purely decorative][].
```
adilsofficial: https://github.com/act-rules/act-rules.github.io/pull/1213/files#r391504210

adilsofficial: If we used word "purely decorative" here in failed examples then we should use the same worh in passed and inapplicable example to make the descriptions consistent.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<img src="/test-assets/shared/w3c-logo.png" alt="" />
+```
+
+#### Failed Example 2
+
+This `img` element with `aria-hidden` set to `true` is not [purely decorative][].
```
adilsofficial: Need more desciption may be something like that:
"This `img` element is visible and not included in accessibility tree because of `aria-hidden="true"` but is not purely decorative."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<img src="/test-assets/shared/w3c-logo.png" aria-hidden="true" alt="W3C logo" />
+```
+
+#### Failed Example 3
+
+This `img` element with an [explicit semantic role][] of `none` is not [purely decorative][].
```
adilsofficial: "This img element with an [explicit semantic role][] of `none` is visible and not included in accessibility tree, is not [purely decorative][]."

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+	ctx.font = '30px Arial'
+	ctx.fillText('ACT Rules!', 20, 40)
+</script>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+This `img` is [included in the accessibility tree][] because the `alt` attribute is not empty (`""`).
```
adilsofficial:
```suggestion
This `img` element is [included in the accessibility tree][] and is a non-decorative image because the `alt` attribute is not empty (`""`).
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<img src="/test-assets/shared/w3c-logo.png" alt="W3C" />
+```
+
+#### Inapplicable Example 2
+
+This `img` is neither [visible][] nor [included in the accessibility tree][].
```
adilsofficial:
```suggestion
This `img` element is neither [visible][] nor [included in the accessibility tree][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<img src="/test-assets/shared/w3c-logo.png" style="display:none" alt="" />
+```
+
+#### Inapplicable Example 3
+
+This `img` is not [visible][] because it is positioned off screen.
```
adilsofficial:
```suggestion
This `img` element is not [visible][] because it is positioned off screen.
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<style>
+	img {
+		position: absolute;
+		top: -9999em;
+	}
+</style>
+<img src="/test-assets/shared/w3c-logo.png" alt="" />
```
adilsofficial: aIf the image is non-decorative it should have `alt` attribute with text as a best practice.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
adilsofficial:
```suggestion
This `svg` element is ignored because this element is a descendant of a link that provides its [accessible name][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Inapplicable Example 5
+
+```html
+<svg role="img" aria-label="Hello"></svg>
+```
+
+#### Inapplicable Example 6
+
+This `canvas` is not [visible][] because it is completely transparent.
```
adilsofficial:
```suggestion
This `canvas` element is not [visible][] because it is completely transparent.
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Inapplicable Example 7
+
+```html
+<canvas role="img" aria-label="Hello"></canvas>
+```
+
+#### Inapplicable Example 8
+
+This `img` is [included in the accessibility tree][].
```
adilsofficial:
```suggestion
This `img` element is [visible][] but [included in the accessibility tree][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Test data for passed example 4 and 5:

- VO + Safari: ignored canvas and svg
- VO + Chrome: ignored canvas and svg
- JAWS + Chrome: ignored canvas, and svg
- JAWS + IE: ignored canvas and svg
- NVDA + Firefox: ignore canvas and svg

Seems pretty consistent. I'm going to not count VO + Firefox because it is not a supported browser for VO. I'll have to take out the `br` elements through, they seem to throw off AT as they can get announced as "blank". That might have been what you noticed.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I have some test data for it here, this appears to function consistently as far as I can tell: https://github.com/act-rules/act-rules.github.io/pull/1213#discussion_r394480194

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Doesn't really matter, but I'll change this to use an empty alt with role=img. Either way I do need to push the boundaries a bit of different ways that an image can be hidden. If all of them have an empty alt, this rule isn't testing all the things it needs to.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I agree, fixing this.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Firefox + VO is not something either Firefox of VO supports, I think that's a safe one to ignore. Others seem to work fine though.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Why? This rule isn't about background images.

WilcoFiers: Resolved with inapplicable examples 5, 7 and 9

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -31,18 +31,18 @@ htmlHintIgnore:
...
 
-**Note**: An example of an image ignored because of "named from author" is when it in a `button` element that uses `aria-label` for its accessible name.
+**Note**: An example of an image ignored because of "named from author" is when the image is a descendant of a `button` element that uses `aria-label` for its accessible name.
 
 ## Expectation
 
 Each test target is [purely decorative][].
 
-**note**: It is relatively common for an informative image such as an icon to be marked up as decorative, if the text alternative is adjacent to the image. This is a [conforming alternative version][] for the image. This fails the rule, but meets [conformance requirement 1 of WCAG 2.1](https://www.w3.org/TR/WCAG21/#cc1).
+**Note**: It is relatively common for an informative image such as an icon to be marked up as decorative, if the text alternative is adjacent to the image. This is a [conforming alternative version][] for the image. This fails the rule but meets [conformance requirement 1 of WCAG 2.1](https://www.w3.org/TR/WCAG21/#cc1).
```
Jym77: Shouldn't this note be moved after the second assumption? It looks like they are about the same kind of stuff (namely, assuming that there is no conforming alternative to the image)

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: Inapplicable examples 5 and 7 are missing description.
And there doesn't seem to be any inapplicable example 9.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: You are right it is related, but I think this is most relevant in the expectation. Would prefer to keep it here.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Darnit, my bad :( forgot to push.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -65,48 +65,48 @@ _No accessibility support issues known._
 
 #### Passed Example 1
 
-This decorative `img` element is ignored by assistive technologies because of the `alt` attribute is empty (`""`).
+This `img` element with an empty `alt` attribute which is not included in accessibility tree is decorative.
```
adilsofficial: Shouldn't we use same words for all examples?
```suggestion
This `img` element with an empty `alt` attribute which is not [included in the accessibility tree][] is [purely decorative][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,300 @@
...
+		position: absolute;
+		top: -9999em;
+	}
+</style>
+<img src="/test-assets/shared/w3c-logo.png" alt="" />
+```
+
+#### Inapplicable Example 4
+
+This `svg` element is ignored because it is a child of a link that has provides its [accessible name][].
```
carlosapaduarte:
```suggestion
This `svg` element is ignored because it is a child of a link that provides its [accessible name][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
carlosapaduarte: Because the rule it's about decorative images. And background-images are a way to display decorative images. This would show that the rule does not apply to background-images. 
But I'm not going to insist on this.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
carlosapaduarte: Plus, I'm missing something here. Can't find any document that says that `svg`s get their name from content when they're child of links. And both Firefox and Chrome show the `svg` with an empty accessible name in their accessibility trees.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: I agree there is something funky with the applicability

I read it the same way as @carlosapaduarte, but @WilcoFiers's answer "[these] elements that are included in the accessibility tree but (…) ignored by AT" makes little sense with that interpretation: if these elements are included in the accessibility tree, then they are already inapplicable and don't need to be specifically excluded.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Assumptions
+
+- `svg` elements with a [semantic role][] of `graphics-document` and with an empty (`""`) [accessible name][] are ignored by assistive technologies tested for this rule. If some assistive technology does not ignore these elements, and that assistive technology is required for conformance, passing this rule does not ensure all decorative `svg` elements can be ignored, and the [success criterion 1.1.1 Non-text content][] may still not be satisfied. The same is true for `canvas` elements with no [semantic role][] and an empty (`""`) [accessible name][].
+
+- A web page with informative images without an [accessible name][] may conform to WCAG 2.1 Level A when the information provided by that image is available elsewhere, either on the web page itself or on a [conforming alternative version][] of the page. For example if an equivalent text is adjacent to the image, or if the text alternative is included in the [accessible name][] of a parent element.
```
Jym77, line 51: As a general note (not just for this rule), I'd like assumptions to state clearly and precisely (as unambiguously as possible) what is assumed, before explaining the reasons or consequence for it (we haven't really standardised how assumptions should look like…)

The reason for this is that I believe it would make testing more transparent. If a rule fails but the page conforms to WCAG, we need to be able to precisely pinpoint whether "this assumption is false for this page, so the rule cannot be applied" or "oops, there is a bug in the rule". Having a clear and precise statement of the assumption will help testers reach the same conclusion.

---
In this case, I'd suggest:
```suggestion
## Assumptions

- This rule assumes that `svg` elements with a [semantic role][] of `graphics-document` and with an empty (`""`) [accessible name][] are ignored by assistive technologies. This is the case for all assistive technologies tested for this rule. If some assistive technology does not ignore these elements, and that assistive technology is required for conformance, passing this rule does not ensure all decorative `svg` elements can be ignored, and the [success criterion 1.1.1 Non-text content][] may still not be satisfied. The same is true for `canvas` elements with no [semantic role][] and an empty (`""`) [accessible name][].

- This rule assumes that there is no [conforming alternative version][] of the page or the image being tested. A web page with informative images without an [accessible name][] may conform to WCAG 2.1 Level A when the information provided by that image is available elsewhere, either on the web page itself or on a [conforming alternative version][] of the page. For example if an equivalent text is adjacent to the image, or if the text alternative is included in the [accessible name][] of a parent element.
```

Jym77, line 51: Also, the first assumption is a case of "the rule pass but the SC fails", which is not good but still OK. The second one is a case of "the rule fails but the SC pass" which is bad and which is why that assumption is more needed than the first one.

Not sure there is anything to do about it, however…

Jym77, line 51: As much as I agree we need the "no conforming alternative version of the image" part, I am not sure we need the "no conforming alternative version of the *page*" part.

Ain't all rules assuming that the test subject is what need to be tested in the first place and that the is no conforming alternative version of it?

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,300 @@
...
+This `img` element with an empty `alt` attribute which is not included in accessibility tree is decorative.
+
+```html
+<p>Happy new year!</p>
+<img src="/test-assets/shared/fireworks.jpg" alt="" />
+```
+
+#### Passed Example 2
+
+This `img` element that is ignored by assistive technologies because `aria-hidden` is set to `true` is [purely decorative][].
```
Jym77, line 77: I'd like to have example descriptions stick closer to the applicability/expectation.
```suggestion
This `img` element, that is [visible][] and not [included in the accessibility tree][] (because `aria-hidden` is set to `true`), is [purely decorative][].
```
This goes for more or less all passed examples descriptions.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: I fail to see how examples 5 and 7 illustrate the **ignored svg** and **ignored canvas** exceptions. This may have to do with misunderstanding around the applicability…

* Example 5 is an `svg` which is included in the accessibility tree (with `aria-label`) and thus inapplicable from the start. Additionally, it has an explicit role of `img` while the **ignored svg** exception is about semantic role of `graphics-document`.
* Example 7 is a `canvas` which is included in the accessibility tree (with `aria-label`) and thus inapplicable from the start. Additionally, it has an explicit role while the **ignored canvas** exception is about `canvas` without semantic role.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I disagree. The "Assumption" heading should be sufficient for that in my opinion.

WilcoFiers: The first one is an edge case, but if someone intends for the SVG to be announced because it isn't decorative, and in some cases that does happen, this rule might flag it as a failure even thought it might not be.

WilcoFiers: You are correct, rules generally do not take confirming alternative versions into account. I talked with Brian about this for quite some time. The issue is something like this:

```html
<img src=".." alt="pdf" />Hello world PDF
```

Some folks call this out as a failure. Their argument is that the image is redundant, and thus decorative and fails because it won't be ignored by assistive technologies. That's not the stance we're taking for this rule. Here we're saying that even if the information is redundant, the image is still informative.  If you change the alt to an empty alt, a conforming alternative version is needed. Problem there is CAV is rather vague, and depending on how you read it, might only be applicable to different pages. We're making an assumption in this rule that this isn't the case, that a CAV can be in the same page, and that this interpretation on redundant information not being decorative is correct.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I don't think this is necessary. We don't need to repeat the language from the applicability over and over in the example descriptions. The short version is it gets ignored by AT, which is what it says.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I don't understand what you mean, can you give an example?

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Assuming this is addressed, feel free to unresolve if it isn't.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Jup, rereading this, I see your point. Fixing it.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I think it's because of the applicability. Please give it another read.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Fair. I'll change the content of the image so it's actually a decorative image.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: Yes, this is an example of a case of "conforming alternative version on the same page" case which I agree is needed.

What I'm saying is that the "no conforming alternative version **of the full page**" part is not needed.
Essentially, if we have
```html
page1.html:
<img src="important-image.jpg" alt="" />

page1-conforming-alternative.html:
<img src="important-image.jpg" alt="A descriptive alt text" />
```
Then the rule is going to fail on `page1.html`, but the full thngie is good because there is a conforming alternative version of it, so the failure of the rule is due to the assumption "no conforming alternative version of the page" being false.

But, this is essentially the case for every single rule. If the rule is tested on a page which has a conforming alternative version of it, then yeah, the rule is going to fail.

So far, I considered that
> A web page [failing this rule] may conform to WCAG 2.1 when the information is available elsewhere, on a [conforming alternative version][] of the page.

was essentially a global assumption made by every single rule and doesn't need to be repeated in each of them.
(e.g. "A web page with no `title` may conform to WCAG if there is a conforming alternative version of the page with a `title`", and there is actually no need to put that assumption in the title rules)

That's why I'm suggesting restricting this assumption to
```suggestion
- `svg` elements with a [semantic role][] of `graphics-document` and with an empty (`""`) [accessible name][] are ignored by assistive technologies tested for this rule. If some assistive technology does not ignore these elements, and that assistive technology is required for conformance, passing this rule does not ensure all decorative `svg` elements can be ignored, and the [success criterion 1.1.1 Non-text content][] may still not be satisfied. The same is true for `canvas` elements with no [semantic role][] and an empty (`""`) [accessible name][].

- A web page with informative images without an [accessible name][] may conform to WCAG 2.1 Level A when the information provided by that image is available elsewhere on the web page itself. For example if an equivalent text is adjacent to the image, or if the text alternative is included in the [accessible name][] of a parent element.
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+<img src="/test-assets/shared/w3c-logo.png" role="none" alt="W3C logo" />
+```
+
+#### Failed Example 4
+
+This `svg` element which has a [semantic role][] of `graphics-document` and an empty (`""`) [accessible name][] is not [purely decorative][].
+
+```html
+<p>Best W3C logo:</p>
+<svg viewBox="0 0 512 512">
```
adilsofficial, line 159: The semantic role of `graphics-document` is missing in this code: `role="graphics-document"`

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+## Applicability
+
+Any `img`, `canvas` or `svg` element that is [visible][] and for which one of the following is true:
+
+- **excluded**: The element is not [included in the accessibility tree][]; or
+- **ignored svg**: The element is an `svg` with an empty (`""`) [accessible name][] and a [semantic role][] of `graphics-document`; or
+- **ignored canvas**: The element is a `canvas` with an empty (`""`) [accessible name][] and no [explicit semantic role][]; or
+- **named from author**: The element is a [descendant][] in the [flat tree][] of an element that is [named from author][].
+
+**Exception**: Exclude any `img` element where the [current request][]'s [state][image request state] is not [completely available][]; or
```
Jym77, line 40:
```suggestion
**Exception**: Exclude any `img` element where the [current request][]'s [state][image request state] is not [completely available][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+
+**Note**: An example of an image ignored because of "named from author" is when the image is a descendant of a `button` element that uses `aria-label` for its accessible name.
```
Jym77, line 42:
```suggestion
**Note**: An example of an image ignored because of "**named from author**" is when the image is a descendant of a `button` element that uses `aria-label` for its accessible name.
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+		fill="#ebebeb"
+		d="M142 176.3h114v45.4h-64.2l4.2 46.5h60v45.3H154.4M156.4 336.3H202l3.2 36.3 50.8 13.6v47.4l-93.2-26"
+	/>
+	<path fill="#fff" d="M369.6 176.3H255.8v45.4h109.6M361.3 268.2H255.8v45.4h56l-5.3 59-50.7 13.6v47.2l93-25.8" />
+</svg>
+```
+
+#### Failed Example 5
+
+This `canvas` element which has no [semantic role][] and an empty (`""`) [accessible name][] is not [purely decorative][].
```
adilsofficial, line 175: Do you mean "[semantic role][] of `graphics-document`" here? if yes update this description and add `role="graphics-document"` in the code below as well.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+_No accessibility support issues known._
+
+## Background
+
+- [H67: Using null alt text and no title attribute on img elements for images that AT should ignore](https://www.w3.org/WAI/WCAG21/Techniques/html/H67.html)
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)
+
+## Test Cases
+
+### Passed
```
Jym77, line 67: We need a Passed (or Failed) Example to illustrate the **named from author** case of the applicability.
(or maybe not, it might be the applicability that needs fixing)

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+
+This `svg` element is ignored because it is a child of a link that provides its [accessible name][].
+
+```html
+<a href="//example.org" aria-label="SVG star">
+	<svg height="200" xmlns="http://www.w3.org/2000/svg">
+		<polygon points="100,10 40,180 190,60 10,60 160,180" fill="yellow" />
+	</svg>
+</a>
+```
```
Jym77, line 228: Currently, this is applicable.
I guess the **named from author** part of the applicability was meant to **exclude** stiff, not to **include** them.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+---
+
+## Applicability
+
+Any `img`, `canvas` or `svg` element that is [visible][] and for which one of the following is true:
+
+- **excluded**: The element is not [included in the accessibility tree][]; or
+- **ignored svg**: The element is an `svg` with an empty (`""`) [accessible name][] and a [semantic role][] of `graphics-document`; or
+- **ignored canvas**: The element is a `canvas` with an empty (`""`) [accessible name][] and no [explicit semantic role][]; or
+- **named from author**: The element is a [descendant][] in the [flat tree][] of an element that is [named from author][].
```
Jym77, line 38: Currently, images that are **named from author** ar applicable. This seems to not be what was intended (judging from Inapplicable example 4). If so, Applicability needs to be corrected.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+	height: 260px;
+	background: url(/test-assets/shared/fireworks.jpg) no-repeat;
+"
+></div>
+```
+
+#### Inapplicable Example 10
+
+This `img` element has an `src` attribute which will cause the [image request state][] to be [Broken
+](https://html.spec.whatwg.org/#img-error).
```
Jym77, line 298:
```suggestion
This `img` element has an `src` attribute which will cause the [image request state][] to be [Broken](https://html.spec.whatwg.org/#img-error).
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: Yes, that's good 👍 
(and I also understand why Inapplicable Examples 5 and 7 were not here from the start, they're kinda not completely needed…)

adilsofficial: I agree with @Jym77 about examples 5 and 7. Please take another look at exceptions and examples.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,318 @@
...
+
+**Note**: While it might be better for the PDF icon to be ignored by assistive technologies, because assistive technologies will announce "PDF" twice, the image is not [purely decorative][]. Having assistive technologies ignore it is not required by [Success Criterion 1.1.1 Non-text content][].
+
+```html
+<img src="/test-assets/shared/pdf-icon.png" alt="PDF" /> PDF document
+```
+
+#### Inapplicable Example 9
+
+This is a `div` element with a background image.
```
adilsofficial, line 282: Add more description for this example as it does not make sense that a `div` element with a background image so what exactly makes this example inapplicable? Maybe with a note that this rule does not apply to background images.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: It's the implicit role for `svg` elements.
https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: Fixed the applicability. Please check again.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: Keeping the same typography (bold) than in the initial list to make it obvious that it is the same thing.
(i.e., consistency in typography).

(I won't insist if you feel it's not needed)

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,319 @@
...
+Any `img`, `canvas` or `svg` element that is [visible][] and for which one of the following is true:
+
+- **excluded**: The element is not [included in the accessibility tree][]; or
+- **ignored svg**: The element is an `svg` with an empty (`""`) [accessible name][] and a [semantic role][] of `graphics-document`; or
+- **ignored canvas**: The element is a `canvas` with an empty (`""`) [accessible name][] and no [explicit semantic role][]; or
+
+**Exception**: Elements for which one of the following is true are always inapplicable:
+
+- The element has an [ancestor][] in the [flat tree][] of an element that is [named from author][]; or
+- The element is an `img` element where the [current request][]'s [state][image request state] is not [completely available][].
```
Jym77, line 42: Should we also name these items?
```suggestion
- **named from author** The element has an [ancestor][] in the [flat tree][] of an element that is [named from author][]; or
- **unavailable** The element is an `img` element where the [current request][]'s [state][image request state] is not [completely available][].
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Applicability
+
+Any `img`, `canvas` or `svg` element that is [visible][] and for which one of the following is true:
+
+- **excluded**: The element is not [included in the accessibility tree][]; or
+- **ignored svg**: The element is an `svg` with an empty (`""`) [accessible name][] and a [semantic role][] of `graphics-document`; or
+- **ignored canvas**: The element is a `canvas` with an empty (`""`) [accessible name][] and no [explicit semantic role][]; or
+
+**Exception**: Elements for which one of the following is true are always inapplicable:
```
Jym77, line 39: I find the word "inapplicable" a bit confusing here given that it could also be the outcome of the rule (in which case it is a global outcome).

I do agree, however, that it is technically correct to use it according to https://www.w3.org/TR/act-rules-format/#applicability
> [Applicability] must contain a precise description of the parts of the test subject to which the rule applies.

Not sure this really needs to change… Maybe
```suggestion
**Exception**: This rule never applies to elements for which one of the following is true:
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
Jym77: The first post of this conversation was due to the Applicability.

With the current Applicability, I consider that this conversation has been solved. If @adilsofficial has more to say on the topic, I'll let him handle it 😃

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,319 @@
...
+
+- The element has an [ancestor][] in the [flat tree][] of an element that is [named from author][]; or
```
carlosapaduarte, line 41:
```suggestion
- The element has an [ancestor][] in the [flat tree][] that is [named from author][]; or
```

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
WilcoFiers: I'm going to keep it as is.

_rules/image-not-in-acc-tree-is-decorative-e88epe.md
```diff
@@ -0,0 +1,319 @@
...
+- **excluded**: The element is not [included in the accessibility tree][]; or
+- **ignored svg**: The element is an `svg` with an empty (`""`) [accessible name][] and a [semantic role][] of `graphics-document`; or
+- **ignored canvas**: The element is a `canvas` with an empty (`""`) [accessible name][] and no [explicit semantic role][]; or
+
+**Exception**: This rule never applies to elements for which one of the following is true:
+
+- The element has an [ancestor][] in the [flat tree][] that is [named from author][]; or
+- The element is an `img` element where the [current request][]'s [state][image request state] is not [completely available][].
+
+**Note**: An example of an image ignored because of "named from author" is when the image is a descendant of a `button` element that uses `aria-label` for its accessible name.
```
Jym77, line 44: If items in the list above are not named, you can't refer to"named from author" like that. Especially given that the meaning here is not the meaning in ARIA (here, "named from author" means that one ancestor is "named from author in the ARIA meaning").

Issue #1130, opened 2020-01-29
Applicability of  'language is valid' rules doesn't match spec (de46e4)


dd8:
The HTML standard says:

> Its value must be a valid BCP 47 language tag, or the empty string
https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes

BCP-47 says:
>Whitespace is not permitted in a language tag. 
https://tools.ietf.org/html/bcp47#section-2.1

The W3 validator flags `lang=' '`  as an error.
https://validator.w3.org/nu/?doc=https%3A%2F%2Fact-rules.github.io%2Ftestcases%2Fde46e4%2Fe0a8e368d7aba4e94771903d5e994766f6548348.html

The applicability in the lang is valid rules says
>has a lang attribute that is neither empty ("") nor only ASCII whitespace; and

`lang=''` does match spec, but `lang=' '` is not allowed by spec and causes undefined AT behaviour:

```
<html lang='en'>
    <div lang='fr'>
        <p lang=" ">1 2 3 4 5 <code aria-hidden='true'>lang=' '</code></p>
    </div>
</html>
```

* Voiceover macOS 10.15 with Safari 13 - speaks lang=' ' as English
* NVDA 2019.2.1 (eSpeak) with FF68 - speaks lang='' as English
* NVDA 2019.2.1 (eSpeak) with IE11 - speaks lang=' ' as English
* NVDA 2019.2.1 (eSpeak) with Chrome 79 - lang=' ' as French
* JAWS 2019.1912.1 with FF68 - speaks lang=' ' as French
* JAWS 2019.1912.1 with IE11 - speaks lang=' ' as English
* JAWS 2019.1912.1 with Chrome 79 - speaks lang=' ' as French

Test results here:
https://qa.powermapper.com/Tests/ACT-R/primary-ext-lang.html

I suggest changing the applicability both lang valid to:

>has a lang attribute that is neither empty (""); and

and making inapplicable-example-3 a failure:
https://act-rules.github.io/rules/de46e4#inapplicable-example-3

WilcoFiers:
Not matching the spec doesn't matter much, but that some AT switch back to the page default language is definitely an issue, and (to me) rather surprising!

Thank you for the test data. That really helps make a case here. Do you happen to know if the same behaviour occurs with empty `lang` attributes or with `lang` attributes without value? That seems relevant here too.

dd8:
[...]

PR #1214, opened 2020-03-09
fix: applicability and examples to not consider `whitespace` (de46e4)

jkodu:
Based on following the discussion in #1130 the below changes have been made.

1) Update applicability to not consider whitespace
2) Update examples accordingly.

Closes issue(s):
- https://github.com/act-rules/act-rules.github.io/issues/1130

Need for Final Call:
- 1 week


jkodu:
Final call ends on March 18th, 2020.

PR #1218, opened 2020-03-09
Improve image-alt description

WilcoFiers:
The description made it sound as though decorative images were out of scope, which they are not.

Need for Final Call: nope


_rules/image-accessible-name-23a2a8.md
```diff
@@ -3,7 +3,7 @@ id: 23a2a8
 name: Image has accessible name
 rule_type: atomic
 description: |
-  This rule checks that each image that is not marked as decorative, has an accessible name.
+  This rule checks that each image that either has an accessible name or is marked up as decorative
```
Jym77:
```suggestion
  This rule checks that each image either has an accessible name or is marked up as decorative
```

Issue #1186, opened 2020-02-25
iFrame cae760 test fail 7 is questionable as a fail


mraccess77:
Failed Example 7 for the iframe indicates that it's part of sequential focus navigation has an empty Accessible name.   However, it ha a role of none.  If there is a role of none then it seems like SC 4.1.2 would not apply because it's not a user interface component with role none.    It's in navigation order yes -- but it's acting like static text content.    

iframe title=" " src="/test-assets/SC4-1-2-frame-doc.html" role="none" tabindex="0" iframe

I'd think this might be inapplicable because of the role of none.

Jym77:
https://act-rules.github.io/rules/cae760#failed-example-7

Actually, the `tabindex` conflicts with the `role="none"` and the [Presentational Role Conflict Resolution](https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none) dictates to ignore the `role="none"` and keep the `iframe` with its implicit role. `iframe` has no implicit role in the HTML AAM, however…

And of course, reality is a bit different.
* Chrome includes the `iframe` in its accessibility tree with a role of "iFramePresentational" (vs "Iframe" if the `role="none"` is not here).
* Edge includes the `iframe` in its accessibility tree, with a role of "none" (vs no role if the `role="none"` is not here).
* Firefox includes the `iframe` in its accessibility tree, with a role of "internal frame" (same if the `role="none"`) is not here).

I haven't checked what screenreaders do on top of that.

Related topic: #1177

Jym77:
Some tests on how iframe are (or not) hidden:
https://github.com/act-rules/act-rules.github.io/wiki/Tests-on-Presentational-role-Conflict-resolution#iframe-no-implicit-role-focusable-by-default-not-in-sequential-focus-order

WilcoFiers:
@mraccess77 we talked about this during today's ACT-R call. As Jean-Yves points out, putting role=none on an iframe does nothing for assistive technologies. Unless you put tabindex="-1" on an iframe, it will not be ignored. The group thinks this should be a failure of WCAG.

I'll update the example description though, I think that this can be explained better.

WilcoFiers:
There's an open PR with the fix. Closing the issue.

PR #1219, opened 2020-03-09
Update example descriptions iframe-acc-name rule

WilcoFiers:
Update example descriptions of iframe acc name rule to follow the new format. Also cleaned up some of the outdated / unnecessary examples.

Closes issue(s): #1186

Need for Final Call: 1 week


_rules/iframe-accessible-name-cae760.md
```diff
@@ -52,90 +52,60 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
+This `iframe` element gets its [accessible name][] from the `aria-label` attribute.
 
 ```html
 <iframe aria-label="Advertisement of tours to Great Wall of China" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
 #### Passed Example 3
 
-Usage of `aria-labelledby` attribute to describe the `iframe` content.
+This `iframe` element gets its [accessible name][] from the content of the `div` reerenced with the `aria-labelledby` attribute.
```
```suggestion
This `iframe` element gets its [accessible name][] from the content of the `div` referenced with the `aria-labelledby` attribute.
```

Issue #1197, opened 2020-03-02
Separate CI tasks


WilcoFiers:
There is currently one big job called "npm install & build", it would be better to split this up into three jobs:

- install
- build
- test

jkodu:
PR in place. Closing issue.

PR #1224, opened 2020-03-10
ci: update CI job and dependencies

jkodu:
- Updated dependencies to be `dev` dependencies, as in this repository all the dependencies are used only for running `test` using `jest`.
- Updated CI job to not have a misleading name and only install dev dependencies. (The thinking is, one day when there are non development dependencies, we only install the dev ones, for running test).

> Note: Previous a name of install and build was given to a test job by mistake, due to copypasta. This was misleading.

Closes issue(s):

- https://github.com/act-rules/act-rules.github.io/issues/1197

Need for Final Call:
- NA


.github/workflows/test.yml
```diff
@@ -15,9 +14,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
-      - name: npm install & build
-        run: |
-          npm install
-          npm run test
+      - run: |
+          npm install -D
```
WilcoFiers:
```suggestion
          npm install
```

.github/workflows/test.yml
```diff
@@ -15,9 +14,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
-      - name: npm install & build
```
WilcoFiers: I think you should name the task.

PR #1226, opened 2020-03-10
Add input aspect to definitions

### LINKED ISSUE #1196: Add input aspect to definitions
PR #1189 added an input aspect to a number of rules, because one of its definitions was updated. That made me wonder if it would be helpful to include input aspects in the frontmatter of the definition. This would help authors know what input aspects to use in new rules, and it'd give us a way to check no input aspects needed for the definitions were missing from rules.

Jym77:
Adding input aspects to definitions.
Auto-testing that these aspects are copied in the rules is not part of this PR.

Closes issue(s):

- closes #1196 

Need for Final Call:
This will not require a Final Call (internal meta-data only)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/programmatic-label.md
```diff
@@ -3,6 +3,8 @@ title: Programmatic Label
 key: programmatic-label
 unambiguous: true
 objective: true
+input_aspects:
+  - DOM tree
```
WilcoFiers: Labelled controls can not be hidden with CSS.

```suggestion
  - CSS styling
  - DOM tree
```

pages/glossary/standard-keyboard-navigation.md
```diff
@@ -3,6 +3,8 @@ title: Standard keyboard navigation
 key: standard-keyboard-navigation
 unambiguous: true
 objective: true
+input_aspects:
+  - DOM tree
```
WilcoFiers: I'm not sure how this requires DOM. Might not have any input.
```suggestion
```

pages/glossary/visible-text-content.md
```diff
@@ -3,6 +3,8 @@ title: Visible Text Content
 key: visible-text-content
 unambiguous: true
 objective: true
+input_aspects:
+  - CSS styling
```
WilcoFiers: Talks of node types and attributes, so DOM is needed.

```suggestion
  - CSS styling
  - DOM tree 
```

pages/glossary/programmatic-label.md
Jym77: 🤔 The  visibility of the labelled control is currently handled in the rule's applicability (definition is only used in one rule), and the HTML definition of "labelled control" does not speak about visibility.

We could move the visibility to the definition (and require that the programmatic label is visible), but I don't think that it is necessarily. The current definition can be used for "AT specific" labels, which could be useful in other rules.

pages/glossary/standard-keyboard-navigation.md
Jym77: It needs to know if elements are focusable, so it should include at least the same input aspects.

pages/glossary/accessible-name.md
```diff
@@ -3,6 +3,9 @@ title: Accessible Name
 key: accessible-name
 unambiguous: true
 objective: true
+input_aspects:
+  - Accessibility tree
+  - DOM tree
```
carlosapaduarte: CSS is relevant for the accessible name computation.

```suggestion
  - Accessibility tree
  - CSS styling  
  - DOM tree
```

pages/glossary/visible.md
```diff
@@ -3,6 +3,8 @@ title: Visible
 key: visible
 unambiguous: true
 objective: true
+input_aspects:
```
carlosapaduarte: I'm not sure we can leave "DOM tree" outside. 
For example, with the `<details><summary>` elements, the details become visible when the `<details>` element gains an `open` attribute.

pages/glossary/included-in-the-accessibility-tree.md
```diff
@@ -3,6 +3,8 @@ title: Included in the accessibility tree
 key: included-in-the-accessibility-tree
 unambiguous: true
 objective: true
+input_aspects:
```
adilsofficial: Accessibility tree is a subset of the DOM tree. Shouldn't we add DOM tree as well?

Jym77: I don't think so. The accessibility tree is considered given by whatever UA/AT we're considering (i.e. it is not the job of the rule or the tester to build it).

PR #1228, opened 2020-03-11
orientation-lock fix parser issue in frontmatter

WilcoFiers:
Not sure why tests didn't catch this. @jkodu thoughts?

Need for Final Call: none


jkodu:
Yes @WilcoFiers . Our test with regards to `frontmatter` checks only for existence of keys like `id`, `name`, `input_rules` etc., 

The above is still valid markdown, meaning it is nested sublists. We have to write a new test to check for this, perhaps one that says, `input_aspects` is not a nested list.

PR #1233, opened 2020-03-16
test: validate rules frontmatter and improve assertion on input_aspects

jkodu:
This PR does the below:
1) Unsure why frontmatter tests for rules were commented out, uncommenting the test, so that it runs for each rule.
2) Update assertion of input aspects
	- ensure each aspect specified is of type string 
	- that aspects are not nested object
3) Found that sometimes frontmatter parses nested yml as a single item in an array like below, and have handles the same:
image
image
4) Fixes failing test for frontmatter on a rule markdown file.

Closes issue(s):
- NA

Need for Final Call:
- None & does not need 3 approvals.

### DIFF
--- a/__tests__/frontmatter.js
+++ b/__tests__/frontmatter.js
@@ -57,14 +57,16 @@ function validateRuleFrontmatter({ frontmatter }, metaData) {
 			expect(frontmatter).not.toHaveProperty('input_rules')
 
 			expect(frontmatter).toHaveProperty('input_aspects')
-			expect(Array.isArray(frontmatter.input_aspects)).toBe(true)
-			expect(frontmatter.input_aspects.length > 0).toBe(true)
+			expect(Array.isArray(frontmatter.input_aspects), `input_aspects should be a list`).toBe(true)
+			expect(frontmatter.input_aspects.length > 0, `should have atleast one item in input_aspects`).toBe(true)
+
 			expect(
 				frontmatter.input_aspects.every(aspect => {
 					const isTypeString = typeof aspect === 'string'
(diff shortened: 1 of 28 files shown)


__tests__/frontmatter.js
```diff
@@ -53,8 +54,18 @@ function validateRuleFrontmatter({ frontmatter }, metaData) {
...
 		test('has optional property `input_aspects` when `rule_type = atomic`', () => {
-			expect(frontmatter).toHaveProperty('input_aspects')
 			expect(frontmatter).not.toHaveProperty('input_rules')
+
+			expect(frontmatter).toHaveProperty('input_aspects')
+			expect(Array.isArray(frontmatter.input_aspects)).toBe(true)
+			expect(frontmatter.input_aspects.length > 0).toBe(true)
+			expect(
+				frontmatter.input_aspects.every(aspect => {
+					const isTypeString = Object.prototype.toString.call(aspect) === '[object String]'
```
Jym77: Shouldn’t that be?
```suggestion
					const isTypeString = typeof aspect === 'string'
```

__tests__/frontmatter.js
```diff
@@ -53,8 +54,18 @@ function validateRuleFrontmatter({ frontmatter }, metaData) {
...
+					const isNotMultipleAspects = /\s-\s/.test(aspect) === false // ensure it is not concatantated aspects like `"DOM Tree - CSS Styling"`
+					return isTypeString && isNotMultipleAspects
+				})
+			).toBe(true)
```
Jym77: What error messages are going to be displayed if these tests fail? Are they understandable?

jkodu: I have added custom messages, example below:

image

PR #1234, opened 2020-03-16
Updating "Image filename is accessible name for image" for W3C publication

carlosapaduarte:
Preparing [Image filename is accessible name for image](https://act-rules.github.io/rules/9eb3f6) for W3C publication.

- Example descriptions updated
- Other minor editorial changes

Need for Final Call:
No need for Final Call

---

## How to Review And Approve

### DIFF
--- a/_rules/image-filename-as-accessible-name-9eb3f6.md
+++ b/_rules/image-filename-as-accessible-name-9eb3f6.md
@@ -31,11 +31,11 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to any HTML `input` element with a [`type`](https://html.spec.whatwg.org/#states-of-the-type-attribute) of `image`, or any HTML element with the [semantic role](#semantic-role) of `img`, that is [included in the accessibility tree][], and has an [accessible name][] that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing [whitespace](#whitespace) should be ignored.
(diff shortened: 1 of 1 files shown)


_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -31,11 +31,11 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to any HTML `input` element with a [`type`](https://html.spec.whatwg.org/#states-of-the-type-attribute) of `image`, or any HTML element with the [semantic role](#semantic-role) of `img`, that is [included in the accessibility tree][], and has an [accessible name][] that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing [whitespace](#whitespace) should be ignored.
+The rule applies to any HTML `input` element with a [`type`][type] of `image`, or any HTML element with the [semantic role][] of `img`, that is [included in the accessibility tree][], and has an [accessible name][] that is equivalent to the [filename][] specified in the `src` attribute. Difference in letter casing, and forward and trailing [whitespace][] should be ignored.
```
adilsofficial:
```suggestion
The rule applies to any HTML `input` element with a [`type`][type] of `image`, or any HTML element with the [semantic role][] of `img`, that is [included in the accessibility tree][], and has an [accessible name][] that is equivalent to the [filename][] specified in the `src` attribute. Difference in letter casing, forward, and trailing [whitespace][] should be ignored.
```

carlosapaduarte:
Merging and closing (only changes to examples descriptions - no need for final call). Then submiting to W3C.

PR #1235, opened 2020-03-16
Update meta-viewport-b4f0c3.md

carlosapaduarte:
Preparing [`meta` `viewport` does not prevent zoom](https://act-rules.github.io/rules/b4f0c3) for W3C publication.

- Moved Failed Example 5 to Passed Example 5 because negative values for the `maximum-scale` property result in it being dropped, therefore no value is set for `maximum-scale`
- Updated example descriptions
- Other minor editorial changes

Need for Final Call:

---

## How to Review And Approve


carlosapaduarte:
@audreymaniez @jkodu You are the authors of [meta viewport does not prevent zoom](https://act-rules.github.io/rules/b4f0c3). While preparing the rule for submission to W3C for publication, I noticed that Failed Example 5 had a negative maximum-scale value. According to the [spec](https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale) these values are dropped. In my understanding (and tests with Chrome on iPhone) this means the property is not set and therefore the example should pass. Can you guys please check if I'm misunderstanding something?

_rules/meta-viewport-b4f0c3.md
```diff
@@ -68,7 +68,7 @@ The `meta name="viewport"` element does not define the `maximum-scale` and `user
 
 #### Passed Example 2
 
-The `meta name="viewport"` element defines the `user-scalable=yes` so the user can still zoom in.
+This `meta` element with a `name="viewport"` and `content` attributes specifies `user-scalable=yes` and does not specify the `maximum-scale` value.
```
WilcoFiers: I think you've inverted cause and effect on these. Try expressing them like this instead:

```suggestion
This  viewport `meta` element does not prevent user scaling because it has `user-scalable` set to `yes`.
```

The same is true for the rest of the example descriptions. Please see what you can do to improve them.

WilcoFiers:
Dismissed @jkodu's review. Only one person per org can approved a PR.

carlosapaduarte:
Final call ends May 14

PR #1237, opened 2020-03-17
Update example descriptions for rule - '`aria-*` attribute is defined in WAI-ARIA' (5f99a7)

jkodu:
Update example descriptions to follow the new format, in preparation for submitting to W3C.

Closes issue(s):
- NA

Need for Final Call:
- Nope

### DIFF
--- a/_rules/aria-attr-defined-5f99a7.md
+++ b/_rules/aria-attr-defined-5f99a7.md
@@ -86,7 +86,7 @@ This `li` element with a role of `menuitemcheckbox` has an `aria-*` attribute wh
 
 #### Failed Example 2
 
-This `div` element with a role of `searchbox` has an `aria-*` attribute which is not defined in [WAI-ARIA Specifications][].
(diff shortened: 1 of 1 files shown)


_rules/aria-attr-defined-5f99a7.md
```diff
@@ -78,15 +78,15 @@ Multiple valid ARIA 1.1 attributes `aria-*` are specified on element `input` wit
...
+This `li` element with a role of `menuitemcheckbox` has an `aria-*` attribute which is not defined in [WAI-ARIA Specifications][].
 
 ```html
 <li role="menuitemcheckbox" aria-not-checked="true">List Item</li>
 ```
 
 #### Failed Example 2
 
-`aria-labelled` is not a defined attribute in ARIA 1.1.
+This `div` element with a role of `searchbox` has an `aria-*` attribute which is not defined in [WAI-ARIA Specifications][].
```
Jym77:
```suggestion
This `div` element with a role of `searchbox` has an `aria-*` attribute (`aria-labelled`) which is not defined in [WAI-ARIA Specifications][].
```
Because the other one (`aria-placeholder`) is OK.

PR #1239, opened 2020-03-24
color-contrast: script issue in passed example 10

WilcoFiers:
Passed example 10 should use `.innerHTML` and not `.textContent`, otherwise the adjusted color contrast isn't applied correctly and the example fails.

Need for Final Call: 1 week


WilcoFiers:
Dismissing @joao-vicente's review, since Carlos already reviewed. Only one person per org can approve a PR.

Issue #883, opened 2019-09-11
Test for (no) angle bracket around element names


Jym77:
The convention in ACT-R is to not put angle brackets around element names (`title`, not `<title>`) but it is not always respected. We should add a test for this.

jkodu:
WIP branch - https://github.com/act-rules/act-rules.github.io/pull/new/test-html-tags-angle-bracket

jkodu:
PR in place. Closing issue.

PR #1240, opened 2020-03-25
[WIP] test: check if non code snippet HTML tags are not surrounded by angle brackets

jkodu:
> Note: Marking this as WIP, given the number of failures to fix.

Closes issue(s):
- https://github.com/act-rules/act-rules.github.io/issues/883

Need for Final Call:
- NA


_rules/explicit-SVG-image-has-name-7d6734.md
```diff
@@ -36,13 +36,13 @@ This rule assumes that the presence of one of the roles outlined in the applicab
...
+The [HTML Accessibility API Mappings](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings) specify that the `svg` element has an implicit role of `graphics-document`. However browser support for the `graphics-document` role and the [SVG Accessibility API Mappings][] is inconsistent.
 
 This rule is limited to the explicit use of roles, as a clear indication that content should convey meaning, until the [SVG Accessibility API Mappings][] is more stable and browser support is more consistent.
 
-Browser and assistive technology support for SVG `<title>` and `<desc>` elements is currently inconsistent. Using WAI ARIA in combination with the `img` role for non-decorative `<svg>` elements significantly improves accessibility browser support.
+Browser and assistive technology support for SVG `title` and `<desc>` elements is currently inconsistent. Using WAI ARIA in combination with the `img` role for non-decorative `svg` elements significantly improves accessibility browser support.
```
Jym77: Somehow, `<desc>` has not been caught, which indicates maybe a problem with the test…

_rules/heading-accessible-name-ffd0e9.md
```diff
@@ -39,7 +39,7 @@ _There are currently no assumptions._
...
 
-**Note:** Completely empty headings (`<h1></h1>`) seem to be consistently ignored by assistive technologies. However, they fail [Technique H42: Using h1-h6 to identify headings][tech h42] (by using heading markup for content which is not heading). Moreover, they may be rendered on screen (by breaking flow content, or because of custom styling), thus causing concerns for sighted users. Therefore, this rule also fails on these.
+**Note:** Completely empty headings (`h1`) seem to be consistently ignored by assistive technologies. However, they fail [Technique H42: Using h1-h6 to identify headings][tech h42] (by using heading markup for content which is not heading). Moreover, they may be rendered on screen (by breaking flow content, or because of custom styling), thus causing concerns for sighted users. Therefore, this rule also fails on these.
```
Jym77: This one is a bit different. The goal here is to illustrate what a "completely empty heading" is, and that is not just the tag but the element with nothing in it…

_rules/explicit-SVG-image-has-name-7d6734.md
jkodu: I use the library `html-tags` -> https://www.npmjs.com/package/html-tags to figure out if a given piece of string is a html tag. `desc` seems to not be here - https://github.com/sindresorhus/html-tags/blob/master/html-tags.json

__tests__/html-tag-no-angle-bracket.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * Helper fn to assert a given text for usage of angle brackets
+ * @param {String} text string
+ */
+function testForAngleBracketsAroundTags(items) {
+	if (items.length) {
+		test.each(items)('%s has no <> bracket', item => {
+			const actual = htmlTagsWithAngleBrackets.includes(item)
```
WilcoFiers, line 66: and then this whould be: 

```suggestion
			expect(getTextSurroundedByAngleBrackets(body), msg).toHaveLength(0)
```

jkodu:
Marking as WIP, as trying to construct a robust regex that, matches opening html tags that is not followed by closing tags or is self closing. Suggestions welcome.

jkodu:
Closing this PR as need to find a better way to work with HTML and regular expressions.

Jym77:
@jkodu can we at least try to extract the corrections that were made to the rules within that PR?

PR #1243, opened 2020-03-26
svg-acc-name: escape "<svg>" string in acc support

WilcoFiers:
This svg element was rendered because it wasn't escaped.

Need for Final Call: none


jkodu:
@Jym77 I have a test for this now - https://github.com/act-rules/act-rules.github.io/pull/1240

PR #1251, opened 2020-03-30
Rule: Text spacing via CSS is below minimum value or not set to `!important` (b8bb68)

jeeyyy:
A rule proposed for year 3. Feedback & reviews welcome.

Closes issue(s):
- NA

Need for Final Call:
- Yes (2 Weeks)


_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Note**: The permissible minimum values are `0.16em` for [word-spacing][], `0.12em` for [letter-spacing][], and `1.5em` for [line-height][] respectively.
+
+## Assumptions
+
+_There are currently no assumptions_
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
```
WilcoFiers, line 41:
```suggestion
While some assistive technologies are able to set [user origin][] styles, others such as browser exteinsions are only able to set style with the [author origin][]. Such assistive technologies can not create styles with a high enough priority to override a `style` attribute with `!important`. If [accessibility support][] does not include assistive technologies that override text spacing through [author origin][], this rule should not be used.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Expectation
+
+The target element does not have an overriding property value with `!important` specified for any of the above applicable CSS properties, except when value is below the permissible minimum respectively:
+
+**Note**: The permissible minimum values are `0.16em` for [word-spacing][], `0.12em` for [letter-spacing][], and `1.5em` for [line-height][] respectively.
+
+## Assumptions
+
+_There are currently no assumptions_
```
WilcoFiers, line 37:
```suggestion
If there is a mechanism available on the page by which text spacing can be adjusted, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,110 @@
...
+
+```html
+<p style="line-height: 1.2em;">
+	The quick brown fox jumps over the lazy dog
+</p>
+```
+
+#### Passed Example 2
+
+This `span` element has a `letter-spacing` of `1.5em !important` which is less than or equal to the permissible minimum.
```
WilcoFiers, line 65:
```suggestion
This `span` element has a `letter-spacing` of `1.5em !important` which is equal to the permissible minimum.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a `line-height` of `1.2em` which is below the permissible minimum.
+
+```html
+<p style="line-height: 1.2em;">
+	The quick brown fox jumps over the lazy dog
```
WilcoFiers, line 59: I think we should use a proper paragraph here, not just dummy text.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,110 @@
...
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+## Test Cases
```
WilcoFiers, line 49: I think this should have passing and failing example of each of the three properties. Have at least one where there are multiple props.

WilcoFiers, line 49: Another thought; might be worth putting in a test for large text; make sure the computed values are actually converted to em, and not just based on px assuming a standard front-size.

_rules/invalid-form-field-value-36b590.md
```diff
@@ -23,17 +23,18 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to each HTML element that has one of the following [semantic roles][semantic role]:
```
WilcoFiers, line 26: This should be in a separate PR. Did you mean to include this?

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,213 @@
...
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][] and has any of the below CSS properties set in the `style` attribute:
```
WilcoFiers, line 23:
```suggestion
This rule applies to any HTML element that is [visible][] and has one of the following CSS properties set by the `style` attribute:
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,213 @@
...
+	<body style="font-size: 24px;">
+		<p style="line-height: 1.25em !important">
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</p>
+	</body>
+</html>
+```
+
+### Inapplicable
```
WilcoFiers, line 181: We should add an inapplicable example where "important" is used on one of those props on a style element.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,213 @@
...
+
+The test target does not have "important" as its [word-spacing][] [priority][], unless the [computed][] [word-spacing][] is `0.16em` or greater.
+
+## Expectation 2
+
+The test target does not have "important" as its [letter-spacing][] [priority][], unless the [computed][] [letter-spacing][] is `0.12em` or greater.
+
+## Expectation 3
+
+The test target does not have "important" as its [line-height][] [priority][], unless the [computed][] [line-height][] is `1.5em` or greater.
```
kasperisager, line 39: All three of these attributes compute to absolute lengths, and hence pixel values, so it doesn't make sense to talk about ems. I'm thinking something like this could make sense instead:

> [...] unless the computed `<property>` is at least `<number>` times the computed `font-size`.

WilcoFiers: The required space is relative to the font size in the SC. EM seems like the right unit for that. How would you propose to do it if not through em? I guess you can express it as a percentage of the font-size, but that's effectively what EM does.

kasperisager: We cannot talk about ems when dealing with pixels, exactly because ems are a relative measure and pixels aren't. It's comparing apples and potatoes. See my suggestion above though ☝️

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,235 @@
...
+- [letter-spacing][]
+- [line-height][]
+
+## Expectation 1
+
+The test target does not have "important" as its [word-spacing][] [priority][], unless the [computed][] [word-spacing][] is greater than `0.16` times the computed [font-size][] of the test target.
+
+## Expectation 2
+
+The test target does not have "important" as its [letter-spacing][] [priority][], unless the [computed][] [letter-spacing][] is greater than `0.12` times the computed [font-size][] of the test target.
```
carlosapaduarte, line 37:
```suggestion
The test target does not have "important" as its [letter-spacing][] [priority][], unless the [computed][] [letter-spacing][] is greater than `0.12` times the [computed][] [font-size][] of the test target.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Expectation 3
+
+The test target does not have "important" as its [line-height][] [priority][], unless the [computed][] [line-height][] is greater than `1.5` times the computed [font-size][] of the test target.
```
carlosapaduarte, line 41:
```suggestion
The test target does not have "important" as its [line-height][] [priority][], unless the [computed][] [line-height][] is greater than `1.5` times the [computed][] [font-size][] of the test target.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,235 @@
...
+
+**Note:** Font size conversations rely on the default pixel size for body (usually 16 pixels), as defined by the [user agent origin][] stylesheet.
+
+## Assumptions
+
+If there is a mechanism available on the page by which text spacing can be adjusted, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
+
+## Accessibility Support
+
+While some assistive technologies are able to set [user origin][] styles, others such as browser extensions are only able to set style with the [author origin][]. Such assistive technologies can not create styles with a high enough priority to override a `style` attribute with `!important` . If accessibility support does not include assistive technologies that override text spacing through [author origin][], this rule should not be used.
```
carlosapaduarte, line 51:
```suggestion
While some assistive technologies are able to set [user origin][] styles, others such as browser extensions are only able to set style with the [author origin][]. Such assistive technologies can not create styles with a high enough priority to override a `style` attribute with `!important`. If accessibility support does not include assistive technologies that override text spacing through [author origin][], this rule should not be used.
```

carlosapaduarte, line 51: I don't understand the need for this paragraph. The SC target's author's content, independently of the assistive technology used in rendering it. Why should this prevent the rule to be used?

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Background
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+## Test Cases
```
carlosapaduarte, line 59: Unless you can show me that the default font size is 16px I recommend setting the font-size attribute to 16px in the style of the elements in the examples that do not use a relative size.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a `line-height` of `20px` (equals `1.25em` ) which is below the permissible minimum, given the default pixel size of the body is 16 pixels.
```
carlosapaduarte, line 65: You might want to add that it does not set !important, so that the user can increase this to 1.5 times the font-size.

Similar comment for the other examples.

jkodu: @carlosapaduarte I do not follow what you mean, can you guide me here?

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -30,25 +28,25 @@ This rule applies to any HTML element that is [visible][] and has one of the fol
...
-The test target does not have "important" as its [letter-spacing][] [priority][], unless the [computed][] [letter-spacing][] is greater than `0.12` times the computed [font-size][] of the test target.
+The test target does not have "important" as its [letter-spacing][] [priority][], unless the [computed][] [letter-spacing][] is greater than `0.12` times the [computed][] [font-size][] of the test target.
 
 ## Expectation 3
 
-The test target does not have "important" as its [line-height][] [priority][], unless the [computed][] [line-height][] is greater than `1.5` times the computed [font-size][] of the test target.
+The test target does not have "important" as its [line-height][] [priority][], unless the [computed][] [line-height][] is greater than `1.5` times the [computed][] [font-size][] of the test target.
 
-**Note:** Font size conversations rely on the default pixel size for body (usually 16 pixels), as defined by the [user agent origin][] stylesheet.
+**Note:** Font size conversions rely on the default pixel size for body (usually 16 pixels), as defined by the [user agent origin][] stylesheet.
```
kasperisager, line 41: This note doesn't seem like it's needed; there's no conversion going on?

_rules/text-spacing-not-important-b8bb68.md
carlosapaduarte: Sorry. Should I have given more context. What I was suggesting is to include more info in the example description. Something like

> This `p` element has a `line-height` of `20px` (equals `1.25em` ) which is below the permissible minimum, given the default pixel size of the body is 16 pixels. Since the priority is not set to `important` the user can increse the `line-height` above the minimum.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,301 @@
...
+
+The test target does not have "important" as its [word-spacing][] [priority][], unless the [computed][] [word-spacing][] is greater than `0.16` times the [computed][] [font-size][] of the test target.
+
+## Expectation 2
+
+The test target does not have "important" as its [letter-spacing][] [priority][], unless the [computed][] [letter-spacing][] is greater than `0.12` times the [computed][] [font-size][] of the test target.
+
+## Expectation 3
+
+The test target does not have "important" as its [line-height][] [priority][], unless the [computed][] [line-height][] is greater than `1.5` times the [computed][] [font-size][] of the test target.
```
kasperisager, line 39: Rather than rely on the `getPropertyPriority()` API, I'd instead refer to the abstract [important flag](https://www.w3.org/TR/cssom/#css-declaration-important-flag). It's also important that we make sure to refer to the important flag of the _inline_ declaration, which is currently not the case. Otherwise, something like this would fail:

```html
<style>
  p { 
    line-height: 1.2 !important;
  }
</style>

<p style="line-height: 1.5">Hello world</p>
```

In fact, it's currently somewhat ambiguous what we mean by "`line-height` priority" as there can be multiple `line-height` declarations for the same test target, as seen above.

_rules/text-spacing-not-important-b8bb68.md
jkodu: @carlosapaduarte see - https://github.com/act-rules/act-rules.github.io/pull/1251/commits/6190fb8b555352aad2bfdfeced360bcd7e7521f8#r404005399, this note was requested to be added per discussion with Wilco.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,287 @@
...
+
+The test target does not have the [important flag][] set for the CSS [word-spacing][] property, unless the [computed][] [word-spacing][] is greater than `0.16` times the [computed][] [font-size][] of the test target.
+
+## Expectation 2
+
+The test target does not have the [important flag][] set for the CSS [letter-spacing][] property, unless the [computed][] [letter-spacing][] is greater than `0.12` times the [computed][] [font-size][] of the test target.
+
+## Expectation 3
+
+The test target does not have the [important flag][] set for the CSS [line-height][] property, unless the [computed][] [line-height][] is greater than `1.5` times the [computed][] [font-size][] of the test target.
```
kasperisager, line 39: This seems pretty close! It's still not clear _which_ `line-height` property we're referring to in reference to the important flag. The earliest [value stage](https://drafts.csswg.org/css-cascade/#value-stages) that makes sense in this context would probably be the [cascaded value](https://drafts.csswg.org/css-cascade/#cascaded), but the [computed value](https://drafts.csswg.org/css-cascade/#computed) seems fine as well.

kasperisager, line 39:
```suggestion
The test target does not have the [important flag][] set for the CSS [line-height][] property, unless the [computed][] [line-height][] is at least `1.5` times the [computed][] [font-size][] of the test target.
```

Otherwise, `line-height: 1.5` would not be sufficient as it would have to be strictly greater than 1.5, "at least" meaning "greater than or equal to".

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,287 @@
...
+- [letter-spacing][]
+- [line-height][]
+
+## Expectation 1
+
+The test target does not have the [important flag][] set for the CSS [word-spacing][] property, unless the [computed][] [word-spacing][] is greater than `0.16` times the [computed][] [font-size][] of the test target.
+
+## Expectation 2
+
+The test target does not have the [important flag][] set for the CSS [letter-spacing][] property, unless the [computed][] [letter-spacing][] is greater than `0.12` times the [computed][] [font-size][] of the test target.
```
kasperisager, line 35:
```suggestion
The test target does not have the [important flag][] set for the CSS [letter-spacing][] property, unless the [computed][] [letter-spacing][] is at least `0.12` times the [computed][] [font-size][] of the test target.
```

Otherwise, `letter-spacing: 0.12em` would not be sufficient as it would have to be strictly greater than 0.12, "at least" meaning "greater than or equal to".

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,287 @@
...
+
+This rule applies to any HTML element that is [visible][] and has one of the following CSS properties set by the `style` attribute:
+
+- [word-spacing][]
+- [letter-spacing][]
+- [line-height][]
+
+## Expectation 1
+
+The test target does not have the [important flag][] set for the CSS [word-spacing][] property, unless the [computed][] [word-spacing][] is greater than `0.16` times the [computed][] [font-size][] of the test target.
```
kasperisager, line 31:
```suggestion
The test target does not have the [important flag][] set for the CSS [word-spacing][] property, unless the [computed][] [word-spacing][] is at least `0.16` times the [computed][] [font-size][] of the test target.
```

Otherwise, `word-spacing: 0.16em` would not be sufficient as it would have to be strictly greater than 0.16, "at least" meaning "greater than or equal to".

_rules/text-spacing-not-important-b8bb68.md
jkodu: @kasperisager I read up the value stages & I am inclined to keep this as computed value. I do not want to mix this with cascaded value as the other 2 expectations use the computed value & also thinking from an implementation perspective, it would make sense to leave this rely on the computed value.

Thoughts welcome. Ta.

kasperisager: To be clear, it's this bit I'm referring to:

> The test target does not have the important flag set for the CSS line-height property.

There are multiple different ways to interpret "line-height property" here. It could be the declared property, of which there might be several. It could also be the cascaded property, of which there's only ever 1.

jkodu: @kasperisager I see your point, but the [applicability][https://github.com/act-rules/act-rules.github.io/pull/1251#discussion_r440244173] of the rule, only applies to what is specified in the `style` attribute of the given element.

kasperisager: That doesn't make the expectations any less ambiguous though.

jkodu: @kasperisager what do you think of this expectation for the line height:

`The test target does not have the [important flag][] set for the [declared value][https://drafts.csswg.org/css-cascade/#declared] of the CSS [computed][] [line-height][] property, unless the [computed][] [line-height][] is at least `1.5` times the [computed][] [font-size][] of the test target.
`

The thinking is given the applicability is for the style attribute specified for the element. It would make sense to use the declared value over the cascaded value.

kasperisager: We're progressing! I think you're right that the declared value is what we need to be looking at, but this also doesn't come without some issues. Let's take it from the top. Property values go through multiple [value stages](https://drafts.csswg.org/css-cascade/#value-stages) before reaching their final value as it will appear when used as part of rendering a page. From start to end, the stages are:

1. [Declared](https://drafts.csswg.org/css-cascade/#declared)
2. [Cascaded](https://drafts.csswg.org/css-cascade/#cascaded)
3. [Specified](https://drafts.csswg.org/css-cascade/#specified)
4. [Computed](https://drafts.csswg.org/css-cascade/#computed)
5. [Used](https://drafts.csswg.org/css-cascade/#used)
6. [Actual](https://drafts.csswg.org/css-cascade/#actual)

As I wrote in https://github.com/act-rules/act-rules.github.io/pull/1251#discussion_r437167976, there can be more than one declared value, but for the rest of the value stages there's at most 1 value for every stage. A small example might help make this clear; consider the following:

```html
<style>
  p {
    line-height: 1.2 !important;
  }
</style>

<p style="line-height: 1.2">Hello world</p>
```

The `<p>` element above has **2** declared values for the `line-height` property: `1.2 !important` and `1.2`. When asked _"is the declared value of the `line-height` property [important](https://drafts.csswg.org/css-cascade/#important)?"_, the answer will of course be _"which declared value?"_, meaning that there's an ambiguity to be resolved. The issue, however, is that the concept of "importance" doesn't really carry over from the declared value as it's sort of resolved as part of cascade. We therefore need to be very specific about _which_ declared value we're referring to if we need to rely on the declared value. I'm guessing we're likely only interested in the one set in the `style` attribute.

jkodu: Wdyt about this - https://github.com/act-rules/act-rules.github.io/pull/1251/files#diff-c8aeb36f7c772db9d9247fd1d52502dfR39

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+
+The test target does not have the [important flag][] set for the [declared value][] of the CSS [line-height][] property that is specified by the `style` attribute of the test target, unless the [computed][] [line-height][] is at least `1.5` times the [computed][] [font-size][] of the test target.
+
+## Assumptions
+
+If there is a mechanism available on the page by which text spacing can be adjusted, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
+
+## Accessibility Support
+
+While some assistive technologies are able to set [user origin][] styles, others such as browser extensions are only able to set style with the [author origin][]. Such assistive technologies can not create styles with a high enough [priority][] to override a `style` attribute with the [important flag][]. If accessibility support does not include assistive technologies that override text spacing through [author origin][], this rule should not be used.
```
carlosapaduarte, line 47:
```suggestion
While some assistive technologies are able to set [user origin][] styles, others, such as browser extensions, are only able to set style with the [author origin][]. Such assistive technologies can not create styles with a high enough [priority][] to override a `style` attribute with the [important flag][]. If accessibility support does not include assistive technologies that override text spacing through [author origin][], this rule should not be used.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a `line-height` of `20px` (equals `1.25em` ) which is below the permissible minimum, given the default pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `line-height` above the permissible minimum value.
```
carlosapaduarte, line 61:
```suggestion
This `p` element has a `line-height` of `20px` (equals `1.25em`) which is below the permissible minimum, given the pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `line-height` above the permissible minimum value.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</p>
+	</body>
+</html>
+```
+
+#### Passed Example 2
+
+This `div` element has a `letter-spacing` of `0.12em` which is equal to the permissible minimum, given the default pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `letter-spacing` above the permissible minimum value.
```
carlosapaduarte, line 81:
```suggestion
This `div` element has a `letter-spacing` of `0.12em` which is equal to the permissible minimum, given the pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `letter-spacing` above the permissible minimum value.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</div>
+	</body>
+</html>
+```
+
+#### Passed Example 3
+
+This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em` ) which is equal to the permissible value, given the default pixel size of the body is 16 pixels.
```
carlosapaduarte, line 101:
```suggestion
This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em`) which is equal to the permissible value, given the pixel size of the body is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</strong>
+	</body>
+</html>
+```
+
+#### Passed Example 4
+
+This `article` element has both `word-spacing` and `line-height` specified, which are both below the permissible minimum, given the default pixel size of the body is 16 pixels. Since the [important flag][] is not set for either the `line-height` or `word-spacing`, the user can increase them to be above the permissible minimum value.
```
carlosapaduarte, line 121:
```suggestion
This `article` element has both `word-spacing` and `line-height` specified, which are both below the permissible minimum, given the pixel size of the body is 16 pixels. Since the [important flag][] is not set for either the `line-height` or `word-spacing`, the user can increase them to be above the permissible minimum value.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+		</article>
+	</body>
+</html>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `p` element has a `line-height` of `30px !important` (equals `1.875em` ) which is above the permissible minimum, given the default pixel size of the body is 16 pixels.
```
carlosapaduarte, line 143:
```suggestion
This `p` element has a `line-height` of `30px !important` (equals `1.875em`) which is above the permissible minimum, given the pixel size of the body is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</p>
+	</body>
+</html>
+```
+
+#### Failed Example 2
+
+This `div` element has a `letter-spacing` of `1em !important` which is above to the permissible minimum, given the default pixel size of the body is 16 pixels.
```
carlosapaduarte, line 163:
```suggestion
This `div` element has a `letter-spacing` of `1em !important` which is above to the permissible minimum, given the pixel size of the body is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</div>
+	</body>
+</html>
+```
+
+#### Failed Example 3
+
+This `strong` element has a `word-spacing` of `8px !important` (equals `0.5em` ) which is above the permissible value, given the default pixel size of the body is 16 pixels.
```
carlosapaduarte, line 183:
```suggestion
This `strong` element has a `word-spacing` of `8px !important` (equals `0.5em` ) which is above the permissible value, given the pixel size of the body is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</strong>
+	</body>
+</html>
+```
+
+#### Failed Example 4
+
+This `article` element has both `letter-spacing` and `line-height` specified, where `line-height` is above the permissible value, given the default pixel size of the body is 16 pixels.
```
carlosapaduarte, line 203:
```suggestion
This `article` element has both `letter-spacing` and `line-height` specified, where `line-height` is above the permissible value, given the pixel size of the body is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</article>
+	</body>
+</html>
+```
+
+#### Failed Example 5
+
+This `p` element has a `line-height` of `1.25em !important` , which is greater than `1.5em` given the body has a default pixel size of 24 pixels.
```
carlosapaduarte, line 223:
```suggestion
This `p` element has a `line-height` of `1.25em !important` , which is greater than `1.5em` given the body has a pixel size of 24 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+		</p>
+	</body>
+</html>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+This `p` element is not [visible][] because of `display: none` .
```
carlosapaduarte, line 240:
```suggestion
This `p` element is not [visible][] because of `display: none`.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+```html
+<p style="position: absolute; top: -999em">
+	The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+	sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+</p>
+```
+
+#### Inapplicable Example 3
+
+This `body` element does not have `style` attribute specified.
```
carlosapaduarte, line 262:
```suggestion
This `body` element does not have a `style` attribute specified.
```

__tests__/spelling-ignore.yml
```diff
@@ -70,13 +70,24 @@
 - labelable
 - merchantability
 - de-hello
-- 14pt
-- 18pt
+- Helvetica
```
EmmaJP, line 73: Does Helvetica need to be in this spelling list twice?

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
+---
+id: b8bb68
+name: Text spacing in `style` attributes is not `!important`
+rule_type: atomic
+description: |
+  This rule checks that text spacing set in the `style` attribute is below permissible minimum value or not set to `!important`
```
EmmaJP, line 6: I had to read Understanding the SC to understand what this rule is trying to do. It think the description could be a little clearer. Perhaps something like:

```suggestion
  This rule checks that text spacing set in the `style` attribute is not set to `!important` or that particular text styles have exceeded the specified metric for retaining content visibility and functionality.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+```html
+<html>
+	<style>
+		body {
+			font-size: 16px;
+		}
+	</style>
+	<body>
+		<p style="line-height: 20px;">
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
```
EmmaJP, line 72: That's a rather grim example of text.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a `line-height` of `20px` (equals `1.25em`) which is below the permissible minimum, given the pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `line-height` above the permissible minimum value.
```
EmmaJP, line 61: I find "permissible minimum" to be a very misleading way of explaining this. The recommended metric is a value below which content visibility and functionality is a requirement, and below which 'important' is not permitted to allow users to over-ride text styles for readability.

The following suggestion would apply to all relevant test cases.

```suggestion
This `p` element has a `line-height` of `20px` (equals `1.25em`) which is below the recommended metric, given the pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `line-height`.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
```
Jym77, line 20: I think the rule would be punchier by checking for the !important flag in the applicability. The expectation only has to check that the computed value is within acceptable range. That would move part of the duplication in expectations to the applicability and remove the duplication.

That would make many of the passed examples inapplicable instead, but that is OK.

Jym77, line 20: Given that the 3 expectations are completely independent, I am not sure we gain much by keeping this a single rule. This could be turned into 3 small rules instead. I think that small rules are more to the point and easier to "swallow" and understand.

This actually goes well with moving !important to applicability:

> Applicability:
> This rule applies to any HTML element that is visible and has a declared value of the word-spacing property that is specified by the style attribute with the important flag set

> Expectation:
> The computed value of the word-spacing property is at least …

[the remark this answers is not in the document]
EmmaJP: I think the idea behind the SC is that enforcing a style that limits the text size and spacing to being smaller than the specified metrics, by using "!important", is bad for accessibility. If the style is already above those metrics it is less impactful to enforce the style. I guess those metrics are a point above which the SC considers content visibility and functionality may be impacted anyway.

EmmaJP: +1 ... that seems like a reasonable alternative approach.

EmmaJP: I think it may be specific enough and small enough once "!important" is part of the applicability to remain as one rule about text spacing styles.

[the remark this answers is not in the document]
EmmaJP: Agree. This example should pass the SC as the line-height is above the threshold and so "!important" doesn't matter.

If "!important" were included in the applicability, then this would become an inapplicable example.

EmmaJP: Actually, looking more closely, I'm not sure "!important" needs to be in the applicability. I think a some other adjustments might help, and maybe splitting into 3 rules if lots of test cases are necessary. If I understand the SC correctly, then something like this perhaps...

**Description** 
This rule checks that the text spacing styles are able to be adjusted, ie. `!important` is not applied to the style, unless the style has already exceeded the specified metric.

**Applicability** 
This rule applies to any HTML element that is visible and has one of the following CSS properties set by the style attribute:
- line-height
- word-spacing
- letter-spacing

**Expectation 1** (for example)
The test target does not have the important flag set for the declared value of the CSS line-height property that is specified by the style attribute of the test target, **or** the computed line-height is at least 1.5 times the computed font-size of the test target.

**Passed Example 1** (for example)
This p element has a line-height less than 1.5 times the font size and does not have the important flag applied, so a user may increase the line-height above the threshold should they wish to.

```
<html>
	<style>
		body { font-size: 1em; }
	</style>
	<body>
		<p style="line-height: 1.25em;">
			I was in darkness, but I took three steps and found myself in paradise. The first step was a good thought, the second, a good word; and the third, a good deed.
		</p>
	</body>
</html>
```

**Passed Example** 2
This p element has a line-height greater than 1.5 times the font size so it is not a requirement for a user to be able to adjust the line-height.

```
<html>
	<style>
		body { font-size: 1em; }
	</style>
	<body>
		<p style="line-height: 1.75em; !important">
			I was in darkness, but I took three steps and found myself in paradise. The first step was a good thought, the second, a good word; and the third, a good deed.
		</p>
	</body>
</html>
```

**Failed Example** 1
This p element has a line-height less than 1.5 times the font size and does have the important flag applied, so a user would be unable to increase the line-height above the threshold should they wish to.

```
<html>
	<style>
		body { font-size: 1em; }
	</style>
	<body>
		<p style="line-height: 1.25em; !important">
			I was in darkness, but I took three steps and found myself in paradise. The first step was a good thought, the second, a good word; and the third, a good deed.
		</p>
	</body>
</html>
```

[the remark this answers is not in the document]
kasperisager: This bit from the Understanding document might help:

> **Author Responsibility**
>
> This SC does not dictate that authors must set all their content to the specified metrics. Rather, it specifies that an author's content has the ability to be set to those metrics without loss of content or functionality. The author requirement is both to not interfere with a user's ability to override the author settings, and to ensure that content thus modified does not break content in the manners shown in figures 1 through 3 in Effects of Not Allowing for Spacing Override.

This rule checks that the inline styles don't interfere with the user's ability to adjust the text spacing attributes.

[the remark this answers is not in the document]
Jym77: @kasperisager that's precisely the kind of explanation that can fit into the Background section and help people figure out why this rule is here.
The rule doesn't (directly) checks that "no loss of content occurs when…" but rather that "authors do not block user's style and enforce a bad style" which is related but slightly different.

@EmmaJP 
>  If the style is already above those metrics it is less impactful to enforce the style.

but the rule description (and examples) say that it's OK if the style is **below** the metrics and has `!important` (although the expectations don't).

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,288 @@
...
+			font-size: 16px;
+		}
+	</style>
+	<body>
+		<p style="line-height: 30px !important">
+			The boy walked down the street in a carefree way, playing without notice of what was about him. He didn't hear the
+			sound of the car as his ball careened into the road. He took a step toward it, and in doing so sealed his fate.
+		</p>
+	</body>
+</html>
```
EmmaJP, line 158: I think Jean-Yves is correct, this example would pass the SC and the expectations of this rule.

jkodu:
@Jym77 @EmmaJP, thanks for the excellent conversation and points on this rule. I have caught up and made changes based on the various comments.

Please review again.

_rules/text-spacing-not-important-b8bb68.md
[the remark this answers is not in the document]
jkodu: The excerpt above is linked as `Understanding the SC` link in Background section. Please let me know if, and how this can be improved. Suggestions welcome.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -20,23 +20,24 @@ acknowledgments:
...
-- [line-height][]
+- [word-spacing][],
+- [letter-spacing][],
+- [line-height][];
+  which are set by the `style` attribute, and have the [important flag][].
 
 ## Expectation 1
 
-The test target does not have the [important flag][] set for the [declared value][] of the CSS [word-spacing][] property that is specified by the `style` attribute of the test target, unless the [computed][] [word-spacing][] is at least `0.16` times the [computed][] [font-size][] of the test target.
+The test target's [computed][] [word-spacing][] is at least `0.16` times it's [computed][] [font-size][].
```
Jym77, line 32:
```suggestion
The test target's [computed][] [word-spacing][] is at least `0.16` times its [computed][] [font-size][].
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -20,23 +20,24 @@ acknowledgments:
...
 
 ## Expectation 2
 
-The test target does not have the [important flag][] set for the [declared value][] of the CSS [letter-spacing][] property that is specified by the `style` attribute of the test target, unless the [computed][] [letter-spacing][] is at least `0.12` times the [computed][] [font-size][] of the test target.
+The test target's [computed][] [letter-spacing][] is at least `0.12` times it's [computed][] [font-size][].
```
Jym77, line 36:
```suggestion
The test target's [computed][] [letter-spacing][] is at least `0.12` times its [computed][] [font-size][].
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -20,23 +20,24 @@ acknowledgments:
...
 
 ## Expectation 3
 
-The test target does not have the [important flag][] set for the [declared value][] of the CSS [line-height][] property that is specified by the `style` attribute of the test target, unless the [computed][] [line-height][] is at least `1.5` times the [computed][] [font-size][] of the test target.
+The test target's [computed][] [line-height][] is at least `1.5` times it's [computed][] [font-size][] of the test target.
```
Jym77, line 40:
```suggestion
The test target's [computed][] [line-height][] is at least `1.5` times its [computed][] [font-size][] of the test target.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -3,7 +3,7 @@ id: b8bb68
 name: Text spacing in `style` attributes is not `!important`
 rule_type: atomic
 description: |
-  This rule checks that text spacing set in the `style` attribute is not set to `!important` or that at least one of the  relevant text styles has exceeded the specified metric for retaining content visibility and functionality.
+  This rule checks that the text spacing styles are able to be adjusted, i.e. `!important` is not applied to the style, unless the style has already exceeded the specified metric for retaining content visibility and functionality.
```
Jym77, line 6: ❤️ 
```suggestion
  This rule checks that the text spacing styles can be adjusted by user, i.e. that `!important` is not applied to the style, unless the style has already exceeded the specified metric for retaining content visibility and functionality.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -20,23 +20,24 @@ acknowledgments:
...
-This rule applies to any HTML element that is [visible][] and has one of the following CSS properties set by the `style` attribute:
+This rule applies to any HTML element that is [visible][] and has a [declared value][] for any of the following CSS properties:
 
-- [word-spacing][]
-- [letter-spacing][]
-- [line-height][]
+- [word-spacing][],
+- [letter-spacing][],
+- [line-height][];
+  which are set by the `style` attribute, and have the [important flag][].
```
Jym77, line 28: Small problem now:
if only "word spacing" has `!important` then the target is applicable and by Expectation 2 needs "letter spacing" to be large enough (even though it doesn't have `!important`).

That's also why I was thinking maybe 3 small rules is better to streamline both applicability and expectations.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -67,18 +68,18 @@ This `p` element has a `line-height` of `20px` (equals `1.25em`) which is below
...
+			The toy brought back fond memories of being lost in the rain forest.
 		</p>
 	</body>
 </html>
 ```
 
 #### Passed Example 2
 
-This `div` element has a `letter-spacing` of `0.12em` which is equal to the permissible minimum, given the pixel size of the body is 16 pixels. Since the [important flag][] is not set, the user can increase the `letter-spacing` above the permissible minimum value.
+This `div` element has a `letter-spacing` of `0.12em` which is equal to the permissible minimum, given the default pixel size is 16 pixels.
```
Jym77, line 82:
```suggestion
This `div` element has a `letter-spacing` of `0.12em` which is equal to the recommended minimum, given the default pixel size is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -87,18 +88,18 @@ This `div` element has a `letter-spacing` of `0.12em` which is equal to the perm
...
+			The toy brought back fond memories of being lost in the rain forest.
 		</div>
 	</body>
 </html>
 ```
 
 #### Passed Example 3
 
-This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em`) which is equal to the permissible value, given the pixel size of the body is 16 pixels.
+This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em` ) which is equal to the permissible value, given the default pixel size is 16 pixels.
```
Jym77, line 102:
```suggestion
This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em`) which is equal to the recommended minimum, given the default pixel size is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -107,40 +108,20 @@ This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12e
...
-	</body>
-</html>
-```
-
 ### Failed
 
 #### Failed Example 1
 
-This `p` element has a `line-height` of `30px !important` (equals `1.875em`) which is above the permissible minimum, given the pixel size of the body is 16 pixels.
+This `p` element has a `line-height` of `10px !important` (equals `0.625em` ) which is below the permissible minimum, given the default pixel size of the body is 16 pixels.
```
Jym77, line 124:
```suggestion
This `p` element has a `line-height` of `10px !important` (equals `0.625em` ) which is below the recommended minimum, given the default pixel size of the body is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -149,18 +130,18 @@ This `p` element has a `line-height` of `30px !important` (equals `1.875em`) whi
...
+			The toy brought back fond memories of being lost in the rain forest.
 		</p>
 	</body>
 </html>
 ```
 
 #### Failed Example 2
 
-This `div` element has a `letter-spacing` of `1em !important` which is above to the permissible minimum, given the pixel size of the body is 16 pixels.
+This `div` element has a `letter-spacing` of `1.5px !important` (equals `0.09375em`) which is below to the permissible minimum, given the default pixel size is 16 pixels.
```
Jym77, line 144:
```suggestion
This `div` element has a `letter-spacing` of `1.5px !important` (equals `0.09375em`) which is below to the recommended minimum, given the default pixel size is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -169,18 +150,18 @@ This `div` element has a `letter-spacing` of `1em !important` which is above to
...
+			The toy brought back fond memories of being lost in the rain forest.
 		</div>
 	</body>
 </html>
 ```
 
 #### Failed Example 3
 
-This `strong` element has a `word-spacing` of `8px !important` (equals `0.5em` ) which is above the permissible value, given the pixel size of the body is 16 pixels.
+This `strong` element has a `word-spacing` of `2.5px !important` (equals `0.15625em` ) which is below the permissible value, given the default pixel size is 16 pixels.
```
Jym77, line 164:
```suggestion
This `strong` element has a `word-spacing` of `2.5px !important` (equals `0.15625em` ) which is below the recommended minimum, given the default pixel size is 16 pixels.
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -189,60 +170,24 @@ This `strong` element has a `word-spacing` of `8px !important` (equals `0.5em` )
...
-	</body>
-</html>
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-This `p` element is not [visible][] because of `display: none`.
+This `p` element is not [visible][] because of `display: none` .
```
Jym77, line 186:
```suggestion
This `p` element is not [visible][] because of `display: none`.
```

_rules/text-spacing-not-important-b8bb68.md
jkodu: @Jym77 I think this can be solved by a play with words in the expectation. How about the below:

`The test target's computed word spacing if having the important flag is at least 0.16 times it's computed font size`.

It makes sense to keep the rules related to text spacing together.

Jym77: Yes, I'm having mixed feeling on my own suggestion to split…
On one hand, I do agree that that keeping these together makes sense as they are very similar and closely related.
On the other hand, if the rule ends up with an applicability of "A or B or C" and three expectations of "If A, then A' ", "If B, then B' ", "If C, then C' ", this sounds so much like 3 rules artificially glued together…

Also, with expectations like "if the important flag is set on XXX, then the value of XXX is at least xx", there is no real need to have the important flag in the applicability.

Which maybe was indeed not needed (my bad). I feel the new wording you're suggesting for expectations is easier to understand than the previous one… Maybe we can go even further:

Applicability: any element where one of this properties is set.
Expectation 1: For each test target, either word spacing is not set, or it does not have the important flag, or its value is at least xx
Expectation : …

WDYT?

jkodu: @Jym77 Yeah I was thinking the same, bring back important to the expectation. We are going back to where the rule was :). But we are getting better.

Jym77: 👍

jkodu: @Jym77 pushed an update. Will resolve this conversation & await review.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -25,19 +25,19 @@ This rule applies to any HTML element that is [visible][] and has a [declared va
 - [word-spacing][],
 - [letter-spacing][],
 - [line-height][];
-  which are set by the `style` attribute, and have the [important flag][].
+  which are set by the `style` attribute.
 
 ## Expectation 1
 
-The test target's [computed][] [word-spacing][] is at least `0.16` times it's [computed][] [font-size][].
+The test target's [computed][] [word-spacing][] either does not have the [important flag][] or is at least `0.16` times its [computed][] [font-size][].
```
Jym77, line 32: Doesn't the important flag relate to the declared value, not the computed one?
```suggestion
For the test target's [word-spacing][] property, either it has no [declared value][], or its [declared value][] does not have the [important flag][] set, or its [computed value]][] is at least `0.16` times the [computed value][] of the test target [font-size][] property.
```
(repeat or the other two expectations)

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -29,15 +29,24 @@ This rule applies to any HTML element that is [visible][] and has a [declared va
 
 ## Expectation 1
 
-The test target's [computed][] [word-spacing][] either does not have the [important flag][] or is at least `0.16` times its [computed][] [font-size][].
+For the test target's [word-spacing][] property, either:
+
+- its [declared][] value does not have the [important flag][]; or
+- its [computed][] value is at least `0.16` times its [computed][] [font-size][].
```
Jym77, line 35:
```suggestion
- it has no [declared][] value; or
- its [declared][] value does not have the [important flag][]; or
- its [computed][] value is at least `0.16` times its [computed][] [font-size][].
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -29,15 +29,24 @@ This rule applies to any HTML element that is [visible][] and has a [declared va
...
 
 ## Expectation 2
 
-The test target's [computed][] [letter-spacing][] either does not have the [important flag][] or is at least `0.12` times its [computed][] [font-size][].
+For the test target's [letter-spacing][] property, either:
+
+- its [declared][] value does not have the [important flag][]; or
+- its [computed][] value is at least `0.12` times its [computed][] [font-size][].
```
Jym77, line 42:
```suggestion
- it has no [declared][] value; or
- its [declared][] value does not have the [important flag][]; or
- its [computed][] value is at least `0.12` times its [computed][] [font-size][].
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -29,15 +29,24 @@ This rule applies to any HTML element that is [visible][] and has a [declared va
...
 
 ## Expectation 3
 
-The test target's [computed][] [line-height][] either does not have the [important flag][] or is at least `1.5` times its [computed][] [font-size][] of the test target.
+For the test target's [line-height][] property, either:
+
+- its [declared][] value does not have the [important flag][]; or
+- its [computed][] value is at least `1.5` times its [computed][] [font-size][].
```
Jym77, line 49:
```suggestion
- it has no [declared][] value; or
- its [declared][] value does not have the [important flag][]; or
- its [computed][] value is at least `1.5` times its [computed][] [font-size][].
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,264 @@
...
+- [letter-spacing][],
+- [line-height][];
+  which are set by the `style` attribute.
+
+## Expectation 1
+
+For the test target's [word-spacing][] property, either:
+
+- it has no [declared][] value; or
+- its [declared][] value does not have the [important flag][]; or
```
kasperisager, line 35: This now runs into the issue I outlined here: https://github.com/act-rules/act-rules.github.io/pull/1251#discussion_r443742146. That is, there may be _multiple_ declared values for a given property, but we're only interested in the one in the `style` attribute. This applies to all three expectations.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,264 @@
...
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][] and has a [declared][] value for any of the following CSS properties:
+
+- [word-spacing][],
+- [letter-spacing][],
+- [line-height][];
+  which are specified via the `style` attribute.
```
carlosapaduarte, line 28: I think this reads better, but feel free to discard.

```suggestion
This rule applies to any HTML element that is [visible][] and has a [declared][] value specified via the `style` attribute for any of the following CSS properties:

- [word-spacing][],
- [letter-spacing][],
- [line-height][].
```

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,264 @@
...
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a `line-height` of `32px` (equals `2em` ) which is above the recommended metric, given the default pixel size is 16 pixels.
```
carlosapaduarte, line 74: There is no "default pixel size" in the specs. That is why the example specifies it.
This applies to the descriptions of all passed and failed examples.

```suggestion
This `p` element has a `line-height` of `32px` (equals `2em` ) which is above the recommended metric, given the specified font size is 16 pixels.
```

carlosapaduarte:
Don't get carried away with the 3 approvals... remember you need another one in this instance ;-)

_rules/invalid-form-field-value-36b590.md
```diff
@@ -241,4 +241,4 @@ There are no elements with any of the required [semantic roles][semantic role].
 [required context]: https://www.w3.org/TR/wai-aria/#scope
 [semantic role]: #semantic-role 'Definition of semantic role'
 [text]: https://www.w3.org/TR/WCAG21/#dfn-text
-[visible]: #visible 'Definition of visible'
+[visible]: #visible 'Definition of visible'
```
WilcoFiers, line 244: This shouldn't change. We don't want this commit message to end in the history of the form-field-value rule.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -1,54 +1,62 @@
...
 
-This rule applies to any HTML element that is [visible][] and has one of the following CSS properties set by the `style` attribute:
+This rule applies to any HTML element that is [visible][] and has a [declared][] value specified via the `style` attribute for any of the following CSS properties:
 
-- [word-spacing][]
-- [letter-spacing][]
-- [line-height][]
+- [word-spacing][],
+- [letter-spacing][],
+- [line-height][].
```
WilcoFiers, line 27: i don't think we should use punctuations this way. It doesn't add anything. Why not use [conditional lists](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/design/rule-design.md#listed-conditions)?

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -1,54 +1,62 @@
...
 
 ## Expectation 1
 
-The test target does not have "important" as its [word-spacing][] [priority][], unless the [computed][] [word-spacing][] is greater than `0.16` times the computed [font-size][] of the test target.
+For the test target's [word-spacing][] property, either:
```
WilcoFiers, line 31: This definitely should use conditional lists.

jkodu:
@WilcoFiers I gave this shot to use conditional lists. Thoughts welcome.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -22,33 +22,33 @@ acknowledgments:
...
 
 ## Expectation 1
 
-For the test target's [word-spacing][] property, either:
+For the test target's [word-spacing][] property, one of the following is true:
 
-- it has no [declared][] value; or
-- its [declared][] value specified via the `style` attribute does not have the [important flag][]; or
-- its [computed][] value is at least `0.16` times its [computed][] [font-size][].
+- has no [declared][] value; or
```
WilcoFiers, line 33: This kinda makes me think there should be three separate rules here. We're using conditional logic in the expectations to get around the issue that the applicability is too broad. What would others think about breaking this into 3 rules?

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -22,33 +22,33 @@ acknowledgments:
 
 This rule applies to any HTML element that is [visible][] and has a [declared][] value specified via the `style` attribute for any of the following CSS properties:
 
-- [word-spacing][],
-- [letter-spacing][],
+- [word-spacing][]; or
+- [letter-spacing][]; or
 - [line-height][].
```
WilcoFiers:
```suggestion
- the `style` attribute declares [word-spacing][]; or
- the `style` attribute declares [letter-spacing][]; or
- the `style` attribute declares [line-height][].
```

_rules/text-spacing-not-important-b8bb68.md
jkodu: @WilcoFiers we discussed the possibility to split this rule into 3 in this [conversation](https://github.com/act-rules/act-rules.github.io/pull/1251#discussion_r445585345).

@Jym77 and @EmmaJP lets discuss this again.

Personally I am open to it.

_rules/text-spacing-not-important-b8bb68.md
jkodu: I took your lead & updated the applicability:

```
This rule applies to any HTML element that is [visible][], for which one of the following is true:

- the `style` attribute [declares][declared][] [word-spacing][] CSS property; or
- the `style` attribute [declares][declared][] [letter-spacing][] CSS property; or
- the `style` attribute [declares][declared][] [line-height][] CSS property.
```

_rules/text-spacing-not-important-b8bb68.md
Jym77: I'm also open to it.
From the logical point of view, I agree that using expectations for a "or" case analysis is odd and splitting the rule would be more streamlined.
From a complexity point of view, it would be 3 super simple rules that are a bit "lost" if we look at them individually and keeping them together gives more punch to the result.

---
🤔 with the conditional list in expectations, and the weird case in each "the property has no declared value", I say I'm slightly in favour of splitting.

_rules/text-spacing-not-important-b8bb68.md
```diff
@@ -0,0 +1,263 @@
...
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which one of the following is true:
+
+- the `style` attribute [declares][declared][][word-spacing][] CSS property; or
+- the `style` attribute [declares][declared][][letter-spacing][] CSS property; or
+- the `style` attribute [declares][declared][][line-height][] CSS property.
```
WilcoFiers, line 27: Think we concluded to split this rule, right?

jkodu: Yes that is right & here is the secondary PR, where there are 3 atomic rules addressing test spacing.

_rules/invalid-form-field-value-36b590.md
WilcoFiers: Please fix this.

jeeyyy:
Closing this in favour of https://github.com/act-rules/act-rules.github.io/pull/1251

PR #1255, opened 2020-03-31
Update examples in "Link has accessible name"

carlosapaduarte:
Following feedback from ACT TF in https://github.com/w3c/wcag-act/issues/439

- included example with `<a>` without `href`
- improved examples and descriptions

Need for Final Call: 1 weeks

---
## How to Review And Approve


_rules/link-accessible-name-c487ae.md
```diff
@@ -69,72 +70,72 @@ For `area` elements that have a `href` attribute, but are not nested inside a `m
 
 #### Passed Example 1
```
WilcoFiers, line 71: Since we're updating the examples anyway, can you make sure the accessible name is appropriate for the link and the image?

_rules/link-accessible-name-c487ae.md
```diff
@@ -175,74 +176,74 @@ When `link` is off screen.
...
+<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png" alt=""/></a>
 ```
 
 #### Failed Example 3
 
-Link with icon inserted via font-awesome.
+This `a` element with an icon inserted via font-awesome has an empty [accessible name][].
 
 ```html
 <a href="http://www.w3.org/WAI"><i class="fa fa-download"></i></a>
```
WilcoFiers, line 198: we should load font-awesome in, otherwise this doesn't have an icon.

_rules/link-accessible-name-c487ae.md
```diff
@@ -69,72 +70,72 @@ For `area` elements that have a `href` attribute, but are not nested inside a `m
...
 ```
 
 #### Passed Example 4
 
-[accessible name][] for link via `aria-label`
+This `a` element has an [accessible name][] via `aria-label`
 
 ```html
-<a href="http://www.w3.org/WAI"><img src="#" aria-label="This is a link"/></a>
+<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png" aria-label="Web Accessibility Initiative"/></a>
```
WilcoFiers, line 100: Lets put the logo in this so it isn't an empty link.

_rules/link-accessible-name-c487ae.md
```diff
@@ -69,72 +70,72 @@ For `area` elements that have a `href` attribute, but are not nested inside a `m
...
 ```
 
 #### Passed Example 5
 
-Link named via `title` on link.
+This `a` element has an [accessible name][] via `title`.
 
 ```html
-<a href="http://www.w3.org/WAI" title="This is a link"><img src="#"/></a>
+<a href="http://www.w3.org/WAI" title="Web Accessibility Initiative"><img src="/test-assets/shared/w3c-logo.png"/></a>
```
WilcoFiers, line 108: We should keep the HTML valid if we can;

```suggestion
<a href="http://www.w3.org/WAI" title="Web Accessibility Initiative"><img src="/test-assets/shared/w3c-logo.png" alt=""/></a>
```

_rules/link-accessible-name-c487ae.md
```diff
@@ -69,72 +70,72 @@ For `area` elements that have a `href` attribute, but are not nested inside a `m
...
 ```
 
 #### Passed Example 7
 
-Link with both `image` and `text`.
+This `a` element has an [accessible name][] from its content.
 
 ```html
-<a href="http://www.w3.org/WAI"><img src="#" />This is my link text</a>
+<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png" />Web Accessibility Initiative (WAI)</a>
```
WilcoFiers, line 124:
```suggestion
<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png" alt="" />Web Accessibility Initiative (WAI)</a>
```

_rules/link-accessible-name-c487ae.md
```diff
@@ -153,7 +154,7 @@ When `link` is off screen.
 
 #### Passed Example 10
 
-`area` element with `href` attribute has [accessible name][].
+This `area` element has a [semantic role][] of `link` and an [accessible name][] via `alt`.
 
 ```html
 <img src="planets.gif" width="145" height="126" alt="Planets" usemap="#planetmap" />
```
WilcoFiers, line 160: This src doesn't exist, we should add it to assets and update the src. @jkodu any idea why the tests might not have caught this?

_rules/link-accessible-name-c487ae.md
```diff
@@ -163,86 +164,79 @@ When `link` is off screen.
...
 ### Failed
 
 #### Failed Example 1
 
-Image link with empty [accessible name][].
+This `a` element has an empty [accessible name][].
 
 ```html
-<a href="http://www.w3.org/WAI"><img src="#"/></a>
+<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png"/></a>
```
WilcoFiers, line 174:
```suggestion
<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png" role="none"/></a>
```

_rules/link-accessible-name-c487ae.md
```diff
@@ -300,12 +294,21 @@ Not [included in the accessibility tree][] due to `aria-hidden="true"`.
...
 ```html
 <area shape="rect" coords="0,0,82,126" />
 ```
 
+#### Inapplicable Example 6
+
+This `a` element does not have the role of link because it does not have an `href` attribute.
+
+```html
+<a alt="not a link" />
```
WilcoFiers, line 308: Why does this have an alt? That's not a valid attr for `a` elements.

_rules/link-accessible-name-c487ae.md
carlosapaduarte: Sorry, I missed this one. This means we currently have the rule published and this test case is missing the source for the image map. I couldn't find the gif anywhere in the repo. You're one of the original rule authors. Do you have this image somewhere?

carlosapaduarte:
> What is the source of this image? Are we allowed to use it? Please put something in about the asset in acknowledgments.

Image came from https://photojournal.jpl.nasa.gov/catalog/PIA10231 
We can use it according to https://www.jpl.nasa.gov/imagepolicy/
I've added an entry to the acknowledgments metadata (waiting to see if it passes the tests).

_rules/link-accessible-name-c487ae.md
```diff
@@ -153,99 +162,92 @@ When `link` is off screen.
...
 <a href="http://www.w3.org/WAI" style="left: -9999px; position: absolute;">
-	<img src="#" />
+	<img src="/test-assets/shared/w3c-logo.png" />
 </a>
 ```
 
 #### Failed Example 8
 
-Link is completely empty, but still shows up in focus order, so it should have a non-empty [accessible name][].
+This `a` element, which shows up in the focus order, has an empty [accessible name][]
```
Jym77, line 239: The rule has nothing to do with focus order, making this description a bit weird.
I understand it as "this link has no accessible name. Sure, it is totally empty and so you may think it doesn't need one, but it's still in focus order (or link list from AT), so it needs an accessible name or it will be very confusing for AT users".

I think we should reformulate the description.

_rules/link-accessible-name-c487ae.md
```diff
@@ -254,7 +256,7 @@ Link is completely empty, but still shows up in focus order, so it should have a
 
 #### Failed Example 10
 
-`a` element where [accessible name][] through content is empty.
+This `a` element has an empty [accessible name][] from its content.
 
 ```html
 <a href="http://www.w3.org/WAI"> </a>
```
Jym77, line 262: This is essentially the same example as Failed Example 8. Do we need both?

carlosapaduarte: No, I don't think we do!

_rules/link-accessible-name-c487ae.md
```diff
@@ -71,72 +74,82 @@ The rule assumes that all links are [user interface components](https://www.w3.o
...
+This `a` element has an [accessible name][] via `title`.
 
 ```html
-<a href="http://www.w3.org/WAI" title="This is a link"><img src="#"/></a>
+<a href="http://www.w3.org/WAI" title="Web Accessibility Initiative"
+	><img src="/test-assets/shared/w3c-logo.png" alt=""
+/></a>
 ```
 
+**Note**: There are assistive technologies that do not support using the `title` attribute for an [accessible name][], or in which this feature can be disabled.
```
daniel-montalvo, line 119: There are several of these notes in the rule. Shouldn't we be moving them to the accessibility support section?

I know it applies to this very example above, but I think the note would fit there better.

carlosapaduarte: @daniel-montalvo I've moved the notes to the accessibility support section. Can you please review again?

carlosapaduarte:
Final call ends June 3

PR #1256, opened 2020-03-31
Update accessibility requirements mapping in "Link has accessible name"

carlosapaduarte:
Following feedback from ACT TF in https://github.com/w3c/wcag-act/issues/439

- update accessibility requirements mapping section (and background)

This will not require a Final Call

---
## How to Review And Approve


carlosapaduarte:
Thanks for the update. Closing this PR.

PR #1260, opened 2020-04-03
Rule: Scrollable element is keyboard accessible

WilcoFiers:
This rule checks that scrollable elements can be scrolled by keyboard, by ensuring either it or one of its descendants are included in the sequential focus navigation.

Need for Final Call: 2 weeks


_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,256 @@
...
+
+Any HTML element that has [visible][] [children][] in the [flat tree][] for which one of the following is true:
+
+- It has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right]; or
+- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom]
+
+## Expectation
+
+Each test target is either included in [sequential focus navigation][] or has a [descendant][] in the [flat tree][] that is included in [sequential focus navigation][].
+
```
WilcoFiers, line 36: Open question for reviewers: Should I add a note here to explain why this is needed?

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,256 @@
...
+  authors:
+    - Wilco Fiers
+---
+
+## Applicability
+
+Any HTML element that has [visible][] [children][] in the [flat tree][] for which one of the following is true:
+
+- It has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right]; or
+- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom]
```
kasperisager, line 31:
```suggestion
- It has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right] of the element; or
- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom] of the element
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
kasperisager: That seems reasonable 👍

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+#### Inapplicable Example 4
+
+This [scrollable][] `section` element has no [visible][] content.
+
+```html
+<p>This is what a scrollbar looks like:</p>
+<section style="height: 20px; width: 500px; overflow-x:scroll;">
+	<div style="width: 1000px; height: 1px;"></div>
```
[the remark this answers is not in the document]
jkodu: Here given the content is 1px in height, it is visible and affects rendered pixels, therefore it is not an invisible element as the description says.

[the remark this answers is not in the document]
WilcoFiers: If you set the opacity of this element to 0, there is no visual change on the page. This isn't visible according to the definition of visible that we're using.

pages/glossary/scrollable-element.md
```diff
@@ -0,0 +1,24 @@
...
+objective: true
+---
+
+A scrollable element is an [element][] with a _horizontal scroll distance_ or a _vertical scroll distance_ greater than 0.
+
+- **horizontal scroll distance**: The difference between [scrollWidth][] and [clientWidth] for elements where the [computed][] [overflow-x][] is `auto`, `clip` or `scroll`. Undefined for other elements.
+
+- **vertical scroll distance**: The difference between [scrollHeight][] and [clientHeight] for elements where the [computed][] [overflow-y][] is `auto`, `clip` or `scroll`. Undefined for other elements.
+
+**note**: Elements such as `iframe` which can render a [nested browsing context][] are not _scrollable elements_. The scrollbars on some `iframe` elements come from the content inside the [nested browsing context][].
```
carlosapaduarte, line 14:
```suggestion
**Note**: Elements such as `iframe` which can render a [nested browsing context][] are not _scrollable elements_. The scrollbars on some `iframe` elements come from the content inside the [nested browsing context][].
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,258 @@
...
+Any HTML element that has [visible][] [children][] in the [flat tree][] for which one of the following is true:
+
+- It has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right] of the element; or
+- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom] of the element
+
+## Expectation
+
+Each test target is either included in [sequential focus navigation][] or has a [descendant][] in the [flat tree][] that is included in [sequential focus navigation][].
+
+**Note:** Focus must be on or in a scrollable region, to ensure there is some element from which arrow keys can be used control the scroll position. This can still create an issue if scripts are used to prevent the keyboard events from reaching the scrollable region. This must be tested separately.
```
carlosapaduarte, line 37:
```suggestion
**Note:** Focus must be on or in a scrollable region, to ensure there is some element from which arrow keys can be used to control the scroll position. This can still create an issue if scripts are used to prevent the keyboard events from reaching the scrollable region. This must be tested separately.
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Assumptions
+
+This rule assumes that all [scrollable elements][scrollable] with visible content need to be keyboard accessible. [Scrollable elements][scrollable] that do not need to be keyboard accessible, perhaps because its content is [purely decorative][] or because it scroll can be controlled some other way, may fail this rule but still satisfy [success criterion 2.1.1 Keyboard][].
```
carlosapaduarte, line 41:
```suggestion
This rule assumes that all [scrollable elements][scrollable] with visible content need to be keyboard accessible. [Scrollable elements][scrollable] that do not need to be keyboard accessible, perhaps because their content is [purely decorative][] or because scroll can be controlled some other way, may fail this rule but still satisfy [success criterion 2.1.1 Keyboard][].
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,258 @@
...
+## Applicability
+
+Any HTML element that has [visible][] [children][] in the [flat tree][] for which one of the following is true:
+
+- It has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right] of the element; or
+- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom] of the element
+
+## Expectation
+
+Each test target is either included in [sequential focus navigation][] or has a [descendant][] in the [flat tree][] that is included in [sequential focus navigation][].
```
Jym77, line 35: Can be rephrased with [inclusive descendant](https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant) but not sure it is needed…
```suggestion
Each test target has an [inclusive descendant](https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant) in the [flat tree][] that is included in [sequential focus navigation][].
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,258 @@
...
+_No accessibility support issues known._
+
+## Background
+
+- [Understanding Success Criterion 2.1.1: Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
+- [G202: Ensuring keyboard control for all functionality](https://www.w3.org/WAI/WCAG21/Techniques/general/G202)
+
+## Test Cases
+
+### Passed
```
Jym77, line 54: Should we have a Passed Example with horizontal scrolling?

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -0,0 +1,258 @@
...
+		<a href="//www.w3.org/TR/WCAG21/#abstract">
+			WCAG 2.1 Abstract
+		</a>
+	</p>
+</section>
+```
+
+#### Inapplicable Example 3
+
+This `section` element has a [computed][] [overflow][] of `hidden`.
```
Jym77, line 185:
```suggestion
This `section` element is not [scrollable][] because it has a [computed][] [overflow][] of `hidden`.
```
Applicability is about being scrollable, and only indirectly about overflow.

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
WilcoFiers: I'd prefer to leave it as is; this is clearer.

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
WilcoFiers: I don't think that's necessary. Seems reasonable to assume that if you know to do this vertically you can do it horizontally. We're not building a test suite here, just testing that an implementor understands the concepts in this rule.

PR #1266, opened 2020-04-07
Make rules consistent with WAI-ARIA 1.2

WilcoFiers:
The purpose of this PR is to allow implementors to start adopting ARIA 1.2 without losing the implementation status on any of the ARIA / Role rules.

Some of the things we've done in ARIA rules in the past are no longer consistent with WAI-ARIA 1.2. This PR updates examples so that the ones that are supposed to be valid are valid in both ARIA 1.1 and 1.2, and for those that aren't supposed to be valid, I've made sure they aren't valid in either version of ARIA.

Most notable change is in required states and props. Largely based on our feedback for it, ARIA 1.2 no longer has required props that also have a default value. Some of them are no longer required, and others have the default removed.

I had to rewrite the examples for role-required-states significantly, so I updated the example descriptions on those. Didn't do that for the others as it wasn't necessary and it would mean the PR did two things in one go.

Need for Final Call: 1 week


jkodu:
@WilcoFiers tests now pass after this [change](https://github.com/act-rules/act-rules.github.io/pull/1267) in spelling tests.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -94,7 +94,17 @@ _There are no major accessibility support issues known for this rule._
...
+<div role="list" aria-owns="item1 item2"></div>
+<div id="item1">
+	Apples
+	<div>
+		<div id="item2">
+			Bananas
+			<div></div>
+		</div>
+	</div>
+</div>
```
Jym77, line 107: I fail to see the usefulness of so many `div`s… Would
```suggestion
<h1>Shopping list</h1>
<div role="list" aria-owns="item1 item2"></div>
<div id="item1">Apples</div>
<div id="item2">Bananas</div>
```
be enough to illustrate this example?

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -56,65 +56,44 @@ This rule relies on browsers and assistive technologies to support leaving out [
 
 #### Passed Example 1
 
-Element has required states (no properties required for this role)
+This `checkbox` has the required property `aria-checked`.
 
 ```html
-<div role="option" aria-selected="false"></div>
+<div role="checkbox" aria-checked="false"><div></div></div>
```
Jym77, line 62: I'm not a huge fan of empty `div`s. Can it be removed without breaking the example? Or maybe replaced by something somewhat meaningful.
```suggestion
<div role="checkbox" aria-checked="false"></div>
```

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: seems reasonable.

_rules/role-required-states-and-properties-4e8ab6.md
WilcoFiers: I don't actually know how that ended up like that. Thanks.

PR #1267, opened 2020-04-07
test: ignore `aria-*` keys ending with punctuation

jkodu:
Tests fail when an `aria-*` key ends with punctuation. See - https://github.com/act-rules/act-rules.github.io/pull/1266/checks?check_run_id=567292841

This is because `retext-spell` parses `WordNodes` ending with punctuations wrongly. Reading further is became apparent that this gets complicated very quickly with natural language processing and esp., with words like `e.g.`, `i.e.`, `Mrs. ` etc.,

So `retext-spell` has taken the easier approach of taking punctuations into consideration when a given work is not a part of the dictionary.

Given this usage case, it made sense to simple add all `aria-*` keys with a punctuation of `.` at the end to the ignore list.

The solution is not the cleanest, but neither were other options, like:
1) Remove all punctuations from a given text before checking spelling.
2) Use markdown syntax tree to build a new text node (large task for a simple gain). 

Closes Issue:
- NA

> Note: No need for final call, and will be merged after one approval

### DIFF
--- a/__tests__/spelling.js
+++ b/__tests__/spelling.js
@@ -129,17 +129,14 @@ function getSpellIgnored() {
 		}
 	})
 
-	const ariaKeys = Array.from(ariaQuery.aria.keys())
-	const ignoreAria = [
-		...ariaKeys,
-		/**
-		 * `retext-spell` parses word nodes sometimes with an ending punctuation when the word is not a part of the dictionary.
-		 * See Issue: https://github.com/syntax-tree/nlcst/issues/4
-		 *
-		 * Below, we all aria keys ending with punctuation (.), to the ignore list
-		 */
-		...ariaKeys.map(key => `${key}.`),
-	]
+	/**
+	 * `retext-spell` by default checks spelling of individual words before ensuring validity of a composite word.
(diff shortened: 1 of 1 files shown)


__tests__/spelling.js
```diff
@@ -129,9 +129,19 @@ function getSpellIgnored() {
...
+	const ignoreAria = [
+		...ariaKeys,
+		/**
+		 * `retext-spell` parses word nodes sometimes with an ending punctuation when the word is not a part of the dictionary.
+		 * See Issue: https://github.com/syntax-tree/nlcst/issues/4
+		 *
+		 * Below, we all aria keys ending with punctuation (.), to the ignore list
+		 */
+		...ariaKeys.map(key => `${key}.`),
+	]
```
Jym77, line 142: I'd rather go for the solution that won't require adding every punctuation mark one by one. It's enough to tell retext that the individual components are good.
```suggestion
    const ignoreAria = [...ariaQuery.aria.keys()].map(key => key.replace(/aria-/, "")).concat("aria")
```

Issue #1170, opened 2020-02-13
What does it actually mean to be part of sequential focus navigation? (cae760)


kasperisager:
Running through the test cases changed in #1102, I took note of failed example 7:

```html
<iframe title=" " src="/test-assets/SC4-1-2-frame-doc.html" role="none" tabindex="0"></iframe>
```

This is supposed to fail the rule and this I'm absolutely on board with; the contents of the `<iframe>` can still be tabbed to and so one would need to name the `<iframe>`. However, [Alfa](https://github.com/siteimprove/alfa) doesn't fail this case because it interprets "part of sequential focus navigation" to mean "elements returned from the [sequential navigation search algorithm](https://html.spec.whatwg.org/multipage/interaction.html#sequential-navigation-search-algorithm)". This algorithm is used for determining what should happen when a user presses the tab button on their keyboard; given their current position and a direction, what should the next active focusable area be? Per this algorithm, browsing context containers, which `<iframe>` elements are, redirect focus to their contents, and are therefore never themselves returned from this algorithm. To see the effect of this, try tabbing to the button named "First" in this document and then hit tab:

```html
<button>First</button>
<iframe srcdoc="<button>Second</button>"></iframe>
```

You'll notice that the next active element becomes the button named "Second" and not the `<iframe>` element itself. This holds true even if we attempt to add something like `tabindex="0"` to the `<iframe>` element; the sequential navigation search algorithm doesn't care.

However, `<iframe>` elements _are_ [focusable areas](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area) and are actually even listed of examples as such. This can be verified by doing this:

```js
const iframe = document.querySelector("iframe");
iframe.focus();
console.log(document.activeElement);
```

If you run that code on the document above, you'll also notice that upon pressing tab, the button named "Second" will indeed become the next active element.

So, while a user cannot ever tab directly to an `<iframe>` element, we can programmatically move focus to it. As such, Alfa does not consider it part of sequential focus navigation. However, one could argue that an `<iframe>` element is still part of sequential focus navigation as it _is_ a focusable area that unless it has a negative tab index will be in the [sequential focus navigation _order_](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation-order).

What are peoples thoughts on this? At the very least, there's clearly an ambiguity that needs to be resolved.

kasperisager:
[...]

PR #1269, opened 2020-04-08
Rule update: `iframe` element has accessible name' (cae760)

jkodu:
Updates to the rule based on suggestion feedback from the task force/ survey.

Closes issue(s):

- https://github.com/w3c/wcag-act/issues/435
- https://github.com/act-rules/act-rules.github.io/issues/1170

Need for Final Call:
- 1 week (only a small number of test cases were changed).

### DIFF
--- a/_rules/iframe-accessible-name-cae760.md
+++ b/_rules/iframe-accessible-name-cae760.md
@@ -21,7 +21,7 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to `iframe` elements that are [included in the accessibility tree][] and has contents that can be accessed by [sequential focus navigation][].
+The rule applies to `iframe` elements that are [included in the accessibility tree][] or that are part of [sequential focus navigation][].
 
 **Note:** `frame` element is deprecated, this rule does not consider `frame` or `frameset` elements.
 
--- a/_rules/iframe-accessible-name-cae760.md
+++ b/_rules/iframe-accessible-name-cae760.md
@@ -55,7 +55,7 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
 This `iframe` element gets its [accessible name][] from the `title` attribute.
 
 ```html
-<iframe title="List of Contributors" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+<iframe title="List of some popular Disney characters" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
 #### Passed Example 2
@@ -63,7 +63,7 @@ This `iframe` element gets its [accessible name][] from the `title` attribute.
 This `iframe` element gets its [accessible name][] from the `aria-label` attribute.
 
 ```html
-<iframe aria-label="Advertisement of tours to Great Wall of China" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+<iframe aria-label="List of some popular Disney characters" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
 #### Passed Example 3
@@ -71,7 +71,7 @@ This `iframe` element gets its [accessible name][] from the `aria-label` attribu
 This `iframe` element gets its [accessible name][] from the content of the `div` referenced with the `aria-labelledby` attribute.
 
 ```html
-<div id="frame-title-helper">Watch highlights of the Worldcup</div>
+<div id="frame-title-helper">List of some popular Disney characters</div>
 <iframe aria-labelledby="frame-title-helper" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
@@ -80,7 +80,7 @@ This `iframe` element gets its [accessible name][] from the content of the `div`
 This `iframe` element gets its [accessible name][] from the `title` attribute. Note, that specifying a positive `tabindex` value only alters the order of the [sequential focus navigation][], and as such the contents of the `iframe` element can be accessed.
 
 ```html
-<iframe title="Top 5 Marathon runners" src="/test-assets/SC4-1-2-frame-doc.html" tabindex="5"> </iframe>
+<iframe title="List of some popular Disney characters" src="/test-assets/SC4-1-2-frame-doc.html" tabindex="5"> </iframe>
 ```
 
 ### Failed
@@ -90,7 +90,7 @@ This `iframe` element gets its [accessible name][] from the `title` attribute. N
 This `iframe` element has an empty (`""`) [accessible name][]. The `name` attribute is not used in computing the [accessible name][] of `iframe` elements.
 
 ```html
-<iframe name="List of Contributors" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+<iframe name="List of some popular Disney characters" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
 #### Failed Example 2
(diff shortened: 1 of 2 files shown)


_rules/iframe-accessible-name-cae760.md
```diff
@@ -75,6 +75,14 @@ This `iframe` element gets its [accessible name][] from the content of the `div`
 <iframe aria-labelledby="frame-title-helper" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
+#### Passed Example 4
+
+This `iframe` element gets its [accessible name][] from the `title` attribute. Note, that specifying a positive `tabindex` value only alters the order of the [sequential focus navigation][], and as such the contents of the `iframe` element can be accessed.
+
+```html
+<iframe title="Top 5 Marathon runners" src="/test-assets/SC4-1-2-frame-doc.html" tabindex="5"> </iframe>
```
Jym77, line 83: I know that 4.1.2 does **not** require the name to be descriptive 🙄 but shouldn't we nonetheless try and have descriptive names in our examples? Especially given that we use the exact same doc in the `iframe` but name it once "Top 5 Marathon runners" and another time "Highlights from the World Cup"…

_rules/iframe-accessible-name-cae760.md
```diff
@@ -123,15 +131,23 @@ This page has no `iframe` element.
...
 ```
 
 #### Inapplicable Example 3
 
-This `iframe` is not part of [sequential focus navigation][] because it has `tabindex="-1"` and not [included in the accessibility tree][] because of `role="presentation"`
+This `iframe` is not [included in the accessibility tree][] because of `role="presentation"`.
+
+```html
+<iframe style="display:none;" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
```
Jym77, line 146: The code of this example is the same as the previous (`display:none`) and doesn't match the description.

The description is the same as Failed Example 4 (just swapping `role="none"` for the equivalent `role="presentation"`) making it super weird that one fails and one is inapplicable.

_rules/iframe-accessible-name-cae760.md
```diff
@@ -21,23 +21,23 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to `iframe` elements that are [included in the accessibility tree][] or that are part of [sequential focus navigation][].
+The rule applies to `iframe` elements that are [included in the accessibility tree][] and has contents that can be accessed by [sequential focus navigation][].
```
kengdoj, line 24: I'm not clear why "has contents" is included here. Is it only applicable for non-empty iframes? I think the original text is better. Line 40 with the update seems to cover iframe content better.

Jym77:
@jkodu this seems to close #1170 (see https://github.com/act-rules/act-rules.github.io/issues/1170#issuecomment-617675565).

kasperisager:
> @jkodu this seems to close #1170 (see [#1170 (comment)](https://github.com/act-rules/act-rules.github.io/issues/1170#issuecomment-617675565)).

After 0cca293, it no longer does.

jkodu:
@kengdoj see https://github.com/act-rules/act-rules.github.io/pull/1269#issuecomment-617685398 and the related issue as to why applicability was changed. Thanks

jkodu:
@kasperisager thanks for helping me remember why this change was made. I revert commit https://github.com/act-rules/act-rules.github.io/commit/0cca293f8954f2584e5abb3ccde4467046e5f31d

_rules/iframe-accessible-name-cae760.md
WilcoFiers, line 24: This one keeps getting more complicated... :( I think we made a mistake with the "is in sequential focus navigation". I did the following tests with Voiceover:

| scenario  | VO result |
|---------|------------|
| empty frame | VO ignores it completely
| empty iframe with tabindex=1 | VO ignores it completely
| iframe with focusable content | VO announces the iframe
| iframe with non-focusable content | VO announces the iframe

Interesting caveat; VO in Safari does not look at aria-label. It only announces title. We should put that in Accessibility support.


This tells me what matters here, at least for VO is if the iframe has any content. I tested with an image with alt="", if that's the only thing in the frame, VO ignores it. I think we need to do more testing before we continue with this one.

kasperisager: That's... concerning 😮

jkodu:
Ok closing and will do some AT testing.

_rules/iframe-accessible-name-cae760.md
daniel-montalvo: Adding the following AT & browser combos:

* JAWS & Chrome.
* NVDA & Firefox.

| scenario | VO result | Chrome & JAWS result |  Firefox & NVDA result |
| --- | --- | --- | --- |
| empty frame | VO ignores it completely | Chrome & JAWS do not speak it, although there seems to be a tab stop for the frame | Firefox and NVDA announce iframe title |
| empty iframe with tabindex=1 | VO ignores it completely | Chrome & JAWS do not speak it, although there seems to be a tab stop for the frame | Firefox and NVDA announce iframe title |
| iframe with focusable content | VO announces the iframe | Chrome & JAWS announce the iframe title and contents as you tab to the iframe | Firefox and NVDA announce the iframe title and contents as you tab to the iframe |
| iframe with non-focusable content | VO announces the iframe | JAWS & Chrome announce the iframe title, iframe content needs to be reviewed using browse mode | Firefox and NVDA announce the iframe title and contents as you tab to the iframe |

> Interesting caveat; VO in Safari does not look at aria-label. It only announces title. We should put that in Accessibility support.

Both JAWS and NVDA work well with `title` and `aria-label`
> 
> This tells me what matters here, at least for VO is if the iframe has any content. I tested with an image with alt="", if that's the only thing in the frame, VO ignores it. I think we need to do more testing before we continue with this one.

I may need some visual help in Chrome. As said above, there seems to be tab stops for each of the empty iframes (with and without tabindex value), but I am not sure if this is really happening because there is a frame or because there is a title. When I remove titles from empty iframes, tab stops seem to disappear.

daniel-montalvo: > This one keeps getting more complicated... :( I think we made a mistake with the "is in sequential focus navigation". I did the following tests with Voiceover:
> 
> scenario	VO result
> empty frame	VO ignores it completely
> empty iframe with tabindex=1	VO ignores it completely
> iframe with focusable content	VO announces the iframe
> iframe with non-focusable content	VO announces the iframe
> Interesting caveat; VO in Safari does not look at aria-label. It only announces title. We should put that in Accessibility support.
> 
> This tells me what matters here, at least for VO is if the iframe has any content. I tested with an image with alt="", if that's the only thing in the frame, VO ignores it. I think we need to do more testing before we continue with this one.

Adding the following AT & browser combos:

* JAWS & Chrome.
* NVDA & Firefox.

| scenario | VO result | Chrome & JAWS result |  Firefox & NVDA result |
| --- | --- | --- | --- |
| empty frame | VO ignores it completely | Chrome & JAWS do not speak it, although there seems to be a tab stop for the frame | Firefox and NVDA announce iframe title |
| empty iframe with tabindex=1 | VO ignores it completely | Chrome & JAWS do not speak it, although there seems to be a tab stop for the frame | Firefox and NVDA announce iframe title |
| iframe with focusable content | VO announces the iframe | Chrome & JAWS announce the iframe title and contents as you tab to the iframe | Firefox and NVDA announce the iframe title and contents as you tab to the iframe |
| iframe with non-focusable content | VO announces the iframe | JAWS & Chrome announce the iframe title, iframe content needs to be reviewed using browse mode | Firefox and NVDA announce the iframe title and contents as you tab to the iframe |

> Interesting caveat; VO in Safari does not look at aria-label. It only announces title. We should put that in Accessibility support.

Both JAWS and NVDA work well with `title` and `aria-label`
> 
> This tells me what matters here, at least for VO is if the iframe has any content. I tested with an image with alt="", if that's the only thing in the frame, VO ignores it. I think we need to do more testing before we continue with this one.

As said above, in Chrome there seems to be tab stops for each of the empty iframes (with and without tabindex value), there is no visual clue that those empty frames are focused but there might be some programmatic focus there. This might relate to https://github.com/act-rules/act-rules.github.io/issues/1170#issue-564753539

PR #1270, opened 2020-04-08
test: update ignore list with (dynamically) capitalised word(s)

jkodu:
It will reduce the number of items we keep in the ignore list. Sometimes redundant capitalisations of a given keyword. Eg: https://github.com/act-rules/act-rules.github.io/pull/1260/files#diff-152f1b4344e056476d2babcc5fc186f2R122

Closes issue(s):
- NA

Need for Final Call:
- No

### DIFF
--- a/__tests__/spelling-ignore.yml
+++ b/__tests__/spelling-ignore.yml
@@ -50,11 +50,9 @@
 - testrunner
 - color
 - focusable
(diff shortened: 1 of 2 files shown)


jkodu:
Given this is a test improvement, going to merge this without 3 approvals.

PR #1272, opened 2020-04-20
All table header cells have assigned data cells (d0f69e): Fix Passed example 4

Jym77:
There was a problem with [Passed Example 4](https://act-rules.github.io/rules/d0f69e#passed-example-4).

Due to the empty data cell in the top-left corner, the cells "Breakfast", "Lunch", and "Dinner", are **not** "column headers". [Definition of column header](https://html.spec.whatwg.org/multipage/tables.html#column-header):
> The cell's scope attribute is in the auto state, and there are no data cells in any of the cells covering slots with y-coordinates y .. y+height-1.

(the top left data cell is covering slot (0, 0) with y-coordinate 0, and these header cells have y=0 and height=1).

Therefore, when performing the [internal algorithm for scanning and assigning header cells](https://html.spec.whatwg.org/multipage/tables.html#internal-algorithm-for-scanning-and-assigning-header-cells) for, say, cell "8:00" and reaching cell "Breakfast", step 9.4 case Δx=0 second point:
> If the _current cell_ is not a column header, then let _blocked_ be true.

sets _blocked_ to true and step 5
> If _blocked_ is false, then add the _current cell_ to the _headers list_.

does **not** add the cell to the _headers list_.

Ultimately, this cell is not assigned to any data cell, and the rule fails.

Easy fix is to manually add the `scope="col"` to all these cells.
(other fix could be to replace the top-left `td` by a `th`, but https://www.w3.org/WAI/tutorials/tables/two-headers/#table-with-header-cells-in-the-top-row-and-first-column is using a `td` with explicit scope, so following that)

Closes issue(s):

- N/A

Need for Final Call:
This will require a 1 week Final Call (fixing example)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/table-header-cells-refer-to-data-cells-d0f69e.md
+++ b/_rules/table-header-cells-refer-to-data-cells-d0f69e.md
@@ -131,9 +131,9 @@ Each row and column header element has assigned cells, within the same element h
 	<thead>
 		<tr role="row">
 			<td></td>
-			<th role="columnheader">Breakfast</th>
-			<th role="columnheader">Lunch</th>
-			<th role="columnheader">Dinner</th>
+			<th scope="col" role="columnheader">Breakfast</th>
+			<th scope="col" role="columnheader">Lunch</th>
+			<th scope="col" role="columnheader">Dinner</th>
 		</tr>
 	</thead>
 	<tbody>


Jym77:
Final Call ends on April 30.

Jym77:

PR #1274, opened 2020-04-22
Fix expectation to exclude self-reference

### LINKED ISSUE #1261: Headers attribute specified on a cell refers to cells in the same table element [a25f45]: Problem with failed example 2
[Failed example 2](https://act-rules.github.io/rules/a25f45#failed-example-2) has a cell with a headers attribute referring to itself, and this is the reason why it fails.
```html
<td id="headerBday" headers="headerBday">
```
However, nothing in the expectation says that the ID have to refer to a **different** cell:
> each of which is the ID of an element, that is a cell of the same table as the target element.

I agree that referring to itself is rather bad (notably because self-referencing are removed by the assignation algorithm).
Thus, the rule should probably be updated to specify that the ID must be of a cell different from the one the test target is set on.

Jym77:
Exclude the target itself from the valid `headers` IDs.
This follow assignation algorithm removing the principal cell from the list of headers, and makes Failed Example 2 actually fail.

Closes issue(s):

- closes #1261

Need for Final Call:
This will require a 1 week Final Call (small update to expectation)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve

### DIFF
--- a/_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
+++ b/_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
@@ -30,7 +30,9 @@ This rule applies to any `headers` attribute specified on a [`cell`][] within a 
 
 ## Expectation
 
-Each target attribute is [a set of space separated IDs][], each of which is the ID of an element, that is a [`cell`][] of the same [`table`][] as the target element.
+Each target attribute is [a set of space separated IDs][], each of which is the ID of an element, that is a [`cell`][] of the same [`table`][] as the target element different from the target.
+
+**Note:** `headers` attribute referencing to the cell itself are ignored when [assigning header cells (step 6)](https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-assigning-header-cells).
 
 ## Assumptions
 
@@ -217,7 +219,7 @@ The `td` element has a `headers` attribute referring to an ID that does not exis
 
 #### Failed Example 2
 
-The `td` element has a `headers` attribute referring to it's own ID.
+The `td` element has a `headers` attribute referring to its own ID.
 
 ```html
 <table>
--- a/_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
+++ b/_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
@@ -28,9 +28,13 @@ acknowledgments:
 
 This rule applies to any `headers` attribute specified on a [`cell`][] within a [`table`][] element, where the [`table`][] element is [visible][] and [included in the accessibility tree][].
 
-## Expectation
+## Expectation 1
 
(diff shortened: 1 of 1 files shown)


_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
```diff
@@ -30,7 +30,9 @@ This rule applies to any `headers` attribute specified on a [`cell`][] within a
 
 ## Expectation
 
-Each target attribute is [a set of space separated IDs][], each of which is the ID of an element, that is a [`cell`][] of the same [`table`][] as the target element.
+Each target attribute is [a set of space separated IDs][], each of which is the ID of an element, that is a [`cell`][] of the same [`table`][] as the target element different from the target.
```
WilcoFiers, line 33: I think this should be handled in a separate expectation.

Jym77: Yes, that will be easier to read 👍

kasperisager:
By the way, are we sure we want something like `headers="other self"` to fail as well?

Jym77:
> By the way, are we sure we want something like `headers="other self"` to fail as well?

I assumed so because the rule was failing `headers="other INVALID"`. So, I kept the "a single incorrect headers is bad". I agree that it is a bit different, but for example the header assignation is going to treat them the same way (i.e. ignore them).

kasperisager:
Valid point 👍

carlosapaduarte:
I'm not really convinced that `headers="other self"` should fail. If the `self` is removed by the assigning algorithm we are left with `headers="other"`, which passes.
The `headers="other Invalid"` example is different because you will have an invalid reference. 
I would agree to an expectation that checked the list is not composed by self references only, but if there are other valid references in the list I don't think this fails the SC.

Jym77:
> I'm not really convinced that `headers="other self"` should fail. If the `self` is removed by the assigning algorithm we are left with `headers="other"`, which passes.
> The `headers="other Invalid"` example is different because you will have an invalid reference.
> I would agree to an expectation that checked the list is not composed by self references only, but if there are other valid references in the list I don't think this fails the SC.

For the [headers assignation algorithm](https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-assigning-header-cells), "self" and "invalid" are treated exactly the same way (Step 3, first case, point 2):
> For each token in the id list, if the first element in the Document with an ID equal to the token is a cell in the same table, and that cell is not the principal cell, then add that cell to header list.

Given that they are both explicitly ignored by the specs, I do not see why the rule should treat them differently.

I do agree that maybe the expectation should be "at least one is not invalid/self" rather than "none is invalid/self".

Jym77:
Final Call ends on April 30.

Jym77:

PR #960, opened 2019-10-02
New rule: Auto-updating text content can be paused or stopped

carlosapaduarte:
Need for Final Call:
This will require a 2 weeks Final Call 

---

### **When creating PR:**

- Make sure you requesting to **pull a branch** (right side) to the `develop` branch (left side).

### **After creating PR:**

- Add yourself (and co-authors) as "Assignees" for PR.
- Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- Close the issue that the PR resolves (and make sure the issue is referenced in the top of this comment)
- Optionally request feedback from anyone in particular by assigning them as "Reviewers".

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Final Call period. In case of disagreement, the longer period wins.

PR #1276, opened 2020-04-23
New rule: Device motion based changes to the content can also be created from the user interface

carlosapaduarte:
Atomic rule that is split out from PR #880 following a major revision of that PR.

Need for Final Call:

---

## How to Review And Approve


_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,191 @@
...
+  - 'attr-lowercase'    
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available supporting the same outcomes of the event.
```
Jym77, line 33: I think that UI components needs to be visible/accessible (but do not require accessible name).
> a part of the content that is perceived by users as a single control for a distinct function

I don't think that something that has `display:none` "can be perceived as" anything given that it cannot be perceived at all.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,191 @@
...
+- [G213: Provide conventional controls and an application setting for motion activated input](https://www.w3.org/WAI/WCAG21/Techniques/general/G213.html)
+- [DeviceOrientation Event Specification](https://www.w3.org/TR/orientation-event/)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This [HTML document][] that can be operated through the device's orientation to increase and decrease the value of a slider has the same functionality available through `button` elements.
```
Jym77, line 56: "instrument" is defined as "an HTML element" (https://github.com/act-rules/act-rules.github.io/pull/960/files?file-filters%5B%5D=.js&file-filters%5B%5D=.md#diff-765403c47c39874f2af13d4ec2e178aaR8)

So, technically, the "instrument" is here the `div`, not the **two** `buttons`…

Maybe it's better (more flexible) to change the expectation and require "a set of instruments" or something similar…

(same on other examples)

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: Agreed. Reinstated the expectations and examples.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -28,9 +28,17 @@ htmlHintIgnore:
...
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, an [instrument][] is available supporting the same outcomes of the event.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, a set of one or more [instruments][instrument] is available supporting the same outcomes of the event.
+
+## Expectation 2
+
+The [instrument][] is [visible][].
+
+## Expectation 3
+
+The [instrument][] is [included in the accessibility tree][] with an [accessible name][] that is not empty ("").
```
Jym77, line 41: I think that WCAG "user component" does not require a non-empty accessible name…
Unless we consider that "is perceived as…" requires an accessible name for the "as…" part.

carlosapaduarte: WCAG's definition of user interface component
> a part of the content that is perceived by users as a single control for a distinct function 

WCAG's example of user interface component
> An applet has a "control" that can be used to move through content by line or page or random access. Since each of these would need to have a name and be settable independently, they would each be a "user interface component." 

From the definition, I would argue that a name is needed for a user to perceive the function. From the example, it's clear it needs to have a name, don't you agree?

Jym77: Fair enough.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,316 @@
...
+  - 'attr-lowercase'    
+---
+
+## Applicability
+
+The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, a set of one or more [instruments][instrument] is available supporting the same outcomes of the event.
```
WilcoFiers, line 33: Got a few questions about this:
- "available": what does that mean? In the DOM? On another page? Anywhere on the web? As in, how do we ensure testers will do this consistently?
- "supporting the same outcomes" How do you know that something is the same outcome? Presumably this isn't just limited to "outcomes" that are apparent to the user, but things users might not be aware of too. What does it mean for the outcome of one event to be "the same" as that of another?

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,316 @@
...
+
+## Expectation 2
```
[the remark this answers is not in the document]
carlosapaduarte: As pointed out by @Jym77 the SC requires that the same functionality can be operated through a user interface component. According to WCAG, a user interface component is something that "is perceived by users as a single control for a distinct function". Without expectations 2 and 3 I don't think we can guarantee that.

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: - I've added to expectation 1 that the instrument should be located on the same web page. Since this instrument is supposed to offer the same functionality that is available through device motion, it does not make sense that it is located elsewhere (in this regard it is different from the instrument that is required to disable the device motion).
- Since this is an expectation and not the applicability, there is no need to objectively define "outcome" or "same". I agree that there might be some ambiguity regarding the scope of "outcomes". I've added a note clarifying that **all** outcomes resulting from the device motion should be considered. Given that you also expressed concerns about the ambiguity of "same" (I don't think it is ambiguous), I've replaced it with a synonym: indistinguishable.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -26,19 +26,21 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to an [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+The rule applies to any [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
 
 ## Expectation 1
 
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, a set of one or more [instruments][instrument] is available supporting the same outcomes of the event.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, a set of one or more [instruments][instrument] is available in the same [web page][] supporting outcomes indistinguishable of those from the event.
```
WilcoFiers, line 33: I don't think we can use the word "outcome" here. That's already defined in ACT and has a different meaning from how it is used here.

WilcoFiers, line 33: If the instrument isn't visible at all times, it sounds like it would fail this rule. That doesn't seem like it should fail the SC though. Especially on mobile apps, it's pretty common for functionality to be organised in menus / popups / collapsable groups. Doesn't seem like that should fail the rule.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -26,19 +26,21 @@ htmlHintIgnore:
...
+
+**Note:** The outcomes considered in this expectation are all the outcomes that result from the device motion operated functionality.
```
WilcoFiers, line 35: This seems like it is specifying something about what an "outcome" is. That shouldn't be in a note.

_rules/device-motion-user-interface-7677a9.md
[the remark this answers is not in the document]
WilcoFiers: I disagree. An offscreen skiplink with aria-hidden on it is still a user interface component.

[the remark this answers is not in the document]
carlosapaduarte: Can't disagree with that!

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: Replaced "outcome" with one of its synonyms: "result"

carlosapaduarte: Removed the reference to being available in the same page. Replaced it with the same solution that I used here: https://github.com/act-rules/act-rules.github.io/pull/960#discussion_r423577207

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: Removed this note. Check if the rephrasing that resulted from replacing "outcome" with "result" addresses satisfactorily the requirements that led to this note.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -30,17 +30,16 @@ The rule applies to any [HTML document][] with an associated [Window object][] t
...
 
-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target, a set of one or more [instruments][instrument] is available in the same [web page][] supporting outcomes indistinguishable of those from the event.
-
-**Note:** The outcomes considered in this expectation are all the outcomes that result from the device motion operated functionality.
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] supporting indistinguishable results of those from the event.
 
 ## Expectation 2
 
-Each [instrument][] from expectation 1 is [visible][].
+For each [instrument][] in the set of [instruments][instrument] from Expectation 1, one of the following is true:
```
WilcoFiers, line 37: Minor issue; if there are two instruments to achieve one function, both of them would be applicable. I think you can reformulate this: "for each test target, at least one instrument from expectation 1 ...."

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -30,17 +30,16 @@ The rule applies to any [HTML document][] with an associated [Window object][] t
...
 
-## Expectation 3
+- the [instrument][] is in the same [web page][] of the test target; or
+- instructions for the user to locate the [instrument][] are in the same [web page][] of the test target.
```
WilcoFiers, line 40: I don't think WCAG requires that there are instructions on every page.

_rules/device-motion-user-interface-7677a9.md
Jym77: I see the point but I think it's a bit more complex to formulate.
Take for example a value that can be changed (up and down) by tilting the phone. There can be only 1 registered event for it, but 2 buttons for covering the feature (one for +, one for -). Then **both** buttons need to pass Expectation 2.

On the other hand, if there are 3 buttons to increase the slider, then, yes, only one of these need to pass Expectation 2.

Which might lead into a rabbit hole if the device motion event has a lot of possible outcomes and components to cover these are doing weird partial grouping that overlap 🙃

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: @WilcoFiers Yes. Same update resorting to #960 definition of 'clearly labeled location'

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: @WilcoFiers @Jym77 if there are 2 instruments (i1 and i2) that do the same thing we can identify 3 sets that meet expectation 1 (s1 = {i1}, s2 = {i2}, s3 = {i1, i2}). If i1 is "findable" and i2 is not "findable" then s1 would pass expectation and the rule while the other sets would not.

If both i1 and i2 are needed to support the same functionality of the event then we only have s3 and both i1 and i2 need to be "findable".

So, because of the second scenario, we need all instruments to pass expectation 2. But in the first scenario that might cause the rule to fail when s2 and s3 are considered. I think we can address this with an assumption that the instruments in the set (if there is more than one) fulfil different parts of the event's result.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -37,14 +37,15 @@ For each registered [device orientation event][device orientation] or [device mo
...
-- instructions for the user to locate the [instrument][] are in the same [web page][] of the test target.
+- the [instrument][] can be found in a [clearly labeled location][].
 
 **Note:** If the set of instruments has more than one instrument, not every instrument of the set needs to be located in the same location.
 
 ## Assumptions
 
 - The motion to operate the device is not used through an [accessibility supported][] interface, which is listed as a valid exception to SC 2.5.4.
 - The motion is not [essential][] for the functionality it triggers, which is listed as a valid exception to SC 2.5.4.
+- If there is more than one [instrument][] in the set, this rule assumes that each [instrument][] fulfils a different part of the event's result. If more than one [instrument][] fulfils the same part of the result, then only one of those [instruments][instrument] needs to meet expectation 2. In this instance, this rule will fail while success criterion 2.5.4 might still be satisfied.
```
Jym77, line 48:
```suggestion
- If there is more than one [instrument][] in the set found by Expectation 1, this rule assumes that each [instrument][] fulfils a different part of the event's result. If more than one [instrument][] fulfils the same part of the result, then only one of those [instruments][instrument] needs to meet expectation 2. In this instance, this rule will fail while success criterion 2.5.4 might still be satisfied.
```
+ link to SC.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -138,6 +139,65 @@ This [HTML document][] that can be operated by rotating the device to increase a
...
+	</head>
+
+	<body onload="activateSlider();">
+		<pre class="output"></pre>
+
+		<h1>Slider Motion Sensor Example</h1>
+
+		<p>
+			Open this slider on a device with a motion sensor, such as a smart phone or tablet. Rotate the device to adjust
+			the slider value. Active the control panel to access decrease and increase buttons that also adjust the value.
```
Jym77, line 173:
```suggestion
			the slider value. Open the control panel to access decrease and increase buttons that also adjust the value.
```
"Active" doesn't sound good (not a verb). Maybe "open", "use", "activate" ?

_rules/device-motion-user-interface-7677a9.md
Jym77: I think you're illustrating only part of the problem I see (and this will lead to a solution)…

Expectation 1 can find several sets. Expectation 2 is about "**the** set from Expectation 1", i.e. implicitly assuming that there is only 1.
I think Expectation 2 should be updated to something like "at least one set from Expectation 1". Needs some heavier rephrasing to get the sentence to be correct 😉 This would solve the cases you explain (Expectation 1 finds s1, s2 and s3 and only s1 passes Expectation 2, which is fine) without need for assumption.


(Weird) case where the assumption is not enough:
3 possible results of the event: r₁, r₂ and r₃.
4 instruments: i₁ does r₁, i₂ does r₂ + r₃, i₃ does r₁+r₂ and i₄ does r₃. 
Then there are 2 sets that match the assumption: {i₁, i₂} and {i₃, i₄} but only one of them needs to be findable.
{i₂, i₃} does not match the assumption, but having it findable (while neither i₁ nor i₄ is) is nonetheless good.

With suggested fix, Expectation 1 would find {i₁, i₂}, {i₃, i₄} and {i₂, i₃} and Expectation 2 would want at least one of them to be findable.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,259 @@
...
+  - 'attr-lowercase'
+---
+
+## Applicability
+
+The rule applies to any [HTML document][] with an associated [Window object][] that has an [event listener list][] with one or more [event listeners][event listener] for [device orientation events][device orientation] or [device motion events][device motion].
+
+## Expectation 1
+
+For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target there exists a set of one or more [instruments][instrument] supporting indistinguishable results of those from the event.
```
WilcoFiers, line 33: Same problem as I pointed out for auto-updating. The word "exists" is unspecified. We need to be able to answer "true" or "false" on each expectation individually.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -0,0 +1,259 @@
...
+## Expectation 2
+
+For at least one set of [instruments][instrument] from Expectation 1, all of the [instruments][instrument] in such set meet one of the following conditions:
+
+- the [instrument][] is in the same [web page][] of the test target; or
+- the [instrument][] can be found in a [clearly labeled location][].
+
+**Note:** If the set of instruments has more than one instrument, not every instrument of the set needs to be located in the same location.
+
+## Assumptions
```
WilcoFiers, line 44: You're making the assumption here that every page with these event handlers attached uses them to operate some functionality. That isn't necessarily true. For example if you're playing a game that uses device motion, the event handlers are likely attached before the game is started. The user may have to click "play" before the actual "functionality" appears. The instrument doesn't need to be available from the start, as long as it's available as soon as you're actually playing the game.

I think this is something that needs to be dealt with in this rule, and not something that can just be assumed. I think the cause of this is the difference between our own definition of HTML web page, which is the page in a single state, and WCAG's definition of Web page, which is that page in any state it can possibly be in.

Jym77: Hmmm… The expectation is that there are instruments to support the same results as the event.
If the event has no result (e.g. before pressing "Play"), then we need instruments to support no result, which may arguably be OK with no instruments at all…

Maybe all that is needed is to have Expectation 1 be explicit that if the events do nothing, no instrument is actually needed.

carlosapaduarte: Similar to #880 I'll update the expectation to include the possibility of the events not having any result on the test target.

_rules/device-motion-user-interface-7677a9.md
carlosapaduarte: Combined expectations

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -28,18 +28,12 @@ htmlHintIgnore:
...
 
-## Expectation 2
-
-For at least one set of [instruments][instrument] from Expectation 1, all of the [instruments][instrument] in such set meet one of the following conditions:
-
-- the [instrument][] is in the same [web page][] of the test target; or
-- the [instrument][] can be found in a [clearly labeled location][].
-
-**Note:** If the set of instruments has more than one instrument, not every instrument of the set needs to be located in the same location.
+- **no result:** The registered event does not have results perceivable by the user; or
```
WilcoFiers, line 35:
```suggestion
- **no result:** Triggering the registered event does not cause a [visible change][] to the [web page][]; or
```

Perceivable by the user can be ambiguous. Which user? What's perceivable to an AT user is different from what is perceivable to someone without. I think this works better, but we need to define "visual change". We can't use "visible" here as that relies on transparency, but the definition could be fairly similar. Something like this perhaps:

> Visible changes in content occur when the rendered pixels change in any port of the document that is currently within the viewport or that can be brought into the viewport via scrolling.

carlosapaduarte:
@WilcoFiers @Jym77 I've created a new definition for "changes in content" trying to capture changes to visible content, audible content and accessibility tree content. Then I updated the expectation based on that (both for no changes and for same result changes). Also updated the name and description of the rule accordingly.

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,18 @@
...
+title: Changes in content
+key: changes-in-content
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+Changes in the content of a [web page][] occur when at least one of the following occurs:
```
Jym77, line 11:
```suggestion
A _change in the content_ of a [web page][] occurs when at least one of the following occurs:
```

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,18 @@
...
+
+- **visible changes:** Visible changes in content occur when the rendered pixels change in any part of the document that is currently within the viewport or that can be brought into the viewport via scrolling; or
+- **accessibility tree changes:** Changes in the content of the accessibility tree occur when any state, property or event of a node representing an [accessible object][] of the tree changes, or when any node is inserted in, or removed from the tree; or
+- **audible changes:** Audible changes in content occur when the audio rendered by the [web page][] change.
```
Jym77, line 15:
```suggestion
- **visible changes:** the rendered pixels change in any part of the document that is currently within the viewport or that can be brought into the viewport via scrolling; or
- **accessibility tree changes:** any state, property or event of a node representing an [accessible object][] of the accessibility tree changes, or any node is inserted in, or removed from the accessibility tree; or
- **audible changes:** the audio rendered by the [web page][] changes.
```

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Changes in content
+key: changes-in-content
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
```
Jym77, line 8:
```suggestion
input_aspects:
  - Accessibility tree
  - Audio output
  - CSS styling
  - DOM tree
```

pages/glossary/changes-in-content.md
Jym77, line 15: "Viewport" and "accessibility tree" may need links to specs.

Jym77, line 15: "The audio changes" => when my computer plays some music, the audio keeps changing (I'm not listening to _4'33''_ that often). I have a hunch that this is not the kind of changes you want to include.
OTOH, given that these changes in content are only considered as reaction to some action, this is probably OK (if an app requires me to move the phone to keep playing music, it definitely fails that rule…)

carlosapaduarte: We haven't been doing this in other rules, but why not :-)

carlosapaduarte: You're right. The definition doesn't handle time-based media as it is. If the app allows me to move forward or backwards an audio track by moving the device then it should fail. But when the audio is playing the audio is always changing (unless it's playing 4'33'' :-) )
This also happens if it used to control a video. The rendered pixels are changing as part of the video playing (in addition to any audio the video has).

I'll try to rewrite, introducing a reference to compare the output against in order to identify changes.

pages/glossary/changes-in-content.md
```diff
@@ -4,15 +4,25 @@ key: changes-in-content
...
 objective: true
 input_aspects:
+  - Accessibility tree
+  - Audio output
   - CSS styling
   - DOM tree
 ---
 
-Changes in the content of a [web page][] occur when at least one of the following occurs:
+A _change in the content_ of a [web page][] caused by an event occurs when, by comparing the [web page][] before and after the event firing, at least one of the following occurs:
```
Jym77, line 13: This does assume that firing the event is indeed causing the change.

If there is a page with audio playing non-stop (so, permanent change in content), then any event will be consider as a cause for this (because comparing the page before and after firing the event creates an **audible change**)

So, if there is a non-stop audio and a device motion event that does **not** cause any change in the content (e.g. it update a database in the backend), the definition will nonetheless consider the **audible change** as a "change in the content caused by" the device motion and the rule will fail while 2.5.4 doesn't (non-stop audio is bad for 1.4.2, of course).

Jym77, line 13: And timing strikes here: "after the event firing" => how long after? What happens if there is a delay between firing the event and causing the change? (that was a bit present before and becomes somewhat more obvious with this formulation…)
This is not too bad as not detecting a "change caused by an event" result in passing the rule (due to **no changes**) while failing the SC, but that does not break the fail-on-fail relationship.

pages/glossary/changes-in-content.md
```diff
@@ -4,15 +4,25 @@ key: changes-in-content
...
 
-- **visible changes:** Visible changes in content occur when the rendered pixels change in any part of the document that is currently within the viewport or that can be brought into the viewport via scrolling; or
-- **accessibility tree changes:** Changes in the content of the accessibility tree occur when any state, property or event of a node representing an [accessible object][] of the tree changes, or when any node is inserted in, or removed from the tree; or
-- **audible changes:** Audible changes in content occur when the audio rendered by the [web page][] change.
+- **visible changes:** the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling; or
+- **accessibility tree changes:** any state, property or event of a node representing an [accessible object][] of the [accessibility tree][] changes, or any node is inserted in, or removed from the [accessibility tree][]; or
+- **audible changes:** the audio rendered by the [web page][] changes.
```
Jym77, line 17: I'm actually not sure whether "rendered" is the correct word to use for audio (or whether it is purely visual).

https://html.spec.whatwg.org/multipage/rendering.html
https://html.spec.whatwg.org/multipage/rendering.html#being-rendered
https://html.spec.whatwg.org/multipage/rendering.html#embedded-content-rendering-rules

This part is non normative, but the second link strongly suggests that we should only use "rendered" for visual stuff, not for audio.

pages/glossary/changes-in-content.md
carlosapaduarte: I shouldn't have used "event" because it was obvious that was going to be confused with dom events. I was using "event" with the generic meaning of "origin of the change" whatever that might be (most probably a dom event!)

Nevertheless, in your example, according to this definition, there would be no "change in content". There is a paragraph after the list explaining that when time-based media is at play, the comparison is not between what was before the event and what is after the event, but instead, between what would be if there was no event and what is when there is an event.

carlosapaduarte: I also already had an assumption for this situation: "For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event."

pages/glossary/changes-in-content.md
carlosapaduarte: The W3C Web Audio API uses "audio rendering" has does almost everyone that works in the field :-)
But I agree that the HTML spec does not, so I'll change to "reproduced"

carlosapaduarte:
@Jym77 Please check my replies to your comments. I have done a single change because I believe the other comments had already been addressed. But do let me know if you don't agree.

pages/glossary/changes-in-content.md
Jym77: * "event": the use both of the word "event" and "firing" does make the confusion with DOM events likely. Not really important due to the rest.
* I have indeed missed the rest of the def 😥

Jym77: Missed that assumption indeed 😥

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,28 @@
...
+
+A _change in the content_ of a [web page][] caused by an event occurs when, by comparing the [web page][] before and after the event firing, at least one of the following occurs:
+
+- **visible changes:** the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling; or
+- **accessibility tree changes:** any state, property or event of a node representing an [accessible object][] of the [accessibility tree][] changes, or any node is inserted in, or removed from the [accessibility tree][]; or
+- **audible changes:** the audio reproduced by the [web page][] changes.
+
+If the [web page][] is reproducing time-based media, rendered pixels and reproduced audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered and reproduced if the event is not fired, with the ones that are rendered if the event is fired.
+
+For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
```
Jym77, line 21: Wouldn't the second assumption de facto exclude time-based media? (thus making the first one useless)
"the page does not change if there is no such event" => no time based media is played.

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,28 @@
...
+
+For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
```
Jym77, line 23: I don't know whether this assumption will hold in practice (due to either micro-delays that can happen, or design choice to delay a bit the reaction).
But I guess it is OK given that this is a manual check and that I guess testers won't look at the result within the next micro-seconds…

Jym77, line 23: I think the assumptions:
1. should be made more proeminent, e.g. with an "Assumptions" section in the def.
2. should be copied in any rule using that definition, given that if the definition gives bad result, then the rule will also give bad result (probably together with a "conclusion" like "if this does not hold, then the def will produce nonsense and the rule will fail because of that without the SC failing" as we usually have.

(and I think that doing 1. also helps remembering to do 2.)

We don't really have assumptions in definitions so far, so there is no real agreement on how to handle them and no guidelines, so this is mostly personal opinion on how to handle them…

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,28 @@
...
+  - Audio output
+  - CSS styling
+  - DOM tree
+---
+
+A _change in the content_ of a [web page][] caused by an event occurs when, by comparing the [web page][] before and after the event firing, at least one of the following occurs:
+
+- **visible changes:** the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling; or
+- **accessibility tree changes:** any state, property or event of a node representing an [accessible object][] of the [accessibility tree][] changes, or any node is inserted in, or removed from the [accessibility tree][]; or
+- **audible changes:** the audio reproduced by the [web page][] changes.
```
Jym77, line 17: I have completely missed that "render" was also used for audio in W3C lingo 😥 
I'm backing up on previous comment, "render" is the correct word to use here. Sorry about that 😓

pages/glossary/changes-in-content.md
carlosapaduarte: I don't think so. Take this scenario
1 - The user activates play and the web page starts playing audio in the background (there was a change in content but for the purpose of this rule the event that caused it did not originate from device motion)
2 - The user flicks the device to the right to move to the next track (there was another change in content - this time originated by a device motion event - even though the audio was already playing)

The first assumption kicks in in the second part of this scenario and it is what allows us to detect the change. 

To be honest, I'm still not totally convinced that we need the second one. I can't really come up with an example of something that will cause changes to the content without being event-based.

pages/glossary/changes-in-content.md
carlosapaduarte: That was my reasoning. This is going to be a manual check, therefore we don't really need to concern oursevels with micro-delays.

carlosapaduarte: I like the suggestion. Going to implement it :-)

pages/glossary/changes-in-content.md
carlosapaduarte: No worries!

pages/glossary/changes-in-content.md
```diff
@@ -14,13 +14,13 @@ A _change in the content_ of a [web page][] caused by an event occurs when, by c
...
 
-For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
-
-For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+- For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
+- For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
```
Jym77, line 23:
```suggestion
- This definition assumes that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event. If there is a significant delay between firing the event and the changes happening, this definition may not detect them.
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -39,6 +39,9 @@ For each registered [device orientation event][device orientation] or [device mo
 
 - The motion to operate the device is not used through an [accessibility supported][] interface, which is listed as a valid exception to [SC 2.5.4][sc2.5.4].
 - The motion is not [essential][] for the functionality it triggers, which is listed as a valid exception to [SC 2.5.4][sc2.5.4].
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+- For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
+- For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
```
Jym77, line 44:
```suggestion
- This rule assumes that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event. If there is a significant delay between firing the event and the changes happening, they may not be detected as [changes in content][] and the rule may pass but [SC 2.5.4][sc2.5.4] is not satisfied.
```

pages/glossary/changes-in-content.md
```diff
@@ -14,13 +14,13 @@ A _change in the content_ of a [web page][] caused by an event occurs when, by c
...
+**Assumptions:**
 
-For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
-
-For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+- For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
```
Jym77, line 22:
```suggestion
- This definition assumes that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event. If this is not the case, changes may be attributed to the wrong event.
```

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -39,6 +39,9 @@ For each registered [device orientation event][device orientation] or [device mo
 
 - The motion to operate the device is not used through an [accessibility supported][] interface, which is listed as a valid exception to [SC 2.5.4][sc2.5.4].
 - The motion is not [essential][] for the functionality it triggers, which is listed as a valid exception to [SC 2.5.4][sc2.5.4].
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+- For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
```
Jym77, line 43:
```suggestion
- This rule assumes that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event. If this is not the case, changes may be attributed to the wrong event and the rule may fail while [SC 2.5.4][sc2.5.4] is still satisfied.
```

pages/glossary/changes-in-content.md
```diff
@@ -14,13 +14,13 @@ A _change in the content_ of a [web page][] caused by an event occurs when, by c
...
+- **audible changes:** the audio rendered by the [web page][] changes.
 
-If the [web page][] is reproducing time-based media, rendered pixels and reproduced audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered and reproduced if the event is not fired, with the ones that are rendered if the event is fired.
+**Assumptions:**
 
-For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
-
-For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
```
Jym77, line 21: I don't think this is an assumption.
As in, it's not something that we suppose and that make things go wrong if it doesn't hold.
It's a precision on how to handle a specific situation.

_rules/device-motion-user-interface-7677a9.md
```diff
@@ -39,6 +39,9 @@ For each registered [device orientation event][device orientation] or [device mo
 
 - The motion to operate the device is not used through an [accessibility supported][] interface, which is listed as a valid exception to [SC 2.5.4][sc2.5.4].
 - The motion is not [essential][] for the functionality it triggers, which is listed as a valid exception to [SC 2.5.4][sc2.5.4].
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
```
Jym77, line 42: I think this is not an assumption and doesn't need to be copied in the rule.

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,28 @@
...
+unambiguous: true
+objective: true
+input_aspects:
+  - Accessibility tree
+  - Audio output
+  - CSS styling
+  - DOM tree
+---
+
+A _change in the content_ of a [web page][] caused by an event occurs when, by comparing the [web page][] before and after the event firing, at least one of the following occurs:
```
Jym77, line 13: There is something in the causality that creates confusion. As in, there are two different concepts mixed here:
1. the page changes in one of the 3 ways indicated.
2. the page changes as a reaction to an event.

This does creates confusion in the assumptions. Notably:
* time-based media => it causes changes but not changes due to an event, hence the need to specify "if there's time based media, ignore it".
* changes only caused by the event => which is a bit weird as you pointed out, but becomes clearer and/or redundant if de-coupled from the even causing it. Currently, any "change in the content" has to be caused by an event (or at least perceived as such) which make the assumption that there are no changes not caused by event weird… because this use the other meaning of change…

Maybe it is worth splitting the definition in two ("changes" and "changes caused by an event") or at least to somehow streamline the difference between the two concepts.

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,28 @@
...
+
+- **visible changes:** the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling; or
+- **accessibility tree changes:** any state, property or event of a node representing an [accessible object][] of the [accessibility tree][] changes, or any node is inserted in, or removed from the [accessibility tree][]; or
+- **audible changes:** the audio rendered by the [web page][] changes.
+
+**Assumptions:**
+
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+- For the purpose of this definition, it is assumed that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event.
```
Jym77, line 22: That assumption seems very strong in its current state.
If two different events create changes, the assumption means that they can't happen (almost) simultaneously.

If I come back to the scenario you describe in https://github.com/act-rules/act-rules.github.io/pull/1276#discussion_r440710851 then once the time based media start, anything else (such as tilting to go to next song) will fail this assumption (there is another origin of changes). This is more or less caught by she previous point (ignoring changes that would have occurred anyway).

We both seem uneasy with this assumption (and/or the combination with the time based media note above). So we probably need to do something about it… I am not sure that completely removing it is the way to go…

---
In this definition, "event" can be almost anything (not just a DOM event). So I agree that changes that happen will be caused by some event, possibly far away (e.g. playing a 10h video will cause changes for 10h).

We need a way to figure out that a given change is caused by a given event. This is currently the note about the time based media, but could apply to other changes (e.g. caused by an asynchronous request completing, this can potentially create change long after the event that caused it, unless we consider that the request completion is the "event" (and not sending the request)).

This does illustrate that there is some degree of subjectivity (hopefully not ambiguity) in the definition. "event" not being a DOM event creates place for interpretation, hence subjectivity. This is not a problem but the frontmatter should be updated accordingly.

---
I thing we're pretty solid on what a "change" is (i.e., objective, unambiguous). But not so on what an "event" is (if not DOM event) and consequently on the causality. Maybe trying to get a better definition of "event" will help us figure out problems with this assumption?

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,28 @@
...
+- For the purpose of this definition, it is assumed that the changes happen instantly after the event firing and therefore the comparison between the page before and after the event firing can be made at any time after the firing of the event.
```
Jym77, line 23: Also, maybe sort of assuming that the event itself is instantaneous 🤔 
But that might be part of what we call "event" here.
So, an asynchronous request could be two events (making the request, getting an answer), or 1 non-instantaneous (making he request), or maybe 4 (starting to send the request, finishing sending, starting to receive the answer, finishing it).

---
Let's say that I can (maybe by tilting the phone) send a request for a bunch of data. After some times, I start receiving that data. It takes some time to download everything because I'm on an old EDGE network 📶 (or just because there is a lot of data) Once download is done, the phone takes some time processing it (e.g., computing some aggregates). Once this processing is done, the page is updated instantaneously (or at least faster than I can notice).

How many events are there in this scenario? Which one is causing the change in the page? Does the rule apply (or rather requires an UI control)?

pages/glossary/changes-in-content.md
carlosapaduarte: I hope that you agree that fixing this to dom events solves the problem :-)

pages/glossary/changes-in-content.md
carlosapaduarte: But this definition is about "change in the content of a web page caused by an event"

I don't think it helps to have a definition of just change (because change implies comparing something, which implies having to define what is going to be compared).

However, I do think it will help if we focus on dom events, given that this definition if being used in rules that deal with dom events. Hopefully, this will make it easier to address the issues.

pages/glossary/changes-in-content.md
carlosapaduarte: This is, for me, the biggest issue with the current definition... even moving to dom events we can have a device motion event that triggers an asynchronous call that will resolve only after 1 minute (or whatever other duration).
The only other option I see, is using the "10 minute" workaround. What do you think @Jym77?

carlosapaduarte:
@Jym77 I've updated the definition (and rule) to use an arbitrary time limit. Since this is looking at device-motion originated events, I used 1 minute instead of 10 minutes. But, given it's arbitrary, I'll be happy to change it to whatever value the group feels reasonable.

I'm also updating the other device motion rule accordingly to the changes in the definition of "changes in content".

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,31 @@
...
+
+- **visible changes:** the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling; or
+- **accessibility tree changes:** any state, property or event of a node representing an [accessible object][] of the [accessibility tree][] changes, or any node is inserted in, or removed from the [accessibility tree][]; or
+- **audible changes:** the audio rendered by the [web page][] changes.
+
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+
+**Assumptions:**
+
+- This definition assumes that the event causing the changes is the only origin of changes to the [web page][] and therefore the page does not change if there is no such event. If this is not the case, changes may be attributed to the wrong event.
```
Jym77, line 23: I am still uneasy about the combination of these two.
I totally get the idea (I think) and agree with it, but the formulation seems contradictory…

It looks like we say on one hand what to do if there are other changes happening and on the other hand that no other changes are happening…

Of course, this is because the assumption is about "changes caused by something (DOM event or else in that case)" while the note above is about "ongoing changes that are not caused by anything anymore".

Given that the assumption is copied in the rules, I'd say it is the part that need clarification.

Maybe something like
> This definition/rule assumes that there are no changes in the content caused by another event.

pages/glossary/changes-in-content.md
```diff
@@ -0,0 +1,31 @@
...
+unambiguous: true
+objective: true
+input_aspects:
+  - Accessibility tree
+  - Audio output
+  - CSS styling
+  - DOM tree
+---
+
+A _change in the content_ of a [web page][] caused by an [event][] occurs when, by comparing the [web page][] before and 1 minute after the event [firing][], at least one of the following occurs:
```
Jym77, line 13: I'm considering
```suggestion
A _change in the content of a [web page][] caused by an [event][]_ occurs when, by comparing the [web page][] before and 1 minute after the event [firing][], at least one of the following occurs:
```
to emphasise that the event causing it is an important part of the definition (vs, say, time-based media).
Not super happy about this because the "of a web page" in the middle is diluting the intention of emphasising "change in the content caused by an event"…

carlosapaduarte: I think I got a way to address that

pages/glossary/changes-in-content.md
carlosapaduarte: I see your point and I agree that it can be better formulated. I'll update based on your suggestion.

carlosapaduarte:
Final call ends July 15

carlosapaduarte:
Final call ended. Merging

PR #1281, opened 2020-04-24
Element with `role` attribute has required states and properties (4e8ab6): Update failed examples

### LINKED ISSUE #1173: None of the failed examples for 'Element with role attribute has required states and properties' are failures
`aria-controls` should only be set for `role=combobox` when the combobox popup is visible and ` aria-expanded=true`:
https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-6

So these aren't a failures because the default state is ` aria-expanded=false`:
https://act-rules.github.io/rules/4e8ab6#failed-example-2

Jym77:
The failed examples where not showing accessibility issues because `aria-controls` is only needed when `aria-expanded` is true.

Closes issue(s):

- closes #1173

Need for Final Call:
This will require a 1 week Final Call (updating examples).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
Final Call ends on May 1st.

Jym77:

PR #1283, opened 2020-04-24
`id` attribute value is unique (3ea0c8): Turn failed examples into actual accessibility issues

### LINKED ISSUE #1181: Show proper a11y issues in duplicate ID rule [3ea0c8]
The examples all use `div` elements, which are not known to cause accessibility problems. It would be better to have known accessibility problems as failure examples for this rule.

Jym77:
Use duplicates `id`s in `aria-labelledby` so that they are actually causing an accessibility issue.

Closes issue(s):

- closes #1181

Need for Final Call:
This will require a 1 week Final Call (updating examples)

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


Jym77:
Final Call ends on May 1st.

Jym77:

PR #1284, opened 2020-04-28
"Image filename is accessible name for image" (9eb3f6): Process TF feedback –the easy part

Jym77:
Processing feedback of the TF on the rule "Image filename is accessible name for image".
https://github.com/w3c/wcag-act/issues/447

Focusing here on:
- explaining the intention of the rule (note after expectation);
- showcasing the use of language (new Passed Example 4);
- generally improving all examples with both better images and a slightly more diverse set of cases.

Updating the applicability according to open issues is done in #1285.

**Note:** This does change all the examples and will thus invalidate all implementations. We'll probably need new implementation data before submitting to TF again. In any case, #1285 will also need updated implementation data…

Closes issue(s):

- N/A.

Need for Final Call:
This will require a 1 week Final Call (changing all examples).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -61,45 +68,88 @@ _There are no major accessibility support issues known for this rule._
 
 #### Passed Example 1
 
-This `img` element has an [accessible name][] equivalent to the filename. The [accessible name][] accurately describes the image.
+This `img` element has an [accessible name][] equivalent to the filename (ignoring letter casing). The [accessible name][] accurately describes the image.
 
 ```html
 <html lang="en">
-	<img src="https://www.w3.org/WAI/demos/bad/img/w3c" alt="w3c" />
+	<img src="/test-assets/image-filename-as-accessible-name-9eb3f6/nyhavn" alt="Nyhavn" />
```
WilcoFiers, line 75: Does this actually work? I'm concerned the file you're referencing here isn't going to get served with the right HTTP headers, and so this example won't work. @jkodu any thoughts on how we might make this work?

_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -61,45 +68,88 @@ _There are no major accessibility support issues known for this rule._
...
+
+This image button has an [accessible name][] equivalent to the filename. The [accessible name][] accurately describes the purpose of the button.
+
+```html
+<html lang="en">
+	<input type="image" src="test-assets/image-filename-as-accessible-name-9eb3f6/login" alt="login" />
+</html>
+```
+
+#### Passed Example 4
```
WilcoFiers, line 101: This example seems unnecessary. Why did you add this?

_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -61,45 +68,88 @@ _There are no major accessibility support issues known for this rule._
...
 <html lang="en">
-	<img src="https://www.w3.org/WAI/demos/bad/img/w3c.png" alt="w3c.png" />
+	<img src="/test-assets/image-filename-as-accessible-name-9eb3f6/paris" alt="Paris" />
 </html>
 ```
 
 #### Failed Example 2
 
-This `input` element with a `type` of `image` has a [semantic role][] of `img` and an [accessible name][] matching the filename. The presence of the file extension in the [accessible name][] is redundant and results in the [accessible name][] not accurately describing the image.
+This `img` element has [accessible name][] matching the image filename. The name is just a checksum and does not describe the image.
```
WilcoFiers, line 125: What's the purpose of this example? What does it add that isn't already covered?

_rules/image-filename-as-accessible-name-9eb3f6.md
Jym77: My tests indicate this works in Chrome, Edge, and Firefox (the image is displayed correctly).
Probably thanks to magic number in the file.

_rules/image-filename-as-accessible-name-9eb3f6.md
Jym77: I think we need a passed example with an image button, given that the rule is targeting them.

_rules/image-filename-as-accessible-name-9eb3f6.md
Jym77: I agree that it is similar to the previous one.
Such a case was mention in the TF review by @maryjom so I thought it might be good to have it.
It showcases a common case of very bad filename (while the first failed example is a bit unlikely but is better at showing that the rule requires a **descriptive** accessible name, not just a **meaningful** one).

WilcoFiers: OK sure, why not I suppose.

Jym77:
Final Call ends on May 14.

PR #1285, opened 2020-04-28
"Image filename is accessible name for image" (9eb3f6): Process TF feedback –update applicability

### LINKED ISSUE #465: "Filename is valid accessible name" (9eb3f6) should limit its applicability
The rule currently applies to all `<input type="image">` elements in addition to any element with a role of `img` and then proceeds to talk about the `src` attribute in its expectations. However, the `src` attribute only has meaning in context of `<input type="image">` and `` elements. As such, the rule should not be applicable to a case such as this:

```html
<div role="img" aria-label="foo.png" src="foo.png"></div>
```

### LINKED ISSUE #1244: image filename, consider srcset / picture + source as well [9eb3f6]
The rule [Image filename is accessible name for image](https://act-rules.github.io/rules/9eb3f6) is limited to the `src` attribute. It could be useful to extend this rule to use srcset attributes, and to consider the `source` element if used in a `picture`.

Jym77:
Update the applicability of "Image filename is accessible name for image" by:
- only consider `img` elements, not anything with a role of `img` (where `src` has no meaning) (still keeping image buttons);
- consider any image in the source set, not just the `img` attribute;
- explicitly remove empty names that are causing weird issues.

**Note:** The change in applicability will invalidate all implementations. We'll need to get implementation data anew before re-submitting to TF.

**Note:** test are failing until #1284 is merged.

Closes issue(s):

- closes #465 
- closes #1244 

Need for Final Call:
This will require a 1 week Final Call (updating applicability).

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -149,10 +208,25 @@ This `img` element has an [accessible name][] which is not equivalent to the fil
 </html>
 ```
 
+#### Inapplicable Example 5
+
+This `img` element has an empty [accessible name][].
+
+```html
+<html lang="en">
+	<img src="?" aria-label="" />
```
WilcoFiers, line 217: I don't quite understand this example. It seems like this might fail SC 1.1.1; aria-label won't cause it to get ignored by AT.

Jym77: Yes, you're right. This example was pointed out in https://github.com/act-rules/act-rules.github.io/issues/465#issuecomment-477103962 to explain why the rule should ignore empty accessible names. But it is indeed failing the SC and thus should not be here…

Jym77:
> Looks good, just please fix the failing test.

Will be fixed once #1284 (which just went to Final Call) is merged.

_rules/id-value-unique-3ea0c8.md
```diff
@@ -152,14 +152,12 @@ The `xml:id` attribute is not considered by this rule.
...
 
-
 #### Inapplicable Example 3
 
 These `id` attributes have an empty value.
 
 ```html
-<span id="">Hello</span>
-<span id="">world!</span>
+<span id="">Hello</span> <span id="">world!</span>
```
WilcoFiers, line 160: Can you take out these prettier changes? This effects an example unrelated to the PR.

Jym77: This is actually somewhat difficult. Prettier is run automatically as a commit hook.
Pushing a non-prettied file actually requires going through a weird process to fore ignore it.

_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -28,26 +28,34 @@ input_aspects:
...
 ---
 
 ## Applicability
 
-The rule applies to any HTML element with the [semantic role][] of `img` or any HTML `input` element with a [`type`][type] of `image` when each of the following is true:
+The rule applies to any HTML `img` element or any HTML `input` element in the [Image Button][] state when each of the following is true:
 
 - the image is [included in the accessibility tree][]; and
-- the image has an [accessible name][] that is equivalent to the [filename][] specified in the `src` attribute. Difference in letter casing, leading and trailing [whitespace][] should be ignored.
+- the image has a non-empty [accessible name][] that is equivalent to the [filename][] of at least one of the [image sources][] in its [source set][] (for `img` elements), or to the [filename][] specified in its `src` attribute (for [Image Buttons][image button]). Difference in letter casing, leading and trailing [whitespace][] should be ignored.
```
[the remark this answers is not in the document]
Jym77: I agree that it doesn't read well…

I tried to use the orthogonal enumeration. I am not sure this read much better, notably because the expectation still has a part specific to source sets.


The cleanest way to split this would be to make separate rules. It made sense to group them when we were looking at the same thing (the `src` attribute), but now it make little sense to have them together (as we see with the wonky formulations we have to use).

_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -28,26 +28,35 @@ input_aspects:
...
 ## Applicability
 
-The rule applies to any HTML element with the [semantic role][] of `img` or any HTML `input` element with a [`type`][type] of `image` when each of the following is true:
+The rule applies to:
 
-- the image is [included in the accessibility tree][]; and
-- the image has an [accessible name][] that is equivalent to the [filename][] specified in the `src` attribute. Difference in letter casing, leading and trailing [whitespace][] should be ignored.
+- any HTML `img` element which is [included in the accessibility tree][], and has a non-empty [accessible name][] that is equivalent to the [filename][] of at least one of the [image sources][] in its [source set][]; and
+- any HTML `input` element in the [Image Button][] state which is [included in the accessibility tree][], and
+  has a non-empty [accessible name][] that is equivalent to the [filename][] specified in its `src` attribute.
```
carlosapaduarte, line 43:
```suggestion
- any HTML `input` element in the [Image Button][] state which is [included in the accessibility tree][], and has a non-empty [accessible name][] that is equivalent to the [filename][] specified in its `src` attribute.
```

_rules/id-value-unique-3ea0c8.md
WilcoFiers: The commit hook only goes through files that were changed in the commit.

_rules/image-filename-as-accessible-name-9eb3f6.md
WilcoFiers, line 43: I think you should lift the repetitive stuff into this section.

```suggestion
The rule applies to any HTML [embedded](https://html.spec.whatwg.org/#embedded-content-2) element that is [included in the accessibility tree][] and has a non-empty (`""`) [accessible name][], for which one of the following is true:

- **img**: the element is an `img` with an [accessible name][] that is equivalent to the [filename][] of at least one of the [image sources][] in its [source set][]; or
- **input image**: the element is an `input` element in the [Image Button][] state with an [accessible name][] that is equivalent to the [filename][] specified in its `src` attribute.
```

Jym77:
> * First one: I think we could eliminate it, as this is a part of the spec that we are explaining. I am hesitant that we should be explaining these in the rules themselves.

Yes. I added it because in the specs, the link from "source set" (at the start of 4.8.4.3) to the algorithm to update it (4.8.4.3.7) is not obvious. It took me some time to make that link myself, and I felt that linking only to "source set", while correct, is not really helping.

> * The other ones I see them as nice to have, but I also think we can get rid of them. In my vview, what they are explaining is going to become evident once people take a look at the examples.

Yes. I added them trying to prevent an upcoming "these are bad examples we shouldn't use them" issue.

I agree that they are not essential (basically, no note is essential, otherwise its content should not be in a note but in the normative text…) and will remove them.

daniel-montalvo:
> 
> 
> > * First one: I think we could eliminate it, as this is a part of the spec that we are explaining. I am hesitant that we should be explaining these in the rules themselves.
> 
> Yes. I added it because in the specs, the link from "source set" (at the start of 4.8.4.3) to the algorithm to update it (4.8.4.3.7) is not obvious. It took me some time to make that link myself, and I felt that linking only to "source set", while correct, is not really helping.
> 
> > * The other ones I see them as nice to have, but I also think we can get rid of them. In my vview, what they are explaining is going to become evident once people take a look at the examples.
> 
> Yes. I added them trying to prevent an upcoming "these are bad examples we shouldn't use them" issue.
> 
> I agree that they are not essential (basically, no note is essential, otherwise its content should not be in a note but in the normative text…) and will remove them.

Good! Will approve once done.

Jym77:
@daniel-montalvo I've removed the notes.

Jym77:
Final Call ends on May 20.

Jym77:

PR #1290, opened 2020-04-28
Semantic role break circularity

### LINKED ISSUE #1230: Circular reference in the definition of semantic role
As pointed out by @WilcoFiers (https://github.com/act-rules/act-rules.github.io/pull/1189#discussion_r390910846) our definition of "semantic role" depends on the definitions of "explicit/implicit role" but both these depends on the definition of "semantic role"…

---
I think this isn’t too much of a problem because the circular reference is technical more than logical. That is, "semantic role" is both the semantic association (a concept) and the actual role of an element (a technical thingie) that are unfortunately named the same way. The dependence goes that way:
* The concept of "semantic role" is defined:
> A semantic role is a semantic association that indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
* The technical terms "explicit/implicit role" are defined and point to the concept since they are way to flesh it out:
> The explicit semantic role is the semantic role of an element set by its role attribute.
* The technical term "semantic role" is defined, and it points to "explicit/implicit role" as it can be any of these and depends on these terms:
> The semantic role of an element is its explicit semantic role if it has any, otherwise, the implicit semantic role is used.

So, all in all, I think this isn’t too bad. This still needs to be solved and properly unfolded.

---
The less intrusive way I see to solve it is to introduce a definition of "semantic association", or something similar, for the concept. That way "explicit/implicit role" could refer to this. This would not change any of the technical terms, hence not change what we actually use in the rules.

Another way to solve this could be to inline the definition of "explicit/implicit role" inside the definition of "semantic role" (back to what we had last summer, essentially). This would however make it somewhat harder to directly refer to the "explicit/implicit roles", something we now do quite often, so I’d rather avoid that…

---
Thoughts?
I’ll probably try and write a PR for this once #1189 is done (to avoid 2 PRs on roles at the same time…)

Jym77:
Break the circularity between definition of roles by adding "semantic association" as an abstract notion which the "semantic role" is one way to implement.

Also makes the "explicit" and "implicit" parts more proeminent in their corresponding definitions.

Closes issue(s):

- closes #1230

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/semantic-role.md
```diff
@@ -7,13 +7,13 @@ input_aspects:
...
 ---
 
-A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic association that indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
+A _semantic association_, also called [role][], indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
 
-Valid semantic roles are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications](https://act-rules.github.io/glossary/#wai-aria-specifications).
+Valid semantic associations are defined by standards. For web content and applications these are defined within the various [WAI-ARIA specifications][].
 
-The _semantic role_ of an element is determined by the first of these cases that applies:
+The _semantic role_ of an element is one way to give it a semantic association. It is determined by the first of these cases that applies:
```
carlosapaduarte, line 14: If it is "one way" what are the other ways?

pages/glossary/semantic-role.md
```diff
@@ -7,13 +7,13 @@ input_aspects:
   - DOM tree
 ---
 
-A _semantic [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role)_ is a semantic association that indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
+A _semantic association_, also called [role][], indicates an object's type. This allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type.
```
carlosapaduarte, line 10: This feels strange. The definition is of semantic role. However, the reader is presented with the definition of semantic association (which is also called semantic role). But 2 paragraphs below it is said that "The semantic role of an element is one way to give it a semantic association". Therefore, the semantic role is one way to give it the element a semantic role!!!

It seems to me that it would be much easier if we used just "semantic association", "explicit semantic role" and "implicit semantic role", and dropped the "semantic role" from our definitions. In the explicit and implicit definitions that's already the case, and they are much easier to read and understand than this one.

Jym77: * We need a definition of "semantic role". It is the result of a rather complex construction (conflict, explicit, implicit). We need a name for the result of this construction so we can reference it in rules.
* There are, indeed, two things named "role" (in ARIA). One is the concept (which I try here to rename as "semantic association", maybe not the best name). One is the construction that we make here. I tried to keep both inside the same definition as they are closely related, maybe that was a mistake and there should be two separate definitions instead.
* The logical dependencies are as follows:
  + A semantic association is something.
  + There exist different way of giving a semantic association to an element. Notably: explicit role, implicit role, and the complex construction called "semantic role".

  If the "semantic role" and "semantic association" are not somehow separated, then there is the (current) circularity of "a semantic role can be an explicit role which is a semantic role".


Would you say "semantic association" needs to be extracted into a separate definition? That is likely going to be a definition which is only used inside definitions… Not a huge problem, though…

pages/glossary/semantic-role.md
Jym77: Both "explicit role" and "implicit role" are other ways (used in "semantic role" but nonetheless different.

We could also imagine a system with, say, the HTTP payload containing a table of xpath/semantic association in JSON format. That would be another way.

pages/glossary/explicit-role.md
```diff
@@ -7,12 +7,17 @@ input_aspects:
...
 
 #### Accessibility Support
 
 Some browsers and assistive technologies treat the tokens of the `role` attribute as case-sensitive. Unless lowercase letters are used for the value of the `role` attribute, not all user agents will be able to interpret the tokens correctly. [ARIA in HTML (working draft)](https://www.w3.org/TR/html-aria/) also specifies that [authors must use lowercase letters for the `role` and `aria-*` attributes](https://www.w3.org/TR/html-aria/#case-sensitivity).
+
+[accessibility support base line]: https://www.w3.org/TR/WCAG-EM/#step1c 'Definition of accessibility support base line'
+[role attribute]: https://www.w3.org/TR/role-attribute/ 'Specification of the role attribute'
+[semantic association]: #semantic-role 'Definition of semantic association'
```
WilcoFiers, line 22: I'm wondering if this should instead link out to the role definition in ARIA. If "semantic association" is meant to represent something different from "semantic role", it can't be the same definition.

pages/glossary/semantic-role.md
```diff
@@ -7,15 +7,15 @@ input_aspects:
...
```
WilcoFiers, line 10: Adding semantic association made this more complicated then I think this needs to be. Why not just define explicit role based on the role attribute, implicit role based on AAM, and semantic role on explicit or implicit role?

Jym77: Yes, maybe the good way to do it is to kill this notion rather than try to make it obvious.

Jym77:
@carlosapaduarte, @WilcoFiers I've removed the notion of "semantic association" which was indeed not helping…
The definitions are a bit more dry (we don't tell why we need them) but more direct and to the point. I think this looks better now.
Thanks.

PR #1291, opened 2020-04-28
Remove marked decorative

### LINKED ISSUE #1231: Remove definition of "marked as decorative"?
The definition of "marked as decorative" is "role of `none` or `presentation`".

I feel that it is a bit misleading due to the presentational role conflict resolution. That is, `` should (by specs) have a role of `img` and thus is **not** marked as decorative. For me, the term "marked as decorative" is more about the developer’s intention (and thus the explicit role) and such an image should be considered marked as decorative but the conflict resolution comes on top of it.

---
I think we should remove the current definition (it is only really used in one rule anyway) and inline it where needed.

When/if we need it, I’d rather see a definition of "marked as decorative" as meaning "explicit role of `none` or `presentation`, or image with empty `alt`". Thus it would correspond to the intention of the developer (I am, intentionally, **marking** this image).

---
I’ll probably remove the current definition once #1189 is merged.

Jym77:
The definition of "marked as decorative" was:
- a bit misleading due to presentationnal conflict resolution (setting `role="none"` is not enough);
- not really matching its name. I feel that "marked as" is indicating intend while the definition was based on the result;
- only used in one place.

So, killing it and inlining the relevant parts.

Closes issue(s):

- closes #1231

Need for Final Call:
This will not require a Final Call (inlining a definition without changing outcomes).
---

### **When creating PR:**

### **After creating PR:**

## How to Review And Approve


pages/glossary/implicit-role.md
```diff
@@ -10,3 +10,7 @@ input_aspects:
 The _Implicit Semantic Role_ is the [semantic role](#semantic-role) of each element that is used when no valid [explicit semantic role](#explicit-role) is specified. Elements with no [role attribute](https://html.spec.whatwg.org/#attr-aria-role), or with a [role attribute](https://www.w3.org/TR/role-attribute/) containing no valid token, are mapped to their implicit role.
 
 Implicit roles for HTML and SVG, are documented in the [HTML accessibility API mappings (working draft)](https://www.w3.org/TR/html-aam/), the [ARIA in HTML (working draft)](https://www.w3.org/TR/html-aria/) documentation, and the [SVG accessibility API mappings (working draft)](https://www.w3.org/TR/svg-aam/).
+
+#### Accessibility Support
+
+- Images with an empty `alt` attribute should have an implicit role of `presentation`, according to the [HTML Accessibility API Mapping (work in progress)](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings). However, there are several popular browsers that do not treat images with empty `alt` attribute as having a role of `presentation` but instead add the `img` element to the accessibility tree with a role of either `img` or `graphic`.
```
daniel-montalvo, line 16:
```suggestion
- Images with an empty `alt` attribute should have an implicit role of `presentation`, according to the [HTML Accessibility API Mapping (work in progress)](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings). However, there are several popular browsers that do not treat images with empty `alt` attribute as having a role of `presentation`. Instead, they add the `img` element to the accessibility tree with a role of either `img` or `graphic`.
```

I think this makes these sentences a bit easier to process.

PR #1293, opened 2020-04-29
Replace section of content by implicit section of content

Jym77:
#1159 was too ambitious in trying to solve the ambiguity of "section of content" (https://github.com/act-rules/act-rules.github.io/pull/1085#discussion_r364145387) and was trying to mix in a new notion of visual section of content.

This new PR is more focused and only solves one problem with minimum change to the rule using the def.

If we want a rule to check headings of visual sections of content, we can define them and write a separate rule later.

Need for Final Call:
This will require a 1 week Final Call (changing a definition)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


pages/glossary/implicit-section-of-content.md
```diff
@@ -0,0 +1,34 @@
...
+key: implicit-section-of-content
+unambiguous: true
+objective: true
+oudatedLinksIgnore:
+  # ignore below links from `link-is-outdated` test
+  - 'https://www.w3.org/TR/html52/dom.html#heading-content'
+  - 'https://www.w3.org/TR/html52/common-idioms-without-dedicated-elements.html#common-idioms-without-dedicated-elements'
+---
+
+An _implicit section of content_ is all the nodes of the [flat tree][], in tree order, between an initial element with a [semantic role][] of [`heading`][heading] and the next element (in tree order) which has both a [semantic role][] of [`heading`][heading] and an [`aria-level`][aria-level] property lower or equal to that of the initial [`heading`][heading] element. The initial element is included, not the final one.
```
WilcoFiers, line 12: I feel like I've left this comment a few times before. In my opinion "the next heading" is not sufficient to describe where the next section starts. The semantics aren't leading in deciding what something "is", the design of a page does. Semantics should represent the visual information, not the other way around. In WCAG we don't fail designs because it doesn't match the semantics, you fail the semantics because it doesn't match the design.

If everyone else thinks i'm wrong about this, feel free to dismiss my comment and move on.

Jym77: I think that you are perfectly right in saying that "visual" sections do exist and should be taken into account. **BUT** I feel this would be topic for another PR, or maybe even another rule for the following reasons:

1. I think that having elements with a role of `heading` which do not describe their implicit section of content **is** a WCAG failure. Maybe not for sighted users, but for users of AT who can list headings and jump to the interesting one, thus need them to describe the content. Therefore, I think the rule as it is now is useful and detect an actual accessibility issue. Maybe not perfect, but useful.
2. The rule has always been targeting elements with a role of `heading`. Thus, it has always ignored visual sections and visual headings. It has also always targeted elements that are in the accessibility tree, but not visible (Passed example 6) and thus never was concerned by visual headings and sections. This doesn't mean we can't or shouldn't change it, but it means that considering visual section is more than just fixing the current definition.
3. The fact that this PR essentially doesn't change the rule that use the definition is a strong indication for me that it is an atomic PR making the minimum change and that changing the rule should be topic of another PR.

Now, and in relation to the first point, I do agree that having visual headings that don't describe their visual section of content is also a WCAG failure. And as you point out, we can't assume that the programmatic and visual structures are identical. Therefore, it would be nice to also handle visual sections. To me it is however a different topic than the current one. I do think it should be a completely separate rule because both kinds of section need a descriptive heading, and trying to mix them lead to a very complex rule. I kept the name "**implicit** section of content" to leave room (to us) for more kind of sections of content, and to indicate (to readers) that there could be other kinds of sections of content.

**TLDR:** Descriptive headings for visual sections of content are needed, but I think as a separate rule.

WilcoFiers: This has been going on for months, and I don't think there is anything I can say that I haven't already covered. Here's the basic argument. Based on this, lets have a vote in this week's CG call so we can resolve this.


1. There aren't different "types" of headings or sections as far as WCAG is concerned. There are headings, and there are sections and they can either have semantic markup, or they can not.
2. SC 1.3.1 requires that all headings be marked up as headings. It does not require that sections are marked up as sections. Therefore assumptions based on headings being marked up correctly are more acceptable than those that assume sections are.
3. This rule relies on the assumption that sections are marked up "implicitly", isn't required by WCAG, and isn't the only way sections can be marked up.

Therefore, in my opinion this is not an acceptable assumption for this rule.

Please :+1: if you agree, and :-1: if you disagree with this.

Jym77: > 1. There aren't different "types" of headings or sections as far as WCAG is concerned. There are headings, and there are sections and they can either have semantic markup, or they can not.

"Headings that have semantic markup" and "heading that don't have semantic markup" are, _de facto_, two "types" of headings. I agree that WCAG treat them the same way, but it doesn't mean we have to. This rule is, and has always been, a partial check for WCAG that is only considering "heading that have semantic markup" (and has always only considered those).

> 2. SC 1.3.1 requires that all headings be marked up as headings. It does not require that sections are marked up as sections. Therefore assumptions based on headings being marked up correctly are more acceptable than those that assume sections are.

I don't fully agree. I think that sections are part of "structure" that is required to match by 1.3.1. Furthermore, 2.4.10 does require sections to start with headings (in the generic sense of the term). Granted, it's only AAA.

> 3. This rule relies on the assumption that sections are marked up "implicitly", isn't required by WCAG, and isn't the only way sections can be marked up.

I disagree. I think that cases that fail this rule do fail 2.4.6. True, it is possible to pass this rule and nonetheless fail 2.4.6, but that is a very frequent case for our rules and does not need an assumption.

If I take your previous codepen: https://codepen.io/wilcofiers/pen/abOvWRB where the "sections defined by visual cues" and "section computed from `role="heading"` structure" don't match, I think that it fails 2.4.6. Because this is what Chrome+NVDA reads (newlines added by me):
> Hello world  heading    level 1  
Welcome to Lorem  heading    level 2  
Lorem ipsum, dolor sit amet consectetur adipisicing elit. Tenetur excepturi nobis illum ad officia hic, nesciunt distinctio quos iusto saepe dicta, incidunt modi vero et enim veritatis esse! Quae, fuga?  
Welcome to Ipsum  heading    level 2  
Lorem ipsum, dolor sit amet consectetur adipisicing elit. Tenetur excepturi nobis illum ad officia hic, nesciunt distinctio quos iusto saepe dicta, incidunt modi vero et enim veritatis esse! Quae, fuga?  Lorem ipsum dolor sit amet consectetur adipisicing elit. Molestiae porro voluptates repellat ducimus necessitatibus blanditiis quia ipsa? Quis libero hic voluptates aliquid, corporis quidem asperiores cumque, a officiis cum possimus.

Screen reader users have no clue that the "Welcome to Ipsum" section ends early. For them, the last bit is all part of the section that started at that heading.


Again, I am not saying that "visual sections" don't exists or that we shouldn't consider them. I am saying that considering them is another matter, topic for another PR or likely even another rule; and that failing this rule is indeed failing 2.4.6.

daniel-montalvo:
I don't really understand the objections here. There might be merely visual sections that are not marked up semantically, there might be merely semantic sections that are difficult to distinguish visually.
The way I take passed example 6 of this rule, where we have a heading that is positioned off-screen, is that we resort to this technique because we have no other design technique (in that particular example) that can actually serve to mark the beginning of a new section of content both visually and semantically.

I do think this is the right space in which we can expand a bit more than what WCAG gives us in terms of sections of content. I agree that work on visual sections of content needs also to be defined, but probably in another pull request/rule.

I understand your point @WilcoFiers that semantics respond to design, but I also think not all what can be conveyed by design can be conveyed by semantics. In other words, semantics conveys only a subset of meanings that design conveys. This is why in my opinion we do need to pay attention to these two realities separately.

Jym77:
```html
<h1>Foo</h1>
Lorem Ipsum
<div class="styled-as-h1">Bar</div>
Lorem Ipsum
```
Would failed rule.
It fails 1.3.1 but likely not 2.4.6

EmmaJP:
I had this open for some reason, and see it remains unresolved. 

WCAG is interesting sometimes. 1.3.1 definitely takes the point of view that the programmatic representation should convey what is represented visually in the presentation. Which certainly backs up Wilco's opinion.

What I find interesting, is that there is not a reverse of this. That the visual presentation should be representative of the programmatic semantics in the code. Which to my mind is how progressive enhancement is done: code semantically meaningful content, style it in a suitable way, enhance it with interactivity. Perhaps that is an implicit aspect of 3.2.

The success criterion for 2.4.6 says it does not require there to be headings or labels (2.4.10 is the closest to a requirement). Nor does 2.4.6 require that headings or labels are marked up correctly (that's 1.3.1). 2.4.6 only cares that if headings and labels do exist, that they are descriptive, though it is not very specific about what they should describe, not even an "associated content".

So, to test if something that could be perceived as a heading (whether coded or otherwise) is descriptive, there has to be an implied something that it is describing to compare it with. That is a huge bucketload of subjective stuff!

I think Jean-Yves is correct that descriptive headings may require multiple rules, but even then there are several crossover scenarios between what is visual and what is coded. And... WCAG only appear to be concerned about the programmatic version correctly representing the visual presentation.

Though you may not like it, I would suggest something more like the following:

> An _implicit section of content_ is a group of nodes from the [flat tree][], in tree order, that visually appear to be associated with each other. Usually between one [`heading`][heading] and the next [`heading`][heading], or within some form of visual demarcation such as a border or background colour.

Jym77:
> Though you may not like it, I would suggest something more like the following:
> 
> > An _implicit section of content_ is a group of nodes from the [flat tree][], in tree order, that visually appear to be associated with each other. Usually between one [`heading`][heading] and the next [`heading`][heading], or within some form of visual demarcation such as a border or background colour.

@EmmaJP, I'm indeed not a fan of the "visual only" approach for several reasons:

---
1. I've tried it here: https://github.com/act-rules/act-rules.github.io/pull/1159/files#diff-196bfebec867169f5b98ca21304c529d and wasn't fully convinced by the result. (still not really unambiguous and objective, feels like there are many cases that were missed, …)

---
2. I'm more and more convinced that a "section of content" is about the content, not the presentation (visual or programmatic). This is supported by:
* @kasperisager comment here: https://github.com/act-rules/act-rules.github.io/issues/1311#issuecomment-629973841 
For me at first, both examples in https://github.com/act-rules/act-rules.github.io/issues/1311#issuecomment-629294586 were **failing** 2.4.6. The first one because AT users will not have information about the change of section and thus the (programmatic) heading "Lime" does not describe the content they perceive following it ("This fruit is green. Lemon. This fruit is yellow"). The second because sighted users don't have the information about the change of sections.
I know agree with Kasper on these examples. There are 2 sections of content in each but they are not properly marked up (either programmatically or visually). The lack of correct markup (whichever kind) does not merge the sections of content.
* See also https://github.com/act-rules/act-rules.github.io/pull/421#issuecomment-637441288 which tells more about "sections of content are about content".

---
3. Specifically for this rule ("heading is descriptive").
It has always been targeting element that are **either** visible **or** included in the accessibility tree. It makes little sense to target elements that are **not visible** if the rule relies only on visual cues… This may have been overlooked when the rule was written and maybe the rule needs more fix than "just" updating the definition…

---
4. Specifically for this rule ("heading is descriptive").
It has always ben targeting "elements with a semantic role of `heading`". With a note about this being a partial check, so I guess it was a conscious design choice and not a random overlook.  Again, it would be very weird to base the rule on visual cues when it specifically selects target on something which is **not** visible.

---
5. Lastly (for now…) 2.4.6 actually does **not** refer to section of content… It's about describing "topic or purpose" but doesn't say "of what". The point here is that with a rule targeting "elements with a role of `heading`" as it is now, the question we need to answer is not "what is a section of content?" but "what does an element with a role of `heading` needs to describe?"
This is a more specific question and the way it's ask push toward having elements with a role of `heading` playing a role in this answer (which might be "the visual stuff this element is part of and sufficiently close to the start"…)

Of course, all that might mean that the fix we need is not to find "what does an element with a role of `heading` needs to describe?" but to change the rule completely…

EmmaJP:
Indeed @Jym77 ... perhaps the pertinent issue is defining what it is that a heading should be considered to describe. 

It isn't always the content that immediately follows. Image, heading/link, description is a fairly common promo order, much though I dislike it. Hero image, heading, text copy is also a fairly common order for articles. Yet there is a subjective understanding, from visual association, that the heading applies to the image before it as much as the text that follows.

It isn't always everything between one heading and the next heading. Sometimes there is unrelated content within that.

The understanding SC and example techniques docs are not much more helpful, even though they do occasionally refer to programmatic sections. However, the [technique](https://www.w3.org/WAI/WCAG21/Techniques/general/G130.html) did trigger a thought... 

If you were to look at all the headings of a webpage, would you have an idea of what content that webpage should include, and does the content match that expectation?

Perhaps that could form the basis for a different way of checking that headings are descriptive that does not require the specific parts of content described by each heading to be precisely determined.

ajanec01:
I may be wrong, but don't think the word _implicit_ will solve anything. The intent is to say that a section has some sort of programmatic relationship ("... all the nodes of the [flat tree]..."). "Implicit section of content" would still refer to a suggested section of content, which may be both programmatic and visual. 

Let's take into consideration the definition of [Implicit WAI-ARIA Sematics](https://www.w3.org/WAI/PF/aria/host_languages#implicit_semantics). It refers to an analogy between, let's say HTML5 semantics and WAI-ARIA semantics which is less ambiguous than a section of content ( a general term).

The point is, if I were reading it for the first time I would assume that "implicit section of content" is referring to something that can be presentational only.

In terms of headings, I don't see much benefit in splitting the [Heading is Descriptive](https://act-rules.github.io/rules/b49b2e) into programmatic and visual headings. I think that rule should have both.

Jym77:
Closing this PR for now. Let's keep discussion in #1311 and #1361 as much as possible.

PR #1294, opened 2020-04-29
button-acc-name: Tweak failed example 2 description

WilcoFiers:
Make it clear that `value` only provides an accessible name when the input element is in one of a few specific states.

Related issue: https://github.com/w3c/wcag-act/issues/451

Need for Final Call: none


Jym77:
@WilcoFiers I suppose you actually wanted to ask us for review rather than assigning us 😀

_rules/button-accessible-name-97a4e1.md
```diff
@@ -132,7 +132,7 @@ This `button` element has no [accessible name][] because it has no content or at
 
 #### Failed Example 2
 
-This `button` element has no [accessible name][]. The `value` attribute does not provide an [accessible name][] for `button` elements, only for `input` elements.
+This `button` element has no [accessible name][]. The `value` attribute does not provide an [accessible name][] for `button` elements, only for `input` elements in the `button`, `submit` or `reset` [state of the `type` attribute](https://html.spec.whatwg.org/multipage/input.html#states-of-the-type-attribute).
```
kengdoj, line 135: Suggest this wording order:

..., only when an `input` element's [state of the `type` attribute](https://html.spec.whatwg.org/multipage/input.html#states-of-the-type-attribute) is `button`, `submit` or `reset` .

Issue #1205, opened 2020-03-04
ARIA state or property has valid value' problems with applicability [6a7281]


dd8:
According to the Accessibility Requirements Mapping 'ARIA state or property has valid value' is checking ARIA validity, but inapplicable example 3 is invalid per the ARIA spec:

https://act-rules.github.io/rules/6a7281#inapplicable-example-3

https://www.w3.org/TR/wai-aria-1.1/#aria-checked

The W3 validator flags this as invalid. 

This affects the applicability which excludes invalid ARIA attributes like `aria-checked` and `aria-checked=''` (but not `aria-checked='  '`)

There is a custom element in the examples, but the HTML spec makes a distinction between HTML elements and custom elements, and the applicability only includes HTML and SVG elements:
https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements

Custom element applicability be fixed by changing 
> and that is specified on an HTML or SVG element

to
> and that is specified on an HTML, SVG or custom element

Jym77:
1. Yes, I agree that no value or empty value is as invalid as any other bad value. We should probably update the applicability to any `aria-*` attribute and check that it has a value and that value is correct. Or add another rule to check that `aria-*` attributes have non-empty value.

2. I don't thing we really define "HTML element" anywhere, so it is not clear whether custom elements (used in HTML) are considered HTML elements or not… Nonetheless, this is the only custom element we use (AFAICT) and the example should maybe just be removed…

kasperisager:
Your point about "HTML element" is a good one. My interpretation is "element in the HTML namespace" which would include custom elements that extend from `HTMLElement`, either because they're autonomous or because they extend built-in elements: https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element. This issue is however evidence of that interpretation not being the only valid one so there's an ambiguity that needs to be resolved.

WilcoFiers:
Talked about this in the CG call. While not ideal, this is allowed in ARIA, we're going to add a note to the example explaining this is invalid HTML and that this will result in the default value.

WilcoFiers:
The ARIA props part of this has been merged. The other part of this issue is a duplicate of https://github.com/act-rules/act-rules.github.io/issues/459, so I'm closing this one.

PR #1295, opened 2020-04-29
aria-props: add a note on HTML validation with aria-current

WilcoFiers:
Add a note explaining that inapplicable example 3 fails the HTML validator, but does not appear to be a conformance issue for WAI-ARIA. It doesn't appear to be invalid HTML either, so this may actually be an issue with the HTML validator.

Related to issue #1205 

Final call: none


_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -6,7 +6,7 @@ description: |
   This rule checks that each ARIA state or property has a valid value.
 accessibility_requirements:
   aria11:state_prop_values:
-    title: ARIA 1.1, 6.3 Values for States and Properties
+    title: ARIA 1.1, 6.3 Valaues for States and Properties
```
jkodu, line 9:
```suggestion
    title: ARIA 1.1, 6.3 Values for States and Properties
```

PR #1296, opened 2020-05-04
Orientation of the page is not restricted using CSS transform property (b33eff): Resolve TF feedback

kasperisager:
This pull request adds a note with an explanation of the expectation in order to address concerns from the TF that the expectation was a little difficult to understand: w3c/wcag-act#445. As I really don't see how we can reasonably shorten the expectation, a note seemed more appropriate. An existing test case has also been adjusted slightly and a new test case added.

I'm guessing this requires a 2 week final call.


_rules/css-restrict-orientation-b33eff.md
```diff
@@ -37,6 +37,8 @@ The rule applies to any HTML element that is [visible](#visible) and has a CSS [
 
 The target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative from the position of the element in `landscape` orientation to the position of the element in `portrait` orientation, and vice versa.
 
+**Note:** Imagine the display of a smartphone with an upwards pointing arrow its center. If we turn the smartphone a quarter turn, that is move from one orientation to the other, we would want the arrow to continue pointing upwards. This is done by rotating it a quarter turn to counter our change in orientation. In effect, the arrow has remained in place and its rotation relative from one orientation to the other is 0 degrees. Now imagine that we rotate the arrow by a quarter turn _only_ when in one orientation and not the other; its rotation relative from one orientation to the other would then be 90 degrees and it would appear stuck, or locked, as we move between orientations. What we've done is effectively counter the smartphones attempt at countering our change in orientation.
```
Jym77, line 40: I still find the note a bit confusing. I think it is because there are a lot of personas (devs, user, smartphone) and it is not clear who the pronoun refer to.
```
**Note:** Imagine the display of a smartphone with an upwards pointing arrow its center. 
If we [=user]
turn the smartphone a quarter turn, that is move from one orientation to the other, 
we [=user ultimately, but also sort of =smartphone deciding this since you can have orientation lock on smartphone and then arrow won't turn]
would want the arrow to continue pointing upwards. 
This is done [=by the smartphone, which is why second "we" in previous sentence feels like smartphone]
by rotating it a quarter turn to counter our change in orientation. 
In effect, the arrow has remained in place and its rotation relative from one orientation to the other is 0 degrees. 
Now imagine that we [=dev?]
rotate the arrow by a quarter turn _only_ when in one orientation and not the other; its rotation relative from one orientation to the other would then be 90 degrees and it would appear stuck, or locked, 
as we [=user] move between orientations. 
What we've [=dev]
done is effectively counter the smartphones [=smartphone]
attempt at countering our [=user's]
change in orientation.
```

Not sure how to express it better…

kasperisager: I'll make sure everything is in active voice and be more clear about the "whos".

kasperisager:
This pull request will be in final call until June 16.

_rules/css-restrict-orientation-b33eff.md
```diff
@@ -37,6 +37,8 @@ The rule applies to any HTML element that is [visible](#visible) and has a CSS [
 
 The target element is neither rotated clockwise nor counter clockwise around the Z-axis at an angle corresponding to 90 degrees relative from the position of the element in `landscape` orientation to the position of the element in `portrait` orientation, and vice versa.
 
+**Note:** Imagine the display of a smartphone with an upwards pointing arrow its center. If a user turns the smartphone a quarter turn, that is move from one orientation to the other, they would want the arrow to continue pointing upwards. The smartphone accomplishes this by rotating the contents of its display a quarter turn to counter the users change in orientation. In effect, the arrow has remained in place and its rotation relative from one orientation to the other is 0 degrees. Now imagine that a developer rotates the arrow by a quarter turn _only_ when in one orientation and not the other; its rotation relative from one orientation to the other would then be 90 degrees and it would appear stuck, or locked, as the user moves between orientations. What the developer has done is effectively counter the smartphones attempt at countering the users change in orientation.
```
EmmaJP, line 40:
```suggestion
**Note:** Imagine the display of a smartphone with an upwards pointing arrow at its center. If a user turns the smartphone a quarter turn, that is a move from one orientation to the other, they would want the arrow to continue pointing upwards. The smartphone accomplishes this by rotating the contents of its display a quarter turn to counter the users change in orientation. In effect, the arrow has remained in place and its rotation relative from one orientation to the other is 0 degrees. Now imagine that a developer rotates the arrow by a quarter turn _only_ when in one orientation and not the other; its rotation relative from one orientation to the other would then be 90 degrees and it would appear stuck, or locked, as the user moves between orientations. What the developer has done is effectively counter the smartphones attempt at countering the users change in orientation.
```

Issue #446, opened 2019-03-19
"Image has accessible name" (23a2a8) has questionable test cases


kasperisager:
The rule currently lists the following case as inapplicable:

```html
<img alt="W3C logo" aria-hidden="true">
```

However, according to the rule, I'm fairly certain that the case above should be a pass. Firstly, the element is applicable as it's an HTML `<img>` element:

> The rule applies to (HTML `<img>` elements) or (any HTML element with the semantic role of `img` that is included in the accessibility tree).

Parentheses are my own to clarify the separation between the statements. The word "that" refers to "any HTML element with the semantic role of `img`" and the rule therefore applies to HTML `<img>` elements regardless of their inclusion in the accessibility tree, which is what we want. Otherwise, `<img>` elements that have a role of `none` or `presentation` would not be able to pass the rule as such elements would not be included in the accessibility per the Core AAM:

> The following elements are not exposed via the accessibility API and user agents MUST NOT include them in the accessibility tree:
>
> - Elements with `none` or `presentation` as the first mappable role in the role attribute. However, their exclusion is conditional and depends on other factors. [...]

As such, the test case I posted above actually _passes_ the rule as the `<img>` element is applicable and has an accessible name. This does not seem to be entirely intentional though. Thoughts?

kasperisager:
I'd very much like to hear your take on this, @ShadowBB.

Brynanders:
So if we updated the [applicability](https://github.com/auto-wcag/auto-wcag/pull/263/files) to separate out each point like so:

> The rule applies to any HTML `input` element with a [`type`](https://www.w3.org/TR/html/sec-forms.html#dom-htmlinputelement-type) of `image`, or any HTML element with the [semantic role](#semantic-role) of `img`
> Applicable elements must be [included in the accessibility tree](#included-in-the-accessibility-tree), and has an [accessible name](#accessible-name) that is equivalent to the [filename](#filename) specified in the `src` attribute. Difference in letter casing, and forward and trailing whitespace should be ignored.

Brynanders:
I was basing the test case on the information provided as part of the definition for included in the accessibility tree that states

> An example of this is a focusable element with an aria-hidden attribute with a value of true.

I interpreted this as aria-hidden="true" by default removes an element from the accessibility tree and is only included if it has been explicitly given focus.

annethyme:
@kasperisager and I just had a chat about this one and has come up with a suggestion for the applicability:

```
This rule applies to:
[...]

PR #1302, opened 2020-05-06
image-accname: fix minor ambiguity in applicability

WilcoFiers:
Closes https://github.com/act-rules/act-rules.github.io/issues/446

Need for Final Call: none

### DIFF
--- a/_rules/image-accessible-name-23a2a8.md
+++ b/_rules/image-accessible-name-23a2a8.md
@@ -36,7 +36,7 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements and HTML element with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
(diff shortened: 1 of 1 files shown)


_rules/image-accessible-name-23a2a8.md
```diff
@@ -36,7 +36,7 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements or any HTML element with the [semantic role][] of `img` that is [included in the accessibility tree][].
+The rule applies to HTML `img` elements and HTML element with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
```
carlosapaduarte, line 39:
```suggestion
The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
```

PR #1309, opened 2020-05-13
Fix presentational conflict resolution

### LINKED ISSUE #1308: Presentational Roles Conflict Resolution test cases are inconsistent
The form control has accessible name rule has 3 examples that are inconsistent with other examples in the same rule.

1) This is a fail:
`<input role="none" />`
https://act-rules.github.io/rules/e086e5#failed-example-8

2) This is inapplicable but is a synonym for failed example 8:
`<input role="presentation" />`
https://act-rules.github.io/rules/e086e5#inapplicable-example-3 

They should both produce identical results because `role=none` is a synonym for `role=presentation`:
https://www.w3.org/TR/wai-aria-1.1/#none

3) This example should be a fail if role conflict resolution is used because the control is focusable and interactive so role=none is ignored:
```
<select role="none">
	<option value="volvo">Volvo</option>
	<option value="saab">Saab</option>
	<option value="opel">Opel</option>
</select>
```
https://act-rules.github.io/rules/e086e5#inapplicable-example-4

Jym77:
Fix a few examples where presentational role conflict resolution triggered and gave results not matching the description.

Closes issue(s):

- closes #1308

Need for Final Call:
This will require a 1 week Final Call (updating a few examples)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/form-control-accessible-name-e086e5.md
```diff
@@ -210,15 +210,15 @@ Hidden to assistive technologies.
...
+<input role="banner" />
 ```
 
 #### Inapplicable Example 4
 
 Option inherits from input, but has a required context role of listbox which inherits from select. We should therefore not consider option as applicable.
 
 ```html
-<select role="none">
+<select>
```
kasperisager, line 221: This will cause the example to fail rather than be inapplicable.

Jym77: Hmm, yes, it now fails on the `select` 🙄

Jym77: Hmm, yes, it now fails on the `select` 🙄

_rules/form-control-accessible-name-e086e5.md
```diff
@@ -210,15 +210,15 @@ Hidden to assistive technologies.
...
+<input role="presentation" disabled />
 ```
 
 #### Inapplicable Example 4
 
 Option inherits from input, but has a required context role of listbox which inherits from select. We should therefore not consider option as applicable.
 
 ```html
-<select role="none">
+<select role="none" disabled>
```
kasperisager, line 221: This now falls victim to presentational role inheritance so the description isn't quite accurate; it is also inapplicable by virtue of the `<option>` elements being presentational (https://w3c.github.io/aria/#presentation). We might be okay with that though 🤷

Jym77: I think this is OK…

---
First, this is what the example was trying to do previously. If we want to fix the inheritance, this would be for another PR.

---
Second, I am not sure that the inheritance actually happens in this case. Chrome does hide the `select` while giving a role of `menuitem` to the `option`s.

I guess the relevant parts about presentational inheritance are
> When an explicit or inherited role of presentation is applied to an element with the implicit semantic of a WAI-ARIA role that has required owned elements, in addition to the element with the explicit role of presentation, the user agent MUST apply an inherited role of presentation to any owned elements that do not have an explicit role defined.

This is OK because `combobox` has no required owned elements.

> Also, when an explicit or inherited role of presentation is applied to a host language element which has required children as defined by the host language specification, in addition to the element with the explicit role of presentation, the user agent MUST apply an inherited role of presentation to any required children that do not have an explicit role defined.

This is indeed not fully OK because `select` content model is "Zero or more option, optgroup, and script-supporting elements." I am not sure why Chrome doesn't inherit the role, however… Maybe it interprets "**zero** or more" as them not really being needed (since zero is OK) which is a bit weird…

---
Anyway, since giving the `select` a role of `none` was the intend of this example, I say that fixing the presentational inheritance might be topic for another PR, especially since it doesn't change the outcome of the test case (and only makes the description inaccurate).

dd8:
There's some discussion going on the meaning of "interactive" in the ARIA "presentation roles conflict resolution"

https://github.com/w3c/aria/issues/1192
https://github.com/w3c/aria/issues/1270

The outcome of this affects how the examples in the PR need changed.

Jym77:
> There's some discussion going on the meaning of "interactive" in the ARIA "presentation roles conflict resolution"
> 
> [w3c/aria#1192](https://github.com/w3c/aria/issues/1192)
> [w3c/aria#1270](https://github.com/w3c/aria/issues/1270)
> 
> The outcome of this affects how the examples in the PR need changed.

I think this is OK. We've sort of lampshaded the issue by looking at what is in the accessibility tree (so we let UA decide what "focusable" means, essentially).

There is a bit of an issue with [This example](https://act-rules.github.io/rules/73f2c2#passed-example-10) because it seems some browsers (at least Chrome) "fix" the tabindex of -1.5 to -1, and thus do remove the element from sequential navigation.
This is an unrelated issue (given that it's already here in the current rule). Easy fix is to use `tabindex=0` instead.

dd8:
Looks like `disabled` with no other attributes does do role conflict resolution, but adding other attributes blocks conflict resolution (e.g. `disabled` with `aria-disabled`) 

https://github.com/w3c/aria/issues/1270#issuecomment-628870333

carlosapaduarte:
@Jym77 Should this have been closed? There are still commits for the form control has accessible name to be merged.
If you reopen I'll add the third approval.

Jym77:
@carlosapaduarte Very good catch!

Jym77:
Final Call will end on June 9.

Jym77:
> Using role="banner" on an input element seems a very contrived way to shoehorn in this example. But as it could be possible, I guess it works.

@EmmaJP The role itself is not very important. The important thing (for this example) is to set up to one which is not a [widget role](https://www.w3.org/TR/wai-aria-1.1/#widget_roles). I'm up for anything else if you have a good idea 💡

Jym77:
Keeping this open to give @EmmaJP a chance to answer https://github.com/act-rules/act-rules.github.io/pull/1309#issuecomment-640426465

Will merge after CG call (so, on June 12).

Jym77:

PR #1324, opened 2020-05-28
Move file to _rules folder

carlosapaduarte:
File for rule "Object element has non-empty accessible name" was in the wrong place. Moving it to the _rules folder.

Need for Final Call:
This will not require a Final Call (change to website: just moving a file with any editing)


jkodu:
Yikes. I am surprised we missed this in the pull request reviews :/

jkodu:
@carlosapaduarte there are failing tests - https://github.com/act-rules/act-rules.github.io/pull/1324/checks?check_run_id=717080152

carlosapaduarte:
@jkodu me too! Can you have a look at the failing test? I can't get what is going on with that

carlosapaduarte:
@jkodu don't know why I couldn't see the full test rusults yesterdat (that's why I couldn't make sense of the report). Now I can and the typo has been fixed. Could you please review again?

PR #1325, opened 2020-05-28
Update expectations in "Error message describes invalid form field value"

### LINKED ISSUE #1298: Error message describes invalid form field value (36b590): Undefined behaviour for form fields without error indicators
Each of the three expectations start with "For each test target with a form field error indicator, [...]" which made me realise during implementation of the rule that it's not currently defined what to do about the test targets that _don't_ have a form field error indicator. I therefore had to assume that they should simply pass.

In hindsight, it also seems a little odd to have the rule be applicable to form fields without error indicators, but that's currently an unfortunate consequence of not having an objective definition of what an error indicator is.

Summoning @carlosapaduarte, @joao-vicente, and @WilcoFiers.

carlosapaduarte:
Updated expectations to take into account form fields that do not have error indicators

Closes issue(s):
- closes #1298 

Need for Final Call:

## How to Review And Approve


WilcoFiers:
@carlosapaduarte can you give this a descriptive title though?

carlosapaduarte:
> @carlosapaduarte can you give this a descriptive title though?

@WilcoFiers I will but I just need to make sure what does the "this" in your sentence refers to :-) Is it this PR?

carlosapaduarte:
Final call ends June 19

_rules/invalid-form-field-value-36b590.md
```diff
@@ -63,7 +63,7 @@ in [text][] that is [visible][].
 
 ## Expectation 3
 
-For each test target with a [form field error indicator][], at least one of the [form field error indicators][form field error indicator] describes:
+Each test target either has no [form field error indicators][form field error indicator], or at least one of the [form field error indicators][form field error indicator] describes:
```
EmmaJP, line 66: Expectation 2 and 3 are identical ... either Expectation 3 is redundant or should describe something else.

carlosapaduarte:
Final call ended. Merging.

PR #1328, opened 2020-06-02
clarify mixup of "subtag" and "tag" in language rules

WilcoFiers:
ACT TF requested I update the title. When I looked into this more closely I realised "subtag" was used in places where "tag" should have been. This should address the issue.

Need for Final Call: none


_rules/html-page-lang-valid-bf051a.md
```diff
@@ -62,15 +62,15 @@ _There are no major accessibility support issues known for this rule._
...
+This `html` element has a `lang` attribute whose value is a [valid][valid language tag] [primary language subtag][].
 
 ```html
 <html lang="fr"></html>
 ```
 
 #### Passed Example 2
 
-This `html` element has a `lang` attribute value that is a [valid language subtag][] even though the [region subtag][] is not.
+This `html` element has a `lang` attribute value that is a [valid language tag][] even though the [region subtag][] is not.
```
Jym77, line 73: It is very weird to say that the attribute has a "valid language tag" with an invalid region subtag. It is mixing the definition of "valid" (ours and BCP 47).
"[value] is a valid language tag" => with our definition of valid.
"the region subtag is not [valid]" => with the BCP definition.


This is likely due to the fact that our definition of "valid language tag" is actually meaning "language tag starting with a valid primary subtag".
I agree that "valid language subtag" was not a good name (it is only checking the primary subtag, not any). But I feel that "valid language tag" is not really better.

pages/glossary/valid-language-tag.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Valid Language Subtag
-key: valid-language-subtag
+title: Valid Language Tag
```
Jym77, line 2:
```suggestion
title: Language Tag with valid primary subtag
```
Not convinced that it is good 🤔 But "valid language **tag**" is really confusing.

_rules/html-page-lang-valid-bf051a.md
WilcoFiers: I think it's better; but I agree that the word "valid" isn't ideal. I thought about that when I worked on this, but couldn't come up with anything to improve on it. Either way I think that's separate from what this PR is addressing, can you open an issue for it?

pages/glossary/valid-language-tag.md
WilcoFiers: I don't like that either. That almost inlines the entire phrase. It isn't a "valid primary subtag" either. We're not checking that it's "valid", we're checking that it exists in the registry.

_rules/html-page-lang-valid-bf051a.md
Jym77: I'm not entirely sure why we'll need 2 PRs to fix a single broken name ("valid language tag"), but sure.

PR #1329, opened 2020-06-03
New rule: Visible focus

Jym77:
New rule for 2.4.7: Visible Focus.

The rule uses the "distinguishing styles" definition from https://github.com/act-rules/act-rules.github.io/blob/inline_link_in_paragraph_is_distinguishable/pages/glossary/distinguishing-styles.md and will need to augment it a bit.

The rule currently has an outstanding issue of not handling correctly the default focus indicator of input fields (blinking cursor) and I am not sure how to solve it (ideas welcome!)

Closes issue(s):
N/A

Need for Final Call:
This will require a 2 weeks Final Call (new rule).

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/focusable-element-has-visible-focus-f4e323.md
```diff
@@ -0,0 +1,352 @@
...
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+The rule applies to any [focusable][] element.
+
+## Expectation
+
+For each target element, there exists a set of [focus indicators][focus indicator] which are not all also [focus indicators][potential focus indicator] for another [focusable][] element (the same for all the indicators).
```
WilcoFiers, line 27: I think you're going to have a very hard time to come up with a complete list of all the different ways something can be focusable. Perhaps focus indicator can be defined as all pixels that change when the element is focused from when the element is not focused.

You could add a second expectation that says each focus indicator must be unique, to replicate this "for another focusable element" thing, but I'd much prefer to have that as an assumption. I don't think I've ever seen that happen, so that seems like a hypothetical we shouldn't spend our time on.

_rules/focusable-element-has-visible-focus-f4e323.md
```diff
@@ -0,0 +1,352 @@
...
+  - CSS styling
+  - DOM tree
+acknowledgments:
+  authors:
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+The rule applies to any [focusable][] element.
```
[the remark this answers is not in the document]
Jym77: > The correct wording here should probably have been _"where the keyboard focus has a visible indicator"_.

I do not understand what you mean here. If the keyboard focus indicator is visible, then the SC is satisfied and there is nothing to test 🤔 


That said, I do agree that restricting the rule to sequential focus navigation makes a lot of sense 👍

_rules/focusable-element-has-visible-focus-f4e323.md
Jym77: I disagree with the need for unicity. Check Passed Example 6 where each element has 2 focus indicator (one on each side) and these focus indicators are shared between two elements. The only unicity condition that applies to this example is that the set of stuff that change is unique.

pages/glossary/distinguishing-styles.md
```diff
@@ -0,0 +1,13 @@
...
+title: Distinguishing styles
+key: distinguishing-styles
+unambiguous: true
+objective: true
+input_apsects:
+  - CSS styling
+  - DOM tree
+---
+
+Placeholder for https://github.com/act-rules/act-rules.github.io/blob/inline_link_in_paragraph_is_distinguishable/pages/glossary/distinguishing-styles.md
```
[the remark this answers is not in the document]
Jym77: Outline is included through https://github.com/act-rules/act-rules.github.io/blob/inline_link_in_paragraph_is_distinguishable/pages/glossary/distinguishing-borders.md 😁 

I admit I haven't followed all discussions in that other rule, but I kinda assume that whatever "visible focus" that one ends up with can be reused here (as in "if it's good there, it should be good here").
Granted, 2.4.7 is very lax in how focus should be shown ("visible", essentially), but 1.4.1 (for which the other rule and def are worked) is also super lax ("color is not the only way").

_rules/focusable-element-has-visible-focus-f4e323.md
```diff
@@ -0,0 +1,390 @@
...
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+The rule applies to any [focusable][] element in a document with several [focusable][] elements when the element is part of [sequential focus navigation][].
+
+## Expectation
+
+For each target element, there exists a set of [focus indicators][focus indicator] which are not all also [focus indicators][focus indicator] for another [focusable][] element (the same for all the indicators).
```
kasperisager, line 27:
```suggestion
For each target element, there exists a set of [focus indicators][focus indicator] which are not all also [focus indicators][focus indicator] for another [focusable][] element.
```

It seemed a little detached 🤔

Jym77: The precision is needed essentially for cases like Passed Example 7 where there are several focusable elements in a row and focus is indicated by lighting a dot on both side of each focusable. Then, each focus indicator is a focus indicator for 2 different elements. So for elements in the middle, all of their focus indicators are also focus indicator for another element, however this is not the same other element for all focus indicator, so this is good.

_rules/sequentially-focusable-element-has-visible-focus-through-styling-f4e323.md
```diff
@@ -0,0 +1,411 @@
+---
+id: f4e323
+name: Element in sequential focus order has visible focus though styling
```
Jym77, line 3: Note to reviewers:
I initially wrote this rule that is too strict for WCAG (2.4.7 is notably weak in its requirements). After discussing with @WilcoFiers, we decided to go for the "minimal version" (which is the other rule).

I was reluctant about throwing away all the work here 😉 So I decided to try and keep it as a "best practice" rule (it doesn't really check for 2.4.11 either). Let me know what you think about it and whether we should keep it or not…
(if we keep it, we'll probably want to move it to another PR for separation of concerns, of course, I'll do that if initial feedback on it is good 😀)

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+<a href="https://act-rules.github.io/">ACT rules</a> <button tabindex="-1">Dummy button</button>
+```
+
+#### Passed Example 3
+
+The first [focusable][] element, is part of [sequential focus navigation][]. The set of pixels that change color when it is [focused][] (the border of the `span` element) is unique (it does not change when any other element is [focused][]).
+
+```html
+<link rel="stylesheet" href="../test-assets/focus-visible/styles.css" />
+<script src="../test-assets/focus-visible/script.js" />
```
carlosapaduarte, line 100:
```suggestion
<script src="../test-assets/focus-visible/script.js"></script>
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+
+The first [focusable][] element is part of [sequential focus navigation]. The set of pixels that change color when it is [focused][] (due to default User Agent's styling) is unique. The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the `tabindex` attribute. Its presence is nonetheless enough to make the first one applicable.
+
+```html
+<a href="https://act-rules.github.io/">ACT rules</a> <button tabindex="-1">Dummy button</button>
+```
+
+#### Passed Example 3
+
+The first [focusable][] element, is part of [sequential focus navigation][]. The set of pixels that change color when it is [focused][] (the border of the `span` element) is unique (it does not change when any other element is [focused][]).
```
carlosapaduarte, line 96:
```suggestion
The first [focusable][] element, is part of [sequential focus navigation][]. The set of pixels that change color when it is [focused][] (the border of the `span` element) is unique (it does not change when any other element is [focused][]). The second element part of [sequential focus navigation][] also as a unique set of color changing pixels when [focused][].
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+<button>Dummy button</button>
+```
+
+#### Passed Example 4
+
+For each of these three [focusable][] elements, the set of pixels changing color when it is [focused][] is unique. For the element with `id` "wcag", each of these pixels also change color when another element is [focused][], but they do not all change color for the same other element. Thus, the set of pixels changing color is also unique for this one.
+
+```html
+<link rel="stylesheet" href="../test-assets/focus-visible/styles.css" />
+<script src="../test-assets/focus-visible/script.js" />
```
carlosapaduarte, line 121:
```suggestion
<script src="../test-assets/focus-visible/script.js"></script>
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+
+The rule applies to any element which is part of [sequential focus navigation][] in a document with several [focusable][] elements.
+
+## Expectation 1
+
+For each target element, there is at least one pixel whose color is different when the element is [focused][] and when it is not.
+
+## Expectation 2
+
+For each target element, and each other [focusable][] element in the document, the set of pixels whose color is different when the test target is [focused][] and when it is not, and the set of pixels whose color is different when the other element is [focused][] and when it is not, are different.
```
carlosapaduarte, line 31:
```suggestion
For each target element and each other [focusable][] element in the document, the set of pixels whose color is different when the test target is [focused][] and when it is not, and the set of pixels whose color is different when the other element is [focused][] and when it is not, are different.
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+### Failed
+
+#### Failed Example 1
+
+None of these [focusable][] elements have any pixel changing color when they are [focused][] because the default styling has been overwritten.
+
+```html
+<a class="no-focus-default" href="https://act-rules.github.io/">ACT rules</a>
+<a class="no-focus-default" href="https://www.w3.org/TR/WCAG21/">WCAG</a>
+```
```
carlosapaduarte, line 162:
```suggestion
```html
<link rel="stylesheet" href="../test-assets/focus-visible/styles.css" />

<a class="no-focus-default" href="https://act-rules.github.io/">ACT rules</a>
<a class="no-focus-default" href="https://www.w3.org/TR/WCAG21/">WCAG</a>
```
```

File: _rules/sequentially-focusable-element-has-visible-focus-oj04fd.md [hunk 11]
```diff
@@ -0,0 +1,232 @@
...
+	>WCAG</a
+>
+<span id="indicator-final" class="indicator solid"></span>
+```
+
+### Failed
+
+#### Failed Example 1
+
+None of these [focusable][] elements have any pixel changing color when they are [focused][] because the default styling has been overwritten.
```
carlosapaduarte, line 157:
```suggestion
None of these [focusable][] elements have any pixel changing color when they are [focused][] because the default styling has been overwritten by a style that removes the outline.
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+```
+
+#### Failed Example 2
+
+The first [focusable][] element is part of [sequential focus navigation][] and has no pixel changing color when it is [focused][]. The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the `tabindex` attribute. Its presence is nonetheless enough to make the first one applicable.
+
+```html
+<a class="no-focus-default" href="https://act-rules.github.io/">ACT rules</a>
+<button tabindex="-1">Dummy button</button>
+```
```
carlosapaduarte, line 171:
```suggestion
```html
<link rel="stylesheet" href="../test-assets/focus-visible/styles.css" />

<a class="no-focus-default" href="https://act-rules.github.io/">ACT rules</a>
<button tabindex="-1">Dummy button</button>
```
```

File: _rules/sequentially-focusable-element-has-visible-focus-through-styling-f4e323.md (hunk 5 above)
Comment (carlosapaduarte) [reply _rules/sequentially-focusable-element-has-visible-focus-through-styling-f4e323.md:3]: I've just browsed this rule, not really reviewed it (because, as you say, we should move it to a different PR if we decide to keep it).

I feel your pain, because of all the effort you must have put into this, but I'm not really sure it will add much value to the other visible focus rule.

File: _rules/sequentially-focusable-element-has-visible-focus-oj04fd.md [hunk 13]
```diff
@@ -0,0 +1,232 @@
...
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+The rule applies to any element which is part of [sequential focus navigation][] in a document with several [focusable][] elements.
+
+## Expectation 1
+
+For each target element, there is at least one pixel whose color is different when the element is [focused][] and when it is not.
```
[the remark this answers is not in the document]
Jym77: I tried to be explain this.
It was surprisingly difficult to say "somewhere on the page", and I am still not sure what I picked is correct (notably for infinite scrolling pages like maps)…

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
+---
+id: oj04fd
+name: Element in sequential focus order has visible focus
+rule_type: atomic
+description: |
+  This rule checks that each element in sequential focus order has some visible focus indication.
+accessibility_requirements:
+  wcag20:2.4.7: # Focus Visible (A)
```
Jym77, line 8: 🤔 
Actually, 2.4.7 is only level AA in WCAG 2.1 but will move to level A in WCAG 2.2.

Not sure how to handle that…

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+
+None of these [focusable][] elements have any pixel changing color when they are [focused][] because the default styling has been overwritten by a style that removes the outline.
+
+````html
+<link rel="stylesheet" href="../test-assets/focus-visible/styles.css" />
+
+<a class="no-focus-default" href="https://act-rules.github.io/">ACT rules</a>
+<a class="no-focus-default" href="https://www.w3.org/TR/WCAG21/">WCAG</a>
+
+#### Failed Example 2 The first [focusable][] element is part of [sequential focus navigation][] and has no pixel
```
WilcoFiers, line 165: Looks like you broke prettier

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+
+The rule applies to any element which is part of [sequential focus navigation][] in a document with at least two [focusable][] elements.
+
+## Expectation 1
+
+For each target element, there is at least one pixel inside the [scrolling area][] of the [viewport][] whose color is different when the element is [focused][] and when it is not.
+
+## Expectation 2
+
+For each target element and each other [focusable][] element in the document, the set of pixels whose color is different when the test target is [focused][] and when it is not, and the set of pixels whose color is different when the other element is [focused][] and when it is not, are different.
```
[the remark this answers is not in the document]
Jym77: I feel that the understanding document is strongly pushing toward unicity with sentences like:
> It must be possible for a person to know which element among multiple elements has the keyboard focus.

if multiple elements have the same focus indicator, it is not possible to know which among them has focus.

> - This Success Criterion helps anyone who relies on the keyboard to operate the page, by letting them visually determine the component on which keyboard operations will interact at any point in time.

uses singular "**the** component" none "some components, one of which has focus".

> - People with attention limitations, short term memory limitations, or limitations in executive processes benefit by being able to discover where the focus is located.

With short term memory, if you can't remember where on the page you are, you'll need focus indicator to be unique in order to be somewhat helpful.

---
I'd rather keep this expectation and add an assumption that unicity is required by WCAG.

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+WCAG does not have any requirement of how big or small focus indicator should be, or how far or near from the [focusable][] element it should be. Thus it is possible to pass this rule and [Success Criterion 2.4.7 Focus Visible][sc247] with barely perceptible changes at the other end of the page. That would however still be an accessibility issue. WCAG 2.2 is working on an extended Success Criterion 2.4.11 specifying how big the focus indicator should be. See the [Understanding Success Criterion 2.4.11: Focus Visible (Enhanced)][usc2411] proposal. All Examples in this rule are voluntarily avoiding very bad practices and thus avoid focus indications that are too small, too far away, too weak, …
+
+WCAG [Understanding Success Criterion 2.4.7: Focus Visible][usc247] explicitly states that "if there is only one keyboard actionable control on the screen, the success criterion would be met". Therefore, this rule only consider documents with several [focusable][] elements.
```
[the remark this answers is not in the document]
Jym77: While not technically normative, the Understanding document still says:
> The Working Group considers the content to be authoritative.

But sure, we can make an assumption that the Understanding document is correct.

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,225 @@
...
+  wcag22:2.4.11: # Focus Appearance (Minimum) (AA)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+  wcag22:2.4.12: # Focus Appearance (Enhanced) (AAA)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
```
Jym77, line 22: Note that this is currently not handled by the build. We'll need to grab WCAG 2.2 data, which is currently not done…

WilcoFiers: Yeah, I'd leave this off until we can do the full migration to WCAG 2.2.

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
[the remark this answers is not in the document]
WilcoFiers: > It must be possible for a person to know which element among multiple elements has the keyboard focus.

That's all well and good, but if it's not in the SC text, then it's not a normative requirement. I'm also not convinced this is something that actually happens. I've definitely never seen an issue like this. The only scenario I can come up with for this is the one above, where I'd believe this creates a false positive.

If you'd like to discuss this on an upcoming call, feel free to put the agenda item label on it, but I don't see a reason to change my vote on this.

WilcoFiers (dismissed):

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -71,23 +55,23 @@ WCAG does not have any requirement of how big or small focus indicator should be
 
 #### Passed Example 1
 
-The first [focusable][] element is part of [sequential focus navigation][]. The set of pixels that change color when it is [focused][] (due to default User Agent's styling) is unique. The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the negative value for the `tabindex` attribute. Its presence is nonetheless enough to make the first one applicable.
+The first [focusable][] element is part of [sequential focus navigation][]. There is at least one pixel that changes color when it is [focused][] (due to default User Agent's styling). The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the negative value for the `tabindex` attribute. Its presence is nonetheless enough to make the first one applicable.
```
WilcoFiers, line 58: I don't think this "at least one pixel" phrase needs to be in these descriptions. We don't want to repeat the expectation. I think the following would be sufficient:

```suggestion
The first [focusable][] element is part of [sequential focus navigation][]. It gets an outline focused because of the User Agent's default styling. The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the negative value for the `tabindex` attribute. Its presence is nonetheless enough to make the first one applicable.
```

The same comment applies to other exampled.

WilcoFiers (changes requested):

WilcoFiers (approved):

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+  - CSS styling
+  - DOM tree
+acknowledgments:
+  authors:
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+The rule applies to any element which is part of [sequential focus navigation][] in a document with at least two [focusable][] elements.
```
carlosapaduarte, line 23: To make sure this applies to HTML documents

```suggestion
The rule applies to any element which is part of [sequential focus navigation][] in an [HTML document](https://html.spec.whatwg.org/#document) with at least two [focusable][] elements.
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Expectation
+
+For each target element, there is at least one device pixel inside the [scrolling area][] of the [viewport][] whose color is different when the element is [focused][] from when it is not.
+
+## Assumptions
+
+This rule assumes that documents with only one focusable element automatically satisfies [Success Criterion 2.4.7 Focus Visible][sc247], as stated in [Understanding Success Criterion 2.4.7: Focus Visible][usc247]: "if there is only one keyboard actionable control on the screen, the success criterion would be met".
```
carlosapaduarte, line 31:
```suggestion
This rule assumes that documents with only one focusable element automatically satisfy [Success Criterion 2.4.7 Focus Visible][sc247], as stated in [Understanding Success Criterion 2.4.7: Focus Visible][usc247]: "if there is only one keyboard actionable control on the screen, the success criterion would be met".
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+Default styling in most user agents is providing perfectly valid focus indication for focusable elements (even those that are not focusable by default), as shown in Passed Examples 1 and 2. Many examples in this rule need to **remove** that indicator in order to illustrate various situations. This is bad practice and should normally be avoided.
+
+WCAG does not have any requirement of how big or small focus indicator should be, or how far or near from the [focusable][] element it should be. Thus it is possible to pass this rule and [Success Criterion 2.4.7 Focus Visible][sc247] with barely perceptible changes at the other end of the page. That would however still be an accessibility issue. WCAG 2.2 includes [Success Criterion 2.4.11 Focus Appearance (Minimum)][sc2411] and [Success Criterion 2.4.12 Focus Appearance (Enhanced)][sc2412] specifying how big the focus indicator should be. All Examples in this rule avoid focus indications that are too small, too far away, too weak, …
```
carlosapaduarte, line 41: Not sure what you're meaning by a "too weak" focus indicator... is it perhaps contrast related?

```suggestion
WCAG does not have any requirement of how big or small focus indicator should be, or how far or near from the [focusable][] element it should be. Thus it is possible to pass this rule and [Success Criterion 2.4.7 Focus Visible][sc247] with barely perceptible changes at the other end of the page. That would however still be an accessibility issue. WCAG 2.2 includes [Success Criterion 2.4.11 Focus Appearance (Minimum)][sc2411] and [Success Criterion 2.4.12 Focus Appearance (Enhanced)][sc2412] specifying how big the focus indicator should be. All Examples in this rule avoid focus indications that are too small, too far away, have low contrast, …
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+- [Understanding Success Criterion 2.4.11: Focus Appearance (Minimum)][usc2411]
+- [Understanding Success Criterion 2.4.12: Focus Appearance (Enhanced)][usc2412]
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The first [focusable][] element is part of [sequential focus navigation][]. It has an outline when it is [focused][] (due to default User Agent's styling). The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the negative value for the `tabindex` attribute. Its presence is nonetheless enough to make the first one applicable.
```
carlosapaduarte, line 58:
```suggestion
The first [focusable][] element is part of [sequential focus navigation][]. It has an outline when it is [focused][] (due to default User Agent's styling). The second [focusable][] element is not applicable because it has been removed from [sequential focus navigation][] due to the negative value for the `tabindex` attribute. Its presence is nonetheless enough to make the rule applicable to the first one.
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+
+This document contains only one [focusable][] element.
+
+```html
+<a href="https://act-rules.github.io/">ACT rules</a>
+```
+
+#### Inapplicable Example 3
+
+None of the [focusable][] elements in this document is part of [sequential focus navigation][].
```
carlosapaduarte, line 164:
```suggestion
None of the [focusable][] elements in this document are part of [sequential focus navigation][].
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+	>WCAG</a
+>
+<span id="indicator-final" class="indicator solid"></span>
+```
+
+### Failed
+
+#### Failed Example 1
+
+The first [focusable][] element does not have any pixel changing color when it is [focused][] because the default styling has been overwritten by a style that removes the outline.
```
carlosapaduarte, line 135: I would place the button in the sequential focus order (i.e. remove the negative tabindex) to make it clear that even if one of the elements pass the rule, it fails because all need to pass it

carlosapaduarte (changes requested):
Some minor requests and typos to fix

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
Jym77: I'm afraid that mixing Passed and Failed targets in the same test case will be messing up the implementation reports.
Currently, we expect one "failed" result for this test case. If we make the other one passing, we'd expect one "fail" and one "passed" result, and thus we also need to figure out which outcome is for which target in order to decide if the implementation is correct, and that makes automated implementations mapping harder, I guess…

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+
+The first [focusable][] element is part of [sequential focus navigation][]. It has an outline when it is [focused][] (due to default User Agent's styling).
+
+```html
+<span tabindex="0">Act rules</span> <button tabindex="-1">Dummy button</button>
+```
+
+#### Passed Example 3
+
+The first [focusable][] element, is part of [sequential focus navigation][]. Its parent `span` puts a border around it when it is [focused][] (the border of the `span` element).
```
carlosapaduarte, line 74:
```suggestion
The first [focusable][] element is part of [sequential focus navigation][]. Its parent `span` puts a border around it when it is [focused][] (the border of the `span` element).
```

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
carlosapaduarte: That's fair

carlosapaduarte (changes requested):
Still found one typo

_rules/object-has-acessible-name-8fc3b6.md
```diff
@@ -174,7 +174,6 @@ There is no `object` element.
 <audio title="Moon speech" src="/test-assets/contrast/example.png"></audio>
 ```
 
-
```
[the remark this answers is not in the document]
Jym77: Because #1488 wasn't merged in yet…

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -0,0 +1,181 @@
...
+
+The rule applies to any element which is part of [sequential focus navigation][] in an [HTML document](https://html.spec.whatwg.org/#document) with at least two [focusable][] elements.
+
+## Expectation
+
+For each target element, there is at least one device pixel inside the [scrolling area][] of the [viewport][] whose color is different when the element is [focused][] from when it is not.
+
+## Assumptions
+
+This rule assumes that documents with only one focusable element automatically satisfy [Success Criterion 2.4.7 Focus Visible][sc247], as stated in [Understanding Success Criterion 2.4.7: Focus Visible][usc247]: "if there is only one keyboard actionable control on the screen, the success criterion would be met".
```
kasperisager, line 31: The keyword here seems to be "keyboard actionable". It therefore seems like there's a disconnect between this assumption and the applicability of the rule as the latter requires at least two "focusable" elements in the document, not at least two "keyboard actionable" elements.

kasperisager (dismissed):

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
carlosapaduarte: And given that @Jym77 has already created the definition for [keyboard actionable](https://github.com/act-rules/act-rules.github.io/blob/bf57d85c3f0514721dc70868506b3519a2e65bda/pages/glossary/keyboard-actionable-element.md) that should be an easy fix 🙂

carlosapaduarte (commented):
Will approve after the changes suggested by @kasperisager are put in place

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
Jym77: Interesting discussion…
I am not convinced that our "keyboard actionable" make sense here. If an element can be tabbed but not clicked, it still require focus indication. Typically, [HTML specs](https://html.spec.whatwg.org/multipage/interaction.html#data-model:sequentially-focusable) have an example where a UA could make bullets (the bullet symbol) in a list could be sequentially focusable as a reading guide, in which case they definitely need an indicator…

OTOH, WCAG (the SC) is only about "a keyboard operable UI component"… Not sure whether these bullets would be considered as a UI component by WCAG… But "operable" could be understand as "tabbable", if the only purpose is to be tabbable (reading guide, …)

Suggestion: put the applicability as "if there are at least 2 sequentially focusable elements" (rather than "at least 2 focusable)"; and add an assumption "this rule assumes that sequentially focusable elements are UI components", which sounds safe.
+ update all examples since the dummy button with `tabindex=-1` does not work any more 😕 

This would also avoid the need for "keyboard actionable" which is annoying to test automatically. So the Applicability could be automated, and the rule also in many cases (e.g., detecting an indicator through styling, …)

@carlosapaduarte @kasperisager WDYT?

kasperisager: Something along the lines of "if the sequential focus navigation order contains more than 1 element" works fine by me 👍

carlosapaduarte: Agreed

carlosapaduarte (approved):

kasperisager (approved):

Jym77:
Final Call ends on November 11.

Jym77:

PR #1332, opened 2020-06-04
New rule: Element marked as decorative is not exposed

Jym77:
New rule, checking that elements marked as decorative have an empty accessible name.
This catch mismatches like `` where the presentationnal role and the accessible name convey contradictory intention (should the image be ignored or not?)

Closes issue(s):

- N/A

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/element-marked-decorative-has-empty-name-46ca7f.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Expectation
+
+Each target element as an empty [accessible name][].
+
+## Assumptions
+
+- This rule assumes that the applicable elements embed [non-text content][]. If this is not the case, the rule may fails while [Success Criterion 1.1.1][sc111] is still satisfied because it is only concerned with [non-text content][]. Note that the mismatch between [marking as decorative][marked as decorative] and providing an [accessible name][] is not necessarily an accessibility issue for text content, but is nonetheless bad practice and should be avoided.
+- This rule assumes that elements which are [marked as decorative][] are [pure decoration][]. If this is not the case, this rule may fail while [Success Criterion 1.1.1][sc111] is still satisfied. Note that [marking as decorative][marked as decorative] an element which is not [pure decoration][] is not necessarily an accessibility issue but is nonetheless bad practice and should be avoided.
```
WilcoFiers, line 41: I'm uncomfortable with this assumption. The fact that you can add `aria-label` to an element to override `role=none` is useful for things like accessibility overlays. You could also pretty easily get into a situation like this with components. Here's an example:

```jsx
const MyImg = props => (props.alt
  ? <img src={props.src} alt={props.alt} />
  : <img src={props.src} role="none" />
)
const HelloWorldImg = () => <MyImg aria-label="hello world" src="..." />
```

It would be a completely different story if some AT wound announce the label, and others would not. That would be a rule purely based on limited accessibility support for presentation conflict resolution. We have other rules like that. That seems fine. It seems like presentation conflict resolution is fairly well supported by AT, without that, I don't think we can conclude this actually is an issue.

Jym77: I do not understand what you mean.

Do you mean that the assumption is too broad and that many many pages will fail it, thus making the rule mostly useless?

---
1.1.1 states (last bullet):
> If non-text content is pure decoration, is used only for visual formatting, or is not presented to users, then it is implemented in a way that it can be ignored by assistive technology.

So it is cristal clear that if something is pure decoration but cannot be ignored, it is an issue.

WilcoFiers: Yes, while the rule itself is sound, I think this assumption is not going to hold up in practice .

kasperisager: > ```js-jsx
> const MyImg = props => (props.alt
>   ? 
>   : 
> )
> const HelloWorldImg = () => <MyImg aria-label="hello world" src="..." />
> ```

This component would never receive the `aria-label` prop though as it isn't passed along. 

I'm personally quite comfortable with this assumption, but that's based on my personal opinion on the matter and not data.

WilcoFiers: Sorry, yer right, bug. Let me fix that. You may want to check with other implements if they have rules like this. Chair hat off; I don't think we'll be implementing such a rule in axe-core, although we'll likely create one that outright fails anything in a conflict resolution as a best practice. It is a terrible thing to have in yer code base. I'm confused why ARIA even allows it. I'm going to file an issue about that.

Jym77: > I don't think we'll be implementing such a rule in axe-core, although we'll likely create one that outright fails anything in a conflict resolution as a best practice.

In its current state, failing things triggering conflict resolution should count as an implementation of the rule, as far as I understand (that doesn't fail any of the Passed/Inapplicable test cases, and that doesn't pass any of the Failed test cases).

Failing every conflict resolution will also make you fail, say, `<span role="none" aria-label="foo">foo</span>` which is inapplicable for the rule because it is not non-text content (so 1.1.1 doesn't care).
We can try and make sure to avoid writing examples like that in order to let tools failing any conflict resolution be counted as implementation for the rule.

I do agree with you that 
>  [conflict resolution] is a terrible thing to have in yer code base. I'm confused why ARIA even allows it. 

so I fully support wider implementations of this rule. The restriction to these specific cases is to make the rule map to 1.1.1 and I am not sure we have a similar SC to map conflict resolution of text content (1.3.1 seems a bit far fetched).

The first assumption explicitly states that conflict resolution on text-content is bad practice but not a WCAG issue, the second one states that conflict resolution is bad practice if done on purpose on non-decorative elements. Both are needed to ensure the mapping to 1.1.1 which is likely needed to make this rule publishable. But both are also "if this assumption is false and the rule fails, the page might pass 1.1.1 but is still crap and need to be fixed"…

That's the main reason why I'm happy with these assumptions: when false, the rule won't flag stuff that are perfectly OK. It will flag stuff that are bad practice but ignored by WCAG. So whether these assumptions hold or not, when the rule fail, the code needs to be fixed.

Jym77:
@WilcoFiers @kasperisager @adilsofficial @carlosapaduarte @ShadowBB 
According to discussion on WAI tools call from Monday June 15, I've updated the rule to be a best practice targeting any elements (not just the ones usually embedding non-text content).

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
```diff
@@ -0,0 +1,150 @@
+---
+id: 46ca7f
+name: Element marked as decorative is not exposed
+rule_type: atomic
+description: |
+  This rule checks that elements marked as decorative are not exposed in the accessibility tree, or with a presentational role.
+accessibility_requirements: # Remove whatever is not applicable
```
WilcoFiers, line 7:
```suggestion
accessibility_requirements: # None, this is a best practice
```

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
```diff
@@ -0,0 +1,150 @@
...
+
+Whenever such a conflict occurs, this indicates at the very least mismatching intentions. Such a conflict should be avoided.
+
+When these conflicts arise on [decorative][] [non-text content][], this is also a failure of [Success Criterion 1.1.1: Non-text Content][sc111] because [decorative][] [non-text content][] must be implemented in a way that allows assistive technologies to ignore it. When this conflicts arise on text content, or on content which is not [decorative][], this is not a failure of WCAG. Therefore this rule is not mapping to any specific WCAG Success Criterion, and is not an accessibility requirement for WCAG.
+
+- [Success Criterion 1.1.1: Non-text Content][sc111]
+- [Technique F38: Failure of Success Criterion 1.1.1 due to not marking up decorative images in HTML in a way that allows assistive technology to ignore them][f38]
+- [Technique F39: Failure of Success Criterion 1.1.1 due to providing a text alternative that is not null (e.g., alt="spacer" or alt="image") for images that should be ignored by assistive technology][f39]
```
WilcoFiers, line 45: I don't think this should be here. I disagree with the assessment that it is a WCAG failure for images, as we discussed before. It might be, depending on the purpose of the image. There are other rules which would catch such issues. I think we are better off staying clear of any reference to WCAG in this rule, to avoid confusing this with actual WCAG issues.

```suggestion
```

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
```diff
@@ -0,0 +1,150 @@
+---
+id: 46ca7f
+name: Element marked as decorative is not exposed
+rule_type: atomic
+description: |
+  This rule checks that elements marked as decorative are not exposed in the accessibility tree, or with a presentational role.
```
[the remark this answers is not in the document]
Jym77: I think we currently avoid links in descriptions. Not sure it's a strong requirement we put on ourselves, but I'd rather keep it that way.

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
Jym77: > I disagree with the assessment that it is a WCAG failure for images, as we discussed before. It might be, depending on the purpose of the image.

Yes. "**if** (some condition), this is a failure". We agreed that it was too big a if to be in an assumption. I feel it is still worth being mentioned. It's in the background, and it does explain a bit why the rule is here.

> There are other rules which would catch such issues.

I do not think we have anything that catches "decorative image not being implemented in a way that allows AT to ignore it".
Which other rule do you think about that would flag ``?

> I think we are better off staying clear of any reference to WCAG in this rule, to avoid confusing this with actual WCAG issues.

I think this does explain why the rule is here. We're not chasing after corner cases in specs to make a rule. This is something that is not **always** a problem (so, no accessibility mapping), but is nonetheless **sometimes** a problem (and we have no data on how often it is a problem). I feel keeping this explain why the rule is here.

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
```diff
@@ -0,0 +1,150 @@
...
+```html
+<img src="test-assets/shared/w3c-logo.png" role="none" />
+```
+
+#### Passed Example 4
+
+This `img` element is [marked as decorative][] through its empty `alt` attribute and has [semantic role][] of `none`.
+
+```html
+<img src="test-assets/shared/w3c-logo.png" alt="" />
```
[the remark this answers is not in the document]
Jym77: I disagree.
The role of `none` comes from the empty `alt` attribute and this is what I wanted to illustrate.

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
```diff
@@ -0,0 +1,144 @@
...
+
+Whenever such a conflict occurs, this indicates at the very least mismatching intentions. Such a conflict should be avoided.
+
+When these conflicts arise on [decorative][] [non-text content][], this is also a failure of [Success Criterion 1.1.1: Non-text Content][sc111] because [decorative][] [non-text content][] must be implemented in a way that allows assistive technologies to ignore it. When this conflicts arise on text content, or on content which is not [decorative][], this is not a failure of WCAG. Therefore this rule is not mapping to any specific WCAG Success Criterion, and is not an accessibility requirement for WCAG.
+
+## Test Cases
```
carlosapaduarte, line 43: This is not something that will prevent me from accepting, but currently, the examples almost hint that this rule is applicable to images that are markd as decorative (all examples use `img` or `svg` elements). Given that the applicability is all elements that marked as decorative, we could have at least an example with text content.

Jym77: Good point. I'll add some.
The rule was initially only for images…

Jym77:
Final Call ends on July 14.

Jym77:

_rules/element-marked-decorative-is-not-exposed-46ca7f.md
```diff
@@ -0,0 +1,150 @@
+---
+id: 46ca7f
+name: Element marked as decorative is not exposed
+rule_type: atomic
+description: |
+  This rule checks that elements marked as decorative either are not included in the accessibility tree, or have a presentational role.
+accessibility_requirements:
+input_aspects:
+  - Accessiblity tree
```
adilsofficial, line 9: Typo
```suggestion
  - Accessibility tree
```

adilsofficial (commented):
typo

PR #1334, opened 2020-06-05
new rule: HTML page language subtag matches default language

WilcoFiers:
This rule checks that the primary language subtag of the page language matches the default language of the page.

Need for Final Call: 2 weeks


_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+This page has a [default language][default page language] of English because most words are English. The [default page language][] matches the language indicated in the `lang` attribute.
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>Gelukkig</title>
+	</head>
+	<body>
+		<p>The Dutch word "gelukkig" has no equivalent in English.</p>
```
Jym77, line 99: I am a bit worried about examples like that (due to the definition of default language).

`The word ale is from "øl"`
Both "the" (alternate spelling of the correct "te"), "ale", "is" and "from" are also Danish words (according to ordnet.dk). So this sentence has actually 5 English words and 5 Danish words, thus no default language. Although it has absolutely no meaning as a sentence in Danish.

Now, this is OK-ish, there is no default language so we don't do anything.

What I'm worried of is that we can tweak such an example enough to have a sentence with, say, 8 words that are English and 10 that are Danish, but just because some English words happen to have a completely different meaning in Danish. The sentence would be undoubtfully English, but the default language would be Danish.

Then, I guess this would be a case where the first assumption is wrong, so that is correct…

Maybe strengthen the first assumption by giving examples of other way to detect language? (letters/words frequency analysis, mutual information based distance, n-grams, …)

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+	</body>
+</html>
+```
+
+#### Failed Example 3
```
Jym77, line 144: Should there be a Passed Example similar to this?

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+		common format that allows any party involved in accessibility testing to document and share their testing procedures
+		in a robust and understandable manner. This enables transparency and harmonization of testing methods, including
+		methods implemented by accessibility test tools.
+	</p>
+</html>
+```
+
+#### Inapplicable Example 4
+
+In page has an undefined [default language][default page language] can not be defined because it can either be English or French.
```
Jym77, line 204:
```suggestion
In page has an undefined [default language][default page language] because it can either be English or French.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+		</blockquote>
+		<p lang="en">
+			This Dutch phrase literally translates into "He went to roost with the chickens", but it means that he went to bed
+			early.
+		</p>
+	</body>
+</html>
+```
+
+### Inapplicable
```
Jym77, line 166: Missing inapplicable examples for:
* invalid lang attribute
* not a top-level browsing context
* not a text/html content type.

Jym77, line 166: Also, can you try and put inapplicable examples in the same order as the items in the applicability?
(not an html element, no lang, invalid lang, not a top-level browsing context, not text/html, no default language)

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This page has a [default language][default page language] of English, which matches the language indicated in the `lang` attribute.
+
+```html
+<!DOCTYPE html>
```
Jym77, line 71: I think the DOCTYPE is not needed when its html, as it will be added automatically (https://github.com/act-rules/act-rules-web/blob/3cb73b7cb5e3a39b6c342fcb7a89ab6551dcbc32/build/create-testcases.js).

pages/glossary/default-page-language.md
```diff
@@ -0,0 +1,36 @@
...
+  - Language
+---
+
+The default page language is the _most common language_ used in a [web page][]. If there are no _words_ in the page, the default language is undefined.
+
+The _most common language_ is determined by counting the number of _words_ in the [web page][] that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. If the number of words is the same, the default language is undefined.
+
+The _words_ used in the following texts should be counted for the _most common language_:
+
+- **page title**: The value of the [document title][] of the [document][] in the [top-level browsing context][].
```
Jym77, line 19: I do not understand why browsing context is needed here. "document title" doesn't seem to need it.

pages/glossary/default-page-language.md
```diff
@@ -0,0 +1,36 @@
...
+- **default language content**: The value of any [text nodes][] and the [accessible name][] of any [elements][] for which all of the following is true:
```
Jym77, line 20: Should that also include accessible description? Or do accessible descriptions always come through text nodes (hum, probably not with the inclusion in the accessibility tree)?

pages/glossary/default-page-language.md
```diff
@@ -0,0 +1,36 @@
...
+  - The [node][] is [included in the accessibility tree][]; and
```
Jym77, line 21: Why including accessible but invisible nodes and not visible but not accessible nodes?

pages/glossary/default-page-language.md
Jym77, line 20: I think there should be examples where the accessible name plays a role in determining the language.
```html
<span id="explanation" aria-hidden=true>this word has no English equivalent</span>
<span aria-labelledby="explanation">gelukkig</span>
```
or something similar.

pages/glossary/default-page-language.md
```diff
@@ -0,0 +1,36 @@
...
+input_aspects:
+  - DOM tree
+  - Accessibility tree
+  - CSS Styling
+  - Language
+---
+
+The default page language is the _most common language_ used in a [web page][]. If there are no _words_ in the page, the default language is undefined.
+
+The _most common language_ is determined by counting the number of _words_ in the [web page][] that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. If the number of words is the same, the default language is undefined.
```
Jym77, line 15: I'm afraid this requires dictionaries (as in, "list of words") in addition to the languages.

Different dictionaries, even for the same language, do not list the exact same words, notably for new words.
For example, as of now, the French dictionary Larousse does not accept "infox" while Le Petit Robert does accept it.
https://www.larousse.fr/dictionnaires/francais/infox
https://dictionnaire.lerobert.com/definition/infox

(and Le Petit Robert only accept it in its 2020 edition, not before).
(and both of these are reference dictionaries in French French, so for the `fr_FR` language)

So, "words that are part of a language" is actually ambiguous => according to which dictionary or dictionaries?




This gets even hairier with regional differences. "armour" is a word in `en_UK` but not in `en_US`. It should be counted as a word in `en` (because the definition only cares for the language, not for the region). But then, both "armour" and "armor" should count for `en`, which requires a list of words for each of the possible subregions.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
+---
+id: ucwvc8
```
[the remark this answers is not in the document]
WilcoFiers: There's no reason for that. I'm not sure how that practice snuck in. That was not the intent. I've been using https://www.random.org/strings/ to create my IDs for a while now.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to any [document element][] if it is an `html` element for which one of the following is true:
+
+- The [document element][] has a `lang` attribute with a value that is a [valid language tag][]; and
+- The [document element][] is in a [top-level browsing context][]; and
+- The [document element][] has a [node document][] with a [content type][] of `text/html`; and
```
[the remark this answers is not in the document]
WilcoFiers: I copied this from the other lang rules. We should change it there too.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,227 @@
...
+- The [document element][] has a defined [default page language][].
+
+## Expectation
+
+For each test target, the [primary language][] of the [valid language tag][] matches the [default page language][] of the test target.
+
+## Assumptions
+
+- The way to determine the default language of a page should be done by counting the number of visible words used in each language. If the default language needs to be derived in some other way, this rule may produce incorrect results.
```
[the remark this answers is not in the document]
WilcoFiers: I think this is the right way to do it. It's an assumption about how to define "default language". I could do something like "The way to determine the default language of a page, as described in WCAG 2 ..." Perhaps that addresses it.

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: Wow, odd, why doesn't it check if there is a doctype before adding one in? :/ I've filed an issue. I'll take them out.

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: I thought of that too. This seems fairly hypothetical though. Real world web pages don't just have a single phrase on them. I'm counting on pages having enough content on them to make things like the same word existing in multiple language not matter.

There are a number of other problems with this approach. The biggest one I have with it is that it's difficult to do accurately. If a page is half English half Danish, you can obviously word count the thing, but grabbing all the "Danish" words from the English section, is difficult to do. I explored some other ideas for this, like looking at sentences or paragraphs and counting those, but there you get into issues of how to define those. Also how do you compare two short paragraphs or two short sentences to a single long one. There's also the issue that some languages concatenate words. The same paragraph in German will have fewer words in it than the English one, just because in English words aren't concatenated.


The reason I went with this approach is because it is going to be easy to do for 99.9% of all pages. You can pretty much just eyeball it in almost every possible case, and in the cases they can't, a simple word count tool will resolve the problem. That extreme case where it is too close to call, yes this isn't ideal, but solving for that will probably take months, and is likely going to add complexity to testing this for the 99.9%. I don't think that's worth it.

pages/glossary/default-page-language.md
WilcoFiers: Because those nodes will not be processed by assistive technologies. The language of content in aria-hidden doesn't much matter, since it's not going to get read out.

pages/glossary/default-page-language.md
WilcoFiers: Good point, yeah I think so.

pages/glossary/default-page-language.md
WilcoFiers: Because the document title of iframed content is never exposed to AT (at least not as far as I'm aware).

pages/glossary/default-page-language.md
WilcoFiers: I think the language input addresses this. Count up all the words you understand to be part of a particular language, that's the number of words. If the dictionary you may use for that is different from the one someone else uses, that's a change that is different because the input is different. I don't think that makes this definition ambiguous.

pages/glossary/default-page-language.md
WilcoFiers: Not too keen on an aria-labelledby one, but an image makes sense. I'll work on something.

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: The other lang rules don't have this either, for a number of reasons:
- invalid lang would not satisfy the SC.
- not a top-level, I suppose we could add a page with no content, and only an iframe, but that'd give a page without a title. Not sure that's a good thing to put in here.
- not text/html; We can't guarantee these code snippets will be served with the correct MIME type. I don't think it's possible for us to do this at this time.

Jym77: Agreed on the first two points.

> * not text/html; We can't guarantee these code snippets will be served with the correct MIME type. I don't think it's possible for us to do this at this time.

For this one, when we use ` ```xhtml ` instead of ` ```html `, we're adding a different DOCTYPE. I should be enough to serve the example with the correct type, I would say…
We do that in https://act-rules.github.io/rules/5b7ae0#inapplicable-example-4

pages/glossary/default-page-language.md
Jym77: Good point. I am convinced.

pages/glossary/default-page-language.md
Jym77: Good point.
Should that be an assumption? (for the rule)
"The rule assumes that `iframe` titles are not exposed to AT and thus do not consider them"

pages/glossary/default-page-language.md
Jym77: Understanding 3.1.1 also list benefits for the visual parts:
> Visual browsers can display characters and scripts correctly. Media players can show captions correctly. 

And so does 3.1.2:
> This is especially important when switching between languages that read from left to right and languages that read from right to left, or when text is rendered in a language that uses a different alphabet. Users with disabilities who know all the languages used in the Web page will be better able to understand the content when each passage is rendered appropriately.

This seems to indicate that a correctly set programmatic language is important both for screen readers and visual display, hence that "visible only" words should apply.


I can imagine a situation where the visual part of the page has been translated, but the accessible hasn't. This would result in the language this definition finds being different whether we count "visible only" and "AT only" words and thus may the rule fail where it shouldn't.

pages/glossary/default-page-language.md
```diff
@@ -0,0 +1,38 @@
...
+
+The _words_ used in the following texts should be counted for the _most common language_:
+
+- **page title**: The value of the [document title][] of the [document][] in the [top-level browsing context][].
+- **text nodes**: The value of any [text nodes][] that is [visible][] or [included in the accessibility tree][]
+- **accessible text**: The [accessible name][] and [accessible description][] of any element included in the [accessibility tree][]
+
+**Exception**: Words in text that comes from nodes that have an [ancestor][] in the [flat tree][] with a non-empty (`""`) `lang` attribute, except for the [root node][].
```
Jym77, line 23:
```suggestion
**Exception**: Do not count words in text that comes from nodes that have an [ancestor][] in the [flat tree][] which has a non-empty (`""`) `lang` attribute and is not the [root node][].
```
The original sentence is OK-ish but I feel it has two problems: 1/ not a sentence (no verb, …) so adding the "do not count" part; 2/ "except for the root node" was separated from what it qualifies ("ancestor in the flat tree"), making it look like it apply to the full nominal group ("nodes that have an ancestor with a non-empty lang attribute").

Not super happy with what I'm suggesting either, feels very "heavy" sentence for something relatively simple…

Jym77, line 23: Do you mean "non-empty lang" or "valid lang"?
i.e., do you mean to exclude `<p lang="xyz">Lorem ipsum</p>`?
I am not sure how invalid lang are handled by UA/AT. If they default to the parent lang, then these should not be ignored from the count.

Jym77, line 23: How is language handled in a case like that:
```html
<div lang="fr"><span id="label">Hello World</span></div>
<img src="foo.jpg" aria-labelledby="label" />
```

With the current definition, "Hello World" is ignored because it has an ancestor with a lang attribute. But how is it handled by AT? Is it going to be read with the ancestor attribute or default to the page language?

Similar question for
```html
<span id="label" lang="fr">Hello World</span>
<img src="foo.jpg" aria-labelledby="label" />
```

WilcoFiers: AT are inconsistent about what they do, some default to the parent, others default to the browser's default language. That's why I think it's best to keep them out. Clearly someone meant for that to be a different language.

WilcoFiers: The text is included because it is part of the accessible name of the img.

Jym77: Then, I do not understand "text that comes from nodes…" the way you seem to (i.e. that need some clarification).

In the examples I've put above, I understand that "Hello World" is text that comes from the `span` node (or more precisely from the text node which is child of the `span`). These nodes do have an ancestor with a non-empty `lang`.

From your answer, I get that you meant that "Hello World" is text that comes from the `img` node, which doesn't have an ancestor with a non-empty `lang`.


Note that the text in the `span` has actually 2 reasons to be counted: it is displayed (and announced), and it is used in an accessible name. One of these reasons is cancelled due to the `lang` attribute, but not the other. That's also a bit counter-intuitive.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,278 @@
...
+
+This page has an undefined [default language][default page language] because it has no content or [document title][].
+
+```html
+<html></html>
+```
+
+#### Inapplicable Example 3
+
+This page has an undefined [default language][default page language] because it has no [document title][] all its content is wrapped in element with a `lang` attribute.
```
carlosapaduarte, line 240:
```suggestion
This page has an undefined [default language][default page language] because it has no [document title][] and all its content is wrapped in an element with a `lang` attribute.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,278 @@
...
+	<body>
+		<img src="/test-assets/shared/fireworks.jpg" alt="Fireworks over Paris" />
+		<p lang="nl">
+			Gelukkig nieuwjaar!
+		</p>
+	</body>
+</html>
+```
+
+### Failed
```
adilsofficial, line 143: I recommend having a failed example with "grandfatheredtag".

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,278 @@
...
+This rule applies to any [document element][] if it is an `html` element for which all of the following are true:
+
+- The [document element][] has a `lang` attribute with a value that is a [valid language tag][]; and
+- The [document element][] is in a [top-level browsing context][]; and
+- The [document element][] has a [content type][] of `text/html`; and
+- The [document element][] has a defined [default page language][].
+
+## Expectation
+
+For each test target, the [primary language][] of the [valid language tag][] matches the [default page language][] of the test target.
```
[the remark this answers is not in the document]
WilcoFiers: This is intentional. I think this reads better. Why do you think this needs to change?

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Assumptions
+
+- This rule assumes that the default human language of a page, as described in WCAG 2, can be determined by counting the number of words used in each language. If the default language needs to be derived in some other way (such as frequency analysis, mutual information based distance, …), this rule may fail while [Success Criterion 3.1.1: Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page) is still satisfied.
```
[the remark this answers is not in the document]
WilcoFiers: This is intentional. This isn't just true for WCAG 2.1, it is true for all of WCAG 2.

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: Grandfather tags are part of the assumptions, we don't need an example for it. I'm also not too sure what the benefit of it would be to the rule.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,278 @@
...
+	</head>
+	<body>
+		<p>Paul put dire comment on tape</p>
+	</body>
+</html>
+```
+
+[valid language tag]: #valid-language-tag
+[default page language]: #default-page-language
+[primary language]: https://tools.ietf.org/html/bcp47#section-2.2.1 'Definition of primary language subtag'
```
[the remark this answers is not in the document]
WilcoFiers: As above, I think this hinders readability of the rule.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,280 @@
+---
+id: ucwvc8
+name: HTML page language subtag matches default language
+rule_type: atomic
+description: |
+  This rule checks that the primary language subtag of the page language matches the default language of the page
```
daniel-montalvo, line 6: Based on https://github.com/act-rules/act-rules.github.io/pull/1334/files#r464284953

Couldn't we omit "subtag" here as well?


```suggestion
  This rule checks that the primary language of the page language matches the default language of the page
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,280 @@
...
+- [BCP 47: Tags for Identifying Languages](https://www.ietf.org/rfc/bcp/bcp47.txt)
+- [The `lang` and `xml:lang` attributes](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This page has a [default language][default page language] of English, which matches the language indicated in the `lang` attribute.
```
daniel-montalvo, line 70: We are not explaining why this page has a default language of English. Adding "as all of its words are English" or something along these lines may help clarity and consistency.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,280 @@
...
+		</p>
+	</body>
+</html>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This page has a [default language][default page language] of English, which is not the language indicated in the `lang` attribute.
```
daniel-montalvo, line 147: We are not explaining why this page has a default language of English. Adding "as all of its words are English" or something along these lines may help clarity and consistency.

_rules/html-page-lang-matches-default-ucwvc8.md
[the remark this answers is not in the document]
adilsofficial: The definition we are referring to here redirects us to the "primary language subtag" and on the other hand, primary language subtag refers to the language not the complete tag with the country. For example, primary language can be `en-GB` and `en-US` but the primary subtag would be `en`, not `GB` as it's a region where it used which specifies the language itself.
I guess I misunderstood that term here, you meant the primary language as general whether simple language subtag or language with other subtags such as `en` or `en-US` respectively.
Please correct me if I'm still misunderstanding it.

_rules/html-page-lang-matches-default-ucwvc8.md
adilsofficial: This description is also a bit another way round with the applicability and expectation of this rule. I guess the description of examples should be according to the [test case design](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/design/rule-design.md#test-cases) that is the example has or does not have or contains the necessary condition to be met and expectation. Indirectly, conditions in the case of a test rule is "Applicability" and then comes "Expectation". 
It does not mean that this description is not sufficient or incorrect but it should be according to the template we defined in [test case design](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/design/rule-design.md#test-cases).
According to the description of the rule plus the applicability and expectation, this rule checks that the primary language subtag of the page language matches the default language of the page. It is similarly the same if we say the default language of this page matches the primary language subtag of the valid language tag of the page but another way round against the description of the rule.
```suggestion
This page has a [primary language][] English ("en") set in the [valid language tag][] which matches the [default page language][] that is English.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -0,0 +1,280 @@
...
+	</head>
+	<body>
+		<p>The Dutch word "gelukkig" has no equivalent in English.</p>
+	</body>
+</html>
+```
+
+#### Passed Example 3
+
+This page has a [default language][default page language] of Dutch because all English text is wrapped in a `p` element with a `lang` attribute set to `en`. Dutch ("nl") is set as the language of the page by the `lang` attribute on the `html` element.
```
adilsofficial, line 106: Examples description varies, shouldn't it be consistent?

_rules/html-page-lang-matches-default-ucwvc8.md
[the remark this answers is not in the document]
WilcoFiers: It's a specifier. A language tag only consists of subtags, so adding it doesn't provide additional meaning. It's like saying "The child node of the node", you might as well say "the child of the node", because all children of a DOM node are themselves nodes. Both phrases mean the same thing, but the first one has unnecessary specificity in it.

If you insist, I'll take on the change, but I don't think it's necessary.

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: See my comment below. You do have to say subtag here. The "primary language of the page" in itself is not the same as the "primary language of the language tag"

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: I don't understand what you mean with this.

_rules/html-page-lang-matches-default-ucwvc8.md
WilcoFiers: I don't think we need to explain that. How definitions work rarely gets explained unless it is essential for understanding the example.

_rules/html-page-lang-matches-default-ucwvc8.md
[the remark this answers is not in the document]
adilsofficial: Well explained. No need to change it. 👍

pages/glossary/default-page-language.md
```diff
@@ -17,11 +17,12 @@ The _most common language_ is determined by counting the number of _words_ in th
 The _words_ used in the following texts should be counted for the _most common language_:
 
 - **page title**: The value of the [document title][] of the [document][] in the [top-level browsing context][].
-- **text nodes**: The value of any [text nodes][] that is [visible][] or [included in the accessibility tree][]
+- **text nodes**: The value of any [text nodes][] that are [visible][] or [included in the accessibility tree][]
 - **accessible text**: The [accessible name][] and [accessible description][] of any element included in the [accessibility tree][]
 
-**Exception**: Do not count words in text that comes from nodes that have an [ancestor][] in the [flat tree][] which has a non-empty (`""`) `lang` attribute and is not the [root node][].
+**Exception**: Do not count words in text that comes from nodes that have an [ancestor][] in the [flat tree][] which has a non-empty (`""`) `lang` attribute and is not the [root node][]. For accessible texts, only ancestors of the node with the accessible name is considered, the ancestry of nodes used in producing the accessible text are ignored, such as those referenced with aria-labelledby.
```
Jym77, line 23: I think we should have a Passed and Failed example with an `aria-labelledby` to illustrate the second sentence.

_rules/html-page-lang-matches-default-ucwvc8.md
daniel-montalvo: Fair enough.

_rules/html-page-lang-matches-default-ucwvc8.md
daniel-montalvo: I get that it is not "essential" to understand the example, but I do think it would help. Mildly thinking these explanations should be added.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -88,7 +92,7 @@ This page has `lang` attribute value of `en` (English), which matches the [defau
 
 #### Passed Example 2
 
-This page has `lang` attribute value of `en` (English), which matches the [default language of the page][default page language]. The default langage is English because all but a few words are English.
+This page has `lang` attribute value of `en` (English), which matches the [default language of the page][default page language]. The default language is English because all but a few words are English.
```
Jym77, line 95:
```suggestion
This page has a `lang` attribute value of `en` (English), which matches the [default language of the page][default page language]. The default language is English because all but a few words are English.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -124,7 +128,7 @@ This page has `lang` attribute value of `nl` (Dutch), which matches the [default
 
 #### Passed Example 4
 
-This page has `lang` attribute value of `en` (English), which matches the [default language of the page][default page language]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `en`.
+This page has `lang` attribute value of `en` (English), which matches the [default language of the page][default page language]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `nl`.
```
Jym77, line 131:
```suggestion
This page has a `lang` attribute value of `en` (English), which matches the [default language of the page][default page language]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `nl`.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -165,7 +169,7 @@ This page has `lang` attribute value of `da` (Danish), which does not matches th
 
 #### Failed Example 2
 
-This page has `lang` attribute value of `nl` (Dutch), which does not matches the [default language of the page][default page language]. The default langage is English because all but a few words are English.
+This page has `lang` attribute value of `nl` (Dutch), which does not matches the [default language of the page][default page language]. The default language is English because all but a few words are English.
```
Jym77, line 172:
```suggestion
This page has a `lang` attribute value of `nl` (Dutch), which does not matches the [default language of the page][default page language]. The default language is English because all but a few words are English.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -201,7 +205,7 @@ This page has `lang` attribute value of `en` (English), which does not matches t
 
 #### Failed Example 4
 
-This page has `lang` attribute value of `nl` (Dutch), which does not matche the [default language of the page][default page language]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `nl`.
+This page has `lang` attribute value of `nl` (Dutch), which does not match the [default language of the page][default page language]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `nl`.
```
Jym77, line 208:
```suggestion
This page has a `lang` attribute value of `nl` (Dutch), which does not match the [default language of the page][default page language]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `nl`.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -217,6 +221,24 @@ This page has `lang` attribute value of `nl` (Dutch), which does not matche the
 </html>
 ```
 
+#### Failed Example 5
+
+This page has `lang` attribute value of `nl` (Dutch), which does not match the [default language of the page][default page language]. The default language is English because the accessible name of the `img` element is English. The `lang` attribute on the `p` element is effectively ignored.
```
Jym77, line 226:
```suggestion
This page has a `lang` attribute value of `nl` (Dutch), which does not match the [default language of the page][default page language]. The default language is English because the accessible name of the `img` element is English. The `lang` attribute on the `p` element is effectively ignored.
```

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -169,7 +169,7 @@ This page has `lang` attribute value of `da` (Danish), which does not matches th
 
 #### Failed Example 2
 
-This page has `lang` attribute value of `nl` (Dutch), which does not matches the [default language of the page][default page language]. The default language is English because all but a few words are English.
+This page has a `lang` attribute value of `nl` (Dutch), which does not matches the [default language of the page][default page language]. The default language is English because all but a few words are English.
```
Jym77, line 172:
```suggestion
This page has a `lang` attribute value of `nl` (Dutch), which does not match the [default language of the page][default page language]. The default language is English because all but a few words are English.
```

daniel-montalvo (approved):

EmmaJP (approved):

PR #1335, opened 2020-06-05
Add labels to input elements

### LINKED ISSUE #1215: Add labels to every input field used as example [73f2c2]
As reported by @daniel-montalvo in https://github.com/act-rules/act-rules.github.io/pull/1189#discussion_r388804284 we have several test cases with `input` elements without a corresponding label. This is normally in unrelated rules (e.g., most examples of [`autocomplete` attribute has valid value](https://act-rules.github.io/rules/73f2c2)).

Should we add a label to these?
The absence of a label does not make the (passed) examples fail the SC, but it is still a bad practice…

Jym77:
Add labels to most `input` elements used in examples, as per CG decision.

I've left out `input` in the following rules:
* https://act-rules.github.io/rules/6cfa84
* https://act-rules.github.io/rules/e6952f
* https://act-rules.github.io/rules/674b10
* https://act-rules.github.io/rules/4e8ab6
because they are all extremely bare minimum `input`s.

I've also tried to avoid adding too many labels (i.e. in case where no label is needed for the example to "work"), please pay special attention in checking that it is the case 😄 

Closes issue(s):

- closes #1215

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


Jym77:
Final Call ends on June 16.

_rules/aria-attr-defined-5f99a7.md
```diff
@@ -71,7 +71,10 @@ This `div` element with a role of `alert` has an `aria-live` attribute which is
 This `input` element with a role of `spinbutton` has multiple `aria-*` attributes specified which are all defined in [WAI-ARIA Specifications][].
 
 ```html
-<input role="spinbutton" aria-valuemax="100" aria-valuemin="0" aria-valuenow="25" type="number" value="25" />
+<label
+	>Enter a number between 0 and 100:
```
EmmaJP, line 75: It seems add to have a new line in the middle of a tag. Minor typo, but should probably be fixed.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -100,55 +100,55 @@ Autocomplete term, only valid for textarea.
 Two autocomplete terms.
 
 ```html
-<input autocomplete="Work EMail" />
+<label>Work email<input autocomplete="Work EMail"/></label>
```
EmmaJP, line 103: Does this autocomplete value reference something? Its unusual camel casing in my experience, so doesn't read like a variable name, or something legit for an autocomplete value.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

_rules/aria-attr-defined-5f99a7.md
Jym77: That prettier being somewhat aggressive and cuting the line due to the long `input` element…
I've replaced this with a `label for=…` to make the result a bit more "logical".

_rules/autocomplete-valid-value-73f2c2.md
Jym77: autocomplete tokens are match case-insensitive, so this is technically OK.
I've nonetheless corrected it as it is indeed a bit confusing and definitely not what the example illustrates.

Jym77:

_rules/autocomplete-valid-value-73f2c2.md
EmmaJP: Why is it not all lower case like the other examples of autocomplete?

Jym77: That, I don't know and deem it an unrelated issue that need (or not) to be tackled in another PR…

PR #1341, opened 2020-06-09
image-button-acc: remove unnecessary note

WilcoFiers:
As suggested by the ACT Task Force. I realise this change will impact other rules too if we go this route. If folks are OK with this, I'll open an issue to do the other ones.

Need for Final Call: none, but I think I'll give it 1 week regardless


_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -39,9 +39,7 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to any HTML `input` element with a `type` attribute in the [`Image Button` state](<https://html.spec.whatwg.org/#image-button-state-(type=image)>), that is [included in the accessibility tree][].
-
-**Note:** The specification of the [`type`](https://html.spec.whatwg.org/#states-of-the-type-attribute) attribute describes in detail how to map the value of the attribute to its corresponding state.
+The rule applies to any HTML `input` element with a `type` attribute set to [`image`](<https://html.spec.whatwg.org/#image-button-state-(type=image)>), that is [included in the accessibility tree][].
```
Jym77, line 42:
```suggestion
The rule applies to any HTML `input` element with a [`type`](https://html.spec.whatwg.org/#states-of-the-type-attribute) attribute in the [`Image Button` state](<https://html.spec.whatwg.org/#image-button-state-(type=image)>), that is [included in the accessibility tree][].
```
"in the Image Button state" is the correct technical term to use. "a `type` attribute set to `image`" might be interpreted as `type="image"` while "in the Image Button state" also include `type="IMage"` and other variations.

WilcoFiers: The TF had multiple people report they could not figure out what this phrase means. We saw this even when we wrote the rule. That's a strong indication that this phrase doesn't pass the "plain language" requirement from the ACT rules format Accessibility experts will know what you're talking about when you say "type attribute set to image". This formulation has been used and approved multiple times.

There is an open issue to address this problem of not being explicit about case sensitivity. Even still I'm not convinced this needs to be addressed. I can not think of a good argument for treating the type attribute as case sensitive, when it is linked to a spec that clearly says the value should be treated as case insensitive. If an interpretation is obviously incorrect it is not ambiguous. That's not to say being explicit about obvious things is a bad thing, but it can not come at the cost of readability, which is the case with this phrase.

Even if it was ambiguous, which I don't believe it is, addressing case sensitivity is out of scope for this PR. It has not been a blocker for other PRs so treating it as a blocker for this one seems out of place.

Jym77: > Even if it was ambiguous, which I don't believe it is, addressing case sensitivity is out of scope for this PR. It has not been a blocker for other PRs so treating it as a blocker for this one seems out of place.

1. The problem is introduced by this PR. I don't see how it can be "out of scope" for any PR to avoid creating new problems.
2. This PRs is about "remov[ing] unnecessary note", not about changing reference to enumerated attribute state by something else in the Applicability. This **change** is out of scope to start with and shouldn't have been here in the first place.
3. If referencing the attribute state is a problem, it should be tackled in all our rules rather than leaving us in an inconsistent situation where we sometimes refer to it and sometimes not.

PR #1342, opened 2020-06-09
page-lang-xml-lang: minimise the number of notes

WilcoFiers:
This was a suggestion from the ACT TF.

Need for Final Call: none


_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -3,7 +3,7 @@ id: 5b7ae0
 name: HTML page `lang` and `xml:lang` attributes have matching values
 rule_type: atomic
 description: |
-  This rule checks that all HTML pages with both a `lang` and `xml:lang` attributes on the root element, have the same primary language subtag.
+  This rule checks that a non-embedded HTML page with both a `lang` and `xml:lang` attributes on the root element, have the same primary language subtag. HTML pages embedded into other documents, such as through `iframe` or `object` elements are not applicable because they are not web pages according to the definition in WCAG.
```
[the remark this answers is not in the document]
Jym77: See https://github.com/act-rules/act-rules.github.io/pull/1340#discussion_r440110846

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -44,6 +40,8 @@ For each test target, the values of the [primary language subtags][], if any exi
 
 - This rule assumes that [grandfathered tags][] are not used as these will not be recognized as [valid language tags](#valid-language-tag).
 
+- The rule assumes that having matching [primary language subtags][] of the `lang` and `xml:lang` attribute, but non-matching [language tags](https://tools.ietf.org/html/bcp47#section-2) overall, will not cause accessibility issues unless there's a sufficiently large difference between the two [language tags](https://tools.ietf.org/html/bcp47#section-2). One notable case is the [language tags](https://tools.ietf.org/html/bcp47#section-2) for Cantonese (`zh-yue`) and Mandarin (`zh-cmn`) where the [primary language subtags][] match, but the [extended language subtags][] don't. Such a case would not fail this rule, but could lead to accessibility issues.
```
Jym77, line 43:
```suggestion
- The rule assumes that having `lang` and `xml:lang` attributes with matching [primary language subtags][] but non-matching [language tags](https://tools.ietf.org/html/bcp47#section-2) overall, will not cause accessibility issues. This is not necessarily the case for all languages. One notable case is the [language tags](https://tools.ietf.org/html/bcp47#section-2) for Cantonese (`zh-yue`) and Mandarin (`zh-cmn`) where the [primary language subtags][] match, but the [extended language subtags][] don't. Such a case would not fail this rule, but could lead to accessibility issues.
```

Jym77: I think that still need to be addressed.

"having matching primary subtags but non-matching tags will not cause accessibility issues unless [it does]" is not an assumption, it's just a fact.
The rule assumes that we are in a case where there isn't "a sufficiently large difference between the two tags".

_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -3,7 +3,7 @@ id: 5b7ae0
 name: HTML page `lang` and `xml:lang` attributes have matching values
 rule_type: atomic
 description: |
-  This rule checks that all HTML pages with both a `lang` and `xml:lang` attributes on the root element, have the same primary language subtag.
+  This rule checks that a non-embedded HTML page with both a `lang` and `xml:lang` attributes on the root element, have the same primary language subtag.
```
carlosapaduarte, line 6:
```suggestion
  This rule checks that both `lang` and `xml:lang` attributes on the root element of a a non-embedded HTML page, have the same primary language subtag.
```

_rules/html-page-lang-xml-lang-match-5b7ae0.md
carlosapaduarte: I agree that this could be better presented. I think it would be okay to say that the rule assumes that having non-matching language tags does not cause accessibility issues. If it does (which is different from ending the sentence with "unless it does"), then the rule might pass but the SC will not be met.

PR #1345, opened 2020-06-09
Add condition list instructions to rule design doc

### LINKED ISSUE #1275: Formatting listed conditions
We have a couple or rules now that do things like:

> Each element that has:
> - property A, or
> - property B

The formatting of these conditional lists is a little inconsistent. I'd like to discuss normalising these. Here is what I think it should be:

## Proposed format

> Some condition... for which (all | one of) the following is true:
> - (**label 1**:) The (X) has condition A; (and | or)
> - (**label 2**:) The (X) has condition B; ( and | or)
> - (**label 3**:) The (X) has condition C.

This means we'd have adjust these types of conditions to use the following rules:

-  Use phrasing like "for which all the following is true" or "for which one of the following is true"; this creates clarity. This can be modified based on use, such as "to achieve one of the following objectives".
-  Labels are optional, not recommended for lists of 2, and recommended for lists of 4 or more conditions
- The subject must be referenced in some way in each conditional. For example by starting with "the element has", or by using phrasing like "The innerText of the element has...". The subject must be a single word that is referenced in the phrase preceding the conditional list.
- Use "; and" or "; or" at the end of each condition, except the last which ends with a period. This reenforces whether this applies to "all" or "one of"

## Guidance on conditional lists

Additionally, i think we should have the following guidelines for using conditional lists

- If there are more than three conditions, use a list, But keep the most important condition(s) in the phrase, not in the list. Don't do "Each element for which ... - that is a link; and - that is some other thing". Instead do "Each element that is a link, for which ... - some condition". This is to establish what type of element / node / etc is being tested, before limiting it down.

- Put things in order of how common they likely are. Even for unordered lists, this helps understand the list 

- Avoid nested conditional lists. Having one condition and inside that condition becomes difficult to read. Instead try to restructure the conditionals. The can be done by either moving some of the conditional logic into a definition, or by phrasing the conditions as a sentence. In expectations, multiple expectations could also be used.

- Group similar concepts into the same list item. For example, if there is a condition for absolute of fixed position, because they are closely related they could be part of the same condition item.

WilcoFiers:
As discussed in https://github.com/act-rules/act-rules.github.io/issues/1275

Closes #1275

Need for Final Call: 1 week


pages/design/rule-design.md
```diff
@@ -150,3 +150,23 @@ The description should:
...
+- Tend towards using a lists if there are more than three conditions. Keep the most important condition(s) in the phrase, not in the list. If something is about `img` elements for example, use "Each `img` element for which all the following is true, instead of putting "`img` element" in the condition list itself.
+- Put things in order of how common they likely are. Even for unordered lists, this helps understand the list
+- Avoid nested conditional lists. These are difficult to read. Instead try to restructure the conditionals. The can be done by either moving some of the list items into the condition phrase, or putting all the subconditions in a single phrase in the condition item. In expectations, use multiple conditions.
+- Group similar concepts into the same list item. For example, if something can have `absolute` or `fixed`, these two are closely related so putting them down as one item helps limit the number of conditions.
```
Jym77, line 172:
```suggestion
- Group similar concepts into the same list item. For example, if something can have `absolute` or `fixed`, these two are closely related so putting them down as one item helps limit the number of conditions.
- Write test cases that check each condition of the list individually, this help understand what that specific condition is needed.
- Refer to the labels of the condition in the test case description to make the link explicit.
- Order the test cases in the same order than the conditions they check, with conditions order to most important to least important, this implies that test cases are also ordered by importance.
```

pages/design/rule-design.md
```diff
@@ -153,20 +153,23 @@ For more details, see [ACT Rules Format: Test Cases](https://www.w3.org/TR/act-r
...
+- Put things in order of how common they likely are. Even for unordered lists, this helps understand the list.
 - Avoid nested conditional lists. These are difficult to read. Instead try to restructure the conditionals. The can be done by either moving some of the list items into the condition phrase, or putting all the subconditions in a single phrase in the condition item. In expectations, use multiple conditions.
 - Group similar concepts into the same list item. For example, if something can have `absolute` or `fixed`, these two are closely related so putting them down as one item helps limit the number of conditions.
+- Write test cases that check each condition of the list individually, this help understand what that specific condition is needed.
+- Refer to the labels of the condition in the test case description to make the link explicit.
+- Order the test cases in the same order than the conditions they check, with conditions order to most important to least important, this implies that test cases are also ordered by importance.
```
Jym77, line 175:
```suggestion
- Order the test cases in the same order than the conditions they check, with conditions ordered from most important to least important, this implies that test cases are also ordered by importance.
```

pages/design/rule-design.md
```diff
@@ -150,3 +150,26 @@ The description should:
...
+> Some condition... for which (all | one of) the following is true:
+> 
+> * (**label 1**:) the (X) has condition A; (and | or)
+> * (**label 2**:) the (X) has condition B; (and | or)
+> * (**label 3**:) the (X) has condition C.
+
+This phrasing is designed to be easily readable, but may not work in every situation. In all cases prefer readability over prescriptive formatting. Additionally to this, keep the following in mind when listing conditions:
+
+- The phrasing "for which (all | one of) the following is" can be modified based on use, such as "to achieve one of the following objectives".
+- Labels are optional, not recommended for lists of 2, and recommended for lists of 4 or more conditions. Keep the labels short, 1 or 2 words. If the label does not help clarify the list, don't use them even for longer lists (for example when the condition is short and the label would just repeat it).
```
carlosapaduarte, line 167: This raised a question for me: is it okay to have labels on only a subset of the list's items?

pages/design/rule-design.md
```diff
@@ -150,3 +150,26 @@ The description should:
...
+- The subject must be referenced in some way in each conditional. For example by starting with "the element has", or by using phrasing like "The innerText of the element has...". The subject must be a single word that is referenced in the phrase preceding the conditional list.
```
carlosapaduarte, line 168: Do we need to enforce the subject to be a "single word"?
I expect there will be situations where it may help to refer to the subject using qualifiers (e.g. something that is applicable to img elements or svg elements, and we have conditions that apply to only to one or the other - it would be useful to reference the "img element" and the "svg element")

PR #1352, opened 2020-06-17
Use technically correct language for attribute values

### LINKED ISSUE #584: Be more clear on the meaning of `attribute=value`
Many rules use the notation `attribute=value` when referring to attributes with certain values. While this may seem intuitive at first, it quite often masks the complexity of the underlying attribute values. 

For example, many attributes in HTML are so-called [enumerated attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute) in which attribute values are mapped to a finite set of possible states. One such attribute is the `type` attribute for the `<input>` element and as such using the notation `type="text"` really means an `<input>` element in the `Text` state (https://html.spec.whatwg.org/multipage/input.html#text-(type=text)-state-and-search-state-(type=search)). However, the values of enumerated attributes are case-insensitive and usually have defaults so both `type=""`, `type="TEXT"`, and `type="foo"` would result in an `<input>` element in the `Text` state.

Some more examples from different rules:

- https://act-rules.github.io/rules/73f2c2#applicability
- https://act-rules.github.io/rules/97a4e1#applicability
- https://act-rules.github.io/rules/bc659a#applicability
- https://act-rules.github.io/rules/6cfa84#applicability
- https://act-rules.github.io/rules/5c01ea#passed-example-2

Jym77:
Fix the use of `attribute=value` or similar formulations.

The new definition is meant to be technically correct while allowing a less cumbersome wording in the rules themselves.
Thus, we can write "an `input` element with a `type` [attribute value] of `image`" instead of "… in the `Image Button` state". The new formulation is more intuitive and closer to the HTML `<input type="image" />`.

I'm wondering whether I should add something in the rule design document, but I am not sure where. This concern both Applicability and Expectation (and possibly test cases descriptions), so it can't really go in any of these Sections. Should we add a section about "general guidelines for writing technically correct language"?

Mostly only changing the Applicabilities and Expectations. I've considered that the rest (notes, test case description, or even assumptions) don't need to be that technically correct and can use the more vernacular `attribute=value` for concision (so I actually haven't checked usage out of Applicabilities/Expectations).

Closes issue(s):

- closes #584

Need for Final Call:
This will require a 2 weeks Final Call (correcting several applicabilites and expectations)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


Jym77:
@WilcoFiers Great idea ❤️ 
I've implemented it. Hopefully this also solves the disagreement we have in https://github.com/act-rules/act-rules.github.io/pull/1341#discussion_r440115225 🤞 

@kasperisager I introduced a new definition and made significant changes. So requesting new review…

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -37,13 +37,12 @@ htmlHintIgnore:
...
-- both `paused` and `muted` attributes whose values are both false, and
+- an `autoplay` [attribute value][] of `true`, and
+- a `muted` [attribute value][] of `false`, and
+- a `paused` attribute whose value is false, and
 - either a `src` attribute or a child `source` element that references content with a duration of more than 3 seconds that contains audio.
 
-**Note:**
-
-The default value of both `paused` and `muted` attributes is `false`.
+**Note**: [`autoplay`](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) and [`muted`](https://html.spec.whatwg.org/multipage/media.html#attr-media-muted) are both HTML attribute set on elements. On the other hand, [`paused`](https://html.spec.whatwg.org/multipage/media.html#dom-media-paused) is a DOM attribute which is not specified in HTML.
```
carlosapaduarte, line 45:
```suggestion
**Note**: [`autoplay`](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) and [`muted`](https://html.spec.whatwg.org/multipage/media.html#attr-media-muted) are both HTML attributes set on elements. On the other hand, [`paused`](https://html.spec.whatwg.org/multipage/media.html#dom-media-paused) is a DOM attribute which is not specified in HTML.
```

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -25,11 +25,12 @@ acknowledgments:
...
 
-- an `autoplay` attribute whose value is true, and
-- both `paused` and `muted` attributes whose values are both false, and
+- an `autoplay` [attribute value][] of `true`, and
+- a `muted` [attribute value][] of `false`, and
+- a `paused` attribute whose value is false, and
 - either a `src` attribute or a child `source` element that references content with a duration of more than 3 seconds that contains audio.
 
-**Note:** The default value of both `paused` and `muted` attributes is `false`.
+**Note**: [`autoplay`](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) and [`muted`](https://html.spec.whatwg.org/multipage/media.html#attr-media-muted) are both HTML attribute set on elements. On the other hand, [`paused`](https://html.spec.whatwg.org/multipage/media.html#dom-media-paused) is a DOM attribute which is not specified in HTML.
```
carlosapaduarte, line 33:
```suggestion
**Note**: [`autoplay`](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) and [`muted`](https://html.spec.whatwg.org/multipage/media.html#attr-media-muted) are both HTML attributes set on elements. On the other hand, [`paused`](https://html.spec.whatwg.org/multipage/media.html#dom-media-paused) is a DOM attribute which is not specified in HTML.
```

_rules/auto-play-audio-has-control-mechanism-4c31df.md
```diff
@@ -29,11 +29,12 @@ htmlHintIgnore:
...
 
-- an `autoplay` attribute whose value is true, and
-- both `paused` and `muted` attributes whose values are both false, and
+- an `autoplay` [attribute value][] of `true`, and
+- a `muted` [attribute value][] of `false`, and
+- a `paused` attribute whose value is false, and
 - either a `src` attribute or a child `source` element that references content with a duration of more than 3 seconds that contains audio.
 
-**Note:** The default value of both `paused` and `muted` attributes is `false`.
+**Note**: [`autoplay`](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) and [`muted`](https://html.spec.whatwg.org/multipage/media.html#attr-media-muted) are both HTML attribute set on elements. On the other hand, [`paused`](https://html.spec.whatwg.org/multipage/media.html#dom-media-paused) is a DOM attribute which is not specified in HTML.
```
carlosapaduarte, line 37:
```suggestion
**Note**: [`autoplay`](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) and [`muted`](https://html.spec.whatwg.org/multipage/media.html#attr-media-muted) are both HTML attributes set on elements. On the other hand, [`paused`](https://html.spec.whatwg.org/multipage/media.html#dom-media-paused) is a DOM attribute which is not specified in HTML.
```

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -55,6 +55,7 @@ For this rule, it is assumed that the `autocomplete` attribute is not used on fo
 
 - While `autocomplete` in a promising technique for supporting personalization in HTML, support for this is fairly limited.
 - Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have a [semantic role][] of `none` and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+- Some user agents treat the value of `aria-disabled` attribute as case-sensitive.
```
carlosapaduarte, line 58:
```suggestion
- Some user agents treat the value of the `aria-disabled` attribute as case-sensitive.
```

_rules/video-only-description-track-ac7dc6.md
```diff
@@ -24,7 +24,7 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to any [non-streaming](#non-streaming-media-element) `video` element [visible][] where the video doesn't contain audio and does contain a `track` element with a `kind="descriptions"` attribute.
+The rule applies to every [non-streaming](#non-streaming-media-element) `video` element that is [visible][] where the video contains audio and a `track` element with a `kind` [attribute value][] of `descriptions`.
```
carlosapaduarte, line 27: You have changed the meaning of the applicability from videos with no audio to videos with audio. I guess that was a mistake (and certainly out of the scope of this PR)
```suggestion
The rule applies to every [non-streaming](#non-streaming-media-element) `video` element that is [visible][] where the video does not contain audio and contains a `track` element with a `kind` [attribute value][] of `descriptions`.
```

pages/glossary/attribute-value.md
```diff
@@ -0,0 +1,28 @@
...
+  - DOM tree
+---
+
+The _attribute value_ of an attribute set on an HTML element is the value that the attribute gets after being parsed and computed according to specifications. It may differ from the value that is actually written in the HTML code due to trimming whitespace or non-digits characters, default values, or case-insensitivity.
+
+Some notable case of attribute value, among others:
+
+- For [enumerated attributes][], the _attribute value_ is either the state of the attribute, or the keyword that maps to it; even for the default states. Thus `<input type="image" />` has an attribute value of either `Image Button` (the state) or `image` (the keyword mapping to it), both formulation having the same meaning; similarly, "an input element with a `type` _attribute value_ of `Text`" can be either `<input type="text" />`, `<input />` (missing value default), or `<input type="invalid" />` (invalid value default).
+- For [boolean attributes][], the _attribute value_ is `true` when the attribute represents the true value and `false` otherwise. Thus `<button disabled>`, `<button disabled="disabled">` and `<button disabled="">` all have a `disabled` _attribute value_ of `true`.
```
carlosapaduarte, line 15:
```suggestion
- For [boolean attributes][], the _attribute value_ is `true` when the presence of the attribute represents the true value and `false` otherwise. Thus `<button disabled>`, `<button disabled="disabled">` and `<button disabled="">` all have a `disabled` _attribute value_ of `true`.
```
I think this makes it easier to understand

Jym77: I don't think this would be correct.
The presence of the attribute **always** represent the true value (per definition), thus
> the attribute value is `true` when the presence of the attribute represents the truth value

is equivalent to
> the attribute value is always `true`

carlosapaduarte: Still, I have to say that I have troubles in understanding the sentence "the attribute value is `true` when the attribute represents the true value". What does it mean for an attribute to represent the true value? Does it have to be equal to true?

What about "the _attribute value_ is `true` when present and `false` otherwise"?

carlosapaduarte: Still, I have to say that I have trouble in understanding the sentence "the attribute value is `true` when the attribute represents the true value". What does it mean for an attribute to represent the true value? Does it have to be equal to true?

What about "the _attribute value_ is `true` when present and `false` otherwise"?

Jym77: Hmmm… 🤔 
Yeah, "represent" is weird but its the word used by the specs…
>  The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

So, "the attribute represents the true value" is essentially equivalent to "the attribute is present".

I wanted to not copy the def from the spec but rather use the result… I guess maybe that wasn't such a great idea…

Jym77:
Final Call ends on July 7.

Jym77:

PR #1356, opened 2020-06-24
Add test case design instructions

### LINKED ISSUE #1306: How strict should our test cases be?
We recently added a few test cases in [Handling Presentational role conflict resolution](https://github.com/act-rules/act-rules.github.io/pull/1189). This has caused our implementation numbers to drop significantly.

As product owner, this isn't high on my list of priorities. I've never once heard that missing support for this edge case caused a problem. My only motivation to want to implement it comes from it being a requirement for an ACT rules.

With my chair hat back on, that feels like we've overstepped our bounds. There are much more pressing issues to be resolved than check to make sure every implementor understands conflict resolution. I came across a similar issue this week, one we haven't addressed yet. I've been hesitant to even raise it because I don't think it's something we should invest time in. (The issue was that we're not taking [presentational children](https://www.w3.org/TR/wai-aria-1.1/#childrenArePresentational) into account anywhere.)

Another example of this is in the color-contrast rule that I wrote. I added an example where text shadow is used to up the contrast ratio just above the minimal requirements. This situation is very uncommon. Unlike the previous one I do know this occasionally causes false positives, but again it is very very rare, and by no means trivial to support.

I want us to have a conversation about how we draw the line when it comes to adding test cases.

WilcoFiers:
I worked on this proposal after discussing it on both the TF and CG call. I think most people are roughly on the same page on this, but I'd love some reviewers on this.

Closes #1306

Need for Final Call: 2 weeks


pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
+# ACT Test Case Design
+
+The goal of test cases in ACT rules is allow implementors to varify that their implementation is sufficiently consistent with the ACT rule. This requires that a rule has test cases for all of the "important" parts of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. A list of test cases full of edge cases is going to ensure greater consistency between implementors, but may result in fewer implementors deciding to adopt the rule.
```
EmmaJP, line 3:
```suggestion
The goal of test cases in ACT rules is to allow implementors to verify that their implementation is sufficiently consistent with the ACT rule. This requires that a rule has test cases for all of the "important" aspects of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. A list of test cases full of edge cases would ensure greater consistency between implementors, but may result in fewer implementors deciding to adopt the rule.
```

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+
+To strike a good balance in test cases, the following principles should be considered:
+
+1. Test every "condition" in the rule
+1. Test one thing at a time
+1. Ensure consistency with accessibility 
+1. Tests should be small
```
EmmaJP, line 10:
```suggestion
1. Keep test cases small
```

Suggestion to keep the grammar perspective consistent between the bullet points rather than to change the meaning, so do adjust if the meaning is changed.

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+At a minimum, every "condition" in the applicability and expectation has to have both positive and negative test cases for it. For example the phrase "`img` and `svg` elements" has two conditions in it; "`img` elements" and "`svg` elements". A rule with such a phrase in the applicability would require the following test cases:
+
+- `img` that passes the rule
+- `svg` that passes the rule
+- `img` that fails the rule
+- `svg` that fails the rule
+- some other element that is inapplicable
+
+Providing these test cases ensures that an implementor that correctly tests `img` elements, but passes every `svg` element can be identified as a partial implementation, and one where it fails all `svg` elements is identified as an inconsistent implementation.
+
```
EmmaJP, line 25: Very nice example!

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+Related to this, failed test cases should only fail the accessibility requirements for one reason. A failed test case for 4.1.2 Name, Role, Value should not have issues with both the accessible name, and the role. If the rule is about accessible names, the role and value must be correct.
+
+## Small Test Cases
+
+A general principle in test case creation is to keep test cases small. Avoid any unnecessary content, as it may cause unexpected side effects in the outcome. For example, a rule about buttons does not require that all test cases have an HTML element with a `lang` attribute and a `title` element. Those elements do not affect the accessibility of a button.
+
+An exception to that, is that ACT Rules should strive to show good accessibility practices. A rule about valid `autocomplete` attributes should still provide labels for the form controls with the `autocomplete` attribute. Those labels both provide context, showing the intended purpose of the input, and they show good practice of providing labels for form controls.
+
```
EmmaJP, line 47: Keeping test cases small is a challenging thing to explain. The various "is descriptive" rules do include the HTML element and lang attribute, but none of the other prescribed wrapping elements such as head, title, and body.

To some extent everything specifically included within a test case should be relevant to the rule, and absent wrapping elements can be assumed to follow best practice.

I think it would be good to state that in some way, though I'm not sure that ^^ is the best phrasing.

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+## Superficial Definition Testing
+
+ACT rules use definition to avoid duplication, and to "hide" away the details. Definitions such as "semantic role", "visible" and "accessible name" are used in many rules, and are fundamental building blocks of ACT. To avoid having a lot of similar test cases in every rule that uses a particular definition, it is best to keep definition testing to a minimum.
```
EmmaJP, line 50:
```suggestion
ACT rules use definitions to avoid duplication, and to "hide" away the details. Definitions such as "semantic role", "visible" and "accessible name" are used in many rules, and are fundamental building blocks of ACT. To avoid having a lot of similar test cases in every rule that uses a particular definition, it is best to keep definition testing to a minimum.
```

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+
+As an example, the "X has non-empty accessible name" rules all have test cases that check for the different ways that element can have an accessible name. For `img`, that is `alt`, `title`, and `aria-label(ledby)` attributes. That creates sufficient coverage to ensure the basics are understood. But those rules do not go into the details of ensuring accessible name computation is done right. Whether an implementor considers `aria-owns` in the accessible name computation is not relevant for "non-empty accessible name" rules.
+
+That is not to say edge cases in definition should never have test cases. If there are **real-world examples** of a particular edge case affecting the outcome of a rule, test cases should be added to the rule to cover that edge case.
```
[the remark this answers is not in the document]
Jym77: Random thinking of probably bad solutions for these problems, hoping that bad ideas can be picked up by others and evolve into good ones…

---
A way to deep test shared definitions without duplicating corner cases in every rules would be to have a fake rule dedicated to that purpose. For example, a rule "element has accessible name" without any accessibility requirements, just for the sake of piling up all the corner cases of accessible name computation.
Problem with that is creating a stupid rule that can be ignored by implementors, and the fact that accessible name computation is not an exact science (as in every single combination of UA+AT solves corner cases differently…)

---
A way to know which definition is of which kind could be to have some "usage guideline" info inside the definition. Either as a label in the frontmatter, or as a somewhat lengthier definition. Thinks like "when using this definition, write test cases for each of the conditions in it" (for something like "hidden state"), or "this definition is thoroughly tested in rule Foo, not need to retest everything in other rules" (for something like background colors), or …
Problem is that these may evolve, require some initial work to be set up, and may require a fair amount of work to maintain.

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+
+It is fairly common for rules to involve technologies that are not fully supported. For example, ARIA roles get added every few years, and it takes a while for new roles to get adopted by assistive technologies. In the case that a particular technology has very little or no support at all, the rules should not have test cases for this.
+
+Once there is significant support for a particular technology, it should be considered for test cases. The best way to introduce technologies that are only partially supported is to exclusively provide failed test cases for them. This makes an implementation that does not support this technology partially consistent.
+
+Technologies that are widely supported can be used in all test cases. In all situations, previous principles of deciding if a test case is needed or not still apply.
+
+## Exceptions
+
+It is important to know that there are exceptions to all of these principles. There may be technical reasons not to include a certain test case. It may not make logical sense for some passed test cases to satisfy all accessibility requirements. Some definitions may not require tests, and others can be argued to need extensive tests.
```
Jym77, line 66: Wrote my comments before reading that 🙄 👏 

Maybe copy this paragraph (or something similar but shorter) at the start, just after the list of principles:
```markdown
…
1. Avoid tests cases contested by (accessibility) support

These are generic principles toward which we aim, but specific situations might call for exceptionally ignoring some of them in a given rule.
```

pages/design/test-cases.md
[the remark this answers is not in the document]
EmmaJP: Definitions, by definition, are a "formal statement of the meaning or significance or a word, phrase" etc. For ACT Rules, essentially explaining an agreed way to understand a word or phrase in respect of the ACT Rules. They aren't really hiding detail, and I don't think they should be. They help ACT Rules to remain more succinct through using the defined word or phrase, rather than explaining it inline each time.

I don't think definitions need tests. However, if something about a definition has implications for a particular rule that may require additional test cases, I do think that should be handled within each rule where that is applicable. Those cases are relevant for the rule, not the definition.

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
+# ACT Test Case Design
+
+The goal of test cases in ACT rules is to allow implementors to verify that their implementation is sufficiently consistent with the ACT rule. This requires that a rule has test cases for all of the "important" aspects of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. A list of test cases full of edge cases would ensure greater consistency between implementors, but may result in fewer implementors deciding to adopt the rule.
```
carlosapaduarte, line 3:
```suggestion
The goal of test cases in ACT rules is to allow implementors to verify that their implementation is consistent with the ACT rule. This requires that a rule has test cases for all of the "important" aspects of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. A list of test cases full of edge cases would ensure greater consistency between implementors, but may result in fewer implementors deciding to adopt the rule.
```

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Superficial Definition Testing
+
+ACT rules use definitions to avoid duplication, and to "hide" away the details. Definitions such as "semantic role", "visible" and "accessible name" are used in many rules, and are fundamental building blocks of ACT. To avoid having a lot of similar test cases in every rule that uses a particular definition, it is best to keep definition testing to a minimum.
+
+As an example, the "X has non-empty accessible name" rules all have test cases that check for the different ways that element can have an accessible name. For `img`, that is `alt`, `title`, and `aria-label(ledby)` attributes. That creates sufficient coverage to ensure the basics are understood. But those rules do not go into the details of ensuring accessible name computation is done right. Whether an implementor considers `aria-owns` in the accessible name computation is not relevant for "non-empty accessible name" rules.
```
carlosapaduarte, line 52:
```suggestion
As an example, the "X has non-empty accessible name" rules all have test cases that check for the different ways that an element can have an accessible name. For `img`, that is `alt`, `title`, and `aria-label(ledby)` attributes. That creates sufficient coverage to ensure the basics are understood. But those rules do not go into the details of ensuring accessible name computation is done right. Whether an implementor considers `aria-owns` in the accessible name computation is not relevant for "non-empty accessible name" rules.
```

pages/design/test-cases.md
```diff
@@ -0,0 +1,68 @@
...
+
+That is not to say edge cases in definition should never have test cases. If there are **real-world examples** of a particular edge case affecting the outcome of a rule, test cases should be added to the rule to cover that edge case.
```
carlosapaduarte, line 54:
```suggestion
That is not to say edge cases in definitions should never have test cases. If there are **real-world examples** of a particular edge case affecting the outcome of a rule, test cases should be added to the rule to cover that edge case.
```

pages/design/test-cases.md
WilcoFiers: Will do.

pages/design/rule-design.md
```diff
@@ -149,4 +149,4 @@ The description should:
 
 > _For example: "This page has a `title` element with content."_
 
-For more details, see [ACT Rules Format: Test Cases](https://www.w3.org/TR/act-rules-format/#test-cases).
+For a detailed description on what to write test cases for see [test case design](test-cases.html). For more details, see [ACT Rules Format: Test Cases](https://www.w3.org/TR/act-rules-format/#test-cases).
```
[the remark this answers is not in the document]
nitedog: I think the suggestion is to replace the "For more details" with something like "For more background" or such. The term "details" does not really fit in this second sentence.

pages/design/test-cases.md
```diff
@@ -0,0 +1,70 @@
+# ACT Test Case Design
+
+The goal of test cases in ACT rules is to allow implementers to verify that their implementation is consistent with the ACT rule. This requires that a rule has test cases for all of the "important" aspects of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. A list of test cases full of edge cases would ensure greater consistency between implementers, but may result in fewer implementers deciding to adopt the rule.
```
nitedog, line 3: This description seems more discouraging than what we have discussed. The entire purpose of ACT is to ensure greater consistency between implementations, so challenging that seems counter-productive. Can we say this more positively, that we encourage the broadest coverage possible but that this needs to be based on real-world issues (as opposed to theoretical situations that do not occur in practice)?

pages/design/test-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+To strike a good balance in test cases, the following principles should be considered:
+
+1. Test every "condition" in the rule
+1. Test one thing at a time
+1. Ensure consistency with accessibility
+1. Keep test cases small
+1. Test definitions superficially
+1. Avoid tests cases contested by (accessibility) support
```
nitedog, line 12: I thought we want to a requirement along the lines of "based on real-world examples" or such?

pages/design/rule-design.md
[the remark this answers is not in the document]
Jym77: @nitedog got it right 😀 My point was that saying "for more details, go see that less detailed document" is not really correct…

pages/design/test-cases.md
```diff
@@ -0,0 +1,70 @@
...
+
+Technologies that are widely supported can be used in all test cases. In all situations, previous principles of deciding if a test case is needed or not still apply.
+
+## Exceptions
+
+It is important to know that there are exceptions to all of these principles. There may be technical reasons not to include a certain test case, or the number of combinations may require an unreasonable number of tests. It may sometimes not make logical sense for some passed test cases to satisfy all accessibility requirements. Some definitions may not require tests, and others can be argued to need extensive tests. Etc.
+
+Letting **real-world** examples drive consistency forward also means rule authors should not get stuck in trying to write test cases for every conceivable situation. Rule authors should strive for a 95% consistency between implementers on first publication. As more examples are discovered, this number will go up over time.
```
nitedog, line 70: I think this comes way too late in the document. I think we should be clear from the start: we do not want to test how the different implementations implement technical specifications, such as HTML, CSS, ARIA, or whatever -- we want to test how the different implementations implement the actual issue represented in the ACT Rule in various realistic code examples. I feel this needs to be one of the primary requirements, that all test cases are driven and backed by real-world examples.

PS: I think the XYZ% consistency part is difficult for rule authors to measure and meet, so that such a requirement seems quite unrealistic for us to require.

pages/design/test-cases.md
WilcoFiers: Updating this now to mention "real-world edge cases" in this first section. I think that should resolve all your comments without significantly changing the document.

pages/design/test-cases.md
WilcoFiers: Added an explanation of the percentage.

jkodu:
@WilcoFiers I changed the title on this like so: https://github.com/act-rules/act-rules.github.io/blob/244c27f2512ce939906edd02226d08c1bc600469/pages/design/test-cases.md#L1

So the spelling plugin does not fail, assuming that it is a [retext repeated acronym](https://github.com/retextjs/retext-redundant-acronyms)

pages/design/test-cases.md
carlosapaduarte: @WilcoFiers you probably forgot to push some commits. I can't find the updated first section. Neither can I find the explanation about the percentages you mention in https://github.com/act-rules/act-rules.github.io/pull/1356/files#r451468720

pages/design/rule-design.md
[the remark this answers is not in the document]
jkodu: @WilcoFiers the above comment needs to be addressed.

pages/design/test-cases.md
jkodu: @WilcoFiers this comment has not been addressed.

WilcoFiers: I did, it's on line 3.

pages/design/test-cases.md
```diff
@@ -0,0 +1,70 @@
+# ACT Rules Test Case Design
+
+The goal of test cases in ACT rules is to allow implementers to verify that their implementation is consistent with the ACT rule. This requires that a rule has test cases for all of the "important" aspects of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. The dicision of wha edge cases to test should be based on issues in **real-world** examples.
```
ajanec01, line 3:
```suggestion
The goal of test cases in ACT rules is to allow implementers to verify that their implementation is consistent with the ACT rule. It requires that a rule has test cases for all of the "important" aspects of it. Beyond that, it is up to the rule authors to determine how extensive to make the list of test cases. The decision of what the edge cases to test will be,  should be based on issues in **real-world** examples.
```

pages/design/rule-design.md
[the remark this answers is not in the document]
EmmaJP: Agree, and suggest "For further reference, see..."

EmmaJP (approved):

adilsofficial (approved):

PR #1357, opened 2020-06-24
form-controls-acc-name: Modernise the rule

WilcoFiers:
- Removed unnecessary notes
- Update the example descriptions
- Remove examples that are inconsistent with the SC
- Ensure all examples have proper labels

Work for the ACT TF: https://github.com/w3c/wcag-act/issues/432

Need for Final Call: 1 week


_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -97,61 +103,54 @@ _There are currently no assumptions_
...
 
 ```html
+<div>last name</div>
 <input />
 ```
 
 #### Failed Example 2
 
-[Non-focusable](#focusable) element still needs an [accessible name][].
+This [non-focusable](#focusable) `input` element has no attribute to give it an [accessible name][].
```
Jym77, line 126: "non-fousable" is not really correct. It's out of sequential focus navigation, but still programmatically focusable.

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -44,21 +37,29 @@ _There are currently no assumptions_
...
+- [H65: Using the title attribute to identify form controls when the label element cannot be used](https://www.w3.org/WAI/WCAG21/Techniques/html/H65)
 
 ## Test Cases
 
 ### Passed
 
 #### Passed Example 1
 
-[Implicit role](#implicit-role) with implicit label.
+This `input` element has an [accessible name][] because of its implicit label.
```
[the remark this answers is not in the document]
ajanec01: I'd say programmatic is more accurate. Implicit could be a suggested visual label which is not programmatically associated.

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -97,61 +103,54 @@ _There are currently no assumptions_
 
 #### Passed Example 5
 
-[Explicit role](#explicit-role).
+This element with a `combobox` [role][semantic role] has an [accessible name][] because of its `aria-label` attribute.
```
ajanec01, line 106: Isn't the first role link redundant?

WilcoFiers: Why would it be?

_rules/form-control-non-empty-accessible-name-e086e5.md
WilcoFiers: I'm replacing this with a disabled element. Makes for a better example IMO.

_rules/form-control-non-empty-accessible-name-e086e5.md
ajanec01: Ignore my comment, I didn't realise that "[role]" was a link to semantic role definition... sorry!

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -44,21 +37,29 @@ _There are currently no assumptions_
...
 - Several assistive technologies have a functionality to list all form fields on a page, including the `disabled` ones. Therefore this rule is still applicable to `disabled` form fields. If an assistive technology consistently ignores `disabled` form fields in all its interactions, then it is possible to have a `disabled` form field with no accessible name without creating accessibility issues for the user.
 - Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have one of the applicable [semantic roles][] and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+- Element with the `option` role are not tested in this rule because options do not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If these elements are presented as user interface components, these need to be tested separate from this rule.
```
carlosapaduarte, line 40:
```suggestion
- Elements with the `option` role are not tested in this rule because they do not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If these elements are presented as user interface components, these need to be tested separately from this rule.
```

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -44,21 +37,29 @@ _There are currently no assumptions_
...
 
 ## Background
 
-- [Understanding Success Criterion 3.3.2: Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions)
-- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)
+The list of roles in the applicability is derived by taking all the roles from [WAI-ARIA Specifications](#wai-aria-specifications) that:
+
+- have a [semantic roles][] that inherits from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and
```
carlosapaduarte, line 46:
```suggestion
- have [semantic roles][] that inherit from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and
```

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -44,21 +37,29 @@ _There are currently no assumptions_
...
+- does not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
```
carlosapaduarte, line 47:
```suggestion
- do not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
```

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -97,61 +102,54 @@ _There are currently no assumptions_
...
 
 ```html
-<input tabindex="-1" />
+<input disabled />
 ```
 
 #### Failed Example 3
 
-`aria-label` with empty text string
+This element with a `combobox` [role][semantic role] has an empty (`""`) [accessible name][] because the space in the `aria-label` [attribute value][] is trimmed.
```
carlosapaduarte, line 133: There is no element with a combobox role in this example

Jym77:
> @WilcoFiers the tests are failing because the `#attribute-value` glossary item is missing. Perhaps forgot to commit it?

It's in Final Cal ending tomorrow: #1352 
So should be good.

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -24,13 +24,6 @@ acknowledgments:
 
 This rule applies to any element that is [included in the accessibility tree](#included-in-the-accessibility-tree), and that has one of the following [semantic roles][]: `checkbox`, `combobox` (`select` elements), `listbox`, `menuitemcheckbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, `switch`, `textbox`.
```
jnurthen, line 25: I question why menuitemcheckbox and menuitemradio appear here and menuitem does not.

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -44,21 +37,29 @@ _There are currently no assumptions_
...
+- Elements with the `option` role are not tested in this rule because they do not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If these elements are presented as user interface components, these need to be tested separately from this rule.
 
 ## Background
 
-- [Understanding Success Criterion 3.3.2: Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions)
-- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)
+The list of roles in the applicability is derived by taking all the roles from [WAI-ARIA Specifications](#wai-aria-specifications) that:
 
-**Note:** This rule does not fail 3.3.2 as there are sufficient techniques within 3.3.2 that don't need the elements to have an [accessible name][]. For example "G131: Providing descriptive labels" **AND** "G162: Positioning labels to maximize predictability of relationships" would be sufficient.
+- have [semantic roles][] that inherit from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and
```
jnurthen, line 46: https://github.com/w3c/aria/pull/1278 removes the multiple inheritance from menuitemcheckbox and menuitemradio such that this will no longer be true once this appears in a spec.

_rules/form-control-non-empty-accessible-name-e086e5.md
WilcoFiers: Fair question. Similarly I think it's worth asking why this include spinbutton but not button. This is unrelated to the PR though, so I'll open a separate issue for it. Definitely something worth looking at.

_rules/form-control-non-empty-accessible-name-e086e5.md
WilcoFiers: Thanks for the heads up! I'll open an issue for it. This'll be actionable once ARIA 1.2 makes it further into rec track.

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -97,61 +102,54 @@ _There are currently no assumptions_
...
-	<input />
-</label>
-```
-
 ### Failed
 
 #### Failed Example 1
 
-No [accessible name][].
+This `input` element has no attribute to give it an [accessible name][].
```
adilsofficial, line 116:
```suggestion
This `input` element does not have an attribute that gives an [accessible name][] to it.
```

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -97,61 +102,54 @@ _There are currently no assumptions_
...
 
 ```html
+<div>last name</div>
 <input />
 ```
 
 #### Failed Example 2
 
-[Non-focusable](#focusable) element still needs an [accessible name][].
+This disabled `input` element has no attribute to give it an [accessible name][].
```
adilsofficial, line 125:
```suggestion
This disabled `input` element does not have an attribute that gives an [accessible name][] to it.
```

_rules/form-control-non-empty-accessible-name-e086e5.md
```diff
@@ -162,58 +160,34 @@ The implicit label is not supported on `div` elements.
...
-```html
-<input role="none" />
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-Hidden to everyone.
+This `input` element is not [included in the accessibility tree][] because of its `style` attribute.
```
adilsofficial, line 174: It's not actually because of `style` attribute alone but because of `display: none;`
```suggestion
This `input` element is not [included in the accessibility tree][] because of its `style` attribute with `display:none;`.
```

ajanec01: Or... "display property value set to none".

PR #1358, opened 2020-06-25
rule-design: Update random ID generator

### LINKED ISSUE #1339: How come our IDs don't use [g-z]?
I noticed this going through our list of rules this week, having all the URLs in one long list. With very few exceptions, the IDs use either a number, or the letters A through F. It strikes me that that's a hex range, and that when we're creating IDs we're not using the full [0-9a-z] range.

Looking more closely, the only rules that I found actually had letters in the G-Z range were ones I authored.

To be specific, there is a list of 50 rules here https://github.com/act-rules/act-rules.github.io/issues/1120#issuecomment-585812132, only 4 of them use chars in G-Z in their ID.

WilcoFiers:
Replace the hex generator with an alphanumeric number generator.

- closes #1339

Need for Final Call: None

### DIFF
--- a/pages/design/rule-design.md
+++ b/pages/design/rule-design.md
@@ -28,7 +28,7 @@ For more details, see [ACT Rules Format: Structure](https://www.w3.org/TR/act-ru
 
 ### Rule identifier
 
-The 6-character identifier is created using a random alpha-numeric generator, such as [random.org/strings](https://www.random.org/strings/?num=1&len=6&digits=on&loweralpha=on&unique=on&format=html&rnd=new). The identifier remains the same if the rule name changes.
+The 6-character identifier is created using a random alpha-numeric generator, such as [random.org/strings](https://www.random.org/strings/?num=1&len=6&digits=on&loweralpha=on&format=html&rnd=new). The identifier remains the same if the rule name changes.
 
 ### Rule name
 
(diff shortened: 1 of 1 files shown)


pages/design/rule-design.md
```diff
@@ -28,7 +28,7 @@ For more details, see [ACT Rules Format: Structure](https://www.w3.org/TR/act-ru
 
 ### Rule identifier
 
-The 6-digit identifier is created using a random hex generator, such as https://www.browserling.com/tools/random-hex. The identifier remains the same if the rule name changes.
+The 6-character identifier is created using a random alpha-numeric generator, such as [random.org/strings](https://www.random.org/strings/?num=1&len=6&digits=on&loweralpha=on&unique=on&format=html&rnd=new). The identifier remains the same if the rule name changes.
```
Jym77, line 31:
```suggestion
The 6-character identifier is created using a random alpha-numeric generator, such as [random.org/strings](https://www.random.org/strings/?num=1&len=6&digits=on&loweralpha=on&format=html&rnd=new). The identifier remains the same if the rule name changes.
```
Unicity of the individual characters is not needed.

jkodu:
I assume this is for future rules that is to be written & we do not intend on changing ids of existing rules.

PR #1359, opened 2020-06-25
Replace usage of attribute=value in the applicability of "Element with aria-hidden has no focusable content" (6cfa84)

carlosapaduarte:
This rule uses `aria-hidden="true"` in the applicability. 
Replacing it as suggested in https://github.com/act-rules/act-rules.github.io/issues/584#issuecomment-527828361

Need for Final Call:
This will require a 1 week Final Call (rephrasing the applicability)

---

### **When merging a PR:**

## How to Review And Approve


carlosapaduarte:
You're right @kasperisager 
I was going through my backlog and missed that it had already been taken care of!

Closing this PR!

PR #1360, opened 2020-06-26
Remove "font awesome" example from Link as non-empty accessible name (c487ae)

### LINKED ISSUE #1353: Font awesome icon has accessible name [c487ae]
Font awesome icons insert the unicode of the icon using css (content for :before)

This means that failed example 3 in rule [Link has non-empty accessible name](https://act-rules.github.io/rules/c487ae) should in fact pass.

Firefox gives an accessible name to the link in that example. Chrome does not. We're tending to agree with Firefox on this. What is everyone's else opinion?

carlosapaduarte:
<< Describe the changes >>

With TF agreement, remove the font awesome example. The CSS content property causes some user agents to give the icon an accessible name, while others don't.

Closes issue(s):

- closes #1353 

Need for Final Call:
This will require a 1 week Final Call (removal of a test case)

---

### **When merging a PR:**

## How to Review And Approve


carlosapaduarte:
@jkodu Thanks for the approval but I'll remove it because this already had @WilcoFiers' approval

carlosapaduarte:
Final call ended. Merging.

Issue #1305, opened 2020-05-07
Reconsider how many notes we have in rules


WilcoFiers:
There are a lot of notes in rules, some rules have 4 or 5 rules for two lines of applicability and expectation. We should rethink how frequent we resort to using notes, and if there are better solutions.

Jym77:
@EmmaJP check the recent PRs about removing some notes.
#1340, #1341, #1342, #1343, #1344.

Jym77:
As mentioned during the CG call, one point that becomes more visible with this policy is the difference between two kinds of assumptions:
* assumptions where, if it doesn't hold the rule may fail while the SC pass (e.g. "language is programatically set by `lang` attribute" => if language is set by HTTP headers, the rule will fail but the SC will pass). These are super important as the fail-on-fail relationship is at the core of the format and we must be cristal clear on the reasons why it may not hold.
* assumptions where, if it doesn't hold the rule may pass while the SC fails (e.g. "there can be more than one `title` element and it can be outside `head`" => if it's not true and WCAG actually means that 2 `title`s is bad, then the rule will pass on pages with 2 `title`s where the SC fails). This are OK-ish as almost all our rules have a "passing: further testing needed". It is good to document  the possible gaps, but this does not create false positives which we absolutely want to avoid.

Moving some notes to assumptions (eg https://github.com/act-rules/act-rules.github.io/pull/1340/files#diff-0778ae27333525135efc58a1e903eadeR52) creates a lot of the second case assumptions (as these tend to be in notes previously).

I'd like us to make sure whenever we write an assumption whether it is a first or second case assumption. We're usually pretty good at that by adding sentences like "if this is not the case, the rule may fail while the SC pass" or (as in https://github.com/act-rules/act-rules.github.io/pull/1340/files#diff-0778ae27333525135efc58a1e903eadeR52) "if this is not the case, failing the rule still means the SC fails".
Let's keep writing these explicit "what happens if false" conclusion to assumptions 😄

Jym77:
Another way to remove (some) notes is to move the content to the "Background" section.

So far, we've mostly used it for list of references but, according to the ACT format (emphasis mine):
> An ACT Rule may contain information about **the background for the development of the rule**, or references to relevant reading.

So, I feel that many "explanatory only" notes can move to the Background section. This is a place for non-normative text and for explaining some of the reasoning.

[...]

PR #1365, opened 2020-06-30
Update aria-required-owned-element-bc4a75.md before publishing

ajanec01:
Three notes from the applicability section moved to the background section. The first note with the `tablist` element removed. All changes made in response to Issue #1305. 

This can be merged with 3 approvals.


_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -64,6 +56,12 @@ If the [explicit semantic role][] on the target element is incorrectly used, and
 - [Required Owned Element](https://www.w3.org/TR/wai-aria-1.1/#mustContain)
 - [Owned Element](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element)
 
+HTML elements with an [implicit semantic role][] corresponding with [explicit semantic role][] are not tested in this rule because some of these elements have additional requirements. For example, an element that has an implicit semantic role that is identical to its explicit semantic role is a `ul` element that has `role="list"`. However, there is no native HTML element to "group" `li` elements in a `ul` element.
```
jkodu, line 59: I may be wrong here, but the below notes are reasoning why some things are excluded by this rule.

I do not think that belongs in the background section?

ajanec01: I've followed the pattern for PR #1342 where the information of HTML elements within `iframe` not being assessed is moved to the background section.  It's also giving information on why the rule has been written this way and not another- [comment](https://github.com/act-rules/act-rules.github.io/issues/1305#issuecomment-644622078) by @Jym77 .

Happy to change if you still feel like it should stay in the Applicability section.

PR #1372, opened 2020-07-06
"ARIA state or property is permitted" (5c01ea): Make examples compliant with ARIA 1.2

### LINKED ISSUE #1370: ARIA state or property is permitted - Passed examples 10 and 11 fail with ARIA 1.2
https://act-rules.github.io/rules/5c01ea#passed-example-10
https://act-rules.github.io/rules/5c01ea#passed-example-11

Passed examples 10 and 11 are (technical) passes with ARIA 1.1 but failures with ARIA 1.2, because `aria-expanded` is no longer allowed on `role=complementary`

https://www.w3.org/TR/wai-aria-1.1/#complementary
https://www.w3.org/TR/wai-aria-1.2/#complementary

Allowing `aria-expanded` on `role=complementary` in ARIA 1.1 was a spec bug so these aren't good pass examples:
https://github.com/w3c/aria/issues/681

Jym77:
Clean up presentational conflict examples and use an element/attribute combination which is allowed in ARIA 1.2…

Closes issue(s):

- closes #1370 

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -132,18 +132,10 @@ The `aria-label` [state][] is [global][] and thus [inherited][] for all [semanti
 
 #### Passed Example 10
 
-This `aside` element has an [explicit role][] of `none`. However, the [global][] [property][] `aria-label` is specified. Thus it has a [semantic role][] of `complementary` due to [Presentational Roles Conflict Resolution][]. The `aria-expanded` [state][] is [inherited][] for the `complementary` role.
+This `button` element has an [explicit role][] of `none`. However, it is [focusable][] (by default). Thus it has a [semantic role][] of `button` due to [Presentational Roles Conflict Resolution][]. The `aria-pressed` [state][] is [supported][] for the `button` role.
```
WilcoFiers, line 135: I think the relevance of a conflict resolution test to this rule is a little questionable. Definitely not as part of the passed examples I'd say. That seems too strict to me. I think it's better to take the role off of this.

Jym77: 🤔 
1. Without the `role="none"` this becomes the same as Passed Example 1 and should as well be removed.
2. If all Presentation conflicts examples are removed, then we end up with a weird Accessibility Support warning that it differs from UA to UA, that ties to nothing in the rule. Yet the Accessibility Support is true and needed, users of different UA/AT **will** experiment different contents.
3. I do not see how to put a Failed Example that rely on the conflict for that case. Something like `<button role="none" aria-sort="">` is not good because it fails whether the conflict is handle (role of `button`) or not (role of `none`). Given that any `aria-*` attribute failing the rule on a non-`none` role will also fail it on a `none` role, I don't see how to make a Failed Example with the conflict.

Jym77:

PR #1376, opened 2020-07-07
fix: limit publish workflow to run against master branch

jkodu:
The publish workflow, does a checkout of the master branch and applies the new website changes that needs to be updated.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,9 +6,6 @@ name: publish
 
 on:
   repository_dispatch:
-  pull_request:
-    branches:
-      - develop
 
 jobs:
   publish:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,17 +6,22 @@ name: publish
 
 on:
   repository_dispatch:
+  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -12,11 +12,13 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: 'master'
       - uses: actions/setup-node@v1
       - run: |
           git config user.name "${{ secrets.USER_NAME }}"
           git config user.email "${{ secrets.USER_EMAIL }}"
-          git clone -b master "https://${{ secrets.USER_PAT }}@github.com/act-rules/act-rules-web"
+          git clone -b master "https://82fc6b9d6b970ad0ace26f922dcb90dba83477bf@github.com/act-rules/act-rules-web"
```
Jym77, line 21: Is that a secret that is now included in the URL?

jkodu: @Jym what happens when a secret is pushed by mistake is that GitHub revokes the token.
So I regenerated a new one after. All good.

PR #1377, opened 2020-07-07
chore: update publish github workflow

jkodu:
Update publish workflow to run only against `master` branch


jkodu:
Given the nature of this change. Merging without approvals.

PR #1380, opened 2020-07-08
page-has-title: improve example description

WilcoFiers:
Change suggested by AWK on a survey from AGWG.

Need for Final Call: none

### DIFF
--- a/_rules/html-page-title-2779a5.md
+++ b/_rules/html-page-title-2779a5.md
@@ -155,7 +155,7 @@ This page has a `title` element that is empty.
 
 #### Failed Example 3
 
-This page does not have a `title` element for the whole page.
+This page does not have a `title` element. The `title` element in the `iframe` does not function as the title for the entire page.
 
 ```html
 <html>
--- a/_rules/html-page-title-2779a5.md
+++ b/_rules/html-page-title-2779a5.md
@@ -155,7 +155,7 @@ This page has a `title` element that is empty.
 
 #### Failed Example 3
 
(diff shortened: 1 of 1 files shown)


_rules/html-page-title-2779a5.md
```diff
@@ -155,7 +155,7 @@ This page has a `title` element that is empty.
 
 #### Failed Example 3
 
-This page does not have a `title` element for the whole page.
+This page does not have a `title` element. The `title` element in the `iframe` does not function as the title for the entire page.
```
ajanec01, line 158: Shouldn't the `iframe` element contain the `title` element to match the description?

WilcoFiers: No, the page inside the iframe does not need to have a title. As far as I know there is nothing in WCAG that requires it.

ajanec01: Sorry, I was having a brain dead moment when writing the initial comment! The idea is to change the description slightly to say that the `title` element is derived from the source document of the `iframe` element. At the moment in kind of reads as there is a `title` element nested inside the `iframe` element. Feel free to leave it.

Jym77: I think I agree wit @ajanec01 
```suggestion
This page does not have a `title` element. The `title` element in the content of the `iframe` does not function as the title for the entire page.
```
might be more explicit that we are not referring to a `<iframe title="…">` but to the content of the test assets file.

PR #1381, opened 2020-07-08
image-has-acc-name: applicability should not use the acc tree

WilcoFiers:
This feedback came from the ACT Task Force

The applicability of this rule was inconsistent with several of its test cases. I think the test cases here are right. The test cases show the rule is inapplicable when aria-hidden was used, but be applicable when the role is none or presentation (implicit or explicit).

Need for Final Call: 1 week


_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -26,6 +26,8 @@ input_aspects:
   - CSS Styling
 acknowledgments:
   authors:
+    - Wilco Fiers
+  previous_authors:
```
WilcoFiers, line 30: I thought this was necessary because of the applicability change, would like a yay or nay on that.

cpandhi: Yay!

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -36,7 +38,11 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
+The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, for which all the following is true about it and all its [ancestors][] in the [flat tree][]:
```
carlosapaduarte, line 41:
```suggestion
The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, for which all the following is true about them and all their [ancestors][] in the [flat tree][]:
```

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -228,10 +234,35 @@ This element is neither an `img` element nor has a role of `img`.
 <div aria-label="W3C logo"></div>
```
ajanec01, line 234: Would it be worth adding the `background-image` property for consistency?

WilcoFiers: Yaknow what, I don't think this is a good example. Putting aria-label on a random `div` does very little. I'm going to take this out.

_rules/image-non-empty-accessible-name-23a2a8.md
Jym77: No real opinion on that.
Maybe we should have some clear guidelines on the topic at some point 😄

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -36,7 +38,11 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
+The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, for which all the following is true about them and all their [ancestors][] in the [flat tree][]:
+
+- the element does not have an `aria-hidden` [attribute value][] of "true"; and
+- the element does not have a [computed][] [display][] property of "none"; and
+- the element does not have a [computed][] [visibility][] property of "hidden".
```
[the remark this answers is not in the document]
WilcoFiers: Did it slightly different, but good idea. I don't think the example descriptions need to change though.

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -214,24 +216,36 @@ This element has a [semantic role][] of `img`, but is not [included in the acces
...
 ```html
-<div aria-label="W3C logo"></div>
+<div style="display: none">
+	<img src="/test-assets/shared/w3c-logo.png" />
+</div>
+```
+
+#### Inapplicable Example 5
+
+This `img` element is hidden with `visibility: hidden`.
```
Jym77, line 237:
```suggestion
This `img` element is hidden because its parent has `visibility: hidden`.
```

WilcoFiers: I think that's an unnecessary detail that is obvious from the example itself.

PR #1300, opened 2020-05-06
image-accessible-name: improve examples

WilcoFiers:
Made a couple of changes here:

- add actual srcs (thanks @dd8, I borrowed some code from #1208)
- remove a bad passing example
- update the example descriptions

Need for Final Call: 1 week

PR #1382, opened 2020-07-08
Image non empty acc name - revised PR

dd8:
The PR replaces https://github.com/act-rules/act-rules.github.io/pull/1208

Reason for new PR: rule has been renamed, and many changes from original PR were merged in https://github.com/act-rules/act-rules.github.io/pull/1300

This PR:

- restricts the applicability to elements that render an image (to make `` and `<div style='background-image:none'>` inapplicable because they have no visual rendering, and are not voiced)
- adds `` as an inapplicable example
- adds additional failed examples like ``

Closes issue(s):

- closes  issue described here https://github.com/act-rules/act-rules.github.io/issues/1124#issuecomment-578042900

Need for Final Call:
This will require a 2 week Final Call

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -36,7 +36,12 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
+The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img` which render an image and are [included in the accessibility tree][].
```
WilcoFiers, line 39:
```suggestion
The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img` which render an image and which are [included in the accessibility tree][].
```

WilcoFiers (dismissed):
This looks great. Just a small editorial suggestion.

_rules/image-non-empty-accessible-name-23a2a8.md
dd8: No problem - I'll give everyone a chance to comment then push an update

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -146,6 +151,33 @@ This `img` element inside a `div` positioned off screen has an [implicit role][]
...
+#### Passed Example 11
+
+This `picture` element displays a `srcset` and has an [accessible name][] because of the `alt` attribute.
+
+```html
+<picture>
+    <source srcset="/test-assets/shared/w3c-logo.png" />
+    <img alt="W3C logo" width="72" height="48" />
+</picture>
+```
```
Jym77, line 179: Technically, the rule is applicable to the `img` element, not to the `picture` one.
It doesn't really change anything as the `picture` element is just here to give the `source` to the `img`, but description should be about the `img` element.

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -188,6 +220,36 @@ This `img` element has an [explicit role][] of `none`. However, it is [focusable
 <img role="none" tabindex="0" src="/test-assets/shared/w3c-logo.png" />
 ```
 
+#### Failed Example 6
+
+This `img` element displays a CSS `background-image`, has an empty [accessible name][] and an [implicit role][] of `img` because it is missing 
+an empty `alt` attribute.
```
Jym77, line 226: We don't cut lines in the markdown part…
```suggestion
This `img` element displays a CSS `background-image`, has an empty [accessible name][] and an [implicit role][] of `img` because it is missing an empty `alt` attribute.
```

Jym77, line 226: I am not sure that "because it is missing an empty `alt` attribute" is really needed in the description (maybe it is instead missing an `aria-label` attribute to give it an accessible name).
```suggestion
This `img` element displays a CSS `background-image`, has an empty [accessible name][] and an [implicit role][] of `img`.
```

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -188,6 +220,36 @@ This `img` element has an [explicit role][] of `none`. However, it is [focusable
...
+
+```html
+<img style="width:72px; height:48px; background-image: url(/test-assets/shared/w3c-logo.png)" />
+```
+
+#### Failed Example 7
+
+This `img` element displays a `srcset`, has an empty [accessible name][] and an [implicit role][] of `img` because it is missing 
+an empty `alt` attribute.
```
Jym77, line 235:
```suggestion
This `img` element displays a `srcset`, has an empty [accessible name][] and an [implicit role][] of `img`.
```

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -188,6 +220,36 @@ This `img` element has an [explicit role][] of `none`. However, it is [focusable
...
+
+```html
+<img srcset="/test-assets/shared/w3c-logo.png" />
+```
+
+#### Failed Example 8
+
+This `picture` element displays a `srcset`, has an empty [accessible name][] and an [implicit role][] of `img` because it is missing 
+an empty `alt` attribute.
```
Jym77, line 244:
```suggestion
This `picture` element displays a `srcset`, has an empty [accessible name][] and an [implicit role][] of `img`.
```

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -36,7 +36,12 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
+The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img` which render an image and are [included in the accessibility tree][].
+
+HTML elements render an image if any of the following are true:
+
+- The element is an `img` element where the [current request][]'s [state][image request state] is [completely available][].
+- The element has a CSS `background-image` that does not [draw nothing](https://www.w3.org/TR/css-backgrounds-3/#the-background-image).
```
Jym77, line 44: Maybe it is worth to extract this into a separate definition? Maybe not…

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -228,10 +290,22 @@ This element is neither an `img` element nor has a role of `img`.
...
 ```
 
+#### Inapplicable Example 5
+
+This `img` element has no `src` attribute so does not render an image.
+
+```html
+<img />
+```
+
```
Jym77, line 300: I think we need Inapplicable Examples for the other cases of not rendering an image:
* image not fully loaded (see https://act-rules.github.io/rules/e88epe#inapplicable-example-10) OTOH, we start having a few of these springled in various rules, so maybe we don't want to duplicate too much of the test (@WilcoFiers WDYT?)
* CSS `background-image` that doesn't draw anything  (I don't even know how to do that…)

Jym77 (dismissed):
Looks pretty good. Some minor comments.

_rules/image-non-empty-accessible-name-23a2a8.md
dd8: I've changed both `picture` examples to say "This `img` element displays"

_rules/image-non-empty-accessible-name-23a2a8.md
dd8: If it needs reused anywhere then it should be a separate definition, but how about keeping it inline until then

_rules/image-non-empty-accessible-name-23a2a8.md
dd8: I've added 2 new Inapplicable Examples. I've also referenced the HTML definition of 'represents nothing' in the examples without `alt` and `src` since I remember one of the 508 people F2F saying they thought this should fail.

dd8:
I think that's all the requested changes addressed. I think it's important to reference the HTML [represents nothing](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element:attr-img-src-9) definition somewhere in the rule, but I'm not convinced doing it from the examples is the right place.

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -228,10 +287,38 @@ This element is neither an `img` element nor has a role of `img`.
 <div aria-label="W3C logo"></div>
 ```
 
+#### Inapplicable Example 5
+
+This `img` element has no `src` attribute and no `alt` attribute so does not render an image and [represents nothing][].
+
+```html
+<img />
```
WilcoFiers, line 295: Looking at this with fresh eyes. I don't think this is correct. These actually should fail SC 1.1.1. Here's my reasoning:

1. These img elements are content, because they are nodes in the accessibility tree
2. There is no text here, so if 1 is true, these quality as non-text content
3. Since these are invisible, the "Decoration, Formatting, Invisible" applies
4. Since these aren't marked up with `aria-hidden=true`,`role=none` or `alt=""`, assistive technologies will not consistently ignore them
5. Because these are not ignored by assistive technologies, the exception isn't met, and so 1.1.1 is not satisfied


I think I'd prefer to keep this rule as is to be honest. That ARIA in HTML explicitly says that an `img` with no `alt` attribute has a semantic role of `img` suggests to me this proposal not how the W3C expects this to be interpreted, even if some AT will ignore some of these (depending on the verbosity setting). It might be better just to mention this in the accessibility support section.

WilcoFiers (dismissed):

_rules/image-non-empty-accessible-name-23a2a8.md
dd8: Looking at the `` example (an image with no `src` and no `alt`). This isn't rendered on screen in Chrome/FF/Safari. It cannot render non-text content since it has no `src`.

It is exposed as `role=img` in Chrome/Firefox. The image does not appear in the accessibility tree in Safari.

There are two contradictory specs:

- html-map maps all `img` elements without `alt` to `role=img` (even if they have no src)
- the HTML spec says an `img` with no `alt` and no `src` represents nothing (i.e. has no semantics)

The html-aam mapping is here:
https://www.w3.org/TR/html-aam-1.0/#el-img

The HTML spec says:

> If the [src](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-src) attribute is not set and either the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute is set to the empty string or the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute is not set at all
The element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) nothing.

https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element

The HTML spec defines "represents nothing" as meaning the element has no semantics.

Note sure what do do here:

- you can make the argument `` **should not** fail 1.1.1 since there's no non-text content
- you can also make the argument `` **should** fail 1.1.1 since it's exposed as an role=img by html-aam

Jym77: I think we should raise the issue to the HTML AAM 🤔 
To see if `img` with no source should still have an image role or not.

dd8: Issue filed at https://github.com/w3c/html-aam/issues/439

Jym77: @WilcoFiers With [ARIA in HTML](https://www.w3.org/TR/html-aria/#el-img-no-name) saying that images with no name also have no role, does that change your opinion on this point?
(somehow, [HTML AAM](https://w3c.github.io/html-aam/#el-img) hasn't followed suit yet)

Jym77: As per https://github.com/w3c/aria/pull/2221 closing that issue, ARIA is moving toward giving these `img` elements with no/empty `src` an implicit role of `none`.

@WilcoFiers Does that solve your concerns about this PR?

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -146,6 +151,33 @@ This `img` element inside a `div` positioned off screen has an [implicit role][]
 </div>
 ```
 
+#### Passed Example 9
+
+This `img` element displays a CSS `background-image` and has an [accessible name][] because of the `alt` attribute.
```
zlayaAvocado, line 156: Does it need a separate example with a `background: url(./w3c.png)`? Or does this assume that the computed css value if we set a `background` will be `background-image`?

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -228,10 +287,38 @@ This element is neither an `img` element nor has a role of `img`.
 <div aria-label="W3C logo"></div>
 ```
 
+#### Inapplicable Example 5
+
+This `img` element has no `src` attribute and no `alt` attribute so does not render an image and [represents nothing][].
```
zlayaAvocado, line 292:
```suggestion
This `img` element has no `src` attribute and no `alt` attribute or accessible name so does not render an image and [represents nothing][].
```

just clarifying it from https://github.com/w3c/aria/pull/2221/files#diff-5c3c8936e81538898c83ae11f5271e9ae9324925713a3b03aca58cbd2a3c3014R3133

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -228,10 +287,38 @@ This element is neither an `img` element nor has a role of `img`.
...
+```
+
+#### Inapplicable Example 7
+
+This `img` element has an `src` attribute which will cause the [image request state][] to be [Broken](https://html.spec.whatwg.org/#img-error).
+
+```html
+<img src="/test-assets/does-not-exist.png" />
+```
+
```
zlayaAvocado, line 313: @WilcoFiers can you weight on this? It doesn't seem like it's in the exception of WCAG 1.1.1 and I am not sure if it's inapplicable. 
1. It will still render and will be mapped to the image role since `src` is not empty
2. Some screenreaders might announce a file name instead if there is a `src` attribute (even if it's not loading) but no `alt` provided, which would not be the text alternative for the image

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -36,7 +36,12 @@ htmlHintIgnore:
 
 ## Applicability
 
-The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img`, except for elements that are not [included in the accessibility tree][].
+The rule applies to HTML `img` elements and HTML elements with the [semantic role][] of `img` which render an image and which are [included in the accessibility tree][].
+
+HTML elements render an image if any of the following are true:
+
+- The element is an `img` element where the [current request][]'s [state][image request state] is [completely available][].
+- The element has a CSS `background-image` that does not [draw nothing](https://www.w3.org/TR/css-backgrounds-3/#the-background-image).
```
zlayaAvocado, line 44: In the "draws nothing" description it's mentioned about the background size: "An image that is empty (zero width or zero height) likewise counts as a layer but draws nothing"

Would the following example be also non applicable then, or only the image with the original size of 0 width and height?
```
   <img
      style="
        width: 72px;
        height: 48px;
        background-image: url('./w3c.png'); // image size is 50px by 50px
        background-size: 0; // change background size to 0
      "
    />
```

zlayaAvocado (changes requested):

_rules/image-non-empty-accessible-name-23a2a8.md
WilcoFiers: Interesting problem. After re-reading WCAG definitions a couple times I think that yes, this fails 1.1.1. I think this element falls under non-text content. Content isn't defined in WCAG, but I'd say since this is both in the accessibility tree, and visible to people, I can't see a reason to say this wouldn't be content. It's non-text content because everything that isn't text content is non-text content. 

SC 1.1.1 applies to all non-text content, we can debate whether this is decorative or not. But in either case not having an alt does fail WCAG. So I agree with you, this should not be inapplicable.

PR #1384, opened 2020-07-09
Update rule to address assistive technology inconsistencies

### LINKED ISSUE #1367: iframe has non-empty name - problems with examples and applicability [cae760]

jeeyyy:
Addressed issues identified, detailed in below issue.

Closes issue(s):

- closes #1367

Need for Final Call:
- 2 weeks

### DIFF
--- a/_rules/iframe-non-empty-accessible-name-cae760.md
+++ b/_rules/iframe-non-empty-accessible-name-cae760.md
@@ -31,7 +31,7 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Assumptions
 
-If an `iframe` is not perceived by the user as a single control, it does not qualify as a [user interface component][] under WCAG 2. In such a scenario, failing this rule would not fail [success criterion 4.1.2](https://www.w3.org/TR/WCAG21/#name-role-value). Unless the `iframe` is both removed from the accessibility tree and removed from [sequential focus navigation][], they usually are considered to be [user interface components][user interface component].
+If an `iframe` is not perceived by the user as a single control, it does not qualify as a [user interface component][] under WCAG 2. In such a scenario, failing this rule would not fail [success criterion 4.1.2](https://www.w3.org/TR/WCAG21/#name-role-value). Unless the `iframe` is removed from the accessibility tree they usually are considered to be [user interface components][user interface component].
 
 ## Accessibility Support
 
@@ -112,6 +112,14 @@ This `iframe` element has an empty (`""`) [accessible name][] because the `title
 
 ```html
 <iframe title=" " src="/test-assets/SC4-1-2-frame-doc.html" role="none"> </iframe>
+```
+
+#### Failed Example 5
+
+This `iframe` element is taken out of [sequential focus navigation][] using `tabindex`.
+
+```html
+<iframe tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
 ### Inapplicable
--- a/_rules/iframe-non-empty-accessible-name-cae760.md
+++ b/_rules/iframe-non-empty-accessible-name-cae760.md
@@ -38,10 +38,10 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
 - Some browsers include `iframe` elements in the [sequential focus navigation][]. This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard. When an `iframe` is removed from the accessibility tree, this rule is still applicable for those browsers, unless the `iframe` is explicitly removed from [sequential focus navigation][] (by having the `tabindex` attribute set to a negative value).
 
 - Browser and assistive technology support for `iframe` elements is currently **inconsistent**. Some examples of inconsistencies include, but are not limited to:
-  - When assistive technoligies are being set up to ignore the `title` attribute, which means that to some users the `title` attribute will not act as an [accessible name][].
-  - Some assistive technologies ignores `aria-label` and only announces `title` attribute as an [accessible name][].
+  - Some assistive technologies can be set up to exclude the `title` attribute from the [accessible name][] of `iframe` elements
+  - Some assistive technologies ignores `aria-label` and only announces `title` attribute as an [accessible name][] of `iframe` elements.
   - Some assistive technologies ignore empty `iframe` elements, regardless of if they are [focusable][] or if they have an [accessible name][].
-  - Some assistive technologies ignore non empty `iframe` elements that are both [focusable][] and have an [accessible name][].
+  - Some assistive technologies always create an empty tab-stop for the `iframe`, even if they have an [accessible name][].
 
 ## Background
 
--- a/_rules/iframe-non-empty-accessible-name-cae760.md
+++ b/_rules/iframe-non-empty-accessible-name-cae760.md
@@ -21,7 +21,7 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to `iframe` elements that are [included in the accessibility tree][].
+The rule applies to `iframe` elements.
 
 **Note:** `frame` element is deprecated, this rule does not consider `frame` or `frameset` elements.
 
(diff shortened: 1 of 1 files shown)


_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -131,17 +132,11 @@ This `iframe` is not [included in the accessibility tree][] because of setting a
 <iframe style="display:none;" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
 ```
 
-#### Inapplicable Example 3
-
-This `iframe` element has a negative `tabindex` and therefore is not included in the [sequential focus navigation][].
```
jkodu, line 136: @dd8 per your suggestion, instead of moving this to failed example, I removed it, as it was similar to failed example 1 if moved to failed category. Thanks.

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,10 +37,11 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
 
-- Browser and assistive technology support for `iframe` elements is currently **inconsistent**. Some examples of inconsistencies include (but are not limited to):
-  - Assistive technologies being set up to ignore the `title` attribute, which means that to some users the `title` attribute will not act as an [accessible name][],
-  - There is a known combination of a popular browser and assistive technology that ignores `aria-label` and only announces `title` attribute as an [accessible name][]
-  - Some assistive technologies ignore empty `iframe` elements, regardless of if they are focusable or if they have an accessible name.
+- Browser and assistive technology support for `iframe` elements is currently **inconsistent**. Some examples of inconsistencies include, but are not limited to:
+  - When assistive technoligies are being set up to ignore the `title` attribute, which means that to some users the `title` attribute will not act as an [accessible name][].
```
WilcoFiers, line 41:
```suggestion
  - Some assistive technologies can be set up to exclude the `title` attribute from the [accessible name][] of `iframe` elements
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,10 +37,11 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
+  - Some assistive technologies ignores `aria-label` and only announces `title` attribute as an [accessible name][].
```
WilcoFiers, line 42:
```suggestion
  - Some assistive technologies ignores `aria-label` and only announces `title` attribute as an [accessible name][] of `iframe` elements.
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,10 +37,11 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
+  - Some assistive technologies ignore empty `iframe` elements, regardless of if they are [focusable][] or if they have an [accessible name][].
+  - Some assistive technologies ignore non empty `iframe` elements that are both [focusable][] and have an [accessible name][].
```
WilcoFiers, line 44:
```suggestion
  - Some assistive technologies always create an empty tab-stop for the `iframe`, even if they have an [accessible name][].
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -21,7 +21,7 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to `iframe` elements that are [included in the accessibility tree][] and that can be accessed by [sequential focus navigation][].
+The rule applies to `iframe` elements that are [included in the accessibility tree][].
```
WilcoFiers, line 24: The assumption is out of date and needs to be updated too.

_rules/iframe-non-empty-accessible-name-cae760.md
dd8: Yep, I think it should be added as a fail, otherwise implementations that only test sequential focus navigation will show as correct and complete for all the tests.

_rules/iframe-non-empty-accessible-name-cae760.md
WilcoFiers, line 24: I'm not convinced that "included in the accessibility tree" covers it here. In trying this out, the following iframe still shows up in the accessibility tree for Chrome and Firefox:

```html
<iframe role="none" tabindex="-1"></iframe>
```

Even if I put aria-hidden="true" on it, Chrome still puts it in the accessibility tree (Firefox does not). Common ways of taking things out of the accessibility tree don't seem to work for iframes. I think that's something the rule should account for.

The other thing I noticed is that despite that iframes with role="none" are included in the accessibility tree, Voiceover treats them completely differently. They're not treated as regions to enter into, but instead treated as part of the regular content. That makes a lot of sense, but does away with this idea that whether or not it is included in the accessibility tree matters. 

That example above... despite the fact that very unintuitively it is included in the accessibility tree, doesn't seem like it actually needs an accessible name. VO skips it because of role=none, NVDA skips it because of tabindex="-1". I'm not sure what JAWS does with it. Definitely something worth testing.

Either way I think we should explore doing this rule without "included in the accessibility tree". Accessibility trees seem to have unique rules for iframes, which as far as I can tell aren't spec'ed out. Instead of hoping someone does enough testing to figure it out, we should probably define what we found in this rule so that everyone can do this consistently.

jeeyyy: To add to above findings, for above example:

NVDA/Firefox - has iframe, can get to it and reads "about:blank".
JAWS/IE11 - does not see the iframe

jeeyyy: This argument is very valid, given the disparity in how `iframe` is interpreted by various AT's. 

Given this context, I have updated this rule to not use `included in accessibility tree`, which is the step in the right direction.

Added a generic accessibility note, but open to expanding on that if necessary.

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -21,7 +21,7 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to `iframe` elements that are [included in the accessibility tree][] and that can be accessed by [sequential focus navigation][].
+The rule applies to `iframe` elements.
```
WilcoFiers, line 24: Why would `display:none` need to be tested for this?

WilcoFiers (changes requested):

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -31,16 +31,18 @@ Each target element has an [accessible name][] that is not empty (`""`).
...
 
-- Browser and assistive technology support for `iframe` elements is currently **inconsistent**. Some examples of inconsistencies include (but are not limited to):
-  - Assistive technologies being set up to ignore the `title` attribute, which means that to some users the `title` attribute will not act as an [accessible name][],
-  - There is a known combination of a popular browser and assistive technology that ignores `aria-label` and only announces `title` attribute as an [accessible name][]
-  - Some assistive technologies ignore empty `iframe` elements, regardless of if they are focusable or if they have an accessible name.
+- Browser and assistive technology support for `iframe` elements is currently **inconsistent**. Some examples of inconsistencies include, but are not limited to:
+  - Some assistive technologies can be set up to exclude the `title` attribute from the [accessible name][] of `iframe` elements
+  - Some assistive technologies ignores `aria-label` and only announces `title` attribute as an [accessible name][] of `iframe` elements.
```
daniel-montalvo, line 42:
```suggestion
  - Some assistive technologies ignore `aria-label` and only announce `title` attribute as an [accessible name][] of `iframe` elements.
```

daniel-montalvo (changes requested):

jeeyyy:
@WilcoFiers as discussed are you taking over this one?

jeeyyy:
Closing this as @WilcoFiers is taking it over.

PR #1393, opened 2020-07-16
Cleaning up rule c249d5 "Device motion based changes to the content can be disabled"

carlosapaduarte:
- Removing note that repeated what is already stated in an assumption
- Changing the time intervals to match what is in the definition of "changes in content"

Need for Final Call:

---

## How to Review And Approve


_rules/device-motion-disabled-c249d5.md
```diff
@@ -32,10 +32,8 @@ The rule applies to an [HTML document][] with an associated [Window object][] th
...
-
-**Note:** The 10 minutes time span is an arbitrary limit which is not included in WCAG. Results that happen after this period will not fail this rule but may nonetheless fail [Success Criterion 2.5.4: Motion Actuation][sc2.5.4]. The accessibility problem tends to be less severe for longer time periods, and without a time limit, testing this rule consistently would be impractical.
+- **no changes:** The registered event does not cause [changes to the content][changes in content] of the [web page][] within a 1 minutes time span of the [event firing][event firing]; or
+- **disabled:** There is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] of the registered event or can be found in a [clearly labeled location][] from that [web page][]. The set of [instruments][instrument] can be used to prevent [changes to the content][changes in content] of the [web page][] resulting from the event within a 1 minutes time span of the [event firing][].
```
Jym77, line 36: Not completely related to this PR, but this item could benefit from https://github.com/act-rules/act-rules.github.io/blob/dca29f7ac2cbde4023d389448b483872a18e40ae/pages/glossary/set-of-clearly-labelled-instruments.md and https://github.com/act-rules/act-rules.github.io/blob/dca29f7ac2cbde4023d389448b483872a18e40ae/pages/glossary/block-event.md

Using "block event" might also help to solve #1394 without getting an awful text:
"There is on [set of clearly labelled instruments][] to [block the event][] for at least YYY minutes after activating the instruments"
or something in that line.

carlosapaduarte: I completely agree with the proposed changes.
I guess we can wait for those definitions to be merged and then update the "device motion" rules.

Jym77: Sounds like a plan.

daniel-montalvo (approved):

_rules/device-motion-disabled-c249d5.md
Jym77: We can use the "new" definitions for these notions.
```suggestion
- **disabled:** There is at least one [set of clearly labelled instruments][] to [block the event][blocked event][] for at least 1 minute.
```
(plus adding the refs)

Jym77 (changes requested):

Jym77 (approved):

carlosapaduarte:
Merging editorial changes

PR #1396, opened 2020-07-17
New rule: Content triggered on hover is hoverable (ep1s13)

carlosapaduarte:
Rule partially checking SC 1.4.13 (the hoverable part of SC 1.4.13)

Blocked: waiting for decision on how to handle states in rules

---

## How to Review And Approve


carlosapaduarte:
Ready for review. Suggestions for more examples are welcome!

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,107 @@
...
+There are no [visible changes in content][] while the target element is [hovered][] or the mouse pointer remains inside a rectangle defined by the following coordinates:
+
+- (**top**:) the lowest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**left**:) the lowest value of the x coordinate of any pixel affected by the [visible changes in content][]; and
+- (**bottom**:) the highest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**right**:) the highest value of the x coordinate of any pixel affected by the [visible changes in content][].
+
+## Assumptions
+
+- The user does not dismiss the tooltip causing the changes in content by pressing a key on the keyboard. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
```
adilsofficial, line 37: comment on "... the tooltip causing the changes in content" changes caused by hover does not change the original content which is already presented on a page but displays additional content instead.

adilsofficial (changes requested):
Good work. There is a difference between "visible changes of content" and "triggered additional content/content triggered on hover" which we should consider. The example is presented in the comment of the expectation section.

pages/glossary/visible-changes-in-content.md
```diff
@@ -0,0 +1,22 @@
...
+  - DOM tree
+---
+
+A _[event][] originated change in the content_ of a [web page][] occurs when, by comparing the [web page][] before and after the event [firing][] the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling.
+
+- If the [web page][] is rendering time-based media, rendered pixels and audio will be changing as part of the playback. The comparison in this instance should compare the pixels and audio that are rendered if the event is not fired, with the ones that are rendered if the event is fired.
+
+**Assumptions:**
+
+- This definition assumes that there are no changes in the content of the [web page][] caused by another [event][]. If this is not the case, changes may be attributed to the wrong event.
```
kasperisager, line 17: This is a real issue that I would be a little wary of assuming doesn't occur. Especially in the context of automated testing, correlating events to content changes is extremely tricky as you'll inevitably have to deal with interference from non-event based content changes

kasperisager (changes requested):

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,107 @@
+---
+id: ep1s13
+name: Additional content triggered on hover is hoverable
+rule_type: atomic
+description: |
+  This rule checks that any additional content that was triggered by hovering an element remains available to the user while the mouse pointer is over that content
```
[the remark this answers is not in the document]
carlosapaduarte: Do you mean that the rule should check if it is possible to move the pointer from the trigger to content, without the content disappearing? 
The SC itself states "This condition generally implies that the additional content overlaps or is positioned adjacent to the target." Would a similar assumption be good for you?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,107 @@
...
+  - CSS Styling
+  - Mouse pointer
+acknowledgments:
+  authors:
+    - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any element that when [hovered][] causes [visible changes in content][].
```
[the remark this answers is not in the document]
carlosapaduarte: But the definition of "visible changes in content" specifies that these changes are not specific to the element. They are any changes to the web page rendered pixels. If by hovering the target a pop-up appears, then there will be changes in the rendered pixels. And the definition does not state these changes are permanent, and neither to the structure of the page (the change might just be making visible an element that was hidden). 
Are you suggesting that we repeat here what is in the definition?
Or would replacing "visible changes in content" by "visible changes in the content of the page" be enough?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Expectation
+
+There are no [visible changes in content][] while the target element is [hovered][] or the mouse pointer remains inside a rectangle defined by the following coordinates:
```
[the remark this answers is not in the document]
carlosapaduarte: I think I finally understand your point! 
Would changing the name (and description in accordance) of the rule from "Additional content triggered on hover is hoverable" to "Content that becomes visible on hover is hoverable"?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: Once again: the definition of visible changes in content refers to changes in the pixel of the page (irrespectively of which elements are added or removed or of having their properties changed).
I went with pixels being changed because that is something objective and avoids the "confusion" that comes from trying to identify the origin of the change.

pages/glossary/visible-changes-in-content.md
carlosapaduarte: I agree, but I couldn't come up with a way that would unequivocally identify the changes that are triggered by an event when there are multiple events causing changes. Would love to get a suggestion if you have one :-)

And in the scope of WAI-Tools were saying that this is a manual rule, so I assume that this assumption was easier to write in this context...

carlosapaduarte:
@adilsofficial and @kasperisager I've answered your comments, but I haven't made any changes to the rule. I'm looking forward to your replies, so I can decide how to proceed.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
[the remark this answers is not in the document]
adilsofficial: Yes, I guess that would be enough to mention "... content of the page" but specify that cause visible changes in content in the form of additional content? 

I'm a bit concern that do we consider changes such as "a change in the size or color of the content" caused by hover as a visible change in this rule?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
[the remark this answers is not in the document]
adilsofficial: I guess the name is according to description but my concern is with something else :-D, as there are three intents in this SC that are "Dismissible", "Hoverable", "Persistent". This rule name and description are more relevant to "Hoverable" (that is the triggered content is hoverable without disappearing if the user moves their pointer off the trigger, in short, it should be possible for the user to move the pointer to triggered content without disappearing).  However, the applicability and expectation are relevant to "Persistent" (that is once triggered content appears, should remain visible until the user removes hover or focus from the trigger and additional content, user dismiss the additional content or the information conveyed by the additional content becomes invalid) which actually you described in expectation.

In addition, you referred F95 in the background section which also states that "the pointer can be moved over the new content without the additional content disappearing" which is again related to the "hoverable" intent of the SC.

My concern is that we should be specified about which aspect of WCAG SC this rule is about by both name, description, applicability, and expectation.

I hope you understand my concern now.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
adilsofficial: Agree.

adilsofficial (dismissed):
One more round @carlosapaduarte, good work 👍

pages/glossary/visible-changes-in-content.md
```diff
@@ -0,0 +1,22 @@
...
+title: Visible changes in content
+key: visible-changes-in-content
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+A _[event][] originated change in the content_ of a [web page][] occurs when, by comparing the [web page][] before and after the event [firing][] the rendered pixels change in any part of the document that is currently within the [viewport][] or that can be brought into the [viewport][] via scrolling.
```
kasperisager, line 11: In extension of https://github.com/act-rules/act-rules.github.io/pull/1396/files#r471423985, I'd argue that the wording "before and after" isn't objective. As you mention the rule being optimised for manual implementations, "before and after" will be different from the perspective of a human observer than it will from a software observer. Whereas the software observer will be able to determine visual changes caused by the firing of an event on the frame render immediately following the event, a human observer will only be able to determine visual changes caused by the firing of an event after _several_ frame renders. The definition of course seems perfectly unambiguous; both of these observations happen _after_ the firing of the event. However, they're relative to the observer and therefore not objective.

kasperisager (dismissed):

pages/glossary/visible-changes-in-content.md
kasperisager: I've experimented with correlating events and mutation changes using frame renders as a heuristic in the past, and it worked reasonably well. That did lead me to this realisation though: https://github.com/act-rules/act-rules.github.io/pull/1396#pullrequestreview-492408701

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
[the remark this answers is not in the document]
carlosapaduarte: I'll update the description to clearly state that it applies to the element and the additional content. 
And I'll also include the assumption about being possible to move the pointer to the additional content

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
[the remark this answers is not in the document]
carlosapaduarte: The rule only addresses the "Hoverable" part of the SC. Checking the parts related to "Dismissible" and "Persistent" seemed to me too complex to address now.

I don't really know how to make that more explicit. The rule name (only) uses hoverable. In the description, (only) hovering is used. In the applicability and expectation (only) hovered is used.

The accessibility requirements section does not state that the SC is met if the rule passes.

Do you have any suggestions on how to improve the writing?

pages/glossary/visible-changes-in-content.md
carlosapaduarte: @kasperisager I've been trying to find a definition for "frame renders" and I haven't been able. 

Another possibility I came up is to say there is a change if the event originates at least one reflow or repaint of the document. Problems:
- I also can't find a definition of reflow or repaint in the specs because this is terminology for how the user agents render a web page
- I'm not sure how an evaluator would be able to establish a direct relationship between the event and the reflow of repaint happening after the event was triggered 

Any more thoughts on this?

kasperisager: I think that'd be the right direction to take. You might be able to use the concept of "animation frame callback" for this: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#animation-frames. I'm thinking something along the lines of "on or after the next run of animation frame callbacks".

carlosapaduarte: I was also looking at that, but I'm not convinced that all reflows or repaints imply an "animation frame callback". The definition begins with "Some objects include the AnimationFrameProvider interface mixin." I couldn't find anywhere specifying which object are those. I'll try to dig a bit deeper.

kasperisager: The processing model for the event loop does seem to imply that animation frame callbacks are run whenever the rendering is updated: https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model:run-the-animation-frame-callbacks-3

carlosapaduarte:
@kasperisager After spending several hours going through the event loop, I've redone the "visible changes in content" in a way that removes all assumptions. Care to have a look? :-)

kasperisager:
Neat, I think we're getting closer! You're likely better equipped to answer this than me: Why are we only interested in events that result in animation frame callbacks? That is, event handlers may also modify the document structure directly and those changes would then become visible on the next update of the rendering, circumventing an animation frame callback. Modern frameworks do generally tend to to batch rendering operations in animation frame callbacks to maintain smooth rendering, though that's likely not the general rule.

carlosapaduarte:
I doubt I'm better equipped since only yesterday I started looking at this seriously!
I was probably misled by step 4 of the update rendering procedure where the unnecessary rendering Documents are decided. The only objectively described condition in there relates to animation frame callbacks. Of course, there is also the condition "the user agent believes that updating the rendering of the Doc would have no visible effect" which lets all other rendering sources through. I'll try to rewrite this is a less targeted manner later today.

carlosapaduarte:
@kasperisager I've given another shot at the definition, this time including also step 15 of the "update the rendering" algorithm, thus, I expect, covering all rendering updates that can change the content of page (and ignoring rendering updates that do not change the content - like resize, scrolling, etc.)
Looking forward to your feedback

kasperisager (dismissed):
That looks solid, kudos! 💯

pages/glossary/visible-changes-in-content.md
```diff
@@ -0,0 +1,27 @@
...
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+An [event][] originates _visible changes in the content_ of a [web page][] when all the following are true
+
+- the [event][] [queues a task][queuing] in the [task source][] of an [event loop][] of a [Document][] of the [web page][]; and
+- the [processing][] of that [task][] causes an [update to the rendering][update the rendering] of the [Document][]; and
```
WilcoFiers, line 14: What does it mean that a task causes an update? If the task sets a timeout that eventually leads to an update, does it cause that update, or is this only about changes in that occur in the same tick? 

I don't think this is all that testable. How would anyone know this is true? Are you expecting someone to step through the debugger to figure this out?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,108 @@
...
```
WilcoFiers, line 28: This would cause hover animations to fail.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,108 @@
...
+- (**top**:) the lowest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**left**:) the lowest value of the x coordinate of any pixel affected by the [visible changes in content][]; and
+- (**bottom**:) the highest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**right**:) the highest value of the x coordinate of any pixel affected by the [visible changes in content][].
+
+## Assumptions
+
+- The user does not dismiss the tooltip causing the changes in content by pressing a key on the keyboard. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
+- The content displayed on the rectangle with the changes in content does not become irrelevant. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
+- The element and the rectangle defined by the additional content are adjacent, so that it is possible to move the mouse pointer from one to the other without causing the additional content to disappear.
```
WilcoFiers, line 39: I think this is a problematic assumption. I'd be more comfortable if we made this part of the applicability.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,108 @@
...
+## Expectation
+
+There are no [visible changes in content][] while the target element is [hovered][] or the mouse pointer remains inside a rectangle defined by the following coordinates:
+
+- (**top**:) the lowest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**left**:) the lowest value of the x coordinate of any pixel affected by the [visible changes in content][]; and
+- (**bottom**:) the highest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**right**:) the highest value of the x coordinate of any pixel affected by the [visible changes in content][].
+
+## Assumptions
```
WilcoFiers, line 35: You're missing an important assumption; that any visible change triggers additional content to become visible. I don't think that's a safe assumption to make.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,108 @@
...
+The rule applies to any element that when [hovered][] causes [visible changes in content][].
+
+## Expectation
+
+There are no [visible changes in content][] while the target element is [hovered][] or the mouse pointer remains inside a rectangle defined by the following coordinates:
+
+- (**top**:) the lowest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**left**:) the lowest value of the x coordinate of any pixel affected by the [visible changes in content][]; and
+- (**bottom**:) the highest value of the y coordinate of any pixel affected by the [visible changes in content][]; and
+- (**right**:) the highest value of the x coordinate of any pixel affected by the [visible changes in content][].
```
WilcoFiers, line 33: 1. Which visible changes? I think you need to specify that this is about the content change from the applicability.
2. What about popups that aren't rectangular?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,108 @@
...
+This button element causes [visible changes in content][] by presenting a tooltip when [hovered][]. The tooltip remains displayed, resulting in no further [visible changes in content][] while the mouse pointer does not leave its boundaries or the button's boundaries.
+
+```html
+<link rel="stylesheet" type="text/css" href="/test-assets/ep1s13/styles.css" />
+<script src="/test-assets/ep1s13/scripts.js"></script>
+
+<body onload="bindEvents({tooltipRemains: true})">
+	<div class="tooltip-container">
+		<button aria-labelledby="tooltip">
+			<span aria-hidden>WCAG</span>
```
WilcoFiers, line 65: This seems unnecessary.

```suggestion
			<span>WCAG</span>
```

WilcoFiers (dismissed):

pages/glossary/visible-changes-in-content.md
carlosapaduarte: That task is in the queue of the event loop of the document, being managed by the browser. If the task was created by a timeout that was set an hour ago is irrelevant for this definition. The timeout is the event that places the task in the queue. Which means that the hover event was not the one that placed the (rendering change) task on the queue.

The definition evolved from pixels changed in the viewport caused by some event that I can't be certain of (and thus would require an assumption that no other event was causing the change - which was challenged by Kasper), to this version of the definition so that it is testable.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: Indeed. I don't think we can avoid that unless we stop looking at visible changes to the content. I'll add an assumption.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: Good suggestion. Will do.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: I'm not sure I understand what you mean. If it's about the "additional" part, there is no requirement about it. The visible changes in content definition just requires that the rendered pixels change. That can be because new content is added, but also because existing content is changed.

If this was not what you were talking about, can you rephrase it?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: 1. As part of the fix to https://github.com/act-rules/act-rules.github.io/pull/1396#discussion_r509163422 I've moved this to a new definition. Check to see if this is still an issue?

2. When writing this I've tested with popups that are not visually rectangular (with rounded corners, for instance) but I always found the hoverable area of the popup to be the rectangle defined in this way.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: The button is labelled by the tooltip content. That's why is the span is aria-hidden.

carlosapaduarte:
@kasperisager With the changes requested I've changed the applicability (to got rid of one assumption). Check to see if you're okay with that.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -21,22 +21,17 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to any element that when [hovered][] causes [visible changes in content][].
+The rule applies to any element that when [hovered][] causes [visible changes in content][] and the [area affected by the changes][area affected by visible changes] is adjacent or overlaps the target content.
```
WilcoFiers, line 24: Sorry, but what does "adjacent or overlaps" mean? I think we need to define that in terms of bounding box positions.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
WilcoFiers: Your definition just means "visible change", not just in content, but anything really. If the color of a link on hover changes, this definition way make that a "visible change in content", while I'd argue it isn't a change in content at all. The problem is that animations on hover will fail this, because they constantly cause a visible change.

WilcoFiers (dismissed):

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: When writing it, I was wondering if I should do that. Guess that I should have. Will do it now.

carlosapaduarte: I'll take the opportunity to add another inapplicable example illustrating this new condition.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: What you're saying is correct, and it was done on purpose. I couldn't come up with a more effective (that would also be objective) way to detect changes in content. I don't know how we can address the problem of animations without using an assumption (which I had already included). 

But I do agree that this rule being applicable when hovering a link causes the link to change color or becoming underlined is too much. And that should not be taken care of with an assumption. So I've updated the applicability to exclude instances where the changes only happen inside the bounding box of the target element. In those instances it's impossible to have the pointer outside the element when it's over the tooltip anyway!

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -31,6 +34,7 @@ There are no [visible changes in content][] while the target element is [hovered
 
 - The user does not dismiss the tooltip causing the changes in content by pressing a key on the keyboard. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
 - The content displayed on the rectangle with the changes in content does not become irrelevant. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
+- The changes in content resulting from the hovering are
```
WilcoFiers, line 37: Unfinished sentence here.

WilcoFiers (dismissed):
Getting close.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: Ups... forgot to delete that

kasperisager (dismissed):

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
ShadowBB: I agree with Wilco that it is unnecessary. The tooltip overrides the accessible name completely so there is no need to use aria-hidden to avoid duplication.

ShadowBB (dismissed):
I see a large problem with sub-sub menu's.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: OK. Taking it out.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,140 @@
...
+## Applicability
+
+The rule applies to any element that when [hovered][] causes [visible changes in content][] and for the [area affected by the changes][area affected by visible changes] all the following are true:
+
+- the area is [adjacent][adjacent areas] or [overlaps][overlapping areas] the [bounding box][] of the target element; and
+- the area is not [contained][] inside the [bounding box][] of the target element.
+
+## Expectation
+
+There are no [visible changes in content][] while the target element is [hovered][] or the mouse pointer remains inside the [area affected by the changes][area affected by visible changes].
```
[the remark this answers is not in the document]
carlosapaduarte: This was a good catch. See if you agree with the revised expectation and new example.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -28,13 +28,14 @@ The rule applies to any element that when [hovered][] causes [visible changes in
 
 ## Expectation
 
-There are no [visible changes in content][] while the target element is [hovered][] or the mouse pointer remains inside the [area affected by the changes][area affected by visible changes].
+There are no [visible changes in content][] in the [area affected by the changes][area affected by visible changes] while the target element is [hovered][] or the mouse pointer remains inside that area.
```
ShadowBB, line 31: I thought about this solution too. But a visual indication on the menuitem that a submenu underneath that menuitem has been opened is rather common in these situations right?

Is you have 

Contact us V

If you hover over that you get

Contact us >
Via phone V
Via email V
Via fax V

And then you hover over email and you get 

Contact us >
- Via phone V
- Via email >
- - Our main branche
- - Our secondary brance
- - Our tertiary brance
- Via fax V

This would still fail this rule right? Because the area is definitely effected, even if the sub-sub menu doesn't open in the middle of the sub menu but off to the side, the little arrow denoting that "email" has been opened is still changed pixels right?

Can we write this rule so content is not allowed to disappear? Only changed or added to?

ShadowBB (dismissed):
I don't think this fully fixes it.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: Yes, that example would fail this rule (that is why I've added the last assumption when I made this change).

I couldn't find a way to make this rule work with distinctions about modifying content (disappearing, changing, adding). The problem is how you objectively identify that. This is why "visible changes in content" considers changes in rendered pixels and not changes of the content or properties of elements.

For example: if to change what is rendered on a page you delete node X that is a child of node Y, then create node Z and append it to node Y, from the user perspective you changed content, but from the DOM's perspective you removed and added content. In this example, content disappeared, which would fail your formulation. 

Would the existing assumption work for you? Or do you find it too strong?

pages/glossary/visible-changes-in-content.md
WilcoFiers: As I understand it, the task queue resolves all tasks as soon as possible. If I use a timeout, it doesn't go into the task queue. Depending on the JS engine, there might be a task to check if any timeouts need to be queued up, but the actual timeout handler doesn't go to the task queue. I don't think what you're saying is correct.

More importantly, the event loop and task queue are not included in the DOM tree. This definition is testing thing that aren't part of the input of the rule. That's not allowed in ACT rules. Can you explain to me how someone would go about testing this? In what testing environment does the browser give you a way to inspect the task queue?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,161 @@
...
+## Applicability
+
+The rule applies to any element that when [hovered][] causes [visible changes in content][] and for the [area affected by the changes][area affected by visible changes] all the following are true:
+
+- the area is [adjacent][adjacent areas] or [overlaps][overlapping areas] the [bounding box][] of the target element; and
+- the area is not [contained][] inside the [bounding box][] of the target element.
+
+## Expectation
+
+There are no [visible changes in content][] in the [area affected by the changes][area affected by visible changes] while the target element is [hovered][] or the mouse pointer remains inside that area.
```
WilcoFiers, line 31: I think this needs some kind of time component to ensure this is tested consistently. If I check this for a 5 seconds, and someone else tests it for a minute, we may get different results.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Assumptions
+
+- The user does not dismiss the tooltip causing the changes in content by pressing a key on the keyboard. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
+- The content displayed on the rectangle with the changes in content does not become irrelevant. In this instance the rule will fail while [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
+- The additional content displayed on hover does not include animations. An animation would cause visible changes to the content of the page and therefore fail this rule. However, if the animation is displayed only while the mouse pointer hovers above the target element or the rectangle with the changes, [success criterion 1.4.13: Content on Hover or Focus][sc1.4.13] might be satisfied.
```
WilcoFiers, line 37: I don't think this is a reasonable assumption. Animations, especially fade-in and face-out animations are common place on the web.

WilcoFiers (dismissed):

pages/glossary/visible-changes-in-content.md
carlosapaduarte: You're correct. I don't think we can include "Browser rendering engine" (or any other engine) as an input aspect. 

Which means we can't use this. And therefore we're back to the initial problem of identifying what changes result from the hovering event. This was initially was handled with an assumption. But @kasperisager pointed out it was too strong of an assumption too make.

So, it's back to the drawing board... and I can't clearly see a way to (objectively) tie the event to the visible changes on a page, so if you have any suggestions they're welcome

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: Fade-out animations are irrelevant to this rule. Would a delay before checking the expectation (to give time for the fade-in animation to complete) be acceptable to you?

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
carlosapaduarte: I don't have nothing against it.

carlosapaduarte:
Summary of major changes for this revision:

- Definition of visible changes to content goes back to changes in pixels of the page (addresses https://github.com/act-rules/act-rules.github.io/pull/1396#discussion_r530323158)
- Added a condition to the applicability that the area affected by the changes must always be same every time the user hovers the element while the user does interact with other elements in the page. By asking for this consistent behaviour the causal connection between the content appearing and the hovering of the element should be established.
- Added a 5 second period before assessing changes in applicability to account for fade-in animations (addresses https://github.com/act-rules/act-rules.github.io/pull/1396#discussion_r530324906)
- Added a 1 minute period for checking if the additional content does not vanish as expectation 2 (addresses https://github.com/act-rules/act-rules.github.io/pull/1396#discussion_r530324226)

@WilcoFiers have these addressed your concerns?
@kasperisager is this enough for you to ensure the connection between hovering event and content appearing on the page?
@ShadowBB still waiting for your reply to https://github.com/act-rules/act-rules.github.io/pull/1396#discussion_r527113813

pages/glossary/adjacent-areas.md
```diff
@@ -0,0 +1,20 @@
...
+title: Adjacent areas
+key: adjacent-areas
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+Web page areas A and B are _adjacent areas_ if at least one pixel from area A is at the distance of 1 pixel from one pixel in area B.
```
WilcoFiers, line 11: I'm not sold on using "pixels" in these area definitions. What does it mean for an area to have pixels? Perhaps "point" is more appropriate for these definitions.

_rules/additional-content-triggered-on-hover-is-hoverable-ep1s13.md
```diff
@@ -0,0 +1,170 @@
...
+  - CSS Styling
+  - Mouse pointer
+acknowledgments:
+  authors:
+    - Carlos Duarte
+---
+
+## Applicability
+
+The rule applies to any element that when [hovered][] causes [visible changes in content][] and for the [area affected by the visible changes][area affected by visible changes], determined after 5 seconds of [hovering][hovered] the element, all the following are true:
```
WilcoFiers, line 24: I think this shows the problem with this rule; this SC doesn't apply to presentational changes, just to changes in content, whereas the definition you are using includes presentational changes.

```suggestion
The rule applies to any element that when [hovered][] causes [visible changes in presentation or content][] and for the [area affected by the visible changes][area affected by visible changes], determined after 5 seconds of [hovering][hovered] the element, all the following are true:
```

WilcoFiers (dismissed):
I would like to have a call about this rule. I don't think we know how to test that something is a change of content, and I don't think we can get away with just assuming every visual change is a change of content either. I feel like we need to do some brainstorming to figure out how to address this.

carlosapaduarte:
@WilcoFiers @kasperisager @ShadowBB 
This one is ready for another round of reviews. 
The applicability and expectations changed substantially from the previous version. Now it tries to identify content that became visible as a result of the hovering (new nodes plus nodes that change visibility) and that become invisible when the hovering ends. Expectations are that the content is adjacent to the hovered element (so that the use can move the pointer there without dismissing it) and that it does not go away for a testable amount of time (1 minute) while the pointer is at the center of the new content (once again, to ensure consistency while testing).
Please, have another go at it!

pages/glossary/overlap.md
```diff
@@ -0,0 +1,15 @@
...
+title: Overlap
+key: overlap
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+[Bounding boxes][] A and B _overlap_ if at least one point inside box A has the same [left][left coordinate] and [top coordinates][top coordinate] of one point inside box B.
```
WilcoFiers, line 11: Points don't have a left / to pcoordinate. I would suggest you define this as any corner of element A is between the left and right, and top and bottom edges of element B, or visa versa.

pages/glossary/center.md
```diff
@@ -0,0 +1,22 @@
...
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+The _center_ of a [bounding box][] is the point given by the following coordinates:
+
+- The [x coordinate][] is equal to the [left coordinate][] of the [bounding box][] plus half of the [width][] of the [bounding box][] (rounded down); and
+- The [y coordinate][] is equal to the [top coordinate][] of the [bounding box][] plus half of the [height][] of the [bounding box][] (rounded down).
```
WilcoFiers, line 14: I think this is enough:
```suggestion
- The [x coordinate][] is equal to the [left coordinate][] of the [bounding box][] plus half its [width][]; and
- The [y coordinate][] is equal to the [top coordinate][] of the [bounding box][] plus half its [height][].
```

pages/glossary/adjacent.md
```diff
@@ -0,0 +1,21 @@
...
+title: Adjacent
+key: adjacent
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+[Bounding boxes][] A and B are _adjacent_ if at least one point from box A is at the distance of 1 point from one point in box B.
```
WilcoFiers, line 11:
```suggestion
[Bounding boxes][] A and B are _adjacent_ if at least one point from box A is at the distance of 1 [CSS pixel[] from one point in box B.
```

WilcoFiers, line 11: Problem with this definition is that it also includes overlapping elements.

pages/glossary/adjacent.md
```diff
@@ -0,0 +1,21 @@
...
+
+The distance between two points is obtained by adding the horizontal distance between the two points with the vertical distance between the two points.
```
WilcoFiers, line 13: That doesn't sound right. So something with x:0,y0 has a distance of 2 to x:1,y:1? I don't think we should do that.

pages/glossary/content-that-becomes-visible.md
```diff
@@ -0,0 +1,25 @@
...
+title: Content that becomes visible
+key: content-that-becomes-visible
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+The _content that becomes visible_ is the [root][] element, if one exists, of the [tree][] that contains all elements that, as result of an [event][] [firing][], meet any of the following:
```
WilcoFiers, line 11: I'm not too sure what is going on here. How can a tree not have a root element? Do you maybe mean to find the closest common ancestor? How do I know that something is the "result of an event firing"? I don't think we should try to figure out if something is or isn't the result of an event. I think it's sufficient to determine that the content appears within half a second of firing the event, and that the content part of a region that updates without user input.

pages/glossary/content-that-becomes-visible.md
```diff
@@ -0,0 +1,25 @@
...
+
+- the element has [visible text content][] and is added to the [web page][] where the [event][] was [fired][firing]; or
+- the element [attributes][]' [values][] are modified in a way that cause some or all of its [text nodes][] to become [visible][].
```
WilcoFiers, line 14: Visible text content isn't bound to the closest ancestor. Any text nodes added to the page results in the HTML element having visible text content added.

How about something like:

```suggestion
has visible text content that did not exist on the page prior to the event firing.
```

WilcoFiers (dismissed):

pages/glossary/overlap.md
carlosapaduarte: Correct. Thanks for the suggestion.

pages/glossary/adjacent.md
carlosapaduarte: Fixed

pages/glossary/adjacent.md
carlosapaduarte: What don't you think we should do? Consider (0,0) and (1,1) to not be adjacent or that the distance between them should not be 2? Or both?

In my opinion they are not adjacent, but its just an opinion and I don't feel too strong about that.

But I do feel the distance is 2 pixels (I have to move 1 pixel right and 1 pixel down to get from one point to the other). Pixels must be integer, therefore it cannot be the square root of 2.

pages/glossary/content-that-becomes-visible.md
carlosapaduarte: The "if one exists" was meant to apply to the tree, not the root element. My bad.

And I'll update to use the half a second instead of resulting from. That's definitely more testable!

pages/glossary/content-that-becomes-visible.md
carlosapaduarte: We don't need to worry about the bounding because this tree only contains new elements (therefore elements that were there already are not considered) and elements that have had their attributes modified (therefore if their text nodes changed without a change in their attributes, they are not considered)

carlosapaduarte:
@WilcoFiers this is ready for another round

pages/glossary/content-that-becomes-visible.md
```diff
@@ -0,0 +1,25 @@
...
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+The _content that becomes visible_ is the [root][] element of the [tree][], if a [tree][] exists, that contains all elements that, after half a second of an [event][] [firing][], meet any of the following:
+
+- the element did not exist on the [web page][] prior to the [event][] [firing][] and has [visible text content][]; or
```
WilcoFiers, line 13: I think there is a fundamental problem in this. If I have an event rerenders the page, all elements are new, even if nothing changed. `body.innerHTML = body.innerHTML` will do that. Modern frameworks try to avoid unnecessary rerenders, but there is a lot of code on the web that rerenders components, irrespective of if something actually changed.

I think we need to define this in a way that works with rerenders. I took a stab at doing that just for text nodes, here's what I've come up with. That isn't enough, but I hope it gets helps move this forward.


```md
A text node is considered to be _visible after an event_ if both of the following of the following are true:

- A tick before the event, there was no [visible][] text node with the same node value; and
- 500 milliseconds after the [event][] [firing][], there the text node is [visible][].
```

WilcoFiers (dismissed):

pages/glossary/adjacent.md
carlosapaduarte: I was trying to come up with a distance metric that would always result in whole pixels. But, you're right, for distance we don't need to require that. I'll delete the definition of distance from this definition.

pages/glossary/center.md
```diff
@@ -10,8 +10,8 @@ input_aspects:
 
 The _center_ of a [bounding box][] is the point given by the following coordinates:
 
-- The [x coordinate][] is equal to the [left coordinate][] of the [bounding box][] plus half of the [width][] of the [bounding box][] (rounded down); and
-- The [y coordinate][] is equal to the [top coordinate][] of the [bounding box][] plus half of the [height][] of the [bounding box][] (rounded down).
+- The [x coordinate][] is equal to the [left coordinate][] of the [bounding box][] plus half its [width][] (rounded down); and
+- The [y coordinate][] is equal to the [top coordinate][] of the [bounding box][] plus half its [height][] (rounded down).
```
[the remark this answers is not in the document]
carlosapaduarte: I was ensuring once again that coordinates were whole numbers. But they don't need to be (they're doubles).

pages/glossary/content-that-becomes-visible.md
carlosapaduarte: I don't think that works because it requires that there are no text nodes with the same node value. If the new content includes a text node with the same value of an already present text node, then it will miss the new content.
I'll try to think of a solution for this...

carlosapaduarte: Following your suggestion, check if you think that what I came up with works

pages/glossary/content-that-becomes-visible.md
```diff
@@ -0,0 +1,21 @@
...
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+The _content that becomes visible_ after an [event][] [firing][] is any element for which all of the following is true:
+
+- 500 milliseconds after the [event][] [firing][] the element has [visible text content][]; and
```
WilcoFiers, line 13: We need to capture an assumption that content on hover will appear within 500ms.

pages/glossary/content-that-becomes-visible.md
```diff
@@ -0,0 +1,21 @@
...
+- a tick before the [event][] [firing][] there was no element with the same [descendant text content][] that the element has after the [event][] [firing][]; and
```
WilcoFiers, line 14: Seems to me like this would apply to elements that update their content too. If I have a button that changes its text on hover, it 1. has visible content after the event. and 2. does not have the same descendant text content. Additionally, descendant text content does not take shadow DOM into account.

pages/glossary/adjacent.md
```diff
@@ -0,0 +1,15 @@
...
+title: Adjacent
+key: adjacent
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+[Bounding boxes][] A and B are _adjacent_ if they do not [overlap][] and at least one point from box A is at the Euclidean distance of 1 [CSS pixel][] from one point in box B.
```
WilcoFiers, line 11: I'm a little nervous about this being exactly 1px with no margin of error. I think we should include anything less than 2, and greater than 0.

WilcoFiers (changes requested):

kasperisager (approved):

PR #1399, opened 2020-07-22
Prepare for TF "Object element has non-empty accessible name" (8fc3b6)

carlosapaduarte:
Preparing "Object element has non-empty accessible name" for TF Survey.

Only minor editorial changes have been made and 1 note moved from the expectation to the background.

Need for Final Call:

---

## How to Review And Approve


carlosapaduarte:
Editorial changes. No final call needed. Merging.

PR #1406, opened 2020-08-03
Put definition examples in separate pages

WilcoFiers:
Closes issue(s): I swear there is one somewhere, but I can't for the life of me find it.

Need for Final Call: 1 week

### DIFF
--- a/pages/glossary/included-in-the-accessibility-tree.md
+++ b/pages/glossary/included-in-the-accessibility-tree.md
@@ -11,45 +11,6 @@ Elements included in the accessibility tree of platform specific accessibility A
 
 The general rules for when elements are included in the accessibility tree are defined in the [core accessibility API mappings](https://www.w3.org/TR/core-aam/). For native markup languages, such as HTML and SVG, additional rules for when elements are included in the accessibility tree can be found in the [HTML accessibility API mappings (working draft)](https://www.w3.org/TR/html-aam/) and the [SVG accessibility API mappings (working draft)](https://www.w3.org/TR/svg-aam/).
 
+For more details, see [examples of included in the accessibility tree](/pages/examples/included-in-the-accessibility-tree/).
+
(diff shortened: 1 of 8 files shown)


pages/glossary/included-in-the-accessibility-tree.md
```diff
@@ -11,45 +11,6 @@ Elements included in the accessibility tree of platform specific accessibility A
...
-<span id="label" style="display:none">ACT rules</span>
-<span aria-labelledby="label">Accessibility Conformance Testing rules</span>
-```
-
-Although this `input` element is not included in the accessibility tree, it is still [focusable](#focusable), hence users of assistive technologies can still interact with it by sequential keyboard navigation. This may result in confusing situations for such users (and is in direct violation of [the fourth rule of ARIA (working draft)](https://www.w3.org/TR/using-aria/#fourth)).
-
-```html
-<input type="text" aria-hidden="true" name="fname" />
-```
+**Note:** Users of assistive technologies might still be able to interact with elements that are not included in the accessibility tree. An example of this is a [focusable](#focusable) element with an `aria-hidden` attribute with a value of `true`. Such an element could still be interacted with using sequential keyboard navigation regardless of the assistive technologies used, even though the element would not be included in the accessibility tree.
```
Jym77, line 16:
```suggestion
**Note:** Users of assistive technologies might still be able to interact with elements that are not included in the accessibility tree. An example of this is a [focusable](#focusable) element with an `aria-hidden` attribute with a value of `true`. Such an element could still be interacted using sequential keyboard navigation regardless of the assistive technologies used, even though the element would not be included in the accessibility tree.
```

Jym77, line 16: Plus, technically, "focusable" does not imply "in sequential keyboard navigation order", so the text is not fully technically correct, but I say it's OK to keep it that way in this context (example in a note).

PR #1408, opened 2020-08-06
remove techniques from text-contrast acc mapping

WilcoFiers:
Failing this rule does not necessarily fail this technique.

Need for Final Call: 1 week

### DIFF
--- a/_rules/text-contrast-afw4f7.md
+++ b/_rules/text-contrast-afw4f7.md
@@ -15,16 +15,6 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
-  wcag-technique:G18: # Ensuring that a contrast ratio of at least 4.5:1 exists between text (and images of text) and background behind the text
-    forConformance: false
-    failed: not satisfied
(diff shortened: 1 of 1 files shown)


WilcoFiers:
The reason I'm taking them out is because one of them is about text needing a 4.5:1 contrast, and the other about 3:1. They both address half the requirement, but not the entire thing. They don't explicitly say which technique applies to what font weight / size, but that is part of the issue. I don't think these have a clear enough scope to justify mapping this rule to those techniques. They're both listed in the background, I think that is the most appropriate place for them.

Jym77:
> The reason I'm taking them out is because one of them is about text needing a 4.5:1 contrast, and the other about 3:1. They both address half the requirement, but not the entire thing. They don't explicitly say which technique applies to what font weight / size, but that is part of the issue. I don't think these have a clear enough scope to justify mapping this rule to those techniques. They're both listed in the background, I think that is the most appropriate place for them.

Ah, yes, I see.
Failing the rule on small text will indeed not fail G145 since G145 doesn't care about small text…

daniel-montalvo (approved):

PR #1411, opened 2020-08-10
Rename "form control" to "form field"

WilcoFiers:
In order to clarify that this rule does not apply to buttons, the rule no longer uses the phrase "form control" instead it uses "form field" to make it clear this rule does not apply to buttons.

This change was suggested during a call with the ACT TF

Need for Final Call: 1 week


_rules/form-field-non-empty-accessible-name-e086e5.md
```diff
@@ -46,6 +46,8 @@ The list of roles in the applicability is derived by taking all the roles from [
 - have [semantic roles][] that inherit from the [abstract](https://www.w3.org/TR/wai-aria/#abstract_roles) `input` or `select` role, and
 - do not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
 
+Note that this rule does not test certain other control-like roles such as `button` and `menuitem`, because these do not in herit from `input` or `select`. These should be tested separately.
```
Jym77, line 49:
```suggestion
Note that this rule does not test other control-like roles such as `button` and `menuitem`, because these do not inherit from `input` or `select`. These should be tested separately.
```

PR #1412, opened 2020-08-10
Rule: Text spacing via CSS is above minimum value or not set to `!important` (24afc2, 78fd32, 9e45ec)

jeeyyy:
Given the discussions on this PR, a decision was made to split the rule to 3, one for each style that affects text spacing. This PR proposes 3 rules as requested.

> Note: Not focussing on failing tests etc at the moment, before we validate if this is the right approach for this rule.

Closes issue(s):
- NA

Need for Final Call:
- 2 weeks


_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,160 @@
...
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared][][letter-spacing][] CSS property.
```
Jym77, line 23:
```suggestion
This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,160 @@
...
+## Expectation 1
+
+For the test target's [letter-spacing][] property, one of the following is true:
+
+- the [declared][] value specified via the `style` attribute does not have the [important flag][]; or
+- the [computed][] value is at least `0.12` times its [computed][] [font-size][].
+
+## Assumptions
+
+If there is a mechanism available on the page by which letter spacing can be adjusted, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```
Jym77, line 34:
```suggestion
There is no mechanism available on the page to adjust letter spacing. If there is such a mechanism, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```
The assumption is the first sentence. The "If" sentence is a fact that is always true, not an assumption.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,160 @@
...
+
+	<body>
+		The toy brought back fond memories of being lost in the rain forest.
+	</body>
+</html>
+```
+
+#### Inapplicable Example 4
+
+This `p` element has a [letter-spacing][] of `24px` specified via the style attribute, but does not have the [important flag][].
```
Jym77, line 134: The lack of `!important` is not relevant for the Applicability anymore. This should become a Passed Example instead.

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,160 @@
...
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared][][line-height][] CSS property.
```
Jym77, line 23:
```suggestion
This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared] the [line-height][] CSS property.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,160 @@
...
+## Expectation 1
+
+For the test target's [line-height][] property, one of the following is true:
+
+- the [declared][] value specified via the `style` attribute does not have the [important flag][]; or
+- the [computed][] value is at least `1.5` times its [computed][] [font-size][].
+
+## Assumptions
+
+If there is a mechanism available on the page by which line height can be adjusted, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```
Jym77, line 34:
```suggestion
There is no mechanism available on the page to adjust line height. If there is such a mechanism, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```
The assumption is the first sentence. The "If" sentence is a fact that is always true, not an assumption.

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,160 @@
...
+
+	<body>
+		The toy brought back fond memories of being lost in the rain forest.
+	</body>
+</html>
+```
+
+#### Inapplicable Example 4
+
+This `p` element has a [line-height][] of `24px` specified via the style attribute, but does not have the [important flag][].
```
Jym77, line 134: No `!important` is now Passing, not Inapplicable.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,159 @@
...
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared][][word-spacing][] CSS property.
```
Jym77, line 23:
```suggestion
This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared] the [word-spacing][] CSS property.
```

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,159 @@
...
+## Expectation 1
+
+For the test target's [word-spacing][] property, one of the following is true:
+
+- the [declared][] value specified via the `style` attribute does not have the [important flag][]; or
+- the [computed][] value is at least `0.16` times its [computed][] [font-size][].
+
+## Assumptions
+
+If there is a mechanism available on the page by which word spacing can be adjusted, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```
Jym77, line 34:
```suggestion
There is no mechanism available on the page to adjust word spacing. If there is such a mechanism, failing this rule might not mean [success criterion 1.4.12 Text spacing](https://www.w3.org/TR/WCAG21/#text-spacing) is not satisfied.
```
The assumption is the first sentence. The "If" sentence is a fact that is always true, not an assumption.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,159 @@
...
+	</style>
+	<body>
+		The toy brought back fond memories of being lost in the rain forest.
+	</body>
+</html>
+```
+
+#### Inapplicable Example 4
+
+This `p` element has a [word-spacing][] of `12px` specified via the style attribute, but does not have the [important flag][].
```
Jym77, line 133: Now Passing.

_rules/letter-spacing-not-important-24afc2.md
EmmaJP: I agree. I was going to ask if an example with too small letter-spacing and without the !important flag should be included in the passed examples.

EmmaJP: Also, I'm not a fan of using fixed units in the examples, as that's not really good practice in regard accessibility. A `body { font-size: 1rem }` defaults as 16px in most browsers, and is possibly a better alternative for the examples.

_rules/line-height-not-important-78fd32.md
EmmaJP: Agree again. With similar thoughts on including a passed example with too small a line-height and no !important flag, and using fixed units in the examples.

_rules/word-spacing-not-important-9e45ec.md
EmmaJP: Agree again. With similar thoughts on including a passed example with too small a word-spacing and no !important flag, and using fixed units in the examples.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -67,17 +67,37 @@ This `div` element has a `letter-spacing` of `0.12em` which is equal to the reco
...
+	</body>
+</html>
+```
+
 ### Failed
 
 #### Failed Example 1
 
-This `div` element has a `letter-spacing` of `1.5px !important` (equals `0.09375em`) which is below to the recommended minimum, given the specified font size is 16 pixels.
+This `div` element has a `letter-spacing` of `1.5px !important` (equals `0.09375em`) which is below to the recommended minimum given the specified font size is `1em`.
```
Jym77, line 94: I guess the size of the `div` should also use relative units now that `font-size` does.

_rules/line-height-not-important-78fd32.md
```diff
@@ -71,18 +91,18 @@ This `p` element has a `line-height` of `32px` (equals `2em` ) which is above th
 
 #### Failed Example 1
 
-This `p` element has a `line-height` of `10px !important` (equals `0.625em` ) which is below the recommended minimum, given the specified font size of the body is 16 pixels.
+This `p` element has a `line-height` of `10px !important` (equals `0.625em` ) which is below the recommended minimum, given the specified font size of the body is `1em`.
```
Jym77, line 94:
```suggestion
This `p` element has a `line-height` of `1.875em !important` which is below the recommended minimum, given the specified font size of the body is `1em`.
```
Which is actually above the minimum of 1.5em 🙃

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -49,13 +49,13 @@ While some assistive technologies are able to set [user origin][] styles, others
 
 #### Passed Example 1
 
-This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em`) which is equal to the recommended minimum, given the specified font size is 16 pixels.
+This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12em`) which is equal to the recommended minimum, given the specified font size is `1em`.
```
Jym77, line 52: The `strong` should also have relative units now.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -67,6 +67,26 @@ This `strong` element has a `word-spacing` of `1.92pt !important` (equals `0.12e
 </html>
 ```
 
+#### Passed Example 2
+
+This `p` element has a [word-spacing][] of `12px` specified via the style attribute, which is above the recommended minimum given the specified font size is `1em`.
```
Jym77, line 72: The `p` should use relative units.

jeeyyy: I wanted to keep a few examples in non relative units. Is that ok?

_rules/letter-spacing-not-important-24afc2.md
Jym77: I'll let @EmmaJP decide on that since she was the one who requested relative units in the first place (https://github.com/act-rules/act-rules.github.io/pull/1412#pullrequestreview-466875191)

Jym77 (dismissed):

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,160 @@
...
+    - Jey Nandakumar
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+
+## Expectation 1
+
+For the test target's [letter-spacing][] property, one of the following is true:
```
WilcoFiers, line 27: This conditional list doesn't work well. You can see it if you try to declare what this list is about in each item, which is required in conditional lists:

```
- the [letter-spacing][] property has a [declared][] value specified via the `style` attribute does not have the [important flag][]; or
- the [letter-spacing][] property has a [computed][] value is at least `0.12` times its [computed][] [font-size][].
```

These two items are not about the same thing. One is about what's declared in the attribute, the other is about the computed property. I think the way this should be done is to write them out in a sentence, where the first thing is the requirement, and the second is an exception. Something like this:

```suggestion
For each test target, there is no [important flag][] on the [declared][] value of the [letter-spacing][] property in the `style` attribute, except if the [computed][] value is at least `0.12` times its [computed][] [font-size][].
```

WilcoFiers (changes requested):
Left one comment, but the same applies to the other two rules.

_rules/letter-spacing-not-important-24afc2.md
Jym77: I disagree. I think that "except" clauses read poorly and require more mental effort to understand than "or" clauses.
"A except if B" is the same logical construction as "A or B". Just written in a more convoluted way.

kasperisager: I must agree with @Jym77 on that. In fact, I've rewritten all such exceptions into opposite form for our internal versions of the rules as it makes things so much easier to digest.

_rules/word-spacing-not-important-9e45ec.md
EmmaJP: The suggestion to use relative units was to demonstrate good practice in our examples. If the font-size is using a relative unit, the word-spacing probably should as well.

Perhaps use fixed units for examples where the minimum requirements have been significantly exceeded.

_rules/letter-spacing-not-important-24afc2.md
EmmaJP: I think for this failed example it certainly demonstrates why those fixed units are a problem. I'm unsure if using different units for the letter-spacing and font-size muddies the waters or not. 

Certainly for an examples using the relative units, we should probably clarify that at default zoom 1em would be 16px.

_rules/letter-spacing-not-important-24afc2.md
EmmaJP: Agree the current 'or' version is easier to read, though undecided if that is because of the use of the 'or' or the use of bullets. That said, bullets make for easier reading and probably work better with an 'or'.

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,160 @@
...
+```html
+<html>
+	<style>
+		body {
+			font-size: 1em;
+		}
+	</style>
+
+	<body>
+		<p style="line-height: 2em; !important">
```
EmmaJP, line 63: Should this example not be 1.5em and equal to the recommended metric? Otherwise how does it differ from passed example 2 other than the unit used.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,159 @@
...
+```html
+<html>
+	<style>
+		body {
+			font-size: 1em;
+		}
+	</style>
+
+	<body>
+		<strong style="word-spacing: 0.12em !important;">
```
EmmaJP, line 63: Should this not be 0.16em to be equal to the recommended minimum?

_rules/word-spacing-not-important-9e45ec.md
EmmaJP: I think its nicer to have both units the same in each example.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,159 @@
...
+		</p>
+	</body>
+</html>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `strong` element has a `word-spacing` of `2.5px !important` (equals `0.15625em` ) which is below the recommended minimum, given the specified font size is 16 pixels.
```
EmmaJP, line 94: This description does not match the example

EmmaJP (changes requested):
Some tidying needed to ensure descriptions and examples match, in regard the sizes and units used. 

Also, I like having a relative unit example and a fixed unit example that exceeds the recommended requirements, though I do think it would be best to keep with one type of unit with in each example.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,159 @@
...
+```html
+<html>
+	<style>
+		body {
+			font-size: 16px;
+		}
+	</style>
+
+	<body>
+		<strong style="word-spacing: 8px !important;">
```
EmmaJP, line 105: Regarding earlier comment, I think this needs to change for the maths in the description to be correct...

```suggestion
		<strong style="word-spacing: 2.5px !important;">
```

_rules/letter-spacing-not-important-24afc2.md
WilcoFiers: That doesn't really change that this condition list doesn't include the subject; that it can't because it is about two different things.

WilcoFiers (dismissed):

_rules/letter-spacing-not-important-24afc2.md
EmmaJP: The subject is the test target's letter-spacing property, so both declared value and computed value can be relevant for that.

Am I missing something?

EmmaJP: Does this work...

For the test target, one of the following is true:

- the [declared][] value of the test target's [letter-spacing][] property, specified via the 'style' attribute, does not have the important flag; or
- the [computed][] value of the test target's [letter-spacing][] property is at least `0.12` times its [computed][] [font-size][].

carlosapaduarte: I agree that it is easier to read the 'or' version. And I was going to suggest exactly the same that @EmmaJP did

Jym77: If we absolutely want to repeat and frontload something:

> For the test target [letter-spacing][] property, one of the following is true:
> - the property has a [declared][] value specified via the `style` attribute without the important flag; or
> - the property has a [computed][] value of at least `0.12` times its [computed][] [font-size][].

But I don't think it reads really better. Notably because:
> For the test target [letter-spacing][] property, one of the following is true:
> - its [declared][] value specified via the `style` attribute does not have the important flag; or
> - its [computed][] value is at least `0.12` times its [computed][] [font-size][].

is more readable imho. Or @EmmaJP suggestion.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,160 @@
...
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+
+## Expectation
+
+For the test target, one of the following is true:
+
+- the [declared][] value of its [letter-spacing][] property, specified via the `style` attribute, does not have the [important flag][]; or
+- the [computed][] value of its [letter-spacing][] property is at least `0.12` times its [computed][] [font-size][].
```
WilcoFiers, line 30: I disagree with the way this is expressed, but as it seems most folks prefer this I'll just leave it for others to approve.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,160 @@
+---
+id: 24afc2
+name: Letter spacing in `style` attributes is not `!important`
+rule_type: atomic
+description: |
+  This rule checks that the letter spacing can be adjusted by the user, i.e. that `!important` is not applied to the style, unless the style has already exceeded the specified metric for retaining content visibility and functionality.
```
WilcoFiers, line 6: I'm not a fan of the way this is formulated. The first part is too broad a description of what this rule actually does. I think 
```suggestion
  This rule checks that the style attribute is not used to prevent increasing letter spacing by using `!important`, except if the 0.12 time the font size.
```

_rules/letter-spacing-not-important-24afc2.md
kasperisager: It's literally in the specification:

> Each property declaration applied to an element contributes a declared value for that property associated with the element. 

https://drafts.csswg.org/css-cascade/#declared

_rules/letter-spacing-not-important-24afc2.md
jeeyyy: I prefer keeping the description generic rather than putting out actual metric like 0.12 in the description. If others feel the same about this. I am willing to change.

Am I right in thinking that based on this [comment](https://github.com/act-rules/act-rules.github.io/pull/1412#discussion_r484744367), you are not looking to approve this rule?

WilcoFiers: The reason I think this should change is because the first part of it is so generic, it might as well be used on other rules. This doesn't set the rule apart from other rules that could be written for this SC.

nitedog: This description seems a little "chatty" to me. What is the problem with a more succinct rewording along the lines of what @WilcoFiers is proposing?

_rules/letter-spacing-not-important-24afc2.md
nitedog: @WilcoFiers how would you rephrase this, noting the CSS spec definitions that @kasperisager refers to above?

kasperisager: Here's an actual issue though: It's not clear how to handle `style` attributes where _multiple_ values are declared for the same property. Consider this:

```html
<p style="letter-spacing: 0.12; letter-spacing: 0.1;">
```

Such a `style` attribute doesn't have just _a_ declared value for the `letter-spacing` property, it has _several_. We want the last, valid declaration.

kasperisager (dismissed):
See https://github.com/act-rules/act-rules.github.io/pull/1412#discussion_r500314903.

_rules/letter-spacing-not-important-24afc2.md
jeeyyy: Valid point @kasperisager. I think simply referencing to [cascaded value][https://www.w3.org/TR/css-cascade/#cascaded] instead of declared value for the first point in the expectation should solve for this. I have made the change accordingly.

Jym77: 🤔 the cascaded value can come from somewhere else.
OTOH, not sure whether a the cascade can be won by something else when there is a declared value, so maybe this is indeed good…

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][], for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+
+## Expectation
+
+For the test target, one of the following is true:
+
+- the [cascaded][] value of its [letter-spacing][] property, specified via the `style` attribute, does not have the [important flag][]; or
```
kasperisager, line 29:
```suggestion
- the [cascaded][] value of its [letter-spacing][] property, if specified via the `style` attribute, does not have the [important flag][]; or
```

This applies to the other rules as well. Since the cascaded value may come from a declaration not in the `style` attribute, we're only interested in the important flag of the cascaded value _if_ it comes from a declaration in the `style` attribute.

kasperisager (changes requested):

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,161 @@
+---
+id: 24afc2
+name: Letter spacing in `style` attributes is not `!important`
+rule_type: atomic
+description: |
+	This rule checks that the `style` attribute is not used to prevent adjusting `letter- spacing` by using `!important`, except if it's at least `0.12` times the font size.
```
carlosapaduarte, line 6:
```suggestion
	This rule checks that the `style` attribute is not used to prevent adjusting `letter-spacing` by using `!important`, except if it's at least `0.12` times the font size.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,161 @@
...
+		<div style="letter-spacing: 0.12em !important">
+			The toy brought back fond memories of being lost in the rain forest.
+		</div>
+	</body>
+</html>
+```
+
+#### Passed Example 2
+
+This `p` element has a `letter-spacing` of `24px` specified via the style attribute, which is above to the recommended minimum given the specified font size is 16 pixels.
```
carlosapaduarte, line 72: Was it the goal to have a 24px letter spacing with a 16px font? The spacing is larger than the font :-) I'm not really sure that this is accessible to some people.
2px for letter spacing would be enough for passing the rule

_rules/letter-spacing-not-important-24afc2.md
carlosapaduarte: I don't think this is enough. 

```html
<html>
 	<style>
 		.noSpace {
 			letter-spacing: 0.01em !important;
 		}
 	</style>

 	<body>
 		<div class="noSpace" style="letter-spacing: 0.12em">
 		</div>
 	</body>
 </html>
```

We can't say anything about condition 1 because the cascaded value was not specified through the `style` attribute. Condition 2 is false. We need condition 1 to pass this example. I guess we need to use the declared value in condition 1 instead of the cascaded value.

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,161 @@
...
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a `line-height` of `32px` (equals `2em` ) which is above the recommended metric, given the specified font size is `1em`.
```
carlosapaduarte, line 52:
```suggestion
This `p` element has a `line-height` of `2em` which is above the recommended metric, given the specified font size is `1em`.
```

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,160 @@
+---
+id: 9e45ec
+name: Word spacing in `style` attributes is not `!important`
+rule_type: atomic
+description: |
+	This rule checks that the `style` attribute is not used to prevent adjusting `word-spacing` spacing by using `!important`, except if it's at least `0.16` times the font size.
```
carlosapaduarte, line 6:
```suggestion
	This rule checks that the `style` attribute is not used to prevent adjusting `word-spacing` by using `!important`, except if it's at least `0.16` times the font size.
```

carlosapaduarte (changes requested):

_rules/letter-spacing-not-important-24afc2.md
kasperisager: Good point!

> I guess we need to use the declared value in condition 1 instead of the cascaded value.

Then we're back to square 1 though 🙈 The issue is that there can be multiple declared values and we want the one that wins cascade, if any. Something like this might work to ensure totality:

> - the cascaded value of its `letter-spacing` property is either not specified through the `style` attribute or does not have the important flag set; [...]

carlosapaduarte: Yes, I believe that works 👍

Jym77:
> I'm also uncomfortable with using a fixed unit font size in the examples, as generally that isn't good practice for accessibility.

@EmmaJP I added a note to point out that this is not good in general but needed for some examples.

Jym77:
@carlosapaduarte @kasperisager @WilcoFiers 
Lovely people, I'm having big troubles with that one 🤯 

I went and did the changes needed to solve https://github.com/act-rules/act-rules.github.io/pull/1412#discussion_r500314903 and https://github.com/act-rules/act-rules.github.io/pull/1412#discussion_r508794215 and added related examples, so far so good.

But then, I realised:
1. `line-height` computed value can also be a number (easy fix), or `normal` (OK fix). This is not super pretty, but I believe I got it correctly. (plus it can be specified as a percentage or as a global keyword, but that's only a couple more examples)

2. The main problem.  Both `letter-spacing` and `word-spacing` values are **added** to a default value. That is `word-spacing: 0px` **does not remove all inter-words space**. And it is fairly hard to measure if that default value is enough or not. To make things "better", it seems that this default depends on the font and possibly on the browser (aka 💥). Also, it seems that when the value is `normal` it means "0 but you can change it to justify" (although I'm not seeing any difference between `0` and `normal`, so I'm not sure).

     It's a bit hard to measure but it seems that without setting any style (so `word-spacing` of `normal` which computes to 0) yield a word spacing of ~2mm for a font size of ~6mm, so a .33 ratio which is above the minimal value. So `word-spacing: 0 !important` should not fail the rule (depending on the font, and possibly on the browser…)

Anyway, 2 is a fairly annoying problem. I do not really see how to solve it (short from ditching 2 of the 3 rules). Any idea?

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,227 @@
+---
```
Jym77, line 1: Big problem with this rule. See https://github.com/act-rules/act-rules.github.io/pull/1412#issuecomment-733606665

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,203 @@
+---
```
Jym77, line 1: Big problem with this rule. See https://github.com/act-rules/act-rules.github.io/pull/1412#issuecomment-733606665

Jym77:
> 2. The main problem.  Both `letter-spacing` and `word-spacing` values are **added** to a default value. That is `word-spacing: 0px` **does not remove all inter-words space**. And it is fairly hard to measure if that default value is enough or not. To make things "better", it seems that this default depends on the font and possibly on the browser (aka 💥). Also, it seems that when the value is `normal` it means "0 but you can change it to justify" (although I'm not seeing any difference between `0` and `normal`, so I'm not sure).

Quick talk with Wilco. It seems that we're safe-ish considering that WCAG litterarilly means the `word-spacing` property, and not the resulting space between words (adding space size). Will add an assumption that way…

Jym77:
@carlosapaduarte @kasperisager @WilcoFiers this should be ready for another round.
Added many examples in the way, to illustrate the discussions we've had and the wording of the conditions we're ending up with…

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,322 @@
...
+
+This `p` element has a [computed][] `line-height` of `1.6` which is above the recommended minimum, thus it matches the **above minimum** condition.
+
+```html
+<p style="line-height: 1.6 !important">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 5
```
WilcoFiers, line 112: As above, don't think we need this one.

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,269 @@
...
+
+- This rule assumes that text inside the target is not justified. The actual space between words may be further adjusted to justify text. This results in the space being sometimes larger and sometimes smaller. On lines where inter-words space is _added_, this rule may fail while [Success Criterion 1.4.12 Text Spacing][sc1412] is still satisfied (because the added space tip it over the limit); on lines where inter-letters space is _removed_, failing this rule will fail [Success Criterion 1.4.12 Text Spacing][sc1412]. For elements rendered on multiple lines, it is safe to assume that justification will not always add space. For single line elements (for example, a `span` inside a justified `p`), it may end up in a line with added space. Note that justifying text is a failure of [Success Criterion 1.4.8 Visual Presentation][sc148]
```
WilcoFiers, line 40: From the look of it, justified text adjusts the space character, while still preserving word-spacing. This assumption might not be necessary.

WilcoFiers (changes requested):

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+</style>
+
+<p style="letter-spacing: 3px !important">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 3
+
+This `p` element has two [declared][] values for its `letter-spacing` property. The latest wins the [cascade sort][]. It has a value of `0.15em`, more than 0.12 times the font size and therefore matches the **above minimum** condition.
```
WilcoFiers, line 90:
```suggestion
This `p` element has two [declared][] values for its `letter-spacing` property. The latest wins the [cascade sort][]. It has a value of `0.15em`, more than 0.12 times the font size.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<p style="letter-spacing: 0.1em !important; letter-spacing: 0.15em !important">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 4
+
+This `p` element has two [declared][] values for its `letter-spacing` property. The one which is [important][] wins the [cascade sort][]. It has a value of `0.15em`, more than 0.12 times the font size and therefore matches the **above minimum** condition.
```
WilcoFiers, line 100:
```suggestion
This `p` element has two [declared][] values for its `letter-spacing` property. The one which is [important][] wins the [cascade sort][]. It has a value of `0.15em`, more than 0.12 times the font size.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<p style="letter-spacing: 0.15em !important; letter-spacing: 0.1em">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 5
+
+This `p` element has a not [important][] [computed][] `letter-spacing`, thus it matches the **not important** condition. Even though the value is too small, styles with [author origin][] declared by assistive technologies may win the [cascade sort][] and override it, thus this may satisfy [Success Criterion 1.4.12 Text Spacing][sc1412] and does not fail this rule. This is nonetheless bad practice and sufficient spacing should be used.
```
WilcoFiers, line 110: Kinda think this is enough. The rest is explained in other parts of the rule. This just feels a little repetitive. I would also suggest we make this the first example. The "default" case should but that it isn't important.

```suggestion
This `p` element has a not [important][] [computed][] `letter-spacing`.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<p style="letter-spacing: 0.1em">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 6
+
+This `p` element has two [declared][] values for its `letter-spacing` property (in the style sheet and in the `style` attribute). The one from the style sheet wins the [cascade sort][] because it is [important][]. Since it is not [declared][] via the `style` attribute, it matches the **cascade** condition. Note that neither the **above minimum** (because the [computed][] value is only 0.1 times the font size), nor the **not important** (because the [computed][] value comes from the style sheet and is [important][]) conditions are matched. Even though the value is too small, styles with [author origin][] declared by assistive technologies may win the [cascade sort][] and override it, thus this may satisfy [Success Criterion 1.4.12 Text Spacing][sc1412] and does not fail this rule. This is nonetheless bad practice and sufficient spacing should be used.
```
WilcoFiers, line 120: I think we can trim this down a bit.

```suggestion
This `p` element has two [declared][] values for its `letter-spacing` property. The one from the style sheet wins the [cascade sort][] because it is [important][]. Since it is not [declared][] via the `style` attribute, it matches the **cascade** condition. Note that neither the **above minimum**, nor the **not important** conditions are matched. Even though the value is too small, styles with [author origin][] declared by assistive technologies may win the [cascade sort][] and override it.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+</style>
+
+<p style="letter-spacing: 0.15em">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 7
+
+Both this `p` and `span` elements match the **not important** condition. For the `span`, the [cascaded][] value is `inherit !important`, thus the [computed][] value is the [inherited][] value, that is the [computed][] value of its parent, and it is not [important][]. Note that neither the **above minimum** (because the [computed][] value is only 0.1 times the font size), nor the **cascade** (because the [cascaded][] value comes from the `style` attribute) conditions are matched. Even though the value is too small, styles with [author origin][] declared by assistive technologies may win the [cascade sort][] and override it, thus this may satisfy [Success Criterion 1.4.12 Text Spacing][sc1412] and does not fail this rule. This is nonetheless bad practice and sufficient spacing should be used.
```
WilcoFiers, line 136:
```suggestion
Both this `p` and `span` elements match the **not important** condition. For the `span`, the [cascaded][] value is `inherit !important`, thus the [computed][] value is the [inherited][] value, that is the [computed][] value of its parent, and it is not [important][]. Note that neither the **above minimum**, nor the **cascade** conditions are matched. Even though the value is too small, styles with [author origin][] declared by assistive technologies may win the [cascade sort][] and override it. This is nonetheless bad practice and sufficient spacing should be used.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+		The toy brought back fond memories of being lost in the rain forest.
+	</span>
+</p>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `p` element has a [computed][] `letter-spacing` of `1.6px` (assuming a default `medium` font size of `16px`) which is below the recommended minimum.
```
WilcoFiers, line 162:
```suggestion
This `p` element has a [computed][] `letter-spacing` of 0.1 times the fonts which is below the recommended minimum.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+There is no HTML element.
+
+```svg
+<svg xmlns="http://www.w3.org/2000/svg"></svg>
+```
```
WilcoFiers, line 214: I don't think SVG is necessary in this.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+## Applicability
+
+This rule applies to any HTML element that is [visible][] and for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+
+## Expectation
+
+For each test target, one of the following is true:
+
+- **above minimum**: the [computed][] value of its [letter-spacing][] property is at least 0.12 times the [computed][] value of its [font-size][] property; or
+- **not important**: the [computed][] value of its [letter-spacing][] property is not [important][]; or
```
WilcoFiers, line 31: Would prefer we turn these around. The base case scenario is for someone to not have used !important. I think we should start with that.

```suggestion
- **not important**: the [computed][] value of its [letter-spacing][] property is not [important][]; or
- **above minimum**: the [computed][] value of its [letter-spacing][] property is at least 0.12 times the [computed][] value of its [font-size][] property; or
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,322 @@
...
+## Applicability
+
+This rule applies to any HTML element that is [visible][] and for which the `style` attribute [declares][declared] the [line-height][] CSS property.
+
+## Expectation
+
+For each test target, one of the following is true:
+
+- **above minimum**: the [computed][] value of its [line-height][] property is not `normal`, and is at least `1.5` or 1.5 times the [computed][] value of its [font-size][] property; or
+- **not important**: the [computed][] value of its [line-height][] property is not [important][]; or
```
WilcoFiers, line 31:
```suggestion
- **not important**: the [computed][] value of its [line-height][] property is not [important][]; or
- **above minimum**: the [computed][] value of its [line-height][] property is not `normal`, and is at least `1.5` or 1.5 times the [computed][] value of its [font-size][] property; or
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,322 @@
...
+
+Some examples use a fixed font size to demonstrate specific aspects of the rule (notably in order to provide a [computed][] value for the property). This is in general not a very good practice and should be avoided.
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a [computed][] `line-height` of `32px` (assuming a default `medium` font size of `16px`) which is above the recommended minimum, thus it matches the **above minimum** condition.
```
WilcoFiers, line 62:
```suggestion
This `p` element has a [computed][] `line-height` of 2 times the font size which is above the recommended minimum, thus it matches the **above minimum** condition.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,322 @@
...
+
+```html
+<p style="line-height: 2em !important">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 2
+
+This `p` element has a [computed][] `line-height` of `30px` specified via the style attribute, which is equal to the recommended minimum given the specified font size is 20 pixels, thus it matches the **above minimum** condition.
```
WilcoFiers, line 72:
```suggestion
This `p` element has a [computed][] `line-height` of `30px` specified via the style attribute, which is equal to the recommended minimum given the specified font size is 20 pixels.
```

WilcoFiers (changes requested):
Left my comments on the first rule, but it applies to all three of 'm.

_rules/letter-spacing-not-important-24afc2.md
Jym77: I'm always a bit uncomfortable giving bad practice in a Passed Example without an explicit warning… Tried to terse it a bit, let me know if that's OK.

_rules/letter-spacing-not-important-24afc2.md
Jym77: I debated than one with myself for quite some times before deciding to go that way 🤔 

My reasoning is that the Expectation is about the computed style, so the description should also be. Which I agree is annoying in forcing to assume the default value even though the example would fail anyway…

_rules/letter-spacing-not-important-24afc2.md
Jym77: I think most (all?) our rules targeting HTML elements have a non-HTML element has inapplicable example.

There's a bit of the same with the two invisible examples where maybe one would be enough…

Jym77 (commented):
Tried to dry down the descriptions.
I hope I didn't messed up copy/paste too much…

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Background
+
+When a style from [author origin][] is [declared][] in the `style` attribute with an [important][] declaration, it "wins" the [cascade sort] over any other style from [author origin][], i.e. it cannot be overridden by any of these. On the other hand, if such a style is [declared][] in a style sheet, it can still "lose" the [cascade sort][] to declarations with higher [specificity][] or simply coming from a later style sheet (such as ones injected by assistive technologies). This rule ensures that the element is not in the first case and that the style can be overridden by users, unless it is already above the minimum recommended threshold. Styles (with an [important][] declaration) that are declared with the [user origin][] or [user agent origin][] can win the [cascade sort][] over styles with the [author origin][]. Therefore, if the assistive technology can produce these styles, the rule is not needed.
```
WilcoFiers, line 48:
```suggestion
When a style is [declared][] in the `style` attribute with an [important][] declaration, it "wins" the [cascade sort] over any other style from [author origin][], i.e. it cannot be overridden by any of these. On the other hand, if such a style is [declared][] in a style sheet, it can still "lose" the [cascade sort][] to declarations with higher [specificity][] or simply coming from a later style sheet (such as ones injected by assistive technologies). This rule ensures that the element is not in the first case and that the style can be overridden by users, unless it is already above the minimum recommended threshold. Styles that are declared with the [user origin][] win the [cascade sort][] over styles with the [author origin][].
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to limit confusion.
```
WilcoFiers, line 50:
```suggestion
CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+Some examples use a fixed font size to demonstrate specific aspects of the rule (notably in order to provide a [computed][] value for the property). This is in general not a very good practice and should be avoided.
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `p` element has a **not [important][]** [computed][] `letter-spacing`, so styles with [author origin][] declared by assistive technologies may override it. This is nonetheless bad practice and sufficient spacing should be used.
```
WilcoFiers, line 64: Much of what you're saying is already explained in the background / expectations. This should be enough:

```suggestion
This `p` element has a **not [important][]** [computed][] `letter-spacing`.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+
+```html
+<p style="letter-spacing: 0.15em !important">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 3
+
+This `p` element has a [computed][] [letter-spacing][] of `3px`, which is **above minimum** (of `3px`).
```
WilcoFiers, line 84: 3px isn't "above" 3px.  This "above minimum" is not a good label. I also don't think this should be bold. Our example descriptions shouldn't need to reference the expectations.
```suggestion
This `p` element has a [computed][] [letter-spacing][] of `3px`, which is sufficiently wide that it should not need to be customised.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+</style>
+
+<p style="letter-spacing: 3px !important">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 4
+
+This `p` element has two [declared][] values for its `letter-spacing` property. The latest wins the [cascade sort][]. It has a value of `0.15em`, and is **above minimum**.
```
WilcoFiers, line 100: Same as above, don't think this should be referencing the bulleted lists as though these are things a reader will know off the top of their head.

```suggestion
This `p` element has two [declared][] values for its `letter-spacing` property. The latest wins the [cascade sort][]. It has a value of `0.15em`, and is sufficiently wide sufficiently wide that it should not need to be customised.
```

_rules/letter-spacing-not-important-24afc2.md
WilcoFiers: Not that I'm aware of.... Rules that apply to HTML elements have SVG inapplicable examples. But I think this is a little different.

I do think the two invisible examples are needed. Off-screen is fairly different from display:none.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,269 @@
...
+
+This `p` element does not have a `style` attribute specified.
+
+```html
+<p>
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Inapplicable Example 5
```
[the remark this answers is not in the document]
WilcoFiers: Still this.

WilcoFiers (changes requested):
Comments also apply to the other rules.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+</style>
+
+<p style="letter-spacing: 0.15em">
+	The toy brought back fond memories of being lost in the rain forest.
+</p>
+```
+
+#### Passed Example 7
+
+Both this `p` and `span` elements match the **not important** condition. For the `span`, the [cascaded][] value is `inherit !important`, thus the [computed][] value is the [inherited][] value, that is the [computed][] value of its parent, and it is not [important][]. Note that neither the **wide enough**, nor the **cascade** conditions are matched. Styles with [author origin][] declared by assistive technologies may override this style. This is nonetheless bad practice and sufficient spacing should be used.
```
WilcoFiers, line 136:
```suggestion
The [computed][] `letter-spacing` property of the `span` element is inherited from the `p` element. Because the [computed][] `letter-spacing` property of the `p` element is not [important][], the `span` element has a [computed][] `letter-spacing` property that is not [important][], even though the [declared][] `letter-spacing` property is.
```

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,271 @@
...
+<p style="letter-spacing: 0.1em">
+	<span style="letter-spacing: inherit !important;">
+		The toy brought back fond memories of being lost in the rain forest.
+	</span>
+</p>
+```
+
+#### Passed Example 8
+
+Both this `p` and `span` elements match the **not important** condition. For the `span`, the [cascaded][] value is `unset !important`, which is equivalent as `inherit` since it is an inherited property, thus the [computed][] value is the [inherited][] value, that is the [computed][] value of its parent, and it is not [important][]. Styles with [author origin][] declared by assistive technologies may override this style. This is nonetheless bad practice and sufficient spacing should be used.
```
WilcoFiers, line 148:
```suggestion
The [computed][] `letter-spacing` property of the `span` element is `unset`, which causes it to become inherited from the `p` element. Because the [computed][] `letter-spacing` property of the `p` element is not [important][], the `span` element has a [computed][] `letter-spacing` property that is not [important][], even though the [declared][] `letter-spacing` property is.
```

_rules/letter-spacing-not-important-24afc2.md
WilcoFiers: Fair, I suppose... But then shouldn't this have a text node with a style attribute with a letter-spacing property? Otherwise what are you testing?

_rules/letter-spacing-not-important-24afc2.md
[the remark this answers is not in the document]
WilcoFiers: What does 4 test that 5 doesn't? Neither of them have a style attribute with a letter-spacing property.

WilcoFiers (dismissed):

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -117,7 +117,7 @@ This `p` element has two [declared][] values for its `letter-spacing` property.
 
 #### Passed Example 6
 
-This `p` element has two [declared][] values for its `letter-spacing` property. The one from the style sheet wins the [cascade sort][] because it is [important][]. Since it is not [declared][] via the `style` attribute, it matches the **cascade** condition. Note that neither the **above minimum**, nor the **not important** conditions are matched. Styles with [author origin][] declared by assistive technologies may override this style. This is nonetheless bad practice and sufficient spacing should be used.
+This `p` element has two [declared][] values for its `letter-spacing` property. The one from the style sheet wins the [cascade sort][] because it is [important][]. Since it is not [declared][] via the `style` attribute, it matches the **cascade** condition. Note that neither the **wide enough**, nor the **not important** conditions are matched. Styles with [author origin][] declared by assistive technologies may override this style. This is nonetheless bad practice and sufficient spacing should be used.
```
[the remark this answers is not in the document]
Jym77: > This, and several other descriptions are still referencing the expectation, assuming the reader will know them. Please explain what about this passes, instead of having call-backs to the expectation.

Not referring to the labels of listed conditions is in direct contradiction with our rule design document.
https://act-rules.github.io/pages/design/rule-design/#listed-conditions (last but one bullet)
> Refer to the labels of the condition in the test case description to make the link explicit.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,263 @@
...
+
+CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+Some examples use a fixed font size to demonstrate specific aspects of the rule (notably in order to provide a [computed][] value for the property). Some Passed Examples set the `letter-spacing` property below the minimum, but do so in a way that can be overridden by styles with [author origin][]. These are however not very good practices and should be avoided.
```
WilcoFiers, line 56: I don't think this needs to be here. I don't think there's anything "not very good" about having normal letter-spacing. If anything, the bad practice is writing things in such a way that users can't control font properties. Having wide letter-spacing isn't great, but it somewhat mitigates the need for users to control that prop themselves. I'm not sure we need to put that in the rule though. WCAG says it's OK, we don't need to contradict that.

```suggestion
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -0,0 +1,314 @@
...
+
+CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+Some examples use a fixed font size to demonstrate specific aspects of the rule (notably in order to provide a [computed][] value for the property). Some Passed Examples set the `line-height` property below the minimum, but do so in a way that can be overridden by styles with [author origin][]. These are however not very good practices and should be avoided.
```
WilcoFiers, line 54:
```suggestion
```

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -0,0 +1,263 @@
...
+
+CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+Some examples use a fixed font size to demonstrate specific aspects of the rule (notably in order to provide a [computed][] value for the property). Some Passed Examples set the `word-spacing` property below the minimum, but do so in a way that can be overridden by styles with [author origin][]. These are however not very good practices and should be avoided.
```
WilcoFiers, line 56:
```suggestion
```

WilcoFiers (changes requested):

WilcoFiers (approved):

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -0,0 +1,261 @@
...
+- **wide enough**: the [computed][] value of its [letter-spacing][] property is at least 0.12 times the [computed][] value of its [font-size][] property; or
+- **cascade**: the [cascaded][] value of its [letter-spacing][] property is not a value [declared][] in its `style` attribute.
+
+## Assumptions
+
+- There is no mechanism available on the page to adjust [letter-spacing][]. If there is such a mechanism, it is possible to fail this rule while [Success Criterion 1.4.12 Text Spacing][sc1412] is still satisfied.
+
+- This rule assumes that WCAG's meaning for the "Letter spacing style property" is the value of the CSS `letter-spacing` property rather than the actual space between letters. The value of the CSS property is _added_ to whichever spacing already exist (for example, due to kerning). Thus, the actual space between letters can be larger than the value of the `letter-spacing` property. If [Success Criterion 1.4.12 Text Spacing][sc1412] is concerned by the actual space between letters, then this rule may fail (with the `letter-spacing` property being too small) while the Success Criterion is still satisfied (with the actual space being enough).
```
carlosapaduarte, line 38: > If [Success Criterion 1.4.12 Text Spacing][sc1412] is concerned by the actual space between letters

We might want to reach out for the AG also to clarify this issue

carlosapaduarte (approved):

_rules/letter-spacing-not-important-24afc2.md
Jym77: When writing this, I had a discussion with @WilcoFiers who was firmly into that interpretation, hence the result. Indeed, it might be worth reaching out to AG:
https://github.com/w3c/wcag/issues/1556

Jym77: Early feedbacks seem to agree with this interpretation 😁

kasperisager (approved):
Assuming that others already dug through the examples, I approve of the applicabilities and expectations 💯

Jym77:
Final Call ends on January 4th.

Jym77:

PR #1414, opened 2020-08-11
Add required aria-checked attribute to ARIA required owns passed example 4

tbostic32:
Added the required aria-checked attribute to the divs in ARIA required owns passed example 4 that have `role="menuitemradio"` and `role="menuitemcheckbox"`

Need for Final Call:

---

## How to Review And Approve


tbostic32:
I had to reverse pull into my own fork to get updated, which has left a bit of an ugly commit in the middle, is there a good way to get rid of this? 

Also, unsure if this requires a 1 week final call or could just be counted as an editorial edit.

PR #1419, opened 2020-08-18
make visible text content definition objective

### LINKED ISSUE #975: Ligature fonts fail "Visible label is part of accessible name " (2ee8b8)
Ligature fonts can be used as icons, things like this:

```html
<span aria-label="Remove" style="font-family: \'LigatureSymbols\'">delete</span>
```

In a case like this, "delete" should not be treated as visible text, as the entire word is replaced with an icon.

### LINKED ISSUE #1254: "visible text content" is subjective [2ee8b8]
> Elements that are visible, that are of type text, excluding text content in title or alt attributes, and are not categorized as non text content.

The definition of "non-text content" is subjective, so therefore visible text content is too. This creates a problem for Visible label is part of accessible name (2ee8b8). I think the way to fix this is to take this piece out of the definition and make it part of the expectation. That would mean inapplicable example 5 becomes something that passes instead.

WilcoFiers:
- Because the definition of Visible text content is used in the applicability, it can not be subjective.
- Add a ligature font passed example

Closes #1254, closes #975

Need for Final Call: 1 week

### DIFF
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -38,7 +38,7 @@ This rule applies to any element that has:
 
 ## Expectation
 
-All [text nodes][] in the [visible text content][] of each target element matches, or is contained within the [accessible name][] of the target element, except for [text nodes][] contains [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+All [text nodes][] in the [visible text content][] of each target element matches, or is contained within the [accessible name][] of the target element, except for characters used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
 
 ## Assumptions
 
@@ -87,6 +87,20 @@ This button has [visible][] text that does not need to be included in the [acces
 
 ```html
 <button aria-label="close">X</button>
+```
+
+#### Passed Example 5
+
+This `button` element has the text "search" rendered as an hourglass icon by the font. Because the text is rendered as [non-text content][], the text does not need to be included in the [accessible name][].
+
+```html
+<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet" />
+<style>
+	button {
+		font-family: 'Material Icons';
+	}
+</style>
+<button aria-label="Find">search</button>
 ```
 
 ### Failed
@@ -161,3 +175,4 @@ This link has no [visible text content][].
 [visible text content]: #visible-text-content 'Definition of Visible text content'
 [whitespace]: #whitespace 'Definition of Whitespace'
 [widget roles]: https://www.w3.org/TR/wai-aria-1.1/#widget_roles 'Definition of Widget role'
+[text nodes]: https://dom.spec.whatwg.org/#text 'DOM text, 2020/08/18'
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -38,7 +38,7 @@ This rule applies to any element that has:
 
 ## Expectation
 
-For each target element, all [text nodes][] in the [visible text content][] either matches or is contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+For each target element, all [text nodes][] in the [visible text content][] either match or are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
 
 ## Assumptions
 
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -38,9 +38,7 @@ This rule applies to any element that has:
 
 ## Expectation
 
-The complete [visible text content][] of the target element either matches or is contained within its [accessible name][].
-
-**Note:** Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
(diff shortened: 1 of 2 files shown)


pages/glossary/visible-text-content.md
```diff
@@ -2,12 +2,16 @@
...
 unambiguous: true
-objective: false
+objective: true
 input_aspects:
   - CSS styling
   - DOM tree
 ---
 
-Elements that are [visible](#visible), that are of type text, excluding text content in `title` or `alt` attributes, and are not categorized as [non text content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content).
+The _visible text content_ of an [element][] is a set of all [visible][] [text node][] that are a [descendant][] in the [flat tree][] of the element.
```
Jym77, line 11:
```suggestion
The _visible text content_ of an [element][] is a set of all [visible][] [text nodes][text node] that are [descendants][descendant] in the [flat tree][] of this element.
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -38,9 +38,7 @@ This rule applies to any element that has:
 
 ## Expectation
 
-The complete [visible text content][] of the target element either matches or is contained within its [accessible name][].
-
-**Note:** Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+All [text nodes][] in the [visible text content][] of each target element matches, or is contained within the [accessible name][] of the target element, except for [text nodes][] contains [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
Jym77, line 41:
```suggestion
Each [text node][] in the [visible text content][] of each target element matches, or is contained within, the [accessible name][] of this target element, except for [text nodes][] containing [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
Maybe even un-nesting a bit to avoid the "except":
```suggestion
Each [text node][] that is in the [visible text content][] of each target element and does not contain [non-text content][] matches, or is contained within, the [accessible name][] of the target element. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```

WilcoFiers: I think it's better to use the word "except" to be clear something is an exception, rather then use "or" to chain two conditions together.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -38,9 +38,7 @@ This rule applies to any element that has:
 
 ## Expectation
 
-The complete [visible text content][] of the target element either matches or is contained within its [accessible name][].
-
-**Note:** Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+All [text nodes][] in the [visible text content][] of each target element matches, or is contained within the [accessible name][] of the target element, except for characters used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
Jym77, line 41:
```suggestion
Each [text nodes][] in the [visible text content][] of each target element matches, or is contained within the [accessible name][] of this target element, except for characters used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```

WilcoFiers: Subtle but (IMO) important distinction here. The expectation applies to **each** applicable element, but within that element, **all** text nodes have to meet the condition or follow the exception.

Jym77: The question is whether we want:
* each individual text node matches or is contained.
* all text nodes, grouped together in one single thingie, match or are contained.

I am not sure what it means for "all test nodes" to match a string. We need to somehow gather "all text nodes" into a single stuff that can match the string. Is it concatenation in tree order? And then, the trimming of whitespace, does it occur before or after gathering all text nodes into a single thingie?

Maybe we want to call in @EmmaJP 😄

WilcoFiers: I've tried to adjust the order of the wording. Hope that helps clarify it.

Jym77 (approved):
I still find the sentence super weird, would personally use "each" and am not sure whether "all" should be singular or plural in that case.

But I won't block this further if I'm the only one having problem with this.

Jym77:
Also, why did you put the Final Call label on this PR? Looks like a misclick…
https://github.com/act-rules/act-rules.github.io/pull/1419#event-3669622094

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -38,9 +38,7 @@ This rule applies to any element that has:
 
 ## Expectation
 
-The complete [visible text content][] of the target element either matches or is contained within its [accessible name][].
-
-**Note:** Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+For each target element, all [text nodes][] in the [visible text content][] either matches or is contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
daniel-montalvo, line 41: Shouldn't this be plurar_ I am assuming we mean all of the text nodes in the visible text content either matching or being contained in the accessible name.

```suggestion
For each target element, all [text nodes][] in the [visible text content][] either match or are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```

daniel-montalvo (dismissed):

daniel-montalvo (approved):

jkodu (approved):

EmmaJP (approved):
These changes look good and seem good solutions for the issues raised. 

On a side note, as I haven't come across 'widget' used within rules to date I had to look up precisely what it meant in this context. Perhaps indicating a need to add a glossary definition. It could simply refer to https://www.digitala11y.com/widget-role/ or a similar explanation in the aria docs if there is one.

WilcoFiers:
@EmmaJP I'm not sure what to do with your comment about widget roles. The term "widget role" is linked in this rule to its definition in WAI-ARIA I think that's what you were asking? In which case that's already done.

PR #1385, opened 2020-07-09
[WIP] Rule update: TF feedback amendments (ff89c9)

jeeyyy:
Amend examples and support notes

Closes issue(s):
- closes https://github.com/act-rules/act-rules.github.io/issues/1174

Need for Final Call:
- 2 weeks

PR #1423, opened 2020-08-19
"ARIA required context role" (ff89c9): Prepare for TF review

### LINKED ISSUE #1174: Problems with 2 examples in  'ARIA required context role' (ff89c9)
Passed example 2 causes problems in some AT:

- not read as a list in Safari 13 with VoiceOver on macOS 10.15 (reads list item as plain text)
- List item not announced correctly by JAWS 2019 with IE11 (reads 'list start', 'list end' without reading list item)

It fails 4.1.1 because `ul` requires `li` children which is the likely cause of the problems above.
https://act-rules.github.io/rules/ff89c9#passed-example-2

Failed Example 3 doesn't cause any problems in any AT we tested:

- voiced correctly as a a list with 1 list item in JAWS 2019.1912.1 with IE11, FF68, Chrome 80
- voiced correctly as a a list with 1 list item in NVDA 2019.2.1 with IE11, FF68, Chrome 80
- voiced correctly as a a list with 1 list item in VoiceOver on macOS 10.15.3 with Safari 13.0.5

It's not clear from any of the ARIA specs why this is a failure.
https://act-rules.github.io/rules/ff89c9#failed-example-3

Jym77:
Prepare the rule for TF review:
- moves notes to Background;
- streamline test cases descriptions;
- fix open issue.

Closes issue(s):

- closes #1174 
- closes #1385 (supersedes)

Need for Final Call:
This will require a 1 week Final Call (rephrasing most content)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


jeeyyy (approved):

kasperisager (approved):

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -17,214 +17,239 @@ input_aspects:
...
+This element with an [explicit role][] of `listitem` is not [owned by][] an element with its [required context role][].
 
 ```html
 <div role="listitem">List item 1</div>
 ```
 
 #### Failed Example 2
 
-The `listitem` is owned by the `tabpanel`, because it is the closest ancestor, but `tabpanel` is not the correct [context](https://www.w3.org/TR/wai-aria-1.1/#scope) for `listitem`.
+These elements with an [explicit role][] of `listitem` are not [owned by][] an element with their [required context role][]. They are [owned by][] the `tabpanel` element, because it is the closest ancestor.
```
EmmaJP, line 147: Is there a need to point out that 'tabpanel' applies because it is in the DOM/accessibility tree, where previously 'presentation' was not?

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -17,214 +17,239 @@ input_aspects:
...
 		<div role="listitem">List item 1</div>
+		<div role="listitem">List item 2</div>
 	</div>
 </div>
 ```
 
 #### Failed Example 3
 
-The `listitem` is owned by the `aria-label="menu"` div, rather than the `list`.
+These elements with an [explicit role][] of `listitem` are [owned by][] an element with their [required context role][]. They are instead [owned by][] the `div` with an `aria-label` attribute; even though this `div` has no role, it has a global ARIA attribute and is thus [included in the accessibility tree][].
```
EmmaJP, line 160: Is there a missing 'not' in the first line?

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -17,214 +17,239 @@ input_aspects:
...
-	<li role="listitem">List item 1</li>
-</ul>
+<div role="header" aria-level="1">Hello!</div>
+<p>Welcome to my homepage!</p>
 ```
 
 #### Inapplicable Example 5
 
-The `doc-biblioentry` has a role from the [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0](https://www.w3.org/TR/dpub-aria-1.0/), not the [WAI-ARIA 1.1 Recommendation](https://www.w3.org/TR/wai-aria-1.1/), and it is therefore inapplicable.
+There is no element wit an [explicit role][] different from its [implicit role][]. This `li` element has an [explicit role][] of `listitem` which is identical to its [implicit role][].
```
EmmaJP, line 233:
```suggestion
There is no element with an [explicit role][] different from its [implicit role][]. This `li` element has an [explicit role][] of `listitem` which is identical to its [implicit role][].
```

EmmaJP (approved):
Spotted a couple of typos and had a question that might trigger another minor change. Other than those, looking great.

_rules/aria-required-context-role-ff89c9.md
Jym77: I think this is OK because this is the intuitive way (the accessibility tree is essentially the DOM tree).
Whereas Passed Example 3 is relying on the differences between accessibility tree and DOM and thus somewhat counter-intuitive and requires the explanation.

Jym77:
Final Call ends on September 9.

Jym77:

PR #1425, opened 2020-08-20
"Heading is descriptive" (b49b2e): replace "section of content"

### LINKED ISSUE #1361: "Heading is descriptive" (b49b2e) What do heading need to describe?
This is the latest spin-off from #1311 …

The question about "sections of content" is twofold. One part is the full fledged "what is a section of content?". The other part is "What do we need to do to fix [Heading is descriptive](https://act-rules.github.io/rules/b49b2e)?"

---
SC 2.4.6 (nor its Understanding document) actually does not speaks about "sections of content". It requires headings to "describe topic or purpose" without precising "topic of **what**". Our rule did add the "section of content" as an answer to that question but this is now creating more problems that it solved… (comparatively, 2.4.10 does refer to "sections").

The point here is that in order to fix "Heading is descriptive", which is currently the **only** rule using "section of content", we don't really need to have a full blown up definition of "section of content" that is correct for every single case we can think of. We only need something that patches this very rule. If and when we need a full definition of section of content later, we can evolve the patch or make a full new def.

---
Now, that rule is (and has always been) targeting "elements with a role of `heading` that is either visible or included in the accessibility tree". It is a subset of every thing that WCAG consider as "heading", so the rule is a partial check. Targeting that subset ensures that the applicability is objective and unambiguous without requiring us to have a full fledged definition of "heading" (in the WCAG sense of the term) which we all know will be hard to agree upon…

So, the question becomes: "here's an element with a role of `heading` which is visible or included in the accessibility tree; which part of the page needs to be described by that element?"

I do feel that the suggestion here: https://github.com/act-rules/act-rules.github.io/issues/1311#issuecomment-629989783 is a reasonable answer for that specific question. However, @EmmaJP is making a very good point in https://github.com/act-rules/act-rules.github.io/pull/1293#issuecomment-649504044 that images just before a heading but illustrating the following text are often used, so maybe not so good…

---
I do not think that the "visual cues" approach would work for that specific rule (but it may be good for a more general definition if needed). More details: https://github.com/act-rules/act-rules.github.io/pull/1293#issuecomment-644758580

---
@EmmaJP is also suggesting (https://github.com/act-rules/act-rules.github.io/pull/1293#issuecomment-649504044) to use something like "whatever user perceives as associated to this heading" for this rule. Maybe there is a way to turn that into something workable. This is definitely subjective, which is OK for expectation. I am afraid that it is so subjective that it becomes ambiguous (it's essentially not that different from "a distinctive part of the document" which is our current definition.

---
Of course, it is not clear that the rule must stay the way it is now. The rule can be expanded to handle more than "elements with a role of `heading`" (as suggested by @ajanec01 https://github.com/act-rules/act-rules.github.io/pull/1293#issuecomment-649590561).

The rule might also be split further (I am specifically thinking of having one rule targeting "elements with a role of `heading` that are included in the accessibility tree" and one for "elements with a role of `heading` that are visible" as these might need to describe different things (the first rule would be aimed mostly at AT users, the second at sighted users)).

If these tricks help us solve the issue, they're welcome given how complex this problem has become…

Jym77:
Change the expectation to remove "section of content" and replace it with "the next palpable content" (according to https://www.w3.org/2020/07/23-act-r-minutes.html#item01).

Doing minimal changes to get the fix working. Notably not splitting the rule for visible/accessible headings (it feels much needed now), and not cleaning up the rest of the rule (will be done once we want to submit it to TF review).

Closes issue(s):

- closes #1361 
- closes https://github.com/act-rules/act-rules.github.io/pull/1085#discussion_r364145387

Need for Final Call:
This will require a 1 week Final Call (changing expectation)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/heading-descriptive-b49b2e.md
```diff
@@ -30,7 +30,11 @@ This rule applies to any element with the [semantic role][] of heading that is e
 
 ## Expectation
 
-Each target element describes the topic or purpose of its [section of the content][].
+Each target element describes the topic or purpose of the first [palpable content][] which is:
+
+- after the target, in tree order in the [flat tree][]; and
+- [visible][], if the target is [visible][]; and
+- [included in the accessibility tree][], if the target is [included in the accessibility tree][].
```
Jym77, line 37: This strongly suggest that this rule should be split in two, one for each case…

kasperisager (approved):
I like it! 👏

_rules/heading-descriptive-b49b2e.md
```diff
@@ -30,13 +31,18 @@ This rule applies to any element with the [semantic role][] of heading that is e
...
 
 **Note:** Headings do not need to be lengthy. A word, or even a single character, may suffice.
 
 ## Assumptions
 
-This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the language of each test target can be correctly determined (either programmatically or by analyzing the content), and sufficiently understood.
+- This rule assumes that the [flat tree][] order is close to the reading order, as elements are rendered on the page. Due to positioning, it is possible to render a document in a order that greatly differ from the tree order, in which case the content which is visually associated with a heading might not be the content following it in tree order and this rule might fail while [Success Criterion 2.4.6 Headings and Label][sc246] is still satisfied.
```
kasperisager, line 45: Isn't the rule essentially assuming conformance to 1.3.2 (https://www.w3.org/TR/WCAG/#meaningful-sequence)? In that case, it seems like an even more reasonable assumption to make.

Jym77: Not really, because the programmatically determined order could be different from the DOM order (and match the visual order). For example by tweaking the accessibility tree with `aria-owns`.

kasperisager: Makes sense 👍 I also just looked into CSS Grid and, interestingly enough, grid layouts that re-arrange the logical order of content are non-conforming: https://drafts.csswg.org/css-grid/#order-accessibility

_rules/heading-descriptive-b49b2e.md
```diff
@@ -30,13 +31,18 @@ This rule applies to any element with the [semantic role][] of heading that is e
 
 ## Expectation
 
-Each target element describes the topic or purpose of its [section of the content][].
+Each target element describes the topic or purpose of the first non-[decorative][] [palpable content][] which is:
+
+- after the target, in tree order in the [flat tree][]; and
+- [visible][], if the target is [visible][]; and
+- [included in the accessibility tree][], if the target is [included in the accessibility tree][].
```
WilcoFiers, line 38: This list isn't following the conditional list format. The reason it can't is because its topic changes. I think you can reformulate this in multiple expectations. I prefer doing this over splitting the rule:

```suggestion
Each target element describes the topic or purpose of the first [non-decorative][decorative] [palpable content][] after the target in tree order in the [flat tree][].

## Expectation 2

Each [palpable content][] from expectation 1 is [visible][], except if its target element is not [visible][].

## Expectation 3

Each [palpable content][] from expectation 1 is [included in the accessibility tree][], except if its target element is not [included in the accessibility tree][].
```

WilcoFiers (dismissed):

_rules/heading-descriptive-b49b2e.md
```diff
@@ -30,13 +31,18 @@ This rule applies to any element with the [semantic role][] of heading that is e
 
 ## Expectation
 
-Each target element describes the topic or purpose of its [section of the content][].
+Each target element describes the topic or purpose of the first non-[decorative][] [palpable content][] which is:
```
EmmaJP, line 34: Good reference find on "palpable content"!!

_rules/heading-descriptive-b49b2e.md
EmmaJP: Not if those 'and' conditional are intentional. I read this as:

If the target is visible and included in the accessibility tree, then it should describe the first palpable content that is visible and included in the accessibility tree that comes after the target in the flat tree.

However, what about non-visible headings describing visible content, where both are included in the accessibility tree? This is not an unusual use case, eg. a news front page with a hidden heading 'Top headlines'. The heading isn't needed visually as that is usually apparent from other cues.

_rules/heading-descriptive-b49b2e.md
EmmaJP: Not sure I understand that.

_rules/heading-descriptive-b49b2e.md
EmmaJP: Not to mention the fun that can ensue when flexbox gets involved!

EmmaJP (changes requested):
I think some fine tuning is still needed with the phrasing of the applicability and expectation sections. Other than that, I think this is really starting to shape up well.

_rules/heading-descriptive-b49b2e.md
Jym77: If it is both visible and included in the accessibility tree, then it is "visible or included in the accessibility tree" (the or is inclusive by default). And then it matches both the expectations and need to describe both visible and accessible content (which may be the same or different palpable content).

WilcoFiers (approved):

kasperisager (approved):

_rules/heading-descriptive-b49b2e.md
```diff
@@ -144,11 +150,23 @@ Heading marked up with `h1` element that describes the topic or purpose of its [
 </html>
 ```
 
+#### Passed Example 8
+
+This heading describes the first [palpable content][] after it. The next one (the second `p` element) is not considered by this rule.
```
ajanec01, line 155:
```suggestion
This heading describes the first [palpable content][] after it (the first `p` element). The next [palpable content][](the second `p` element) is not considered by this rule.
```

_rules/heading-descriptive-b49b2e.md
```diff
@@ -192,6 +210,18 @@ Heading marked up with `h1` element that does not describe the topic or purpose
 </html>
 ```
 
+#### Failed Example 5
+
+This heading does not describe the first [palpable content][] after it. The next one (the second `p` element) is not considered by this rule.
```
ajanec01, line 215:
```suggestion
This heading does not describe the first [palpable content][] after it- the first `p` element. The next [palpable content][], the second `p` element, is not considered by this rule.
```

ajanec01 (approved):

Jym77:
Final Call ends on September 23rd.

Jym77:

Issue #1432, opened 2020-08-26
Aria prop has valid value, are existing IDREFs required? [6a7281]


WilcoFiers:
There's an inconsistency in this rule. The expectation first says that IDREF must reference an existing element, but is immediately followed by a note that says this isn't required.

> For value types ID Reference and ID Reference List for WAI-ARIA required properties at least one of the elements with the given ids exists in the same document tree or shadow tree as the element that specifies the target attribute.
> ...
> Note: Only for WAI-ARIA required properties with value types ID Reference and ID Reference List is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.

I think the idea of only making this required if the attribute is required is a reasonable one, but isn't how ARIA is designed. For instance combobox in ARIA 1.2 required aria-controls, regardless of if the popup is visible or not. It seems unreasonable to pass a combobox where the popup is hidden by toggling `display:none` but fail it if it is hidden by injecting and removing the popup from the DOM.

While I think a strict reading of WAI-ARIA suggests referencing non-existent elements isn't allowed, that doesn't seem like the intended behaviour. In my opinion, we should put in an exception for IDREF and IDREFs to exclude them.

jeeyyy:
@WilcoFiers what you mention here makes sense especially with the combobox example.

> `In my opinion, we should put in an exception for IDREF and IDREFs to exclude them`

Are you suggesting that we amend the expectation in such a way that we do not it mandatory for ID(REFS) actually exist?

WilcoFiers:
I think we should not make this required, correct. There are good reasons for IDREFs to point to non-existing elements.

ajanec01:
> ... at least one of the elements with the given ids exists in the same document tree or shadow tree as the element that specifies the target attribute.

On a slightly separate note, I think that the information about referenced `id` values of elements inside a shadow tree needs to be either removed or the accessibility support section needs to be expanded. If we check the [code pen example by Steve Faulkner](https://codepen.io/stevef/pen/Djzga/) then it's clear that relationships based on the `id` values don't work between the DOM and the shadow DOM. 

I don't think there is any value in adding a note to the Accessibility Support section. I think that id relationships between DOM and shadow DOM should be inapplicable, if not a failure.

ajanec01:
Addressed in #1649

PR #1433, opened 2020-08-26
update ARIA state or property has valid value rule to current writing style

jeeyyy:
- Updated applicability 
- Updated the example descriptions

Need for Final Call: 
- 1 week

> Note: There is an outstanding issue to tackle the expectation around ID(REFs) - https://github.com/act-rules/act-rules.github.io/issues/1432


_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,28 @@ acknowledgments:
 
 ## Applicability
 
-Any [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is not empty (""), and that is specified on an HTML or SVG element.
+The rule applies to any HTML or SVG element that has any not empty (`""`) [WAI-ARIA 1.1 state or property][] specified.
 
 ## Expectation
 
-Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type][].
```
Jym77, line 29: That is not correct anymore. With the new Applicability, the test targets are elements, not `aria-*` attributes. HTML/SVG elements do not have a "WAI-ARIA value type".

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,28 @@ acknowledgments:
...
+For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties][] at least one of the elements with the given ids exists in the same [document tree][] or [shadow tree][] as the element that specifies the target attribute.
 
-For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+For value type `URI` the value matches the [generic URI syntax][].
 
-**Note:** Only for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.
+**Note:** Only for [WAI-ARIA required properties][] with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.
 
 **Note:**
 For value type `URI`, this rule does not require that the destination URI exists.
```
Jym77, line 38: The notes could probably move to the Background section. The first one might be expanded a bit to explain why this extra requirement is only for required properties (why is it OK for non-required properties to reference non-existing IDs).
(the second note is fairly obvious: the rule checks the syntax of the URL, but is not a broken link checker)

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,28 @@ acknowledgments:
...
+- This rule catches values that are undefined in [WAI-ARIA Specifications][], and where the resulting behavior in user agents are also undefined in WAI-ARIA. This might lead to accessibility issues, if the intention was to use behavior defined in [WAI-ARIA Specifications][]. When values are used that do not have a defined behavior in [WAI-ARIA Specifications][], the HTML/SVG specification decides what default values should be used, since it is defined here what should happen when an invalid value is used for an attribute. If the default value for invalid attribute values happens to match the author's intention for the value, there will not be an accessibility issue.
```
Jym77, line 42: I'd make this a bit more direct:
```suggestion
- This rule assumes that the default value of the attributes do not match the author's intention. ARIA state and properties have a default value as defined in the [WAI-ARIA Specifications][], and this default is used when the provided value is invalid. In some case, the default value may happen to match the author's intention in using this attribute. In such a case, the rule will fail without creating an accessibility issue.
```

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,28 @@ acknowledgments:
...
 
 ## Accessibility Support
 
-Some user agents treat the value of `aria-*` attribute as case-sensitive (even when these are not ID) while some treat them as case-insensitive.
+- Some user agents treat the value of `aria-*` attribute as case-sensitive (even when these are not ID) while some treat them as case-insensitive.
```
Jym77, line 46:
```suggestion
- Some user agents treat the value of `aria-*` attributes as case-sensitive (even when these are not ID) while some treat them as case-insensitive.
```
```suggestion
- Some user agents treat the value of `aria-*` attribute as case-sensitive (even when these are not ID) while some treat them as case-insensitive.
```

_rules/aria-state-or-property-valid-value-6a7281.md
Jym77, line 42: I'm also not sure we need the bullet list when there is only one item in it.

_rules/aria-state-or-property-valid-value-6a7281.md
Jym77, line 46: I'm also not sure we need the bullet list when there is only one item in it.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,28 @@ acknowledgments:
...
-Any [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is not empty (""), and that is specified on an HTML or SVG element.
+The rule applies to any HTML or SVG element that has any not empty (`""`) [WAI-ARIA 1.1 state or property][] specified.
 
 ## Expectation
 
-Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type][].
 
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
+For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties][] at least one of the elements with the given ids exists in the same [document tree][] or [shadow tree][] as the element that specifies the target attribute.
```
Jym77, line 31: This somehow needs to mention the role of the element the property is on. Because required property depend on role.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -59,39 +59,39 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
+This `div` element has an `aria-pressed` state with valid tristate value.
 
 ```html
 <div role="button" aria-pressed="mixed">An other button</div>
 ```
 
 #### Passed Example 4
 
-`aria-errormessage` property with valid ID reference value
+This `div` element has an `aria-errormessage` property with valid ID reference value.
```
Jym77, line 86: Given that no element with that ID exist here, an explanation about the property not being required is needed.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -59,39 +59,39 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
 
 ```html
 ```
 
 #### Passed Example 5
 
-`aria-owns` property with valid ID reference list value
+This `div` element has an `aria-owns` property with valid ID reference list value.
```
Jym77, line 94: There are 3 `div` in that example, so "**this** `div`" is not well defined.

Plus, add explanation on whether the IDs need to exist or not.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -128,15 +128,15 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
+This `div` element has an `aria-dropeffect` property with valid token list value.
 
 ```html
 <div role="dialog" aria-dropeffect="copy move"></div>
 ```
 
 #### Passed Example 10
 
-`aria-controls`, which is a required property for the role `scrollbar`, has `ID Reference list` that references at least one element existing in the same document tree.
+The second `div` element has an `aria-controls`, which is a required property for the role `scrollbar`, has `ID Reference list` that references at least one element existing in the same document tree.
```
Jym77, line 139: This example feels overly complicated due to the many other `aria-*` attributes in it. Could we achieve the same thing with less attributes? (i.e., make the example atomic testing only one thing)

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -154,79 +154,79 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
+This `div` element has an Element with invalid token list value.
 
 ```html
 <div role="dialog" aria-dropeffect="invalid move"></div>
 ```
 
 #### Failed Example 9
 
-`aria-expanded` state with invalid true/false/undefined value for custom element
+This `div` element has an `aria-expanded` state with invalid `true/false/`undefined` value for custom element.
```
Jym77, line 221:
```suggestion
This `div` element has an `aria-expanded` state with invalid `true/false/undefined` value for custom element.
```
```suggestion
This `div` element has an `aria-expanded` state with invalid `true/false/`undefined` value for custom element.
```

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -154,79 +154,79 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
 
 ```html
 <my-button role="button" aria-expanded="collapsed">My button</my-button>
 ```
 
 #### Failed Example 10
 
-`aria-controls`, which is a required property for the role `scrollbar`, references an element that does not exist in the same document tree.
+This `div` element has an `aria-controls`, which is a required property for the role `scrollbar`, references an element that does not exist in the same document tree.
```
Jym77, line 229: Same as Passed Example 10. Especially here since some of the attributes are OK and some aren't.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,28 @@ acknowledgments:
 
 ## Applicability
 
-Any [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is not empty (""), and that is specified on an HTML or SVG element.
+The rule applies to any HTML or SVG element that has any not empty (`""`) [WAI-ARIA 1.1 state or property][] specified.
```
Jym77, line 25: We actually should have examples (Passed and Failed) with SVG element, since they are applicable.

Jym77 (changes requested):
The change in Applicability is breaking the rule.
The rest are mostly editorials.

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: I see your point. I have updated the applicability so it relies still on the state or property.

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: This should solve this.

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: @Jym77 are you suggesting that we make elements that do not have a role attribute as inapplicable to this rule?

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: It is not a MUST to set other aria-*, except aria-controls, so I have updated accordingly.

jeeyyy:
@Jym77 thanks for the round of review. I am not entirely following your argument around - "not mandatory for non-required properties to reference non-existing IDs". Help appreciated.

Other comments have been addressed.

_rules/aria-state-or-property-valid-value-6a7281.md
Jym77: Not sure what I was thinking about either… 🤷

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,19 @@ acknowledgments:
...
-Any [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) that is not empty (""), and that is specified on an HTML or SVG element.
+The rule applies to any [WAI-ARIA 1.1 state or property][] that is not empty (`""`), and is specified on an HTML or SVG element.
 
 ## Expectation
 
-Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
-
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
-
-For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
```
Jym77, line 33: Not sure why these have suddenly been removed 🤔 

Maybe it would be clearer to have them as second and third expectation?

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -53,45 +44,51 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
 - [WAI-ARIA 1.1, Characteristics of States and Properties, Value](https://www.w3.org/TR/wai-aria/#propcharacteristic_value)
 - [Uniform Resource Identifier (URI): Generic Syntax (RFC 3986)](https://www.ietf.org/rfc/rfc3986.txt)
 
+**Note:** Only for [WAI-ARIA required properties][] with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.
```
Jym77, line 47: I feel we need to say **why** the requirement is sometimes here, sometimes not. Instead of just saying that there is a difference.
```suggestion
Depending on the role of an elements, certain WAI-ARIA properties are or not required. This rule considers that when such a property has a value type of `ID reference` or `ID reference list`, then an element with one of the given ids must exist. On another hand, if a property that is not required refers to an incorrect id, it is less clear that this will cause an accessibility issue (the property is not required on the first place, so the element can live without it, and maybe later interaction with the page will create the correct id and this is the intended behaviour). Therefore, the rule has no extra requirement for properties with type `ID reference` or `ID reference list` that are not required.
```
+ links, and maybe polishing a bit the text.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -53,45 +44,51 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
+
+**Note:**
+For value type `URI`, this rule does not require that the destination URI exists.
```
Jym77, line 50:
```suggestion
For value type `URI`, this rule does not require that the destination URI exists. Testing that the link is not dead is out of scope for the rule and therefore ignored.
```

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -53,45 +44,51 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
 <div role="button" aria-pressed="mixed">An other button</div>
 ```
 
 #### Passed Example 4
 
-`aria-errormessage` property with valid ID reference value
+This `div` element has an `aria-errormessage` property with valid ID reference value.
 
 ```html
+<div id="my-error">Please provide a valid username</div>
```
Jym77, line 85: As a matter of fact, `aria-errormessage` is **not** required by `textbox`.
I think it is good to show that part of the rule in action. I suggest:
* remove the `div id="my-error"`.
* update the description to say that `aria-errormessage` is not required, so we don't care that the id is missing. Maybe adding a mollifying sentence in the line of "presumably, the error message (with the correct id) will be added by scripting when an error is detected".
* re-order the examples to have "required ID" before "not-required ID" so that a logical sequence follows. (and keep "required ID list" close to them).

Jym77 (changes requested):

Jym77 (dismissed):
https://github.com/act-rules/act-rules.github.io/pull/1433#discussion_r484460949 and https://github.com/act-rules/act-rules.github.io/pull/1433#discussion_r484467332 still need to be addressed.

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: @Jym77 see [issue](https://github.com/act-rules/act-rules.github.io/issues/1432), for reasoning as to why this was removed.

Jym77: If we remove that requirement, we need to clean more traces of it elsewhere in the rule… So, at least:
1. In the expectation, precise that "valid value" does not mean "existing IDref" given that the discussion in #1432 is about that.
2. Replace the Background note about "IDref must exist on required property" by something summing up #1432 and explaining why we don't require existing IDref.
3. Remove traces of it from the example descriptions. Not need to have a description about which case the IDref must or not exist since this is not part of the rule anymore.

Jym77 (dismissed):

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: @Jym77, 

For:
1) I disagree, the point of moving this section to background, was not to pollute the expectation with extra details. I believe keeping expectation precise and to the point that adding edge cases is better.
2) Again, I read through the background description & seems just right. I also believe after a conclusion has been reached to #1432, perhaps this can be improved. So leaving as is.
3) I disagree. Good to have longer descriptions than leaving it open.

I will see if similar reviews are arrived from over reviewers, to take any action on them. In the meantime dismissing this.

WilcoFiers: I agree with @Jym77 on this. ARIA 1.1 says an IDREF must exist, and for IDREFs that at least one of them must. Since that's not what this rule (should) require, we need to include an exception for it in the expectation.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -53,45 +44,41 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
 
 ### Passed
 
 #### Passed Example 1
 
-`aria-required` property with valid true/false value
+This `div` element has an `aria-required` property with valid `true/false` value.
 
 ```html
 <div role="textbox" aria-required="true" aria-label="A required textbox"></div>
```
WilcoFiers, line 60: Why are we using divs for this at all? Couldn't we just use native elements here? I think a mix of native and ARIA role elements would be better.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -100,166 +87,160 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
 
 ```html
 <div id="content1">Lorem ipsum...</div>
-<div
-	role="scrollbar"
-	aria-controls="content1 content2"
-	aria-orientation="vertical"
-	aria-valuemax="100"
-	aria-valuemin="0"
-	aria-valuenow="25"
```
WilcoFiers, line 149: Lets leave aria-valuenow on here. In ARIA 1.2 it no longer has a default.

WilcoFiers (changes requested):

_rules/aria-state-or-property-valid-value-6a7281.md
jeeyyy: I have added a simple note.

WilcoFiers: Notes can not contain requirements. Please put this as an exception in the expectation text.

WilcoFiers (changes requested):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -22,28 +22,19 @@ acknowledgments:
...
-
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
-
-For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
-
-**Note:** Only for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.
-
-**Note:**
-For value type `URI`, this rule does not require that the destination URI exists.
+Each test target has a valid value according to its [WAI-ARIA 1.1 value type][]. For [WAI-ARIA required properties][] with [value type][] of `ID reference` and `ID reference list` at least one of the referenced elements with the given ids should exist in the same [document tree][] or [shadow tree][].
```
WilcoFiers, line 29: Can you please review the comments again? This rule stats that these only need to exist if the ARIA attribute is required for that role.

WilcoFiers (changes requested):

WilcoFiers:
Closing, going to start from scratch.

PR #1434, opened 2020-08-27
image has alt: editorial fix on the example description

WilcoFiers:
This change was suggested by Kathy Eng on an ACT TF survey.

Need for Final Call: none

### DIFF
--- a/_rules/image-non-empty-accessible-name-23a2a8.md
+++ b/_rules/image-non-empty-accessible-name-23a2a8.md
@@ -152,7 +152,7 @@ This off screen `img` element has an [implicit role][] of `presentation` because
 
 #### Failed Example 1
 
-This `img` element has an empty [accessible name][] and an [implicit role][] of `img` because it is missing an attribute to give it an [accessible name][].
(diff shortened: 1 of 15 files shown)


_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -152,7 +152,7 @@ This off screen `img` element has an [implicit role][] of `presentation` because
 
 #### Failed Example 1
 
-This `img` element has an empty [accessible name][] and an [implicit role][] of `img` because it is missing an empty `alt` attribute.
+This `img` element has an empty [accessible name][] and an [implicit role][] of `img` because it is missing an attribute to give it an [accessible name][].
```
Jym77, line 155:
```suggestion
This `img` element has an [implicit role][] of `img` and an empty [accessible name][] because it is missing an attribute to give it an [accessible name][].
```
Keeping the "because" clause next to the text it explains.

Jym77 (changes requested):

Jym77 (approved):

kengdoj (approved):
Thanks!

PR #1435, opened 2020-08-27
text-contrast: make passed example 7 a failed example

WilcoFiers:
The ACT TF discussed this example at length, and concluded that because it is not only for aesthetic purposes, that font display is not decorative, and should therefore fail the rule.

Need for Final Call: 1 week


Jym77 (dismissed):
If that text is not decorative, I think we should just drop the example.

It does not add anything from, say, Failed Example 1 (it is illustrating the WCAG def of decoration, not the rule). Plus it creates a test case with both a "passed" and a "failed" outcome, making automatic gathering of implementation result difficult.

WilcoFiers:
@Jym77 What you suggest was discussed on the TF call too. The fact that we got this wrong in the first place is the reason I want to keep this example in. It illustrates a misconception about the way to interpret the definition of "purely decorative". Failed example 1 does not show this.

As for the mapping, I don't see why it would make mapping any harder. The way we do mapping supports this kind of thing.

Jym77 (approved):

kengdoj (approved):
Looks good!

jeeyyy (approved):

PR #1439, opened 2020-09-02
"Role attribute has valid value (674b10): process TF feedback

Jym77:
Processing TF feedback as per https://github.com/w3c/wcag-act/issues/467 :
- Remove accessibility mapping to 4.1.2 since 4.1.2 is only for UI components and any element could have a `role` attribute. Mapping instead to the list of roles in the various ARIA specs.
- As a consequence, removing the assumption that was related to it. That assumption was apparently very problematic…
- Replacing the first Inapplicable example by something with an implicit role, rather than a `div`.
- There is still an outstanding issue in our process about recursive inclusion of glossary in rules pages. This is still broken. We need to fix that fairly soon…

Closes issue(s):

- N/A

Need for Final Call:
This will require a 1 week Final Call (updating one example)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/role-attribute-valid-value-674b10.md
+++ b/_rules/role-attribute-valid-value-674b10.md
@@ -5,11 +5,6 @@ rule_type: atomic
 description: |
   This rule checks that each `role` attribute has a valid value.
 accessibility_requirements:
-  wcag20:4.1.2: # Name, Role, Value (A)
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
   wcag-technique:ARIA4: # Using a WAI-ARIA role to expose the role of a user interface component
     forConformance: false
     failed: not satisfied
@@ -41,7 +36,7 @@ Each test target has at least one token which is a valid value corresponding to 
 
 ## Assumptions
 
-This rule assumes that the [implicit role][] of elements is not enough to satisfy [Success Criterion 4.1.2 Name, Role, Value][sc412]. In case of invalid `role` attribute, the [semantic role][] defaults to the [implicit role]. If this is the correct role for the element, the rule will fail but [Success Criterion 4.1.2 Name, Role, Value][sc412] is still satisfied. For example, the element `<img role="image" src="/test-assets/shared/w3c-logo.png" alt="W3C logo" />` will fail this rule (because `image` is not a valid role) but satisfies [Success Criterion 4.1.2 Name, Role, Value][sc412] because the element defaults to its [implicit role][] of `img`.
+_There are currently no assumptions_
 
 ## Accessibility Support
 
@@ -55,7 +50,6 @@ Further reading:
 
 - [List of WAI-ARIA Roles][wai-aria role] and [List of Graphics ARIA Roles](https://www.w3.org/TR/graphics-aria-1.0/#role_definitions)
 - [Specification of the `role` attribute][role attribute]
-- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
 - [WAI-ARIA 1.1 Categorization of Roles](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization)
 - [WAI-ARIA Roles](https://www.w3.org/TR/wai-aria-1.1/#usage_intro)
 
--- a/_rules/role-attribute-valid-value-674b10.md
+++ b/_rules/role-attribute-valid-value-674b10.md
@@ -111,7 +111,7 @@ This `role` attribute contains two tokens, but none of these tokens is a valid r
 There is no `role` attribute.
 
 ```html
-<div>Some Content</div>
+<img src="/test-assets/shared/w3c-logo.png" alt="W3C logo" />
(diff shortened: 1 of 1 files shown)


EmmaJP (approved):

carlosapaduarte (approved):

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -5,11 +5,6 @@ rule_type: atomic
 description: |
   This rule checks that each `role` attribute has a valid value.
 accessibility_requirements:
-  wcag20:4.1.2: # Name, Role, Value (A)
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
```
WilcoFiers, line 12: I think we should fix the problem that this rule doesn't neatly map to WCAG, instead of not mapping it to WCAG at all. This rule becomes rather useless if we don't map it to any existing standard.

WilcoFiers (changes requested):

_rules/role-attribute-valid-value-674b10.md
Jym77: Unfortunately, I don't see a SC we can easily map to.
Notably, it is not really possible to map to 4.1.2 since any element (not just UI component) can have a `role` and 4.1.2 is only concerned by UI components… and given the very nature of the rule, we cannot restrict it to "elements with a `role` inheriting from `widget`"…

So I've mapped to the ARIA (+modules) list of roles, which is the best I could find.
The `role` attribute can be used to many things. Its spec says that it's up to users to define the list of acceptable roles (https://www.w3.org/TR/role-attribute/#s_role_module_attributes), so ARIA list of roles is the best I could find.

This depends on https://github.com/act-rules/act-rules-web/pull/207 for working.

kasperisager (approved):

Jym77:
Final Call ends on September 23rd.

carlosapaduarte:
I share @WilcoFiers concerns (https://github.com/act-rules/act-rules.github.io/pull/1439#discussion_r484431605)that the rule will become "useless" without a WCAG mapping. 
The way I see it, his suggestion (https://github.com/act-rules/act-rules.github.io/pull/1439#discussion_r489427931) could be, at least partially, implemented:
- One rule mapping to 4.1.2 for elements that are in the sequential focus order (or with a widget role)
- For other elements, I can't really see how we could map this rule to 1.1.1 or 1.3.1. But we could have, for those elements, the rule mapping to ARIA as proposed by @Jym77 in this PR

Jym77:
> * One rule mapping to 4.1.2 for elements that are in the sequential focus order (or with a widget role)

We can't have any selection based on role because the rule is about the validity of the role…
So `<div role="buton">` does **not** have a widget role but is typically the thing we want to flag with this rule.


My take is to leave the rule as it and consider implementing another rule for focusable elements only, mapping to 4.1.2 (with an assumption that focusable implies UI component). I'm rather in favour of flagging the same element by two rules (i.e. flagging a focusable element by this rule and by the one mapping to 4.1.2) rather than trying to have mutually exclusive Applicabilities that end up weird and hard to explain, see #1049 

I consider rules checking that ARIA is correctly used as valuable, even though they don't map to WCAG. Essentially in this case, if a page is full of typos in roles, this is preventing all other rules to work correctly (e.g. `<div role="image">` will be ignored by all image rules…) and is likely creating huge accessibility issues…

And of course, ARIA is also a W3C specification for accessibility.

carlosapaduarte:
I would be happy with such a solution (i.e. one rule for elements in the sequential focus order mapping to 4.1.2 and keeping this one).

Since I'm in academia I have less contact with industry clients than you do, but my limited experience nowadays (with EU monitoring requirements coming into play) is that WCAG conformance is what organizations are looking at. So, ARIA conformance is something "useless" for those organizations (but I agree when you say it is useful!)

Jym77:
> I would be happy with such a solution (i.e. one rule for elements in the sequential focus order mapping to 4.1.2 and keeping this one).
> 
> Since I'm in academia I have less contact with industry clients than you do, but my limited experience nowadays (with EU monitoring requirements coming into play) is that WCAG conformance is what organizations are looking at. So, ARIA conformance is something "useless" for those organizations (but I agree when you say it is useful!)

Yes of course, customers want WCAG AA conformance (even only WCAG 2.0 for the ones under section 508) and maybe care about the rest but don't have resources for achieving it…

Which is definitely a shame. And especially when it comes to ARIA conformance. ARIA conformance is not required by WCAG, because you can satisfy WCAG by using other technologies (theoretically), but it just happen that 100% of the people trying to satisfy WCAG for HTML are using ARIA… I see ARIA conformance as an extension to 4.1.1 if you use ARIA…
That's also what I often say to customers facing colleagues when they ask about these checks, relaying customers questions… By disabling the ARIA conformance checks, there is an high risk of blinding the checker to huge part of the code…
(comparatively, if customers want to disable a check about not skipping heading levels, I don't care that much, this is a best practice but not a huge accessibility issue, and doesn't cascade into other checks…)

Jym77:

PR #1415, opened 2020-08-12
update autocomplete rule to current writing style

WilcoFiers:
- Refactored applicability to be a little more precise
- Remove the unnecessary first expectation
- Updated the example descriptions
- Removed some unnecessary examples

Need for Final Call: 1 week

PR #1440, opened 2020-09-02
Clearing issues related to `autocomplete` attribute has valid value

### LINKED ISSUE #487: The autocomplete test (73f2c2) needs to be combined with a manual test
* Use of an invalid autocomplete attribute by itself is not a WCAG violation.
* Use of a valid autocomplete attribute by itself is not a pass of WCAG
* The first manual step of testing SC 1.3.5 is to determine if the field is requesting information about the user.  If the field is not requesting information about the user then whatever autocomplete is there or not there does not fail WCAG.

Recommendation would be to clarify that this test is only tangentially related to WCAG 2.1 and a separate test must be made in conjunction with the current test that can be automated.  Currently the relevance test is a manual test, but could in the future be tested with machine learning while the validity test can be made machine testable.

### LINKED ISSUE #992: Autocomplete attribute and aria-disabled (73f2c2)
`autocomplete` attribute has valid value
https://act-rules.github.io/rules/73f2c2
has several Applicability checks, one of which is:
The element has a aria-disabled="true" attribute

After some internal debates on implementation of this test it was pointed out that aria-disabled has no actual impact on the behavior of the widget and should not remove it from being tested. In particular was a comment from Bryan Garaventa:

In the case of aria-disabled, the only thing that this attribute does is to set a disabled state flag within the property for that control so it is conveyed for AT users. ATs are not actually supposed to honor anything beyond that. E.G You should still be able to set focus to the form control and interact with in the same way as if it did not include this attribute at all. I believe this is the reasoning why the aria-disabled attribute is being removed from the ‘global attribute’ list within the ARIA spec, and will then be forbidden in accordance with the ARIA-HTML spec. So the disabled attribute alone will be the only valid use of such an attribute on native form fields going forward.

In light of this, it seems like aria-disabled should be removed from the Applicability test exclusion list. As Bryan points out, the control is still going to be there with or without the aria-hidden and so its autocomplete attribute tokens should be checked.

### LINKED ISSUE #1400: Autocomplete, note that it also passes with autocomplete=off on the form (73f2c2)
As in title

```html
<form autocomplete="off">
  <input autocomplete="given-name" />
</form>
```

ajanec01:
- Adding Expectation 3 and changing passed example 2 to close #1400- elements with a form owner that has `autocomplete="off"` satisfy the rule/ sc1.3.5.
- The first part of the assumptions section closes #487-  mentions that the form elements collect information about the user and not anyone else. 
- The last paragraph of the assumptions section closes #992- elements with `aria-disabled` are assumed to be out of the sequential focus navigation and not to be otherwise operable.

Not sure if those changes fit the writing style- especially Expectation 3. Please review.

Also, I added changes made in #1415 so it may look like a lot now but should be easier to follow once #1415 passes the final call and then is being merged.

Need for Final Call:
1 week

### DIFF
--- a/_rules/autocomplete-valid-value-73f2c2.md
+++ b/_rules/autocomplete-valid-value-73f2c2.md
@@ -21,39 +21,35 @@ acknowledgments:
 
 ## Applicability
 
-The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` attribute that is a set of one or more [space separated tokens](https://html.spec.whatwg.org/#set-of-space-separated-tokens), except if one of the following is true:
+The rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` [attribute value][] that is neither empty (`""`) nor only [ASCII whitespace][], except if one of the following is true:
 
-- The element is not [visible](#visible), and not [included in the accessibility tree](#included-in-the-accessibility-tree)
-- The element is an `input` element with a `type` [attribute value][] of either `hidden`, `button`, `submit` or `reset`
-- The element has an `aria-disabled` [attribute value][] of `true`
-- The element is not part of [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation) and has a [semantic role](#semantic-role) that is not a [widget role](https://www.w3.org/TR/wai-aria-1.1/#widget_roles).
+- **hidden**: the element is not [visible][], and not [included in the accessibility tree][]; or
+- **disabled**: the element has an `aria-disabled` [attribute value][] of `true`; or
+- **fixed value**: the element is an `input` element with a `type` [attribute value][] of either `hidden`, `button`, `submit` or `reset`; or
+- **static**: the element is not part of [sequential focus navigation][] and has a [semantic role][] that is not a [widget role][].
 
 ## Expectation 1
 
-The `autocomplete` attribute is a single term, or a space separated list of terms.
+Each test target's `autocomplete` [attribute value][] is a [space separated][] list of one or more tokens that follow the [HTML specification for Autofill detail tokens][], which requires that the token list match the following in the correct order:
+
+1. An optional token that starts with "section-"; then
+2. An optional token of either "shipping" or "billing"; then
+3. An optional token of either "home", "work", "mobile", "fax" or "pager", only if the last token is "email", "impp", "tel" or "tel-*"; then
+4. A required token from the [correct autocomplete field][].
 
 ## Expectation 2
 
-The autocomplete term(s) follow the [HTML specification - Autofill detail tokens](https://html.spec.whatwg.org/#autofill-detail-tokens), which requires that it/they match the following in the correct order:
-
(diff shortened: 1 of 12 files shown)


_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -21,40 +21,48 @@ acknowledgments:
...
-4. Has a [correct autocomplete field](#correct-autocomplete-field) _(required)_
+## Expectation 2
 
-**Note:** Autocomplete terms are case insensitive. When multiple terms are used, they must be used in the correct order.
+Each test target's `autocomplete` [attribute value][] has a [correct autocomplete field][] that is an [appropriate][appropriate field for the form control] for that test target.
 
 ## Expectation 3
 
-The [correct autocomplete field](#correct-autocomplete-field) is an [appropriate field for the form control](#appropriate-field-for-the-form-control).
+Test targets can have a [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) with the `autocomplete` [attribute value][] set to `off`. In such a case, user agents may not autofill the form fields, but those elements are still be programmatically identifiable.
```
EmmaJP, line 46:
```suggestion
Test targets can have a [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) with the `autocomplete` [attribute value][] set to `off`. In such a case, user agents may not autofill the form fields, but those elements may still be programmatically identifiable.
```

EmmaJP: Alternatively:

```suggestion
Test targets can have a [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) with the `autocomplete` [attribute value][] set to `off`. In such a case, user agents may not autofill the form fields, but those elements are still programmatically identifiable.
```

Which is what was intended.

ajanec01: yes!

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -155,119 +149,120 @@ The `input` element does not have a semantic role that is a widget role, but sti
...
+This `autocomplete` [attribute value][] has an unknown term that is not a [correct autocomplete field][].
 
 ```html
 <label>Username<input autocomplete="badterm"/></label>
 ```
 
 #### Failed Example 2
 
-Term `work` not allowed before `photo`.
+This `autocomplete` [attribute value][] has the `work` token on a [correct autocomplete field][], however "work" can not be used with "photo".
```
carlosapaduarte, line 160:
```suggestion
This `autocomplete` [attribute value][] has the `work` token on a [correct autocomplete field][], however `work` can not be used with `photo`.
```

carlosapaduarte (changes requested):

WilcoFiers (approved):

ajanec01:
Hey all

I messed up this pull request. It will be easier to read the changes If I close it and then open another one. The comments made in this pr will be reflected in the new one.

Sorry!

Issue #1410, opened 2020-08-06
Object element has non-empty accessible name - better explanations in failures


moekraft:
Please describe the issue with references and relevant examples where necessary.
https://act-rules.github.io/rules/8fc3b6
It would be helpful in Failure examples 1, 2, and 4 to give a further explanation as to why the example fails. 
1. because title is empty
2. because the span element with id="label" is empty
4. because it does not provide an accessible name in it's title, an aria-label or aria-labelledby

--- or ---

If authoring a proposal for a new rule, choose from one of the issue templates:

- [Atomic Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/atomic-template-empty.md)
- [Composite Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/composite-template-empty.md)

Jym77:
@carlosapaduarte I think this is for you (you're sheperding this rule).

carlosapaduarte:
Solved in PR #1441 
Closing this issue

Issue #1409, opened 2020-08-06
Object element has non-empty accessible name - grammatical error


moekraft:
Please describe the issue with references and relevant examples where necessary.
https://act-rules.github.io/rules/8fc3b6 
Minor grammatical error found in _Accessibility support._ "Non supported" should be "Non-supported".

--- or ---

If authoring a proposal for a new rule, choose from one of the issue templates:

- [Atomic Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/atomic-template-empty.md)
- [Composite Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/composite-template-empty.md)

Jym77:
@carlosapaduarte I think this is for you.

PR #1441, opened 2020-09-04
Resolve TF feedback on "Object has accessible name" 8fc3b6

carlosapaduarte:
Processing TF feedback in w3c/wcag-act#479

- Added assumption about decorative objects
- Removed SC 1.1.1 from the rule's accessibility requirements
- We couldn't find further ways to add the accessible name to an object
- Tested, with VoiceOver, objects that inject HTML documents. The object is announced as a frame. If the object has an accessible name, VO reads it. If the object does not have an accessible name, but the document has a title, this is read by VO (in Chrome and Firefox) as the accessible name (but the object does not have an accessible name in the accessibility trees). If there is no accessible name nor title in the HTML document, the frame is announced without name. In Safari, VO never announces the accessible name.
- Updated some examples to inject other types of objects (images and HTML documents)
- Fallback content does not generate an accessible name, so I don't think we should promote its use with examples
- Corrected grammatical error
- Update failed examples descriptions with further information

Closes issue(s):

- #1409 
- #1410 

Need for Final Call:

This will require a 1 week Final Call (added assumption, small changes to test cases)

---

## How to Review And Approve


_rules/object-has-acessible-name-8fc3b6.md
```diff
@@ -5,11 +5,6 @@ rule_type: atomic
 description: |
   This rule checks that each `object` element  has a non-empty accessible name.
 accessibility_requirements:
-  wcag20:1.1.1: # Non-text Content (A)
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
```
WilcoFiers, line 12: Instead of removing the mapping, I think we should fix that sometimes this rule is failing the wrong SC. I think the simplest way to do this is to either add an assumption, or exclude objects with certain types of content.

WilcoFiers (dismissed):

Jym77 (changes requested):
Seconding @WilcoFiers comment.

carlosapaduarte:
Updated applicability to consider only image, audio or video resources and reinstated SC 1.1.1

_rules/object-has-acessible-name-8fc3b6.md
```diff
@@ -18,7 +23,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any `object` element that is [included in the accessibility tree][].
+This rule applies to any `object` element that is [included in the accessibility tree][] and embeds a resource with an [image MIME type](https://mimesniff.spec.whatwg.org/#image-mime-type) or an [audio or video MIME type](https://mimesniff.spec.whatwg.org/#audio-or-video-mime-type).
```
Jym77, line 26: I like that ❤️ It does raise a few questions…

1. Should we have examples with a `type` attribute?
2. How is the type determined without `type` attribute? Looks like it is step 8 in the algo at https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-object-element which is somewhat complex :-/
The question is especially relevant in "are we sure that the examples are going to be served with the correct MIME type?"
3. Can we add more types to this list (e.g. archives, fonts, …)? This is definitely not needed for the validity of the rule…
I think it might be worth to add a definition of "non-text MIME type" which would them be easy to update if needed (plus, can be reused, and makes the intention super clear in that rule: "an `object` which embeds a [non-text][non-text mime] resource")

Jym77 (dismissed):

_rules/object-has-acessible-name-8fc3b6.md
carlosapaduarte: @Jym77 I'll start by addressing question 2 and then 1 (before question 3)

2. I don't think we should dive into the algorithm (similar to what we do with accessible names, for instance). The browser determines the mime type of the resource (in the same way it builds the accessibility tree). I admit I haven't looked into the details of the process, but by looking at the headers in the HTTP response to the request for the object, the mime type is there and it is the correct one for all the examples we have. 
Furthermore, even when I enter an incorrect mime type using the type attribute, I made two observations: 1) the response headers always have the correct type (ignoring what I entered in the type attribute); 2) the browser always renders the object correctly, apparently also ignoring the type attribute.

1. Given the observations above, I don't really think we need to have examples with the type attribute, because it doesn't seem to be used for any rendering decisions.

3. Not sure it that would be useful. SC 1.1.1 addresses non-text content that is presented to the user. Fonts and archives are not really presented to user, are they? And if they are, I supposed they would be presented as text. By including those, we then would have to write the applicability in a way to leave them out.

I'll wait for your input, to understand if you think we still need to change anything in the rule!

Jym77: I'm OK with that 👍

Jym77 (approved):

_rules/object-has-acessible-name-8fc3b6.md
```diff
@@ -49,32 +49,31 @@ Testing that the [accessible name][] describes the purpose of the `object` eleme
...
 ```
 
 #### Passed Example 2
 
-This `object` element has a non-empty [accessible name][] through its `title` attribute.
+This `object` element which embeds a video resource has a non-empty [accessible name][] through its `title` attribute.
 
 ```html
-<object title="Moon speech" data="/test-assets/moon-audio/moon-speech.mp3"></object>
+<object title="My University" data="/test-assets/rabbit-video/video.mp4"></object>
```
WilcoFiers, line 63: The title on this doesn't seem appropriate for the video.

```suggestion
<object title="Rabbit animated short" data="/test-assets/rabbit-video/video.mp4"></object>
```

WilcoFiers (changes requested):
Minor issue

_rules/object-has-acessible-name-8fc3b6.md
carlosapaduarte: Good catch

WilcoFiers (approved):

_rules/object-has-acessible-name-8fc3b6.md
```diff
@@ -156,16 +155,26 @@ This `object` element is not [included in the accessibility tree][] due to `aria
...
-<object type="image/png" role="presentation" data="/test-assets/contrast/example.png"> </object>
+<object type="image/png" role="presentation" data="/test-assets/contrast/example.png"></object>
 ```
 
 #### Inapplicable Example 5
 
+This `object` element embeds an HTML resource.
+
+```html
+<object title="My University" data="/test-assets/shared/index.html"></object>
```
EmmaJP, line 166: This unusual scenario feels similar to an iframe. Just double checking, but presume for this rule/example it doesn't matter if the HTML resource is non-text content?

EmmaJP (approved):
This seems OK. I've asked a question, but suspect the response won't affect approval.

_rules/object-has-acessible-name-8fc3b6.md
carlosapaduarte: It is similar to an iframe (I seem to remember one of the browsers even treating this as an iframe in the DOM).
The rule was revised to be applicable only to images, audio or video mime types. That's why this was moved to an Inapplicable example. 
Do let me know if you wish to change your approval!

carlosapaduarte:
Final call ends October 5

carlosapaduarte:
Final call ended. Merging

PR #1443, opened 2020-09-07
Autocomplete clear issues

### LINKED ISSUE #487: The autocomplete test (73f2c2) needs to be combined with a manual test

### LINKED ISSUE #992: Autocomplete attribute and aria-disabled (73f2c2)

### LINKED ISSUE #1400: Autocomplete, note that it also passes with autocomplete=off on the form (73f2c2)

ajanec01:
- The first part of the assumptions section closes #487- mentions that the form elements collect information about the user and not anyone else.
- The last paragraph of the assumptions section closes #992- elements with aria-disabled are assumed to be out of the sequential focus navigation and not to be otherwise operable.
- Adding information to the Accessibility Support section, changing passed example 2, and expanding the background section to close #1400- elements with a form owner that has autocomplete="off" satisfy the rule/ sc1.3.5.

Need for Final Call:
1 week


_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -43,15 +43,23 @@ Each test target's `autocomplete` [attribute value][] has a [correct autocomplet
...
 
+The `aria-disabled` state is used on `input` elements which are not part of [sequential focus navigation][] and are not otherwise [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable). It means that this rule may fail elements which satisfy success Criterion [1.3.5 Identify Input Purpose][sc135].
+
 ## Accessibility Support
 
 - While `autocomplete` is a promising technique for supporting personalization in HTML, support for this in assistive technologies is fairly limited.
 - Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have a [semantic role][] of `none` and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
 - Some user agents treat the value of the `aria-disabled` attribute as case-sensitive.
+- Test targets can have a [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) with the `autocomplete` [attribute value][] set to `off`. In such a case, user agents may not autofill the form fields, but those elements are still programmatically identifiable.
```
Jym77, line 62: When you say "use agents *may not*", do you mean "the specs leave that possibility open, UA can do as they wish" (the normative sense of "may") or "the specs explicitly disallow this possibility" (the normative sense of "may not") or "some UA do it one way, some do it another way" (the usual thing we flag in Accessibility Support?

If it's one of the first two, it is a Background info, not an Accessibility Support. If it,s the third (as I understand from #1400 referring different behaviour in Chrome/Firefox), make it explicit.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -34,7 +34,7 @@ Each test target's `autocomplete` [attribute value][] is a [space separated][] l
 
 1. An optional token that starts with "section-"; then
 2. An optional token of either "shipping" or "billing"; then
-3. An optional token of either "home", "work", "mobile", "fax" or "pager", only if the last token is "email", "impp", "tel" or "tel-*"; then
+3. An optional token of either "home", "work", "mobile", "fax" or "pager", only if the last token is "email", "impp", "tel" or "tel-\*"; then
```
Jym77, line 37: I'm not sure why you add a backslash here 🤔 
OTOH, maybe all these tokens (also in other items of the list) might need to be in backticks as they are essentially code.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -75,13 +83,15 @@ This `autocomplete` [attribute value][] only has the required token, and is vali
 
 #### Passed Example 2
 
-This `autocomplete` [attribute value][] only has the required token, and is valid for a `select` element.
+This `autocomplete` [attribute value][] only has the required token, and is valid for a `select` element. Adding the `autocomplete="off"` to the element's [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) does not prevent the `autocomplete` [attribute value][] from being programmatically identifiable.
```
Jym77, line 86:
```suggestion
This `autocomplete` [attribute value][] only has the required token, and is valid for a `select` element. Even through the element's [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) has `autocomplete=off`, thus preventing UA to complete it, it does not prevent the `autocomplete` [attribute value][] from being programmatically identifiable.
```

Jym77 (changes requested):

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: Hey @Jym77 

When I first edited the file, the editor came back with a syntax error so I thought that I needed to escape it. Will change it now.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -43,15 +43,23 @@ Each test target's `autocomplete` [attribute value][] has a [correct autocomplet
...
+The `autocomplete` attribute is not used on form fields that:
+
+- do not correspond to [Input Purposes for User Interface Components](https://www.w3.org/TR/WCAG21/#input-purposes), and
+- do not collect information about the user.
+
+If the `autocomplete` field is used to describe "custom" taxonomy, rather than that described in the list of `input` purposes, or the form fields do not collect information about the user, success Criterion [1.3.5 Identify Input Purpose][sc135] may be satisfied even if this rule failed.
 
 The `type` attribute is used correctly according to the intended purpose of `input` elements. If an incorrect `type` attribute is used for `input` elements, this rule may fail elements that satisfy success Criterion [1.3.5 Identify Input Purpose][sc135]. For example if an `input` element has a `type` of `number`, but is expecting an e-mail address.
 
+The `aria-disabled` state is used on `input` elements which are not part of [sequential focus navigation][] and are not otherwise [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable). It means that this rule may fail elements which satisfy success Criterion [1.3.5 Identify Input Purpose][sc135].
```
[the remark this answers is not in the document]
ajanec01: Correct!

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: Thanks @Jym77!
There is no mention about this in the spec. As far as I can tell, the only word about `autocomplete="off"` on the form owner in the spec refers to how the default values are computed for elements which are missing the `autocomplete` attribute (link below).

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute:attr-fe-autocomplete-7

You're right it refers to the inconsistencies. I'll change that too.

EmmaJP: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion has some good explanations on how autocomplete functions in firefox for these different scenarios.

If I've understood correctly, `autocomplete="off" indicates a browser should not remember what is input to a field/form, and should not automatically fill in the field/form, but it may still suggest previously remembered values for a field.

I think this makes it background understanding, rather than an inconsistency of accessibility support.

EmmaJP (dismissed):
I agree with both the suggestions Jean-Yves has made. Other than those, this seem OK.

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: Thanks @EmmaJP!

It definitely needs re-writing as it causes confusion. I mixed two issues together.

The point about auto-filling/ autocompleting is, in my opinion, more of a side note.  You're right, it should go to the background section.

However, @Jym was right in saying that the accessibility support statement for the `autocomplete` property of `input` should be more explicit. As it stands Firefox returns an empty string whenever the `autocomplete` property is queried. Chrome and Edge, on the other hand, return the correct `autocomplete` value. This is an accessibility support issue as the `autocomplete` attribute in Firefox may not work for AT.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -43,28 +43,30 @@ Each test target's `autocomplete` [attribute value][] has a [correct autocomplet
...
 
 ## Accessibility Support
 
 - While `autocomplete` is a promising technique for supporting personalization in HTML, support for this in assistive technologies is fairly limited.
 - Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have a [semantic role][] of `none` and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
 - Some user agents treat the value of the `aria-disabled` attribute as case-sensitive.
-- Test targets can have a [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) with the `autocomplete` [attribute value][] set to `off`. In such a case, user agents may not autofill the form fields, but those elements are still programmatically identifiable.
+- In some user agents querying the value of the `autocomplete` property returns an empty string ("") even when the attribute was set according to the rule's expectations. It affects assistive technologies which rely on this property to personalise input fields collecting information about the user.
```
Jym77, line 62:
```suggestion
- In some user agents, querying the value of the `autocomplete` property returns an empty string ("") even when the attribute was set according to the rule's expectations. It affects assistive technologies which rely on this property to personalize input fields collecting information about the user.
```
W3C is using US English…

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -43,28 +43,30 @@ Each test target's `autocomplete` [attribute value][] has a [correct autocomplet
...
 
 ## Background
 
 The intent of this rule is to ensure that the `autocomplete` attribute can be used to support personalization. Many users may find it easier to fill out forms if those can be styled or laid out in a way that is familiar to them. Assistive technologies can do this when a form control is marked up in such a way that its purpose can be understood. For instance, assistive technologies could add familiar icons and colors to different fields, making it easier for the user to understand what the form does.
 
+The autocompleting feature of the `autocomplete` attribute benefits many users, but it is not required to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135]. Setting the `autocomplete="off"` on the element's [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) prevents the user agent from completing it, but it does not prevent the `autocomplete` [attribute value][] from being programmatically identifiable.
```
Jym77, line 68:
```suggestion
The autocompleting feature of the `autocomplete` attribute benefits many users, but it is not required to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135]. Setting `autocomplete="off"` on the element's [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) prevents the user agent from completing it, but it does not prevent the `autocomplete` [attribute value][] from being programmatically identifiable.
```

Jym77 (dismissed):
Good work. A couple of minor typos.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -43,20 +43,30 @@ Each test target's `autocomplete` [attribute value][] has a [correct autocomplet
...
 
 ## Background
 
 The intent of this rule is to ensure that the `autocomplete` attribute can be used to support personalization. Many users may find it easier to fill out forms if those can be styled or laid out in a way that is familiar to them. Assistive technologies can do this when a form control is marked up in such a way that its purpose can be understood. For instance, assistive technologies could add familiar icons and colors to different fields, making it easier for the user to understand what the form does.
 
+The autocompleting feature of the `autocomplete` attribute benefits many users, but it is not required to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135]. Setting `autocomplete="off"` on the element's [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) prevents the user agent from completing it, but it does not prevent the `autocomplete` [attribute value][] from being programmatically identifiable.
```
Jym77, line 68:
```suggestion
The auto-completing feature of the `autocomplete` attribute benefits many users, but it is not required to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135]. Setting `autocomplete="off"` on the element's [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) prevents the user agent from completing it, but it does not prevent the `autocomplete` [attribute value][] from being programmatically identifiable.
```
"autocompleting" seems to upset the spell check and make the test fail…

Jym77 (approved):

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: Thanks! Wasn't sure if that was a spell check error.

Jym77: When you click the "show test" error, it brings you to another page where you should be able to navigate and expand the steps (especially the red ones) until you see an hopefully explicit error message… In that case it was https://github.com/act-rules/act-rules.github.io/runs/1091127733?check_suite_focus=true#step:4:34 to find the file with the error and https://github.com/act-rules/act-rules.github.io/runs/1091127733?check_suite_focus=true#step:4:38 to find the actual error.

ajanec01: Gotcha, nice one!

WilcoFiers (approved):

carlosapaduarte (approved):
This looks good.
While you're at it, you can remove the "an" from "that is an appropriate for that test target" in the second expectation.

EmmaJP (approved):

Issue #1386, opened 2020-07-09
Element with aria-hidden has no focusable content - Passed Example 3 is a fail


dd8:
Passed example 3 assumes that all screen reader users access content via the keyboard and can't see the content on screen:
https://act-rules.github.io/rules/6cfa84#passed-example-3

The button is still focusable by clicking or touch - add a button:focus{ border: 1px solid red; } CSS rule in dev tools to see this

This example would cause problems for the following people:

- dyslexics with perfect vision who need a screen reader to voice text on screen
- people with low vision who may be able to see outlines but can't read text

In both cases they would see the button, and could focus it by clicking or touch, but it wouldn't voice anything

--- or ---

If authoring a proposal for a new rule, choose from one of the issue templates:

- [Atomic Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/atomic-template-empty.md)
- [Composite Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/composite-template-empty.md)

WilcoFiers:
OK we didn't get to this in the call, but I think you're right. The way this should have been done is to disable the button. I'll update the rule.

WilcoFiers:
Discussed this on the call today. i'm going to try to fix this by checking if the element can be focused by touch or click. Need to figure out how to define that, but I think that's a good addition to this rule. Thanks for the suggestion Mark!

Issue #643, opened 2019-07-02
aria-hidden-focus (6cfa84): as it applied to modal content (keyboard trap and aria-modal)


mfairchild365:
Consider the following situation:

A modal dialog is displayed and is wrapped in a keyboard trap. All content behind the modal has `aria-hidden="true"` applied to it, but interactive elements are left in the tab order. Additionally, the modal dialog has `aria-modal="true"` set. However,
  * Because of `aria-hidden="true"`, it is not possible to navigate to the parent content in reading mode.
  * Because of the keyboard trap that limits tab and shift+tab to modal content, it is not possible to navigate to interactive content outside of the modal.
  * `aria-modal="true"` limits navigation to the modal content in most screen readers
  * Only content in the modal appears in the screen reader's element list when the modal is open

I can't share an actual example of this (firewalls and whatnot), so here is some pseudo code:

```
<!-- parent content -->
<div aria-hidden="true">
   <p>we have the best stuff</p>
   <button>apply now</button>
</div>

<!-- open modal wrapped in a keyboard trap -->
<div class="keyboard-trap" tabindex="0"></div>
<div role="dialog" aria-hidden="true" aria-labelledby="dialog-title">
  <h2 id="dialog-title">Are you sure?</h2>
  <p>Please make sure that you are sure!</p>
  <button>Close</button>
</div>
<div class="keyboard-trap" tabindex="0"></div>
```

In this example, the "Apply now" button is in the tab order and would fail this rule as currently written, however, to my knowledge, it is impossible to actually navigate to it.

Should this rule still fail?

Referenced rule: https://act-rules.github.io/rules/6cfa84

Update: I found a pattern that is implemented like this in the wild, https://material.angular.io/components/dialog/examples

I tested with I tested with VO/iOS, VO/MacOS, Jaws, and NVDA, and none were able to access content outside of the modal (readig mode, tab or shift+tab, or via the elements list).

timbomckay:
> I referenced this issue in an [axe-core issue](https://github.com/dequelabs/axe-core/issues/1482#issuecomment-523555707) as well.

Seems like there's a bit of discussion on this at `axe-core`. Not sure if this pertains more to ACT Rules or Axe-Core.

carlosapaduarte:
I've tried this out with VO on Safari. 

[...]

PR #1444, opened 2020-09-09
Update aria hidden focus 6cf184

carlosapaduarte:
<< Describe the changes >>
#643 identifies a situation where an element cannot be focused (because of scripts) but would still fail the rule
#1386 identifies a situation where an element is not in the sequential focus navigation but can still be focused with a click or touch event

This PR updates the expectation so that now test targets cannot be focused through any mechanism, fixing the 2 issues.
Examples were updated in accordance. Example descriptions were also updated.

Closes issue(s):
- #643 
- #1386 

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


kasperisager (approved):
LGTM 👍

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -37,7 +37,8 @@ The rule applies to any element with an `aria-hidden` [attribute value][] of `tr
 
 ## Expectation
 
-None of the target elements are part of [sequential focus navigation](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation), nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are part of [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation).
+None of the target elements can be [focused][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that can be [focused][].
```
Jym77, line 40: I am a bit unsure on the formulation "can be focused".
Inlining the definition, this yields "can match the `:focus` pseudo-class" (or the `:focus-within`). Which sounds weird. And then again, the definition of `:focus` is "The `:focus` pseudo-class applies while an element has the focus (accepts keyboard or mouse events, or other forms of input)." (no further links). So "can be focused" actually means "can have the focus" (we'd better shortcut the two indirections and put that directly) and focus is not defined further…

[`:focus`](https://drafts.csswg.org/selectors-4/#the-focus-pseudo) has special case for HTML (we may decide to restrict he rule to HTML) liking to [focusable area](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area)… This has nice perks on inert and disable, but then again, this definition ultimately requires "[tabindex or] the element is determined by the user agent to be focusable" without further link.

So, essentially, an HTML element that "can be focused" is an element that "the UA determined as being focusable". Period. No further explanation… And I do not know if there is a way to list these elements¹.

Which does make some sense from the rule point of view as in "let the UA decide", the same way we take DOM or accessibility tree as input aspects and let the UA build them.

Note that this is fairly different from saying that an element "**is** [focused][]", aka "matches the `:focus` pseudo-class" as this is something we can easily query from the UA's API. It is perfectly determined which element matches `:focus`, not which elements can match it.

---
Also, if I take the code example from #643 (https://stackblitz.com/edit/angular-n7f6qz?file=src%2Fapp%2Fdialog-overview-example.ts) I am not sure what it means that the elements in the back "cannot match the `:focus` pseudo-class".
Does it mean that when I send `.focus()` on them, they do not match `:focus` (after a bit of time needed or processing)? That looks fairly hard to test.
Does it means that I cannot interact with the page in a way that they end up matching `:focus`? That seems not true as my interaction could close the modal and then focus the background element.

So, how do you test (auto or manual) that an element "cannot match the `:focus` pseudo-class"?
Especially because the interaction may change the rest of the page. For example, focusing (whether click or programmatic) the background elements may actually closes the modal and removes the `aria-hidden` on the background. Then, these elements are `aria-hidden` and can be focused, but the focus handler removes the `aria-hidden` and saves the day. The rule would fail because it doesn't take into account the possibility of altering the `aria-hidden` on focus…
(and I can fairly easily imagine cases where clicking out of the modal closes it)

---
Also, I'd argue that elements in the background of this example "can be focused". I'd say that the UA does "determine them as being focusable". It's not the UA that makes them non-focusable, it's the script on top of it. I would say the UA sees an `input`, consider that it is focusable and when it tries to `.focus()` it, gets hijacked by the `focusin` event handler on an ancestor. But the UA has no way² to actually figure out that these elements will never get focus and would consider them as focusable.

---
¹ Keeping this nice list here: https://allyjs.io/data-tables/focusable.html as it might help. But as far as I understand, it is generated by sending `.focus()` to elements and looking at the result…
² Apart from figuring out the semantic of the script, which is undecidable.

Jym77 (changes requested):
I'm having big concerns around the testability of "can be focused" and the actual meaning of it…

_rules/aria-hidden-no-focusable-content-6cfa84.md
kasperisager: I personally find the use of "can match the `:focus` pseudo-class" to be an improvement. Yes, it's document language specific. Yes, it might be different between UAs. However, it's meaning is perfectly clear; an element is focusable if you can _somehow_ make it match the `:focus` pseudo-class, of course without modifying the structure of the document. A first approach to testing this may be the `Element#focus()` method. Or keyboard emulation through browser automation. Or heuristics, which is the approach we would need to take in Alfa.

Jym77: But `Element#focus()`, for example, won't be enough…

```html
<button aria-hidden="true" onfocus="this.setAttribute("aria-hidden", "false")>Hello</button>
```
is `aria-hidden` (hence Applicable), and can be focused (thus failing the Expectation and the rule), but is not an actual problem (since the focus does clear the `aria-hidden`).

This is actually already a problem with the current version of the rule…

kasperisager: That code is very much a problem though because AT will ignore the button and never move focus to it 🤔 I couldn't interact with it at all using just VoiceOver; I had to rely on tab navigation to clear the `aria-hidden` attribute before being able to interact with the button using VoiceOver.

WilcoFiers: Counterproposal: How about instead of making this change, we add a second expectation, that the subject can not be focused through a pointer event, either because it can not be focused, because it is obscured, off screen, or pointer-events:none on it.

carlosapaduarte: @WilcoFiers that does not solve issue #643 because you would still have aria-hidden elements in the sequential focus order, thus failing the rule even if you wouldn't be able to focus those elements.

carlosapaduarte: @Jym77 Would changing the applicability to HTML documents specifically and, in the expectation, use the definition of "focusable" instead of "focused" be okay for you? Something like:

```md
None of the target elements is [focusable](#focusable), nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [focusable](#focusable).
```

Jym77: Digging into the docs and everything is getting hairier and hairier 🙈 

---
Our definition of [focusable](https://act-rules.github.io/glossary/#focusable), or similarly, the synonyms [focusable](https://html.spec.whatwg.org/multipage/interaction.html#focusable) and [focusable area](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area) in the specs, is essentially "let the UA decide".

Notably, specs has case of focusable areas with
> the element is determined by the user agent to be focusable;

followed by example:
> sometimes `<a href="">` (depending on platform conventions).

And later, again a case
> Any other element or part of an element determined by the user agent to be a focusable area, especially to aid with accessibility or to better match platform conventions.

with examples
> A user agent could make all list item bullets sequentially focusable, so that a user can more easily navigate lists.

> Similarly, a user agent could make all elements with title attributes sequentially focusable, so that their advisory information can be accessed.

See also this part of a later note confirming that UAs can act very differently…
>  For example, macOS users can set the user agent to skip non-form control elements, or can skip links when doing sequential focus navigation with just the `Tab` key (as opposed to using both the `Option` and `Tab` keys).

So, "focusable" is just something left at the choice of the UA (and https://allyjs.io/data-tables/focusable.html further confirm that it was a mess 4 years ago). This is not a problem per se (DOM tree, Accessibility tree, … are also at the choice at the UA). This is a bit more of a problem because I do not find any way to query the UA and get a list of focusable elements¹. Comparatively, it is possible to query an UA for the DOM, and (more or less) for the currently focused element (`Document.activeElement`: https://html.spec.whatwg.org/multipage/interaction.html#dom-documentorshadowroot-activeelement). But I do not see a way to get all the focusable elements which makes the full testability of the rule, and reproducibility of the tests, difficult.

Note that "sequential focus order" (or maybe rather "sequentially focusable" which appear to be different 🙄) is imho better as it is (at least manually) query-able simply by tabbing through the page.

At the very least, using "focusable" should only be done together with a big accessibility support warning…
(note that "focusable" is currently used in many rules, but that seems to be only a consequence of being used in notes/examples in "included in the accessibility tree" and "semantic role", so not an actual problem, but should be cleaned nonetheless imho)

¹ actually, there is something around [tabindex ordered focus navigation scope](https://html.spec.whatwg.org/multipage/interaction.html#tabindex-ordered-focus-navigation-scope) and [focus navigation scope](https://html.spec.whatwg.org/multipage/interaction.html#focus-navigation-scope) which seems to be such a list but is not clear that it's query-able. Both [jQueryUI](https://api.jqueryui.com/focusable-selector/) and [ally.js](https://github.com/medialize/ally.js/blob/master/src/is/focus-relevant.js) seem to solve this by having their own list of elements, possibly different from those of UAs…

---
I do not think that "focusable" solves #643.

Consider a case like `<span tabindex="0" onfocus="this.blur(); focusNext()" aria-hidden="true">` which is essentially a stripped down version of the example there.

It has a non-`null` tab index, it is not a shadow host, it is not disabled, it is not inert, it is rendered. Therefore, it does match all the conditions for the first case in the table that "describes what objects can be focusable areas" (https://html.spec.whatwg.org/multipage/interaction.html#focusable-area). Being a focusable area, I do not see how it is not focusable as per our definition. So, it would be focusable and `aria-hidden`, and fail the rule despite not being an actual problem (and being the case we try to solve with this PR…)

---
What about:
Applicability: element in sequential focus order.
Expectation: when the element is `:focused`, it is not `aria-hidden`.
?

That would let go of the "progamatically prevented focus" elements like #643 because even though they are in sequential focus order, they cannot be `:focused` and thus the expectation would always pass (may need to be a bit clear that this is the intended behaviour of the rule, of course).

WilcoFiers: I don't know if this is a problem we need to solve Jean-Yves. Adding an assumption is always an option. Something I encountered just yesterday is that Firefox makes scrollable element focusable, and unless they have tabindex="-1" it adds them to the sequential focus navigation too.

What does need to be solved is that just because an element is focusable, does not mean it can be focused. Mark's comment is about touch. It's not possible to fire a touch / click event on an element that's off screen, or that's completely obscured by an element that takes pointer-events. I think that needs to be solved as part of this.

WilcoFiers (changes requested):

_rules/aria-hidden-no-focusable-content-6cfa84.md
Jym77: > What does need to be solved is that just because an element is focusable, does not mean it can be focused.

Which is indeed a fairly big problem given that our definition of [focusable](https://act-rules.github.io/glossary/#focusable) is quite literarily "can be focused"…
And the specs don't help because the definition of "focusable" there is essentially "let the UA decide". The FF example also shows that. FF decides that some more stuff should be focusable and it's perfectly allowed to do so according to specs…

---
And again, the example in #643, essentially `<span tabindex="0" onfocus="this.blur(); focusNext()" aria-hidden="true">` is matching the HTML definition of "focusable area" (non-null tabindex, not a shadow host, not disabled, not inert, rendered), and thus is very much "focusable" as per our definition. Which in turn make "can be `:focus`" likely matching due to this part of CSS specs:
> There may be document language or implementation specific limits on which elements can acquire :focus. For example, [HTML] defines a list of focusable areas.

which I read as "in HTML, elements that can acquire `:focus` are the so-called focusable areas". This `span` is a focusable area, so I read it as "can acquire :focus" according to CSS def…


---
The problem we face here (and in #643) is scripting that overrides the HTML semantics. We can't really solve it (I think) by relying on pure HTML semantics/behaviour (such as "focusable").

Jym77: I think my problem with this PR is that it tries to use a static mean (HTML & CSS) to solve a dynamic problem (JS).

We essentially can't (automatically) look into the scripting (because anything becomes undecidable), so any attempt at getting a definition of "focusable" that only depends on HTML and CSS is, in my opinion, doomed to consider "focusable" elements that get their focus highjacked by scripting (barebone example being `<button onfocus="this.blur()">` can be more elaborated and real life like the case in #643).

What does that leaves us? I see three OK-ish solutions:
1. Add an assumption that scripting is not mishandling focus. This essentially means that the case in #643 is violating this assumption so the rule produce weird results. This is not super satisfying as real life cases like that one do exist…

2. Have a definition of "focusable" which is more explicit about considering scripting. That essentially makes the check manual. Something like "it is possible to interact with the page in a way that results in the element to match `:focus`" Maybe with a time limit (e.g. need to match `:focus` for more than 5 seconds in a row to make sense). Maybe with a precision that the DOM tree hasn't changed (so that something behind a modal is not "focusable" because closing the modal to focus it would change the DOM).

3. The somewhat lampshading solution I proposed above: have an expectation in the line of "when the element is `:focus`, it is not `aria-hidden`". That does hide a bit the fact that the process is slightly manual (by hiding the **how** to we reach a point where the element is `:focus`).

carlosapaduarte: I went with inspiration from option 2.
- updated the focused definition with the "interaction with the page" and the time limit
- added an assumption to the rule to clarify that the interactions with the page should not change elements that were previously hidden to not being hidden anymore

My main question to everyone: do you find "user interacted with the page" to be objective?

kasperisager (approved):
Still LGTM 👌 I'm a little wary of arbitrary thresholds, but if it achieves the purpose of unambiguity and consensus then it seems like a necessary evil 🤷

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -33,22 +33,22 @@ acknowledgments:
 
 This rule applies to any element with an `aria-hidden` [attribute value][] of `true`.
 
-**Note:** Using `aria-hidden="false"` on a descendant of an element with `aria-hidden="true"` **does not** expose that element. `aria-hidden="true"` hides itself and all its content from assistive technologies.
-
 ## Expectation
 
-None of the target elements are part of [sequential focus navigation](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation), nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are part of [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation).
+None of the target elements can be [focused][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that can be [focused][].
```
WilcoFiers, line 38: I'm going to repeat my comment from https://github.com/act-rules/act-rules.github.io/pull/1444/files

> What does need to be solved is that just because an element is focusable, does not mean it can be focused. Mark's comment is about touch. It's not possible to fire a touch / click event on an element that's off screen, or that's completely obscured by an element that takes pointer-events. I think that needs to be solved as part of this.

WilcoFiers (changes requested):

_rules/aria-hidden-no-focusable-content-6cfa84.md
WilcoFiers: I don't see how that resolved it. Let me take a stab at this one...

WilcoFiers, line 38: How about we do something like this:

```suggestion
None of the target elements are [focusable][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [focusable][].
```

```md
---
title: Focusable
key: focusable
unambiguous: true
objective: true
input_aspects:
  - CSS styling
  - DOM Tree
---

An element is _focusable_ if one or both of the following are true:

- the element is part of [sequential focus navigation][]; or
- the element can be [focused][] with a [single pointer][] user action in the [visible][] region of the [page][].

**Exception**: Elements that lose focus within 100ms of receiving focus are not considered _focusable_.

**Note**: The 100ms time span is an arbitrary limit which is not included in WCAG. Any focus redirect that is slower than 100ms will be clearly noticeable and could be considered a [change of context][] instead.

[change of context]: https://www.w3.org/TR/WCAG21/#dfn-change-of-context
[focused]: #focused
[page]: #html-web-page
[single pointer]: https://www.w3.org/TR/WCAG21/#dfn-single-pointer
[sequential focus navigation]: https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation
[visible]: #visible
```

dd8 (commented):
> The 5 seconds time span is an arbitrary limit which is not included in WCAG

One issue with 5 seconds is that this may impose a maximum speed on crawling and other types of batch processing. For a 50,000 page site waiting 5 seconds for each page = 69 hours

_rules/aria-hidden-no-focusable-content-6cfa84.md
carlosapaduarte: @WilcoFiers When processing this I recalled that we already have a [Focusable definition](https://act-rules.github.io/glossary/#focusable)

Still, checking the HTML spec, your definition looks a lot like the two types of focusability: 
- [Sequentially focusable](https://html.spec.whatwg.org/#sequentially-focusable)
- [Click focusable](https://html.spec.whatwg.org/#click-focusable)

Would you agree with rewriting the expectation as 
```suggestion
None of the target elements are [sequentially focusable](https://html.spec.whatwg.org/#sequentially-focusable) or [click focusable](https://html.spec.whatwg.org/#click-focusable), nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [sequentially focusable][] or [click focusable](https://html.spec.whatwg.org/#click-focusable).
```

WilcoFiers: > A focusable area is said to be click focusable if the user agent determines that it is click focusable

No... that definitely doesn't work for me. Besides, this just again says that anything with negative tabindex is click focusable, which isn't what we want. If you can't actually click it, because it's off screen or obscured, this shouldn't be true.

carlosapaduarte: @WilcoFiers 
Then before committing to changing the definition, just want to make sure you (and everyone else) agree with this course of action:
- update the name of the existing ["Focusable" definition](https://act-rules.github.io/glossary/#focusable) to "Keyboard focusable" and update the rules that used this definition (this should be done in a different PR)
- create the new "Focusable" definition (in this PR)
- update any of the rules that used the old focusable definition to use the new one if that makes sense (rules that looked specifically at keyboard focus won't be updated) (this is another PR after this one has been merged)

carlosapaduarte:
Ready for review again

WilcoFiers (approved):

pages/glossary/focusable.md
```diff
@@ -8,4 +8,15 @@ input_aspects:
   - DOM tree
 ---
 
-Elements that can become the target of keyboard input as described in the [HTML](https://www.w3.org/TR/html) specification of [focusable](https://html.spec.whatwg.org/#focusable-area) and [can be focused](https://html.spec.whatwg.org/#specially-focusable).
+An element is _focusable_ if one or both of the following are true:
+
+- the element is part of [sequential focus navigation][]; or
+- the element has a [tabindex value][] that is not null.
+
+**Exception**: Elements that lose focus immediately after the focus event is fired are not considered _focusable_.
```
Jym77, line 16: What does "immediately" means? Less than 1s? 1ms?
If JS is used to redirect focus, there will necessarily be some time for the redirection to occur (and it might actually depend on the computer and its load).

pages/glossary/focusable.md
```diff
@@ -8,4 +8,15 @@ input_aspects:
...
+
+The [tabindex value][] of an element is the value of the [tabindex attribute][] parsed using the [rules for parsing integers](https://html.spec.whatwg.org/#rules-for-parsing-integers). For the [tabindex value][] to be different from null, the [tabindex attribute][] must be a [valid integer](https://html.spec.whatwg.org/#valid-integer).
```
Jym77, line 18: I actually ran into an issue where a site was using `tabindex="-1-"` (extra `-` at the end), which doesn't parse as a valid integer according to HTML spec, and therefore Alfa was rejecting it, but Chrome doesn't care and treated it as `tabindex="-1"`.
We should at least have some note about that. It seems to only create false negative for now (something would be accepted by the rule due to having an incorrect tabindex, even though it would be an actual problem in Chrome (and maybe others)), so we maybe don't want to do more than a note…

Jym77 (changes requested):

pages/glossary/focusable.md
WilcoFiers: Interesting! Didn't think that could work. Perhaps a little more research here is needed. Figure out exactly what will and will not be accepted.

pages/glossary/focusable.md
WilcoFiers: On the next tick is what we were thinking, just didn't quite know how to define that. This seemed reasonable... but perhaps not. Any suggestions?

Jym77: "next tick" might be too restrictive (not fully sure exactly what it is, however).
`<button onfocus="this.blur(); someElement.focus()">` would loose focuse on "next tick" (I guess), but
`<button onfocus="console.log('Redirecting…'); this.blur(); someElement.focus()">` wouldn't.

And even the first case, I'm not fully sure what a "tick" is, but simply calling a JS function can actually do a lot of things and may take several ticks.

pages/glossary/focusable.md
carlosapaduarte: `tabindex="-1-"` is parsed by the algorithm as `-1` (so, not only Chrome, but all browsers respect the spec)

I've updated the last sentence to be correct, since it wasn't.

pages/glossary/focusable.md
carlosapaduarte: Going back to using a time limit (1 sec) without user intervention

pages/glossary/focusable.md
Jym77: 🤔 Ah, that's actually more twisted (as always…)
The definition of [valid integer](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-integer) is
> A string is a valid integer if it consists of one or more ASCII digits, optionally prefixed with a U+002D HYPHEN-MINUS character (-).

(so `-1-` is not a valid integer)

…

but the [rules for parsing it, point 8](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-integers) just below states:
> Collect a sequence of code points that are ASCII digits from input given position

thus disregarding anything after the first non-digit included (so `123etuirentiuernautier456` parses as `123`)…

pages/glossary/focused.md
```diff
@@ -7,4 +7,6 @@ input_aspects:
   - CSS styling
 ---
 
-An element is _focused_ when that element matches the [`:focus` pseudo-class](https://drafts.csswg.org/selectors-4/#focus-pseudo).
+An element is said to be _focused_ when the element matches the [`:focus` pseudo-class](https://drafts.csswg.org/selectors-4/#focus-pseudo) uninterruptedly for a period of 5 seconds after a user interacted with the page.
```
Jym77, line 10: I'd rather say "without a user interacting with the page" or something similar, since the user could keep interacting

🤔 

Actually, maybe we don't care about it.

pages/glossary/focusable.md
```diff
@@ -13,10 +13,14 @@ An element is _focusable_ if one or both of the following are true:
 - the element is part of [sequential focus navigation][]; or
 - the element has a [tabindex value][] that is not null.
 
-**Exception**: Elements that lose focus immediately after the focus event is fired are not considered _focusable_.
+**Exception**: Elements that lose focus during a period of up to 1 second after gaining focus, without the user interacting with the page the element is on, are not considered _focusable_.
 
-The [tabindex value][] of an element is the value of the [tabindex attribute][] parsed using the [rules for parsing integers](https://html.spec.whatwg.org/#rules-for-parsing-integers). For the [tabindex value][] to be different from null, the [tabindex attribute][] must be a [valid integer](https://html.spec.whatwg.org/#valid-integer).
+Notes:
 
+- The 1 second time span is an arbitrary limit which is not included in WCAG. Given that scripts can manage the focus state of elements, testing the focused state of an element consistently would be impractical without a time limit.
```
WilcoFiers, line 20: This seems like it should go in the background.

pages/glossary/focusable.md
```diff
@@ -13,10 +13,14 @@ An element is _focusable_ if one or both of the following are true:
...
+- The [tabindex value][] of an element is the value of the [tabindex attribute][] parsed using the [rules for parsing integers][]. For the [tabindex value][] to be different from null, it needs to be [parsed][rules for parsing integers] without errors.
```
WilcoFiers, line 21: Is this necessary? What does this add over just linking to the definition? Can we do this in a way that doesn't use a note. Notes suggest content that doesn't impact the outcome, it seems like this could?

pages/glossary/focused.md
```diff
@@ -7,4 +7,6 @@ input_aspects:
...
+
+The 5 seconds time span is an arbitrary limit which is not included in WCAG. Giving the possibility of the focus state of elements being managed through scripts, testing the focused state of an element consistently would be impractical without a time limit.
```
Jym77, line 12: Interestingly:
* this rule doesn't really use the definition of "focused". Only in the assumption, which is OK to be less precise than the rest of the text. I'd say we can keep "focus" undefined in the Assumption without really breaking anything 🤔 
* the definition of "focusable" introduces a 1s limit, this one a 5s limit. They should probably be the same (otherwise something that redirects focus after 3s would be "focusable" but could never be "focused", which is at best misleading).
* "focused" is only used in [Element in sequential focus order has visible focus)(https://act-rules.github.io/rules/oj04fd). Since 2.4.7 does not care about the limit, this makes the rule a bit worse (but only creating false negatives, so it's OK).


I would argue in favour of leaving the definition of "focused" as it is and not define "focus" in the assumption. I think it is OK to have undefined terms in assumptions and leave them to "common knowledge".
(for example in [ARIA required context role](https://act-rules.github.io/rules/ff89c9#assumptions) we have an assumption that explicit roles are "appropriate" for their element without explaining fully what it means)

pages/glossary/focusable.md
```diff
@@ -8,4 +8,19 @@ input_aspects:
...
+An element is _focusable_ if one or both of the following are true:
+
+- the element is part of [sequential focus navigation][]; or
+- the element has a [tabindex value][] that is not null.
+
+**Exception**: Elements that lose focus during a period of up to 1 second after gaining focus, without the user interacting with the page the element is on, are not considered _focusable_.
+
+Notes:
+
+- The 1 second time span is an arbitrary limit which is not included in WCAG. Given that scripts can manage the focus state of elements, testing the focused state of an element consistently would be impractical without a time limit.
```
Jym77, line 20:
```suggestion
- The 1 second time span is an arbitrary limit which is not included in WCAG. Given that scripts can manage the focus state of elements, testing the focusability of an element consistently would be impractical without a time limit.
```

Jym77 (dismissed):
They are two unrelated images that somehow crept into the PR 🙈

pages/glossary/focusable.md
carlosapaduarte: This is taken from the definition (it does not add to the definition). I added it here because people might confuse the tabindex value with the tabindex attribute, and wrongly assume that all that is needed is a non-empty attribute.

Nothing against removing it, if you don't think it adds any value?

pages/glossary/focusable.md
carlosapaduarte: OK. I'm afraid mentioning it there might be convoluted, but we'll see what others think.

pages/glossary/focused.md
carlosapaduarte: I like your suggestion. I think it captures the intention better.

pages/glossary/focused.md
carlosapaduarte: Agree with suggestion.

Jym77 (changes requested):
`test-assets/0va7u6/letter.jpg` is still part of the PR 🤔

pages/glossary/focused.md
```diff
@@ -7,6 +7,6 @@ input_aspects:
   - CSS styling
 ---
 
-An element is said to be _focused_ when the element matches the [`:focus` pseudo-class](https://drafts.csswg.org/selectors-4/#focus-pseudo) uninterruptedly for a period of 5 seconds after a user interacted with the page.
+An element is said to be _focused_ when the element matches the [`:focus` pseudo-class](https://drafts.csswg.org/selectors-4/#focus-pseudo) uninterruptedly for a period of 5 seconds after a user stopped interacting with the page.
```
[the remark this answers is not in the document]
carlosapaduarte: Agreed and updated

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -65,15 +67,15 @@ A [focusable][] element with `aria-hidden="true"` is ignored as part of the read
```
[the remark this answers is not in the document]
carlosapaduarte: Added passed example 4

carlosapaduarte:
Finally removed the incorrectly placed asset

Jym77 (approved):
Minor editorial. It's good to go for me.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -30,26 +30,28 @@ acknowledgments:
...
-
 ## Expectation
 
-None of the target elements are part of [sequential focus navigation](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation), nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are part of [sequential focus navigation](https://html.spec.whatwg.org/#sequential-focus-navigation).
+None of the target elements are [focusable][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [focusable][].
 
 ## Assumptions
 
-_There are currently no assumptions_
+Interacting with the page to focus an element does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog would make previously hidden and not [focusable][] elements to become [focusable][].
```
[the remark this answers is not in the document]
kengdoj:
```suggestion
Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and not [focusable][] elements become [focusable][].
```

kengdoj (approved):

carlosapaduarte:
Call for reviews ends on February 25

carlosapaduarte:
Call for review ended. Merging

PR #1449, opened 2020-09-17
image-filename: move note to background

WilcoFiers:
Need for Final Call: none


Jym77 (approved):

_rules/image-filename-as-accessible-name-9eb3f6.md
```diff
@@ -64,6 +62,8 @@ _There are no major accessibility support issues known for this rule._
 
 ## Background
 
+It is fairly common for CMS or other tools to default the alt-text of an image to its filename if no alt-text is provided. However, these names are usually not descriptive (often due to the presence of the file extension). This rule uses this heuristic to pinpoints cases where the [accessible name][] should be looked at by human testers. This rule does not automatically decide in which case a filename is correct (notably, it does not automatically decide whether adding the file extension is acceptable).
```
jeeyyy, line 65:
```suggestion
It is fairly common for content management systems (CMS) or other tools to default the alt-text of an image to its filename if no alt-text is provided. However, these names are usually not descriptive (often due to the presence of the file extension). This rule uses this heuristic to pinpoint cases where the [accessible name][] should be looked at by human testers. This rule does not automatically decide in which case a filename is correct (notably, it does not automatically decide whether adding the file extension is acceptable).
```

jeeyyy (dismissed):

carlosapaduarte (approved):

HelenBurge (approved):

PR #1452, opened 2020-09-17
visible-label: fix minor whitespace

WilcoFiers:
Passed example 2 doesn't seem like it needs that whitespace character. Example 1 has it covered.

Need for Final Call: 1 week


jeeyyy (dismissed):

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -70,7 +70,7 @@ This link has [visible][] text that, ignoring trailing whitespace, matches the [
 This link has [visible][] text that, ignoring case, matches the [accessible name][].
 
 ```html
-<a href="https://act-rules.github.io/" aria-label="act rules ">ACT rules</a>
+<a href="https://act-rules.github.io/" aria-label="act rules">ACT rules</a>
```
Jym77, line 73: I think I'd prefer to remove the extra space in Passed example 1 and keep it here, updating the description to "ignoring case and trailing spaces".

Reason: keep the first Passed Example as "the simplest way" to pass the rule helps humans reading it. First show the most basic case, then the weird stuff.

Jym77 (changes requested):

Jym77 (approved):

carlosapaduarte (approved):

maryjom (approved):

jeeyyy (approved):

PR #1453, opened 2020-09-17
scrollable-region: move note to the background

WilcoFiers:
Need for Final Call: none


_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -46,6 +44,9 @@ _No accessibility support issues known._
 
 ## Background
 
+To ensure there is some element from which arrow keys can be used to control the scroll position, focus must be on or in a scrollable region. 
+Tf scripts are used to prevent the keyboard events from reaching the scrollable region, this could still cause a keyboard accessibility issue. This must be tested separately.
```
Jym77, line 48:
```suggestion
To ensure there is some element from which arrow keys can be used to control the scroll position, focus must be on or in a scrollable region. If scripts are used to prevent the keyboard events from reaching the scrollable region, this could still cause a keyboard accessibility issue. This must be tested separately.
```

Jym77 (dismissed):

Jym77 (approved):

jeeyyy (approved):

carlosapaduarte (approved):

PR #1464, opened 2020-09-29
iframe-has-title: various improvements

### LINKED ISSUE #1170: What does it actually mean to be part of sequential focus navigation? (cae760)

### LINKED ISSUE #1367: iframe has non-empty name - problems with examples and applicability [cae760]

WilcoFiers:
This PR solves a number of problems with the iframe has accessible name rule.

1. Iframes that are not focusable and are not containers should not be treated as user interface components, and so are inapplicable to this rule. Several of the inapplicable examples were failed inappropriately in the previous iteration of this rule.

2. This rule no longer assumes browsers follow the standard when it comes to iframes. Known quirks are listed in accessibility support. Iframe behavior in browsers is inconsistent to the point that testing according to the standard would result in a lot of passes that do not actually satisfy success criterion 4.1.2. To solve this, the rule is rewritten specifically to work for the three major screen readers used on the web, and the browser they are most compatible with: VoiceOver+Safari, JAWS+Chrome, NVDA+Firefox.

Because of quirky support in accessible name computation and accessibility tree construction, this rule uses more low level concepts like "explicit role" and "disabled state". This also required going from using the accessible name, to using the title attribute alone, since VoiceOver+Safari only uses the title to generate the accessible name of an iframe.

Closes #1367, closes #1170

Need for Final Call: 1 week

## How to Review And Approve


_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe style="visibility:hidden;" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 3
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has an [explicit role][] of `presentation`.
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content inaccessible to keyboard.
```
carlosapaduarte, line 160:
```suggestion
**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe role="presentation" tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 4
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has an [explicit role][] of `none`.
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content inaccessible to keyboard.
```
carlosapaduarte, line 170:
```suggestion
**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe role="none" tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 5
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has an [hidden state][] of `true`.
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content inaccessible to keyboard.
```
carlosapaduarte, line 180:
```suggestion
**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe aria-hidden="true" tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 6
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has no [content][] inside the `iframe`.
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content inaccessible to keyboard.
```
carlosapaduarte, line 190:
```suggestion
**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
```

test-assets/SC4-1-2-frame-doc.html
```diff
@@ -1,9 +1,16 @@
...
-	<li>Rice</li>
-	<li>Bread</li>
-	<li>Unsweetened Granola</li>
-	<li>Basil</li>
-	<li>Honey</li>
-</ul>
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>Grocary List</title>
```
carlosapaduarte, line 4:
```suggestion
		<title>Grocery List</title>
```

carlosapaduarte (changes requested):
Minor editorial changes

carlosapaduarte (approved):

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+This `iframe` element has `title` [attribute value][] that consists of only [ASCII whitespace][].
+
+```html
+<iframe title="  " src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+```
+
+#### Failed Example 4
+
+This `iframe` element should use the `title` attribute, instead of using `aria-label`.
```
Jym77, line 107:
```suggestion
This `iframe` element has no `title` attribute, the `aria-label` attribute is not considered by this rule and many user agents.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe aria-label="Grocery list" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+```
+
+#### Failed Example 5
+
+This `iframe` element with a [semantic role][] of `article` should use the `title` attribute, instead of using `aria-labelledby`.
```
Jym77, line 115:
```suggestion
This `iframe` element with a [semantic role][] of `article` has no `title` attribute, the `aria-labelledby` attribute is not considered by this rule and many user agents.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<div id="frame-title-helper">Grocery List</div>
+<iframe role="article" aria-labelledby="frame-title-helper" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+```
+
+#### Failed Example 6
+
+This `iframe` element should have a `title` attribute, or be removed from [sequential focus navigation][].
```
Jym77, line 124:
```suggestion
This `iframe` element has no `title` attribute, even though it has a presentational role, it is still part of [sequential focus navigation][] and thus require a name.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe src="/test-assets/SC4-1-2-frame-doc.html" role="none"> </iframe>
+```
+
+#### Failed Example 7
+
+This `iframe` element should have a `title` attribute, or be removed as a container by setting an [explicit role][] of `presentation` or `none`, or by setting `aria-hidden` to `true`.
```
Jym77, line 132:
```suggestion
This `iframe` element has no `title` attribute, even though is has been removed from [sequential focus navigation][], it is still exposed as a container (and needs a name) without setting an [explicit role][] of `presentation` or `none`, or setting `aria-hidden` to `true`.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+This `iframe` is not part of [sequential focus navigation][] and has no [content][] because its CSS `display` property is `none`.
```
Jym77, line 142:
```suggestion
This `iframe` is not part of [sequential focus navigation][] and has a [hidden state][] of `true` because its CSS `display` property is `none`.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe style="display:none;" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 2
+
+This `iframe` is not part of [sequential focus navigation][] and has no [content][] because its CSS `visibility` property is `hidden`.
```
Jym77, line 150:
```suggestion
This `iframe` is not part of [sequential focus navigation][] and has a [hidden state][] of `true` because its CSS `visibility` property is `hidden`.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+```html
+<iframe style="visibility:hidden;" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 3
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has an [explicit role][] of `presentation`.
```
Jym77, line 158:
```suggestion
This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`; and it has an [explicit role][] of `presentation`.
```
Not fully clear whether the subject of "has" was the tabindex or the iframe. I think that improves it.

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
+
+```html
+<iframe role="presentation" tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 4
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has an [explicit role][] of `none`.
```
Jym77, line 168:
```suggestion
This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`; and it has an [explicit role][] of `none`.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
+
+```html
+<iframe role="none" tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 5
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has an [hidden state][] of `true`.
```
Jym77, line 178:
```suggestion
This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`; and it has an [hidden state][] of `true`.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Warning:** Using `tabindex="-1"` on an `iframe` makes it, and all its content, inaccessible to keyboard.
+
+```html
+<iframe aria-hidden="true" tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"></iframe>
+```
+
+#### Inapplicable Example 6
+
+This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`, and has no [content][] inside the `iframe`.
```
Jym77, line 188:
```suggestion
This `iframe` is not part of [sequential focus navigation][] because its `tabindex` [attribute value][] is `-1`; and it has no [content][] inside the `iframe`.
```

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+[included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of included in the accessibility tree'
+[ascii whitespace]: https://infra.spec.whatwg.org/#ascii-whitespace
+[sequential focus navigation]: https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation 'HTML definition of Sequential focus navigation'
+[user interface component]: https://www.w3.org/TR/WCAG21/#dfn-user-interface-components
+[accessible name computation]: https://www.w3.org/TR/accname
+[explicit role]: #explicit-role
+[hidden state]: #hidden-state
+[content]: https://html.spec.whatwg.org/multipage/dom.html#palpable-content 'HTML definition of Palpable Content'
```
Jym77, line 203: I think we might call it "palpable content" everywhere in the rule, given that "content" has already a meaning that is more than just the "palpable content".

Also, "palpable content included in the accessibility tree" is almost [perceivable content](https://github.com/act-rules/act-rules.github.io/blob/rule-cf77f2-bypass-blocks/pages/glossary/perceivable-content.md) I've been working on. I'm not sure whether "perceivable content" would fit your needs here… and/or if we should chisel out a definition that fits both needs…

Jym77 (dismissed):

_rules/iframe-non-empty-title-attribute-cae760.md
WilcoFiers: Seeing as this isn't done yet, I'm going to leave this for what it is now. We can always revisit later.

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+    - Wilco Fiers
+---
+
+## Applicability
+
+The rule applies to `iframe` elements, except if the `iframe` is not part of [sequential focus navigation][], and one of the following is true:
+
+- the `iframe` has a [hidden state][] of `true`; or
+- the `iframe` has an [explicit role][] of `presentation` or `none`; or
+- the `iframe` has no [content document][] with [content][] [included in the accessibility tree][].
```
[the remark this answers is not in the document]
WilcoFiers: Good idea, except without the brackets. Those are in the template to indicate this is optional.

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Assumptions
+
+This rule assumes that an `iframe` is only a [user interface component][] if it is a container that assistive technologies can navigate into, or if it is a stop in the [sequential focus navigation][]. Since [4.1.2 Name Role Value][sc412] only applies to [user interface components][user interface component] not all `iframe` elements require a `title` attribute, including some `iframe` elements that are announced by assistive technologies.
+
+## Accessibility Support
+
+There are substantial differences between how various browsers and assistive technologies treat `iframe` elements. This rule is designed to work in a variety of commonly used combinations of them. Some notable differences are:
+
+- Some browsers do not use the [accessible name computation][] to compute the name of an `iframe` element, and instead only announce the `title` attribute.
```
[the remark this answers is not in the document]
WilcoFiers: I put it here because this is neither an assumption about the content, nor an assumption about how to interpret the requirement. The "choice" that is made is to not exclude voiceover+safari from this rule, which I think the rule properly documents. I think this needs to be here.

_rules/iframe-non-empty-title-attribute-cae760.md
```diff
@@ -0,0 +1,208 @@
+---
```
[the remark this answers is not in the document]
WilcoFiers: I have no idea why Github's UI is showing it this way, but as you can see in the commit for it; this file was rename:
https://github.com/act-rules/act-rules.github.io/pull/1464/commits/df5af2c377205db209f2912b5dfd1d431c7f56ff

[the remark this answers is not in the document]
Jym77: The thing is that git has no real notion of "moving" (it deletes and recreate instead, including with `git mv`).
I think this is due to move + too many changes and then git algo consider that it is a different file, not the same, and there is a high risk that it will creep that way in the merge commit (i.e., we'll loose continuity after merge).

We had a similar issue last year (with a label rule, iirc), which we ended up solving by having a PR just to rename the file (so that git picks the correct move in all its logs), and then a PR for doing the actual changes 😞 See #904 and linked PRs/issues.

_rules/iframe-non-empty-title-attribute-cae760.md
Jym77: I still think we should use "palpable content" all over in the text, rather than "content. Because "content" has another meaning.

_rules/iframe-non-empty-title-attribute-cae760.md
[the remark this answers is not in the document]
Jym77: The problem is that the examples with `aria-label` are now false positive for JAWS users.

We're choosing to follow one UA+AT and not the others, which we have to do, but we're making the choice to follow the one that goes directly against the specs: https://www.w3.org/TR/html-aam-1.0/#iframe-element which doesn't sound good.

I'm not sure this solves #1367 and would like to see @dd8 opinion on this.

Jym77 (changes requested):

_rules/iframe-non-empty-title-attribute-cae760.md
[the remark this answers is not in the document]
WilcoFiers: OK I'll open a new PR just to rename the file. We'll have to merge that in before this one then.

_rules/iframe-non-empty-title-attribute-cae760.md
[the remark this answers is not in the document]
WilcoFiers: Not a single one of them respects the specs JY. We have three options here:

1. Follow the spec, ignoring the fact that not a single browser actually works the way it is spec'ed. This was the first version, and the ACT TF we've had pushback on that, including from AG and the ACT TF. So that option is out.

2. Write a rule to the lowest common denominator; if it works anywhere at all, pass it. Which as far as I can see, means iframes never fail because Chrome+JAWS just doesn't treat iframes as UICs. We would just have to deprecate the rule. That's doens't seem like a particularly viable option either.

3. We intentionally leave out some of the major browsers, even though none of them work according to spec. That seems like picking favorites, which doens't seem fair.

4. Write a rule that, ignoring spec, makes sure that iframes are accessible in all major browsers and screen readers. That is this version. If this rule says its a fail, it is definitely a fail somewhere. Not in all of them mind you, but definitely somewhere. This is in line with "accessibility support" as WCAG defines it, and seems to be what Mark was suggesting we do too.

If we want an iframe accessible name rule at all, I think this is the only viable option. It's a little controversial. But it's the least controversial option I could figure out. Open to other ideas, but I'm not on board with picking favorites or ditching the rule.

kasperisager (changes requested):
I also share @Jym77 concerns about singling in on these very specific combinations of UA/AT. I'd be much more comfortable with leaving the specifics of how an accessible name is computed as an unspecified implementation detail and simply refer to it as "whatever label is exposed for an object in the accessibility tree".

_rules/iframe-non-empty-title-attribute-cae760.md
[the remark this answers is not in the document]
Jym77: A solution I would be much more confortable with:
* Keep "each target has a non-empty accessible name" as expectation. At this point we don't really care where the accessible name comes from.
* Remove examples that use `aria-label`, `aria-labelledby` or anything other than `title`. These examples behave differently on different UA/AT, making them fairly annoying for us whether they pass or fail, since we don't want to take sides.
* Add an Accessibility Support in the line of "some UA+AT only use `title` for accessible name of `iframe`, while other also use `aria-label` and `aria-labelledby`, depending on UA+AT, the same target may pass or fail." + maybe strengthen with "it is recommended to use `title` only since this will be correctly handled by all UA+ATs"
* + likely copy the AS note in the def of accessible name, since it is a discrepancy about accessible name computation and behaviours that differ from the specs.

WilcoFiers:
@kasperisager "whatever label is exposed for an object in the accessibility tree" can't be implemented consistently. Rules have to harmonise. We can't publish a rule that is ambiguous about something like if aria-label is sufficient for iframes. And while in all other cases we'd go by the spec, since not a single browser follows the specs when it comes to iframes, that's just not an option.

I didn't pick a couple of UA+AT combinations at random for this. It's my understanding this problem comes from how the accessibility tree is built, which is done by the render engine. There are only three of them actively maintained, Chromium, Webkit and Gecko. All three are represented in this rule.

kasperisager:
> "whatever label is exposed for an object in the accessibility tree" can't be implemented consistently.

Of course it can, though it would seem we have differing opinions on what "consistent" means. Using DOM, I'd answer this question using `Element#computedName`. Using the browser's dev tools, I'd investigate the accessibility properties for the given element. Using a screen reader, I'd check whatever name is announced. All of those will consistently tell me which name is exposed for a given object in the accessibility tree. It's absolutely true that the precise answer might depend on the UA or AT in use, but I see that as a feature, not a bug. If I want to check if the rule holds across _all_ technologies in my support baseline, I need to test using those technologies.

If nothing else, let's agree to disagree 😄 I don't approve of these changes so I'll leave this feedback here for now and unassign myself.

kasperisager:
Also, case in point: We consider the accessibility tree an input aspect as its construction varies across UAs. Does that also go against the goal of harmonisation and consistency? I certainly don't believe so.

WilcoFiers:
I'm going to need info from the ACT Task Force on what to pass and what to fail for iframes before I can work on this again. I can reopen or start anew once I have their info.

PR #1465, opened 2020-09-30
Clear up link-acc-name examples

WilcoFiers:
These changes were requested on yesterday's AGWG call. Here is the survey of it: https://www.w3.org/2002/09/wbs/35422/act-rules-2020-09/results#xq3 (only accessible to AGWG members)

Need for Final Call: 1 week

---

## How to Review And Approve


_rules/link-non-empty-accessible-name-c487ae.md
```diff
@@ -180,26 +180,42 @@ This `area` element has a [semantic role][] of `link` and an [accessible name][]
...
 
 ```html
-<a href="http://www.w3.org/WAI"><img src="/test-assets/shared/w3c-logo.png" role="none"/></a>
+<a href="http://www.w3.org/WAI"></a>
 ```
 
 #### Failed Example 2
 
-This `a` element with a decorative image has an empty [accessible name][].
+This `a` element with an image has an empty [accessible name][]. The image is decorative because it has an empty `alt` [attribute value][].
```
Jym77, line 188: I'm a bit uneasy on the formulation "the image is decorative because it has empty `alt` (or role of `none`)" as it feels like it reverses the causality: the image is decorative because it brings no information. It has an empty `alt` because it is decorative (and need to be marked as such to UA/AT).

But this is irrelevant to this PR, so not really a problem.

Jym77 (approved):

_rules/link-non-empty-accessible-name-c487ae.md
WilcoFiers: How about if I used "presentational" instead of "decorative"?

Jym77: I'm not a huge fan either 🤔 

I see "decorative" or "presentational" as almost synonyms. And I understand them as intrinsic property of the image, independantly of the techniques used to expose it or not.
Also, the rule (Applicability and Expectation) does not mention being decorative as relevant to it, so it is a bit weird to see that popping up in the description.
I'm not sure exactly what was the request from the AGWG (no access to the survey), so it's hard to figure out what would make everybody happy…

Maybe,
> The image is decorative and is marked as such with an empty `alt` attribute value.

I guess the problem we want to illustrate here is that it is possible to use a decorative image as content for a link, but in that case, author must provide a name for the link (since it is not provided by content) for AT users.

What about a fully explicit version like
> This `a` element has an empty accessible name. Its content is a decorative image which is correctly marked as such with an empty `alt` attribute value. However, there is nothing else to give a name to the link resulting in an empty accessible name.

(i.e. the error is not the empty `alt` but the lack of `aria-label` or such on the link to compensate for decorative content, so we don't say that image links can't be decorative; but we say that if they are decorative, then AT users need something)

maryjom: An idea for the description: "This link has an empty accessible name because the link's image is marked as decorative and link text or an ARIA label has not been provided." 

Saying generically "ARIA label" to cover both aria-label and aria-labelledby.  Would this work?

bruce-usab: +1 to @Jym77 suggestion for "The image is decorative and is marked as such with an empty alt attribute value."

WilcoFiers: Going with @Jym77 's (shortest) suggestion.

carlosapaduarte (approved):

jeeyyy (approved):

PR #1467, opened 2020-09-30
explicit-svg-img, add <text> example

WilcoFiers:
This example addresses a question that came up during yesterday's AGWG call. 

Need for Final Call: 1 week

---

## How to Review And Approve


kasperisager (approved):

_rules/explicit-SVG-image-non-empty-accessible-name-7d6734.md
```diff
@@ -133,6 +133,20 @@ This `circle` element has an [explicit role][] of `graphics-symbol` but has no [
 </svg>
 ```
 
+#### Failed Example 4
+
+This `svg` element has an [explicit role][] of `img` has an empty (`""`) [accessible name][]. The SVG `text` element is not used in computing the accessible name.
```
Jym77, line 138:
```suggestion
This `svg` element with an [explicit role][] of `img` has an empty (`""`) [accessible name][]. The SVG `text` element is not used in computing the accessible name.
```
(avoid double verb)

Jym77 (changes requested):

Jym77 (approved):

jeeyyy (approved):

PR #1469, opened 2020-10-05
visible-label-in-acc-name: editorial updates

WilcoFiers:
See https://github.com/w3c/wcag-act/issues/491

Need for Final Call: none

---

## How to Review And Approve

### DIFF
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -50,7 +50,7 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 This rule applies to elements with a [widget role][] that [support name from content][supports name from content]. This includes the following: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
 
-The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that uses text characters as symbols, such as using "x" to mean "close". Because there is no definition for symbolic text characters, this rule uses non-text content instead.
+The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that uses text characters as symbols, such as using "x" to mean "close". This rule considers them as "characters expressing non-text content". Unicode emojis are another example of non-text content, although these are not "symbolic text characters".
 
 - [Understanding Success Criterion 2.5.3: Label in Name][understand253]
 - [G208: Including the text of the visible label as part of the accessible name](https://www.w3.org/WAI/WCAG21/Techniques/general/G208)
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -50,7 +50,7 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 This rule applies to elements with a [widget role][] that [support name from content][supports name from content]. This includes the following: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
 
-The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that text characters as symbols, such as using "x" to mean "close". Because there is no definition for symbolic text characters, this rule uses non-text content instead.
+The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that uses text characters as symbols, such as using "x" to mean "close". Because there is no definition for symbolic text characters, this rule uses non-text content instead.
 
 - [Understanding Success Criterion 2.5.3: Label in Name][understand253]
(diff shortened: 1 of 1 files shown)


_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -48,9 +48,11 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 ## Background
 
-This rule applies to elements with a [widget roles][] that [support name from content][supports name from content]. This includes the following: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
+This rule applies to elements with a [widget role][] that [support name from content][supports name from content]. This includes the following: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
 
-- [Understanding Success Criterion 2.5.3: Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html)
+The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that text characters as symbols, such as using "x" to mean "close". Because there is no definition for symbolic text characters, this rule uses non-text content instead.
```
Jym77, line 53:
```suggestion
The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that uses text characters as symbols, such as using "x" to mean "close". Because there is no definition for symbolic text characters, this rule uses non-text content instead.
```

Jym77, line 53: I'm not a fan of the last sentence. Especially "because there is no definition of foo" looks both a bit lazy (what don't we write the definition if we need it) and not entirely true (the Understanding document of 1.4.5 has an almost definition of this, and even the UC of 2.5.3 is giving a lot of details on why this mean).
Additionally, the Expectation is already saying "except for characters used to express non-text content" which is almost a definition of "symbolic characters"…

I do agree that explaining the difference in wording and giving example makes sense.
What about replacing the last sentence by something like
> This rule considers them as "characters expressing non-text content".

or something in this line?

Maybe also adding that "characters expressing non-text content" also includes emojis (at least for some definition of characters, but emojis are Unicode codepoints so I say they qualify as characters). That would point out that our name is a bit more generic than "symbolic characters" and justify why we use this instead.

Jym77 (changes requested):

_rules/visible-label-in-accessible-name-2ee8b8.md
WilcoFiers: Thanks for the suggestions. I've taken another stab at it. Let me know!

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -48,9 +48,11 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 ## Background
 
-This rule applies to elements with a [widget roles][] that [support name from content][supports name from content]. This includes the following: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
+This rule applies to elements with a [widget role][] that [support name from content][supports name from content]. This includes the following: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
```
Jym77, line 51: "This includes" sounds like the list is not exhaustive. AFAICR, this list is exhaustive, so let's take the opportunity to
```suggestion
This rule applies to elements with a [widget role][] that [support name from content][supports name from content], namely: `button`, `checkbox`, `gridcell`, `link`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch`, `tab`, `treeitem`.
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -48,9 +48,11 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
...
 
-- [Understanding Success Criterion 2.5.3: Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html)
+The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that uses text characters as symbols, such as using "x" to mean "close". This rule considers them as "characters expressing non-text content". Unicode emojis are another example of non-text content, although these are not "symbolic text characters".
```
Jym77, line 53:
```suggestion
The understanding document of [2.5.3 Label in Name][understand253] use the term "symbolic text characters" to refer to a type of [non-text content][] that uses text characters as symbols, such as using "x" to mean "close". This rule considers them as "characters expressing non-text content". Unicode emojis are another example of characters expressing non-text content, although these are not "symbolic text characters".
```

Jym77 (changes requested):

_rules/visible-label-in-accessible-name-2ee8b8.md
WilcoFiers: I don't really agree with that. Much prefer the current proposal.

Jym77 (approved):

carlosapaduarte (approved):

daniel-montalvo (approved):

PR #1473, opened 2020-10-08
"ARIA required context role" and "ARIA required owned elements" (ff89c9 / bc4a75): Update note on DPUB due to upcoming changes in 1.1

Jym77:
The inapplicable example about ARIA DPUB role was failing the requirement because the `doc-biblioentry` role was not owned by its required `doc-bibliography`. This resulted as implementation considering the ARIA DPUB module (such as Alfa) to be rejected as they were failing this example.

Closes issue(s):

- N/A

Need for Final Call:
This will require a 1 week Final Call (updating an example)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -206,14 +206,11 @@ There is no element with an [explicit role][].
 
 #### Inapplicable Example 3
 
-This `section` element with an [explicit role][] of `doc-biblioentry` has a role from the [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0][dpub 1.0], not the [WAI-ARIA 1.1 Recommendation][aria 1.1].
+This `section` element with an [explicit role][] of `doc-abstract` has a role from the [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0][dpub 1.0], not the [WAI-ARIA 1.1 Recommendation][aria 1.1].
 
 ```html
-<section role="doc-bibliography">
-	<h1>Cited Works</h1>
-	<p role="doc-biblioentry" id="b8cab5dd-bc24-459c-9858-7afa9da69b64">
-		John Steinbeck, The Grapes of Wrath (New York: The Viking Press, 1939)
-	</p>
+<section role="doc-abstract" aria-label="Abstract">
+	<p>Accessibility of web content requires semantic information about widgets, structures, and behaviors …</p>
 </section>
 ```
 
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -43,7 +43,7 @@ This rule assumes that the `role` attribute is used to give a [semantic role][] 
 
 ## Background
 
-The applicability of this rule is limited to the [WAI-ARIA 1.1 Recommendation][aria 1.1] roles, since there are unresolved issues with how [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0][dpub 1.0] uses role inheritance to define the [required context roles][], which makes it deviate from the model defined in [WAI-ARIA 1.1][aria 1.1]. The [WAI-ARIA Graphics Module](https://www.w3.org/TR/graphics-aria-1.0/) does not include any [required context roles][].
+The applicability of this rule is limited to the [WAI-ARIA 1.1 Recommendation][aria 1.1] roles. The [WAI-ARIA Graphics Module][] does not include any [required context roles][]. The [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0][dpub 1.0] only has two roles with [required context roles][] (`doc-biblioentry` and `doc-endnote`); both of them have issues with their use of role inheritance, and both of them are deprecated in the [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.1][dpub 1.1] editor's draft.
 
 An example of an element that has an [implicit semantic role][] that is identical to its [explicit semantic role][] is a `<li role="listitem">` element. These elements are not applicable because they have extra requirements and should thus be checked separately.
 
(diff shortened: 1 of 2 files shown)


kasperisager (approved):

carlosapaduarte (approved):

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -211,11 +211,9 @@ This `section` element with an [explicit role][] of `doc-biblioentry` has a role
...
 <section role="doc-bibliography">
 	<h1>Cited Works</h1>
-	<div role="list">
-		<p role="doc-biblioentry" id="b8cab5dd-bc24-459c-9858-7afa9da69b64">
-			John Steinbeck, The Grapes of Wrath (New York: The Viking Press, 1939)
-		</p>
-	</div>
+	<p role="doc-biblioentry" id="b8cab5dd-bc24-459c-9858-7afa9da69b64">
+		John Steinbeck, The Grapes of Wrath (New York: The Viking Press, 1939)
+	</p>
```
WilcoFiers, line 216: This should not be changed. The point of this example is to demonstrate that "owned by" as used in ARIA means any descendant, not just the child. This example comes directly from the DPUB ARIA spec:

https://www.w3.org/TR/dpub-aria-1.0/#doc-biblioentry

This confusion about "owned by" including all descendants, and not just the children and direct references of aria-owns is a known issue in ARIA. See: w3c/aria#1161, w3c/aria#748.

WilcoFiers (changes requested):

_rules/aria-required-context-role-ff89c9.md
Jym77: 🙈 That example has so many problems I want to unsee it…

Fortunately (maybe), `doc-biblioentry` is deprecated in DPUB-ARIA 1.1 (https://w3c.github.io/dpub-aria/#doc-biblioentry) precisely because of the wonkyness it creates with the `list` role (https://github.com/w3c/dpub-aria/issues/15 and https://github.com/w3c/dpub-aria/pull/22). This is very recent changes!

I guess this example should be completely revamped. As far as I can tell, there is no more required context roles in DPUB-ARIA 1.1. I don't think it is a good idea to keep an example showing a known problem in the specs which is already fixed in the next version.

If the point of the example is to show the "descendant" meaning of "owned by", we need to find a non DPUB example.

If the point of the example is to show that DPUB roles are ignored, we can take any, but they have no required context role…

Maybe we should update the Background note saying that DPUB-ARIA 1.1 does not have any required context role (like Graphics ARIA) which justifies our choice to restrict ourselves to ARIA 1.1 roles.

Jym77:
I've updated the background explaining why we focus on ARIA. Now saying that DPUB 1.1 has no role with required context role. Copied that note to "required owned element" rule where the same thing happen. Essentially, the "unresolved issues" that this note mentioned have been resolved by deprecating the offending elements.

The example has been updated accordingly.

Dismissing all reviews, since this is a significant change of scope.

kasperisager (approved):

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -43,7 +43,7 @@ This rule assumes that the `role` attribute is used to give a [semantic role][]
 
 ## Background
 
-The applicability of this rule is limited to the [WAI-ARIA 1.1 Recommendation][aria 1.1] roles, since there are unresolved issues with how [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0][dpub 1.0] uses role inheritance to define the [required context roles][], which makes it deviate from the model defined in [WAI-ARIA 1.1][aria 1.1]. The [WAI-ARIA Graphics Module](https://www.w3.org/TR/graphics-aria-1.0/) does not include any [required context roles][].
+The applicability of this rule is limited to the [WAI-ARIA 1.1 Recommendation][aria 1.1] roles. The [WAI-ARIA Graphics Module][] does not include any [required context roles][]. The [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.0][dpub 1.0] only has two roles with [required context roles][] (`doc-biblioentry` and `doc-endnote`); both of them have issues with their use of role inheritance, and both of them are deprecated in the [Digital Publishing WAI-ARIA Module (DPUB ARIA) 1.1][dpub 1.1] editor's draft.
```
WilcoFiers, line 46: DPUB-ARIA 1.1 hasn't even made it to first draft yet. I don't think we should be referencing it.

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -206,16 +206,11 @@ There is no element with an [explicit role][].
...
 ```html
-<section role="doc-bibliography">
-	<h1>Cited Works</h1>
-	<div role="list">
-		<p role="doc-biblioentry" id="b8cab5dd-bc24-459c-9858-7afa9da69b64">
-			John Steinbeck, The Grapes of Wrath (New York: The Viking Press, 1939)
-		</p>
-	</div>
+<section role="doc-abstract" aria-label="Abstract">
+	<p>Accessibility of web content requires semantic information about widgets, structures, and behaviors …</p>
```
WilcoFiers, line 213: I disagree with changing this. The point of this example is to show a DPUB role that has a required context that is inapplicable.

WilcoFiers (dismissed):
I think this is getting ahead of things, and I don't think the new example is a proper replacement of the current one.

_rules/aria-required-context-role-ff89c9.md
Jym77: But in that example, the `<div role="list">` owns no element with a role of `listitem`. Because role inherintance doesn't work like that and ARIA DPUB 1.0 is broken on that point, and this is known, and this is the reason we advance for excluding DPUB roles. See also https://github.com/w3c/dpub-aria/issues/15 for more details.

So, we have an Inapplicable example that does not satisfy the Accessibility Requirement (1.3.1), which goes directly against our test cases design guidance… https://github.com/act-rules/act-rules.github.io/blob/develop/pages/design/test-cases.md
> This requires that **passed and inapplicable test cases should satisfy the accessibility requirements** of the rule

(emphasise is not mine)

So that example should be killed anyway.

WilcoFiers: Aaahhh. I see what you mean. Sorry, took me a while. Yeah that's a fair point. That `list` is invalid here.

WilcoFiers (approved):

carlosapaduarte (approved):

Jym77:
Final Call ends on November 4.

Jym77:

PR #1476, opened 2020-10-08
Update PR template

### LINKED ISSUE #1468: Can we skip final call on example updates?
The most common reason for us to go through a final call is that an example was updated. The original reason for us to do this was to allow implementors the time to check their implementation against the updated example. I'm not aware of anyone who does that. Certainly I haven't seen any feedback from an implementor who did this.

Most of the comments we get are editorial in nature. They are suggestions for improving the language. I think there is a mismatch between what we put into a final call, and what we are actually getting feedback on.

I bring this up, partially to revisit if we're doing final calls on the right things, and partially because I'd like to see if we can streamline the process a little more. There might be things we can do more efficiently without losing quality.

carlosapaduarte:
Update the PR template so that it's clear that editorial changes to the applicability, expectation or example sections do not require a Final Call period. Also, add a request for PR authors to leave the "How to Review and Approve" section in their PR description.

Closes issue(s):

- closes #1468 

Need for Final Call:
This will not require a Final Call 
---

## How to Review And Approve

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,9 +6,10 @@ Closes issue(s):
 
 Need for Final Call:
 << choose one of the following and remove the rest >>
+<< leave the "How to Review and Approve" section in your pull request description >>
 << check https://act-rules.github.io/pages/design/process/#final-call-aka-call-for-consensus-cfc >>
 This can be merged with 1 approval << choose reason: editorial changes to website/test code, adding new contributor, other (explain). >>
-This will not require a Final Call << choose reason(s): editorial changes, changes to assumptions, background, accessibility support, change to website/test code (not rule), other (explain). >>
+This will not require a Final Call << choose reason(s): editorial changes (including to the applicability, expectation or examples section), changes to assumptions, background, accessibility support, change to website/test code (not rule), other (explain). >>
 This will require a 1 week Final Call << small changes affecting a small number of test cases, if in doubt do not use this. >>
 This will require a 2 weeks Final Call << new rule, or substantial changes affecting a large number of test cases, if in doubt, use this. >>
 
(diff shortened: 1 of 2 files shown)


WilcoFiers (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,9 +6,10 @@ Closes issue(s):
 
 Need for Final Call:
 << choose one of the following and remove the rest >>
+<< leave the "How to Review and Approve" section in your pull request description >>
```
Jym77, line 9: Seems weird to have this as one of the "Need for Final Call" options.

Jym77 (changes requested):

pages/design/process.md
```diff
@@ -159,7 +159,7 @@ This includes, but might not be limited to changes that changes, extends or limi
 
 These changes are considered non-substantial and will not require a "Final call" before being published:
 
-- Editorial changes, that does not change the meaning of a rule (also if it's in the Applicability or Expectations)
+- Editorial changes, that does not change the meaning of a rule (also if it's in the Applicability, Expectations or Examples)
```
daniel-montalvo, line 162:
```suggestion
- Editorial changes, that do not change the meaning of a rule (also if it's in the Applicability, Expectations or Examples)
```

daniel-montalvo (approved):
Just a small plural that I think should be there as we are speaking about changes.

.github/PULL_REQUEST_TEMPLATE.md
carlosapaduarte: Indeed. Moved it to the instructions in the PR Etiquette section

Jym77 (approved):

carlosapaduarte:
Update PR template approved. Merging

PR #1478, opened 2020-10-09
scrollable-keyboard-acc: add accessibility support note

WilcoFiers:
This feedback was provided by the ACT Task Force. Firefox it turns out, will add elements with a scrollbar to the sequential focus navigation.

Need for Final Call: none

---

## How to Review And Approve

### DIFF
--- a/_rules/scrollable-element-keyboard-accessible-0ssw9k.md
+++ b/_rules/scrollable-element-keyboard-accessible-0ssw9k.md
@@ -45,7 +45,7 @@ This rule assumes that all [scrollable elements][scrollable] with visible conten
 
 ## Accessibility Support
 
-Some browsers will automatically make any scrollable element focusable to ensure keyboard accessibility. However, the browser does not included these elements in [sequential focus navigation][] when it has a negative number as a tabindex [attribute value][].
+Some browsers will automatically make any scrollable element focusable to ensure keyboard accessibility. However, the browser does not include these elements in [sequential focus navigation][] when it has a negative number as a tabindex [attribute value][].
 
 
 ## Background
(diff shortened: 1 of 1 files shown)


_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -45,7 +45,7 @@ This rule assumes that all [scrollable elements][scrollable] with visible conten
 
 ## Accessibility Support
 
-_No accessibility support issues known._
+Some browsers will automatically make any scrollable element focusable to ensure keyboard accessibility. Those elements become inaccessible only if they have negative number as a tabindex [attribute value][].
```
maryjom, line 48: Suggest editing this to read: 

However, the browser does not automatically make the scrollable element focusable when it has a negative number as a tabindex [attribute value][].

maryjom (changes requested):
See my comment for suggested changes.

carlosapaduarte (changes requested):
Agree with @maryjom suggestion. Will approve after the change.

carlosapaduarte (approved):

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -45,7 +45,8 @@ This rule assumes that all [scrollable elements][scrollable] with visible conten
 
 ## Accessibility Support
 
-_No accessibility support issues known._
+Some browsers will automatically make any scrollable element focusable to ensure keyboard accessibility. However, the browser does not included these elements in [sequential focus navigation][] when it has a negative number as a tabindex [attribute value][].
```
jeeyyy, line 48: worth linking scrollable element to the glossary item

jeeyyy, line 48:
```suggestion
Some browsers will automatically make any scrollable element focusable to ensure keyboard accessibility. However, the browser does not include these elements in [sequential focus navigation][] when it has a negative number as a tabindex [attribute value][].
```

jeeyyy (dismissed):

Jym77 (approved):

PR #1479, opened 2020-10-13
aria-required-owned: remove implicit exception

### LINKED ISSUE #1413: ARIA required owned, do we need the no implicit requirement? bc4a75
Got some more feedback on the following line, this time from the ACT TF:

> the element has an implicit semantic role that is identical to its explicit semantic role; or

Now that I've seen what the content model rules look like, this might not be necessary anymore. Need to think through the different cases.

WilcoFiers:
Closes #1413

Need for Final Call: 1 week

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-owned-element-bc4a75.md
+++ b/_rules/aria-required-owned-element-bc4a75.md
@@ -48,8 +48,6 @@ If the [explicit semantic role][] on the target element is incorrectly used, and
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


Jym77 (approved):

jeeyyy (dismissed):

carlosapaduarte (dismissed):

Jym77 (changes requested):
The Background note at https://github.com/act-rules/act-rules.github.io/pull/1479/files#diff-b67a2f24a6ac7b3ff85d5593f4f29edc3007d36f94ff5200ae39fd8ae2f0048eL52 must be removed or somehow adapted since it is now incorrect.

Jym77 (approved):

carlosapaduarte (approved):

jeeyyy (approved):

PR #1481, opened 2020-10-13
link is descriptive, remove unnecessary quotes

WilcoFiers:
Need for Final Call: None

---

## How to Review And Approve

### DIFF
--- a/_rules/descriptive-link-5effbb.md
+++ b/_rules/descriptive-link-5effbb.md
@@ -31,7 +31,7 @@ acknowledgments:
 
 This rule applies to any HTML or SVG element which
 
-- has the [semantic role][] of ['link'][link] or a [semantic role][] that inherits from the ['link'][link] role; and
+- has the [semantic role][] of [link][link] or a [semantic role][] that inherits from the [link][link] role; and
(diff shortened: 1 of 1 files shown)


_rules/descriptive-link-5effbb.md
```diff
@@ -31,7 +31,7 @@ acknowledgments:
 
 This rule applies to any HTML or SVG element which
 
-- has the [semantic role][] of ['link'][link] or a [semantic role][] that inherits from the ['link'][link] role; and
+- has the [semantic role][] of [link][link] or a [semantic role][] that inherits from the [link][link] role; and
```
Jym77, line 34: I guess the intention was rather
```suggestion
- has the [semantic role][] of [`link`][link] or a [semantic role][] that inherits from the [`link`][link] role; and
```
Also, this might soon benefit from using https://github.com/act-rules/act-rules.github.io/blob/inline_link_in_paragraph_is_distinguishable/pages/glossary/semantic-link.md (which has the same problem).

Jym77 (approved):

jeeyyy (approved):

ShadowBB (approved):

PR #1483, opened 2020-10-13
Link to related rules in background

### LINKED ISSUE #1402: Link to related rules in the background
This came up recently in the ACT TF. Some rules are closely related to others. Especially if rules are about the same type of element and the same type of requirement, it would be good to link between them.

For example, these three rules should probably all link to each other:
- [HTML page has lang attribute](https://act-rules.github.io/rules/b5c3f8)
- [HTML page lang attribute has valid language tag](https://act-rules.github.io/rules/bf051a)
- [HTML page lang and xml:lang attributes have matching values](https://act-rules.github.io/rules/5b7ae0)

That also serves as an example of something that I think should not be linked. "Element with lang attribute has valid language tag" is about a different element and a different SC, so even though they are "related" I don't think we should link to that.

Opening this as an issue so that we can discuss before we start to do this.

WilcoFiers:
Closes #1402

Need for Final Call: 1 week

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -46,6 +46,7 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 ## Background
 
+- [ARIA state or property has valid value](https://act-rules.github.io/rules/6a7281)
 - [Understanding Success Criterion 4.1.1: Parsing](https://www.w3.org/WAI/WCAG21/Understanding/parsing.html)
 - [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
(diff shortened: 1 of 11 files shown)


Jym77 (approved):

jeeyyy (approved):
This is super useful.

carlosapaduarte (approved):
Just wondering about the criteria (don't know if you're planning on creating subsequent PRs). If you have "link has non-empty accessible name" and "link in context is descriptive" why not do the same for "Image has non-empty accessible name" and "Image accessible name is descriptive"?

Other suggestions:
- "Links with identical accessible names have equivalent purpose" and "Links with identical accessible names and context serve equivalent purpose"
- "Device motion based changes to the content can also be created from the user interface" and "Device motion based changes to the content can be disabled"

If you agree that these would also fit the criteria, I would be happy to open a PR

EmmaJP:
Looks like a good solution. I haven't considered which rules are being linked to each other. However, I wondered if anyone else thought linked rules should come after other background links rather than before.

PR #1010, opened 2019-11-07
New Rule: inline link in paragraph is distinguishable be4d0c

carlosapaduarte:
New rule 

Blocked: waiting for decision on how to handle states in rules

Need for Final Call:
This will require a 2 weeks Final Call << new rule, or substantial changes affecting a large number of test cases, if in doubt, use this. >>

---

### **When creating PR:**

- Make sure you requesting to **pull a branch** (right side) to the `develop` branch (left side).

### **After creating PR:**

- Add yourself (and co-authors) as "Assignees" for PR.
- Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- Close the issue that the PR resolves (and make sure the issue is referenced in the top of this comment)
- Optionally request feedback from anyone in particular by assigning them as "Reviewers".

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Final Call period. In case of disagreement, the longer period wins.

PR #1485, opened 2020-10-14
semantic-link dfn: editorial changes

WilcoFiers:
Minor change, related to #1481

Need for Final Call: none

---

## How to Review And Approve


pages/glossary/semantic-link.md
```diff
@@ -5,7 +5,7 @@ unambiguous: true
 objective: true
 ---
 
-A semantic link is any HTML element which has the [semantic role][] of ['link'][link] or a [semantic role][] that inherits from the ['link'][link] role.
+A semantic link is any HTML element which has the [semantic role][] of [`link`][] or a [semantic role][] that inherits from the [`link`][] role.
```
carlosapaduarte, line 8: I think you need to keep these because [`link`] is not the same as [link]

```suggestion
A semantic link is any HTML element which has the [semantic role][] of [`link`][link] or a [semantic role][] that inherits from the [`link`][link] role.
```

carlosapaduarte (changes requested):

pages/glossary/semantic-link.md
WilcoFiers: Pretty sure that would work, but I'm not bothered either way.

carlosapaduarte (approved):
I also thought it worked, but looking at the markdown viewer in github it doesn't show as a link, so...

Jym77:
Should that rather been done as part of #1010 which is still open and introduces that definition?

Jym77 (changes requested):
Don't onboard all other files

PR #1490, opened 2020-10-16
move iframe-has-name file

WilcoFiers:
This should resolve https://github.com/act-rules/act-rules.github.io/pull/1464/files#r502239615

Need for Final Call: None

---

## How to Review And Approve


carlosapaduarte (approved):

Jym77 (approved):

nitedog (approved):

WilcoFiers:
Going to close this. Don't think we need at right now.

PR #1493, opened 2020-10-21
new rule: Element with presentational children has no focusable content

WilcoFiers:
Every focusable element must be included in the accessible name, so that assistive technologies can announce what they are and what they do. Because descendants of elements with presentational children, such as buttons are not included in the accessibility tree (in some browsers), those elements will fail SC 4.1.2 if they are focusable.

I've tested how much support there is for presentational children. Here's what I found:

- Safari: full support. If you put a heading inside a button, the heading vanishes from the accessibility tree
- Chrome: Almost no support. The only role where all descendants are made presentational is `img`.
- Firefox: No support at all. Firefox never makes children presentational, not even for the `img` role.

Need for Final Call: 2 weeks

---

## How to Review And Approve


_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+#### Failed Example 3
+
+This element with `menuitemcheckbox` role has a checkbox as a child. Because the checkbox is not disabled, it is included in [sequential focus navigation][].
+
+```html
+<ul role="menu">
+	<li role="menuitemcheckbox" aria-checked="true">
+		<input type="checkbox" checked />
```
WilcoFiers, line 116:
```suggestion
		<input type="checkbox" checked role="none" />
```

Jym77:
@WilcoFiers did you mean to assign or to request reviews?

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
+---
+id: 307n5z
+name: Element with presentational children has no focusable content
+rule_type: atomic
+description: |
+  This rule checks that elements with a role that makes it children presentational do not contain focusable elements.
```
Jym77, line 6:
```suggestion
  This rule checks that elements with a role that makes its children presentational do not contain focusable elements.
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
...
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Wilco Fiers
+---
+
+## Applicability
+
+The rule applies to any HTML or SVG element with a [semantic role][] that defines its [children][child] to be [presentational children][].
```
Jym77, line 28:
```suggestion
This rule applies to any HTML or SVG element with a [semantic role][] that defines its [children][child] to be [presentational children][].
```
https://www.w3.org/2020/10/22-act-r-minutes.html#item04

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
...
+_There are currently no assumptions_
+
+## Accessibility Support
+
+Several major browsers ignore the WAI-ARIA requirements on [presentational children][] for most or sometimes all roles. Because some browsers do, and others do not implement presentational children, there can be big differences between browsers.
+
+## Background
+
+Elements with a [semantic role][] that has [presentational children][] will not have any descendants in the accessibility tree. If any of those descendants are included in [sequential focus navigation][], this causes the focus to land on an element that has no corresponding node in the [accessibility tree][]. The result is that there is no programmatic name or role available for assistive technologies. There are other problems that can come from [presentational children][] too. These must be tested separately.
+
```
Jym77, line 45: Wondering whether we should include a list of roles with presentational children here (like we do for some of the form field rules targeting "role that inherit from input/select and have no required context that inherit from input/select".

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
...
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+Neither `button` element has no [descendants][] that are included in [sequential focus navigation][].
+
+```html
+<button>Save</button> <button aria-label="save options" aria-expanded="false">▼</button>
```
Jym77, line 59: Shouldn't that be 2 separate examples?

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
...
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)
+- [WAI-ARIA 1.1 Presentational Children][presentational children]
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+Neither `button` element has no [descendants][] that are included in [sequential focus navigation][].
```
Jym77, line 56:
```suggestion
None of these `button` elements has [descendants][] that are included in [sequential focus navigation][].
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
...
+	</span>
+	<a href="/terms">terms of service</a>
+</p>
+```
+
+#### Passed Example 3
+
+This element with `menuitemcheckbox` role has an `input` element as a descendant. Because the `input` is disabled it is not included in [sequential focus navigation][].
+
+**Note**: The `input` checkbox has a `role` [attribute value][] of `none` to ensure it is ignored by browsers that to not support [presentational children][].
```
Jym77, line 79:
```suggestion
**Note**: The `input` checkbox has a `role` [attribute value][] of `none` to ensure it is ignored by browsers that do not support [presentational children][].
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,142 @@
...
+		Sort by Last Modified
+	</li>
+</ul>
+```
+
+### Failed
+
+#### Failed Example 1
+
+The `button` element has a [child][] `span` element that is included in [sequential focus navigation][].
```
Jym77, line 94:
```suggestion
This `button` element has a [child][] `span` element that is included in [sequential focus navigation][].
```

Jym77 (dismissed):
Looks good. Some minor typos.

_rules/presentational-children-no-focusable-content-307n5z.md
WilcoFiers: Maybe, lets leave this open, see what other reviewers think.

Jym77 (approved):

ShadowBB (approved):
I will approve it without a list, but I think adding the list would benefit the rule:

button
checkbox 
img
math
menuitemcheckbox
menuitemradio
option
progressbar
radio
scrollbar
separator
slider
switch
tab

WilcoFiers:
@Jym77 @ShadowBB I added the list to the background, same way we did in "visible label is part of accessible name".

ShadowBB:
Looks good!

Jym77:
Some more thoughts on this rule. Sorry, half-good-ish, half-fairly-bad news 😕 

1. I think this rule can actually be merged with [Element with aria-hidden has no focusable content](https://act-rules.github.io/rules/6cfa84). Both of them check the same idea: that something which is not in the accessibility tree but can get focused is a bad idea.

   Rather than individually checking different ways of not being in the accessibility tree (`aria-hidden`, presentational children, …) and then needing to add more rules when we find more (and take into account new Accessibility Support for each), we could just upgrade it into one big generic rule: "element not in the accessibility tree is not focusable" (or "focusable element is included in the accessibility tree", it can go both ways depending what we want… (I have a small preference for the later as it likely creates less test targets)).

   Such a generic rule would encompass this one and the `aria-hidden` one, would need less precaution in  the Accessibility Support (if an UA doesn't implement presentational children, we don't care that much, we just look at whatever accessibility tree it provides us), will handle more cases (for example, testing on an UA with bad implementation of presentational conflict resolution would flag stuff, …) and is somewhat more future-proof. So, I think this is a good move to do (the main negative point being that it counts as one less rule for WAI tools projects, but that's minor in my opinion…)

2. Now for the bad stuff… It also means that this rule is hit by #643 Namely, a case like ` <button onfocus="this.blur(); focusNext()">Foo</button></img>` has a `button` which is both in sequential tab order and a presentational child, so it would fail the rule but it is not really a problem because scripting makes sure it never actually gets the focus.

   So, at the very least, we need to already implement the fix worked on in #1444. As you now from the discussion at https://github.com/act-rules/act-rules.github.io/pull/1444#discussion_r486157010 I'm very reluctant about how the fix is currently worked (essentially because it tries to solve by static means (HTML & CSS) a dynamic problem (JS)). Which is why this is bad news, there seems to be a strong blocker on these two rules (or on the merged one) right now 😞

Jym77 (dismissed):
See https://github.com/act-rules/act-rules.github.io/pull/1493#issuecomment-731094293

WilcoFiers:
@Jym77 We'll need to discuss this on Monday's call.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,144 @@
...
+		Sort by Last Modified
+	</li>
+</ul>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This `button` element has a [child][] `span` element that is included in [sequential focus navigation][].
```
daniel-montalvo, line 96:
```suggestion
This `button` element has a [child][] `span` element. Because the `span` element has a `tabindex` attribute with value of `0`, it is included in [sequential focus navigation][].
```
I think it is worth explaining why in the description.

daniel-montalvo (approved):
Just a minor suggestion in one  example description.

Jym77 (dismissed):
There is still the second problem I mentioned in https://github.com/act-rules/act-rules.github.io/pull/1493#issuecomment-731094293:
> 2\. Now for the bad stuff… It also means that this rule is hit by #643 Namely, a case like ` <button onfocus="this.blur(); focusNext()">Foo</button></img>` has a `button` which is both in sequential tab order and a presentational child, so it would fail the rule but it is not really a problem because scripting makes sure it never actually gets the focus.

If a presentational child is in tab-order but with some scripting highjacking that, the rule fails but the SCs are satisfied.
However, I do believe this is super-rare (more than with `aria-hidden`), notably because UAs seem to push the button out of the image anyway (so this needs to be built by scripting in the first place).

I say we should be good with an assumption that "focusable presentational children do not have focus highjacked by scripting" or likewise. (I probably wouldn't have problem if we weren't already trying to fix this issue in the other rule…)

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,144 @@
...
+
+This rule applies to any HTML or SVG element with a [semantic role][] that defines its [children][child] to be [presentational children][].
+
+## Expectation
+
+None of the target elements have [descendants][] in the [flat tree][] that are part of [sequential focus navigation][].
+
+## Assumptions
+
+_There are currently no assumptions_
```
WilcoFiers, line 36:
```suggestion
This rule assumes that elements that are part of [sequential focus navigation][] do not immediately blur, or move focus to another element. Such elements will fail this rule, but may still satisfy success criteria 1.3.1 and 4.1.2.
```

Jym77 (approved):

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,144 @@
...
+  wcag20:1.3.1: # Info and Relationships (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+  wcag20:4.1.2: # Name, Role, Value (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
```
ajanec01, line 17: Would this rule also map to [2.4.3 Focus order?](https://www.w3.org/TR/WCAG21/#focus-order) There is part in the understanding doc that goes:

> Focusable components need to receive focus in **an order that preserves meaning and operability** only when navigation sequences affect meaning and operability.

Although operable, elements which are not announced by the AT are meaningless so they fail the above statement.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,144 @@
...
+
+This rule applies to any HTML or SVG element with a [semantic role][] that defines its [children][child] to be [presentational children][].
+
+## Expectation
+
+None of the target elements have [descendants][] in the [flat tree][] that are part of [sequential focus navigation][].
+
+## Assumptions
+
+This rule assumes that elements that are part of [sequential focus navigation][] do not immediately blur, or move focus to another element. Such elements will fail this rule, but may still satisfy success criteria 1.3.1 and 4.1.2.
```
ajanec01, line 36: As per my previous comment about Focus order. If true, the criterion should be added.
```suggestion
This rule assumes that elements that are part of [sequential focus navigation][] do not immediately blur, or move focus to another element. Such elements will fail this rule, but may still satisfy success criteria 1.3.1, 2.4.3, and 4.1.2.
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Accessibility Support
+
+Several major browsers ignore the WAI-ARIA requirements on [presentational children][] for most or sometimes all roles. Because some browsers do, and others do not implement presentational children, there can be big differences between browsers.
```
ajanec01, line 40: My preference, feel free to ignore it

```suggestion
Several major browsers ignore the WAI-ARIA requirements on [presentational children][] for most or sometimes all roles. Because some browsers do, and others do not implement presentational children, there can be significant differences between browsers.
```

_rules/presentational-children-no-focusable-content-307n5z.md
WilcoFiers: You are right that this effects the meaning of the focusable element, but it isn't the order that is the problem, so I don't think 2.4.3 applies here.

carlosapaduarte (approved):

PR #1509, opened 2020-11-30
new rule: Menuitem has non-empty accessible name

WilcoFiers:
This rule checks that each element with a `menuitem` role has a non-empty accessible name.

Need for Final Call: 2 weeks

## How to Review And Approve

### DIFF
--- a/_rules/menuitem-non-empty-name-m6b1q3.md
+++ b/_rules/menuitem-non-empty-name-m6b1q3.md
@@ -1,0 +1,153 @@
+---
+id: m6b1q3
+name: Menuitem has non-empty accessible name
+rule_type: atomic
+description: |
+  This rule checks that each element with a `menuitem` role has a non-empty accessible name.
+accessibility_requirements:
+  wcag20:4.1.2: # Name, Role, Value (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - Accessibility Tree
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to HTML elements with a `menuitem` [semantic role][] that are [included in the accessibility tree][].
+
+## Expectation
+
+Each target element has an [accessible name][] that is not empty (`""`).
+
+## Assumptions
+
+The rule assumes that all menuitems are [user interface components as defined by WCAG 2](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If an element has a role of `menuitem` that would not be perceived as a single control by users, [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) would not apply and so failing this rule would not result in a conformance issue.
+
+## Accessibility Support
+
+Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have a [semantic role][] of `button` and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+
+## Background
+
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)
+- [ARIA14: Using aria-label to provide an invisible label where a visible label cannot be used](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA14)
+- [ARIA16: Using aria-labelledby to provide a name for user interface controls](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA16)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This element with the `menuitem` role has an [accessible name][] because of its text content.
+
+```html
+<ul role="menu">
+	<li role="menuitem">New file</li>
+</ul>
+```
+
+#### Passed Example 2
+
+This element with the `menuitem` role has an [accessible name][] because of its `aria-label` attribute.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem" aria-label="New file">
+		<i class="fa fa-file"></i>
+	</li>
+</ul>
+```
+
+#### Passed Example 3
+
+This element with the `menuitem` role has an [accessible name][] because its `aria-labelledby` attribute references a `div` element with text content.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem" aria-labelledby="newfile">
+		<i class="fa fa-file"></i>
+		<span id="newfile" hidden>New file</span>
+	</li>
+</ul>
+```
+
+#### Passed Example 4
+
+This element with the `menuitem` role has an [accessible name][] because of its `title` attribute.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem" title="New file">
+		<i class="fa fa-file"></i>
+	</li>
+</ul>
+```
+
+### Failed
+
+#### Failed Example 1
+
+This element with the `menuitem` role has no [accessible name][] because it has no content or attribute that can provide it.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem">
+		<i class="fa fa-file"></i>
+	</li>
+</ul>
+```
+
+#### Failed Example 2
+
+This element with the `menuitem` role has no [accessible name][] because it has no content or attribute that can provide it.
+
+```html
+<html lang="en">
+	<style>
+		.offscreen {
+			position: absolute;
+			left: -9999px;
+			top: -9999px;
+		}
+	</style>
+	<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+	<ul role="menu" class="offscreen">
+		<li role="menuitem">
+			<i class="fa fa-file"></i>
+		</li>
+	</ul>
+</html>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+This element with the `menuitem` role does not need an [accessible name][] because it is not [included in the accessibility tree][].
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu" hidden>
+	<li role="menuitem">
+		<i class="fa fa-file"></i>
+	</li>
+</ul>
+```
+
+[accessible name]: #accessible-name 'Definition of accessible name'
+[included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of included in the accessibility tree'
+[presentational roles conflict resolution]: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none 'Presentational Roles Conflict Resolution'
+[semantic role]: #semantic-role 'Definition of Semantic Role'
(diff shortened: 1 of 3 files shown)


_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to HTML elements with a `menuitem` [semantic role][] that are [included in the accessibility tree][].
```
Jym77, line 24: We seem to have various ways of saying that in the various "has accessible name" rules 🙈 
So, this is OK but we might want to unify all of them a bit at a later point.

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+accessibility_requirements:
+  wcag20:4.1.2: # Name, Role, Value (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - Accessibility Tree
+  - DOM Tree
+  - CSS Styling
```
Jym77, line 16:
```suggestion
  - CSS Styling
  - DOM Tree
```
(sort alphabetically)

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+
+This rule applies to HTML elements with a `menuitem` [semantic role][] that are [included in the accessibility tree][].
+
+## Expectation
+
+Each target element has an [accessible name][] that is not empty (`""`).
+
+## Assumptions
+
+The rule assumes that all menuitems are [user interface components as defined by WCAG 2](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If an element has a role of `menuitem` that would not be perceived as a single control by users, [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) would not apply and so failing this rule would not result in a conformance issue.
```
Jym77, line 32:
```suggestion
This rule assumes that all menuitems are [user interface components as defined by WCAG 2](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If an element has a role of `menuitem` that would not be perceived as a single control by users, [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) would not apply and so failing this rule would not result in a conformance issue.
```

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+	<li role="menuitem">New file</li>
+</ul>
+```
+
+#### Passed Example 2
+
+This element with the `menuitem` role has an [accessible name][] because of its `aria-label` attribute.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
```
Jym77, line 63: I am a bit uncomfortable in linking external resources.
It's not about using FA, it's about having the validity of the examples depend on something we don't really control…

I'd be happier if this was a local .png icon with `alt=""` (for example).
Although, I also agree that FA is definitively use and that this makes for more natural examples… 🤷

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+<ul role="menu">
+	<li role="menuitem" aria-label="New file">
+		<i class="fa fa-file"></i>
+	</li>
+</ul>
+```
+
+#### Passed Example 3
+
+This element with the `menuitem` role has an [accessible name][] because its `aria-labelledby` attribute references a `div` element with text content.
```
Jym77, line 73:
```suggestion
This element with the `menuitem` role has an [accessible name][] because its `aria-labelledby` attribute references a `span` element with text content.
```
Actually, we can probably just go with
```suggestion
This element with the `menuitem` role has an [accessible name][] because its `aria-labelledby` attribute references an element with text content.
```
since the precise element  name doesn't matter.

_rules/menuitem-non-empty-name-m6b1q3.md
Jym77, line 24: This looks extremely focused. Would it make sense to extend to any `widget` role? (or is there some that don't require a name?)

This is definitely a working rule but it feels it's checking something extremely narrow and could easily be extended.

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+	<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+	<ul role="menu" class="offscreen">
+		<li role="menuitem">
+			<i class="fa fa-file"></i>
+		</li>
+	</ul>
+</html>
+```
+
+### Inapplicable
```
Jym77, line 135: For coverage of the Applicability, we need an example with a non-`menuitem` role.

Jym77 (dismissed):
Some minor editorial, a couple of deeper but non-blocking questions.

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+
+#### Passed Example 2
+
+This element with the `menuitem` role has an [accessible name][] because of its `aria-label` attribute.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem" aria-label="New file">
+		<i class="fa fa-file"></i>
```
daniel-montalvo, line 66:
```suggestion
		<i class="fa fa-file" aria-hidden="true"></i>
```

Shall we add `aria-hidden` here? Sometimes this provokes screen readers remain silent when they are put into virtual buffer mode (or review cursor), as you use the arrow keys to move through the FA.

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+
+#### Passed Example 3
+
+This element with the `menuitem` role has an [accessible name][] because its `aria-labelledby` attribute references a `div` element with text content.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem" aria-labelledby="newfile">
+		<i class="fa fa-file"></i>
```
daniel-montalvo, line 79:
```suggestion
		<i class="fa fa-file" aria-hidden="true"></i>
```

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,153 @@
...
+
+#### Passed Example 4
+
+This element with the `menuitem` role has an [accessible name][] because of its `title` attribute.
+
+```html
+<link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" />
+<ul role="menu">
+	<li role="menuitem" title="New file">
+		<i class="fa fa-file"></i>
```
daniel-montalvo, line 93:
```suggestion
		<i class="fa fa-file" aria-hidden="true"></i>
```

daniel-montalvo (dismissed):

_rules/menuitem-non-empty-name-m6b1q3.md
WilcoFiers: I intentionally kept this one a little simpler than the others because there are fewer conditions to consider here.

WilcoFiers: I do think all widgets require an accessible name, but some apply to different SCs. This one can't even be merged with its parent role "command", since that includes links, which are also applicable to 2.4.4.

Jym77: Hmm, I guess we're back to the discussion at #1049 where we decided it was OK to have a rule with wider applicability and lesser mapping, plus an extra rule with narrow applicability for those special cases, even if it ends up with the same element failing 2 rules for the same reason.

See also https://www.w3.org/2020/02/13-act-r-minutes.html#x03 and https://github.com/act-rules/act-rules.github.io/issues/1049#issuecomment-585818622

Not a show blocker for the current PR, however.

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,161 @@
...
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This element with the `menuitem` role has an [accessible name][] because of its text content.
+
+```html
+<ul role="menu">
```
Jym77, line 55: Should we use the equivalent
```suggestion
<menu role="menu">
```
(on all examples)?

Jym77 (approved):
Looks good. Minor and somewhat unrelated comments.

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,161 @@
...
+Each target element has an [accessible name][] that is not empty (`""`).
+
+## Assumptions
+
+This rule assumes that all menuitems are [user interface components as defined by WCAG 2](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If an element has a role of `menuitem` that would not be perceived as a single control by users, [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) would not apply and so failing this rule would not result in a conformance issue.
+
+## Accessibility Support
+
+Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have a [semantic role][] of `button` and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+
```
ajanec01, line 39: Would it be worth to mention the support issues for the role of `menuitem`? I've done a quick test using JAWS and NVDA on FF and Chrome. NVDA only announces menu items in the browsing mode but not when they receive keyboard focus. Also, NVDA seems to support the `title` attribute as accessible name only in the browsing mode.

ajanec01 (dismissed):
Just the accessibility support section I'm not sure about- not a blocker though.

daniel-montalvo (approved):
Worth giving a thought to mention issues raised by @ajanec01  Otherwise approving.

kasperisager (dismissed):

carlosapaduarte (approved):

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -0,0 +1,161 @@
...
+This element with the `menuitem` role has an [accessible name][] because of its `title` attribute.
+
+```html
+<ul role="menu">
+	<li role="menuitem" title="New file">
+		<img src="/test-assets/shared/file.svg" alt="" />
+	</li>
+</ul>
+```
```
Jym77, line 95: Quick test on this example.
Chrome gives the `li` a name of "New File". Firefox gives it a name of "·" (😖 it's the list marker, why default to that…) Edge gives it no name (and thus fails the rule).

I say we should at least have an Accessibility Support note on support for `title`. Maybe remove this example.

Jym77 (changes requested):

_rules/menuitem-non-empty-name-m6b1q3.md
WilcoFiers: I think using `ul` is standard. I can't say I've ever seen someone use the `menu` element.

_rules/menuitem-non-empty-name-m6b1q3.md
WilcoFiers: Oh! I think I know why. The bullet is treated as text content, and so because there is text, the accessible name won't go to the title attribute. I'll update the example. That's a good catch!

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -52,46 +54,46 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
...
 ```html
-<ul role="menu">
-	<li role="menuitem" title="New file">
+<div role="menu">
+	<button role="menuitem" title="New file">
 		<img src="/test-assets/shared/file.svg" alt="" />
-	</li>
-</ul>
+	</button>
+</div>
```
Jym77: Still no name in Edge, but now FF gets it… So I guess this works.

Jym77 (approved):

carlosapaduarte (approved):

kasperisager (approved):

PR #1513, opened 2020-12-03
"Zoomed text node is not clipped with CSS overflow" (59br37): Handle non-length line-height

### LINKED ISSUE #1316: Zoomed text node is not clipped with CSS overflow (59br37): What to do about `normal` line heights?
The rule currently states the following:

> A computed line-height equal to or greater than the height of its bounding box, or in case of a computed overflow of clip, its content box.

During implementation, it struck me that the `line-height` property computes to either the `normal` keyword, a `font-size` multiplier (`<number>`), or a `px` length. The latter two we can easily compare to the height of the bounding or content box, but for the `normal` keyword we have assume a [used](https://www.w3.org/TR/css-cascade-3/#used-value) value as we cannot compare `normal` to a `px` value. I'm guessing that this should either be clarified in the rule or that we should instead refer to the used or [actual](https://www.w3.org/TR/css-cascade-3/#actual-value) value of the `line-height` property.

Summoning @WilcoFiers.

### LINKED ISSUE #1508: "Zoomed text node is not clipped with CSS overflow" (59br37): computed line-height is not always a length
[Expectation of Zoomed text node is not clipped with CSS overflow](https://act-rules.github.io/rules/59br37#expectation) refers to the "computed line-height" and compares it to a length.
However, computed line-height may also be a number or the keyword `normal` and the rule doesn't tell what to do in these case.

We can either change it to "used line-height" (but this is a bit more blurry since it depends on many things) or assume that `normal` is 1.2 (which seems to be a good approximation). Handling numbers is not too hard since it just need to be multiplied by the font size.

Jym77:
Handle cases where the computed `line-height` is not a length and thus cannot be directly compared with `height`.

Closes issue(s):

- closes #1316 
- closes #1508

Need for Final Call:
This will require a 1 week Final Call (clarifying expectation)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


pages/glossary/actual-line-height.md
```diff
@@ -0,0 +1,23 @@
...
+- if the value is a number, then the _actual line height_ is that number multiplied by the [computed value][] of the `font-size` property of the element;
+- if the value is the keyword `normal`, then the _actual line height_ is 1.2 times the [computed value][] of the `font-size` property of the element.
+
+#### Background
+
+This definition is meant to mimic the [used value][] of the `line-height` property. However, in the `normal` case, this value depends on font specific metrics resulting in complex computations. Therefore, this definition uses 1.2 as a reasonable heuristic. [CSS specifications][line-height normal] recommend that the [used value][] is between 1.0 and 1.2 and major browsers are effectively using values close to 1.2.
+
+[computed value]: https://www.w3.org/TR/css-cascade-4/#computed 'CSS Cascading and Inheritance Level 4 (Working draft) - Computed Values'
+[line-height normal]: https://drafts.csswg.org/css2/#valdef-line-height-normal "CSS 2.2 (Editor's draft) - normal line-height"
+[used value]: https://www.w3.org/TR/css-cascade-4/#used 'CSS Cascading and Inheritance Level 4 (Working draft) - Used Values'
```
kasperisager, line 23: This definition overloads the concept of "actual values" from CSS (https://www.w3.org/TR/css-cascade/#actual), which I'm not entirely comfortable with.

kasperisager (changes requested):

pages/glossary/actual-line-height.md
Jym77: Good point.
I wanted a name different from "used value" since it is not the same thing, but taking another existing name is as bad 🙈 

What about "effective line height"?

kasperisager: I'd keep it simple and simply reference "used value" as that's the one we're really interested in.

Jym77: The problem is that it's not really the used value. When the `line-height` is `normal`, the used value can be many different things and we take 1.2 to keep it simple and because it is not too far from the used value in practice. But this is not really the used value (or not always).

Or you mean the rule could reference "used value" and let implementers decide what they want to do with that?

kasperisager: > Or you mean the rule could reference "used value" and let implementers decide what they want to do with that?

Exactly. From the a rule specification perspective, we don't really care about the exact value of the given property, nor even the fact that it might differ between UAs. We just care about specifying something that is unambiguous.

To get the used `line-height` value, browser APIs such as `Node#clientHeight` can be used. Implementations without access to browser APIs can assume a used value when faced with `line-height: normal` or do something else; that's up to them.

WilcoFiers (approved):

kasperisager (approved):

Jym77:
@kasperisager @WilcoFiers can you recheck Passed Example 4?
It is assuming that the used value of `normal` is 1.2 which might be wrong in that specific case and could result in tools fetching the real used value to fail it…
Not sure if it needs to be removed, or maybe just to force the font in the style and make sure the used value is 1.2 for that one (which is not a trivial task, it's a bit annoying to dig into font metrics).

_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -112,6 +114,27 @@ This [text node][] is restricted to a single line, by setting a `line-height` th
...
 
+Assuming a [used][] [line-height][] of 1.2 when `normal`, this [text node][] is restricted to a single line. A link to a full version of the poem is also provided.
+
+```html
+<style>
+	.wordClip {
+		overflow: hidden;
+		word-wrap: break-word;
+		height: 18px;
+		line-height: normal;
```
kasperisager, line 125: Ah, that's a good point... I'd be fine with removing the example altogether given that it could cause consistency issues.

carlosapaduarte: I agree that it's best to leave passed example 4 out.

carlosapaduarte (changes requested):

WilcoFiers (approved):

kasperisager (approved):

carlosapaduarte (approved):

Jym77:
Final Call ends on December 16.

Jym77:

PR #1515, opened 2020-12-04
new rule: iframe with negative tabindex has no interactive elements

WilcoFiers:
This rule checks that `iframe` elements with a negative `tabindex` attribute value contain no interactive elements.

Need for Final Call: 2 weeks
## How to Review And Approve

### DIFF
--- a/_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
+++ b/_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
@@ -30,7 +30,7 @@ This rule applies to any `iframe` element that has a negative number as a `tabin
 
 ## Expectation
 
-None of the test target's [nested browsing context][] includes elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
+For each test target, the [nested browsing context][] does not include elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][]. 
 
 ## Assumptions
 
--- a/_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
+++ b/_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
@@ -1,0 +1,102 @@
+---
+id: akn7bn
+name: iframe with negative tabindex has no interactive elements
+rule_type: atomic
+description: |
+  This rule checks that `iframe` elements with a negative `tabindex` attribute value contain no interactive elements.
+accessibility_requirements:
+  wcag20:2.1.1: # Keyboard (A)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+  wcag-technique:G202: # Ensuring keyboard control for all functionality
+    forConformance: false
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Brian Bors
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+
+## Expectation
+
+None of the test target's [nested browsing context][] includes elements that are [visible][] and part of the [sequential focus navigation][].
+
+## Assumptions
+
+This rule assumes that interactive content inside `iframe` elements are used to provide functionality. If the interactive content does not provide functionality, for example a button that does nothing when clicked, [success criterion 2.1.1][sc211] may be satisfied, even if the rule is failed.
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or some other negative number, it becomes impossible to move the focus into the [browsing context][nested browsing context] of the `iframe` element. Even though its content is still included in the [sequential focus navigation][], there is no way to move the focus to any of the items in the `iframe` using standard keyboard navigation.
+
+- [Understanding Success Criterion 2.1.1: Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard)
+- [WCAG Technique G202: Ensuring keyboard control for all functionality](https://www.w3.org/WAI/WCAG21/Techniques/general/G202)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `iframe` element contains no content that is part of [sequential focus navigation][].
+
+```html
+<iframe tabindex="-1" srcdoc="<h1>Hello world</h1>"></iframe>
+```
+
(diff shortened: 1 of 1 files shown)


ajanec01 (approved):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -0,0 +1,102 @@
...
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+
+## Expectation
+
+None of the test target's [nested browsing context][] includes elements that are [visible][] and part of the [sequential focus navigation][].
```
kasperisager, line 33: "includes elements" needs to be qualified to ensure unambiguity. I'm guessing we're looking for "has descendants in the flat tree"?

kasperisager (dismissed):

carlosapaduarte (changes requested):
Besides @kasperisager's comment I believe an inapplicable example without a tabindex attribute could also be useful.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -0,0 +1,104 @@
...
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+
+## Expectation
+
+None of the test target's [nested browsing context][] includes elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" if its [owner document][] is the [container document][] of the [nested browsing context][].
```
carlosapaduarte, line 33:
```suggestion
None of the test target's [nested browsing context][] includes elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
```

carlosapaduarte (approved):
Just an edit suggestion to improve the clarity, but I'm approving this anyway.

kasperisager (dismissed):

wcreedle:
Forgive me if this is not the place to ask this example (first time dipping my toes in the water), but how would this rule handle an example like this?

Issue: there is a non conforming video type that has pertinent information to the user. As a workaround the page
1. sets negative tabindex on the video
2. provides an accessible transcript in front of the video to provide the same information.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -0,0 +1,104 @@
...
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+
+## Expectation
+
+None of the test target's [nested browsing context][] includes elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
```
Jym77, line 33: As far as I understand, an `iframe` only has one nested browsing context, not several (so "none of" is not needed).

Jym77, line 33: Linked to previous comment, I understand the expectation as not going recursively into nested `iframe`, because "owner document" only look at the parent, not the top-level ancestor.
I don't know if this is intentional or not.

Jym77 (dismissed):

WilcoFiers:
@wcreedle Thanks for the input. This is a little complicated, but I'll try to explain why I think this is correct. WCAG allows a page to conform, even if a success criterion is not satisfied, as long as a conforming alternative version is provided. This rule tests specifically for the success criterion; so the rule finds things that do not satisfy the success criterion, but even with content that does not satisfy the SC, a page may still conform if an alternative version is provided.

wcreedle:
Thank you, this helps a lot.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
WilcoFiers: It's not recursive, no. Because `iframe` elements are themselves focusable, an empty `iframe` inside your `iframe[tabindex="-1"]` would fail this rule unless it too had a negative tabindex. That is captured in the assumption. I think that's reasonable. Iframes need to be focusable in order to allow scrolling with keyboard.

WilcoFiers: Fair point, I'll fix that.

Jym77: 🤔 
```html
<iframe tabindex="-1" srcdoc="<iframe tabindex='-1' srcdoc='<a href=…>W3C</a>' />" />
```
(not filling the `href` to avoid quote protection nightmare)
Then the link is not keyboard accessible (the SC is not satisfied) but the rule would pass because it only looks at the outer `iframe` and is happy to see that none of its elements are part of sequential focus navigation.

And I don't think this is caught by the assumption since the iframe (both inner and outer) do indeed provide functionality.

This is not a blocker, though since it means the rule passes too often, which is usual.

carlosapaduarte (changes requested):
Editorial

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: Ah no, my bad. The inner iframe should be applicable and fail.
It is not completely explicit in the applicability, however…

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -0,0 +1,104 @@
...
+    - Wilco Fiers
+---
+
+## Applicability
+
+This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+
+## Expectation
+
+For each test target, the [nested browsing context][] does not include elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
```
ajanec01, line 33: Wouldn't that first line be more appropriate in the applicability section? At the moment the applicability says all `<iframe tabindex="-1">` whereas it's more about `<iframe tabindex="-1">` with widgets.

WilcoFiers: I'm not sure. I suppose it can be done in a bunch of different ways. We could also completely flip it around and make this applicable to any focusable element in an iframe, and then have an expectation that the iframe doesn't have a tabindex.... I think I prefer it this way around, but if you have a strong preference I don't mind changing it.

ajanec01: No strong preference. Valid point.

ajanec01 (approved):

kasperisager (approved):

carlosapaduarte (approved):

Jym77 (commented):
Still good. Only commenting since we already have one SI approval.

WilcoFiers:
Putting it back into final call, ending on January 29th.

Issue #1520, opened 2020-12-16
"Element with `lang` attribute has valid language tag" (de46e4): Should it check accessible names?


Jym77:
[Element with lang attribute has valid language tag](https://act-rules.github.io/rules/de46e4) only checks element with visible/accessible text nodes. This means that
```html
<div lang="invalid"><img src="…" alt="lorem ipsum" /></div>
```
is not tested because it has no descendant text nodes.

It's not a problem per se but could be an improvement (or a companion rule) to also check elements with descendants having an accessible name.

PR #1521, opened 2020-12-16
New rule: HTML element language subtag matches language

Jym77:
New rule for 3.1.2 checking that the `lang` attribute is correct.
Plus a bit of polishing of related rules to unify some wordings and conditions.

Closes issue(s):

- N/A

Need for Final Call:
This will require a 2 weeks Final Call (new rule).

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,339 @@
...
+
+- This rule assumes that [grandfathered tags][] are not used as these will not be recognized as [valid language tags][].
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+This rule ignores elements that are nested inside `iframe`. `iframe` are intended to provide a layer of isolation, and it is not clear how language is inherited in nested browsing contexts. Incorrect `lang` attribute inside `iframe` are still likely to cause accessibility issues even though this rule won't flag them.
```
Jym77, line 60: I remember us having problems around language and `iframe`. I don't remember the details 😞 
Playing it safe here, can easily be changed later.

_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,339 @@
...
+- [Understanding Success Criterion 3.1.2: Language of Page][usc312]
+- [H58: Using language attributes to identify changes in the human language](https://www.w3.org/WAI/WCAG21/Techniques/html/H58)
+- [BCP 47: Tags for Identifying Languages](https://www.ietf.org/rfc/bcp/bcp47.txt)
+- [The `lang` and `xml:lang` attributes](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes)
+
+In all examples, the `html` element has itself a `lang` attribute in order to make sure that the examples do not fail [Success Criterion 3.1.1 Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page). These `html` elements are, however, never applicable because they are not descendants of a `body` element, and the example descriptions do not mention them further.
+
+Similarly, all examples contain a short text in the page language without any `lang` attribute in order to make the `lang` attribute on other elements useful. Especially, Technique [H58: Using language attributes to identify changes in the human language](https://www.w3.org/WAI/WCAG21/Techniques/html/H58) is about changes of language.
```
Jym77, line 71: Actually not fully sure this is really needed… What do reviewer think?

_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,339 @@
...
+
+## Test Cases
```
Jym77, line 73: Examples feel fairly contrived with nested `lang` attributes… Not sure it's possible to do much better given the nature of the rule 🤷 

Especially tried to make sure that all targets in a given example pass/fail the rule together to avoid mixed results and make tool validation easier.

_rules/element-lang-valid-de46e4.md
```diff
@@ -27,9 +27,9 @@ acknowledgments:
 
 ## Applicability
 
-This rules applies to any HTML element with a `lang` [attribute value][] that is not empty (`""`) and for which all of the following is true:
+This rule applies to any HTML element with a `lang` [attribute value][] that is not empty (`""`) and for which all of the following is true:
 
-- the element is a [descendant][] in the [flat tree][] of a `body` element; and
+- the element is an [inclusive descendant][] in the [flat tree][] of a `body` element; and
```
Jym77, line 32: I think it is OK for this rule to check `lang` on a `body` element. Added it to match the Applicability of the new rule.

_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,337 @@
...
+<html lang="nl">
+	<head>
+		<title>Met de kippen op stok</title>
+	</head>
+	<body>
+		<blockquote>
+			<p>"Hij ging met de kippen op stok"</p>
+		</blockquote>
+		<p lang="en">
+			The Dutch phrase "Hij ging met de kippen op stok" literally translates into "He went to roost with the chickens",
```
carlosapaduarte, line 109: I would fail 3.1.2 here. There is a passage where the correct language cannot be programmatically determined.

_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,337 @@
...
+				>literally translates into "He went to roost with the chickens", but it means that he went to bed early.</span
+			>
+		</p>
+	</body>
+</html>
+```
+
+#### Failed Example 4
+
+This `div` element has a `lang` attribute value of `fr` (French), which does not match its [default language][]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `fr`. The first `p` element also has an incorrect `lang` attribute.
```
carlosapaduarte, line 226:
```suggestion
This `div` element has a `lang` attribute value of `fr` (French), which does not match its [default language][]. The default language is English because the accessible texts are English, and all other text is in a `p` element with a `lang` attribute value of `nl` which is also an incorrect `lang` attribute.
```

_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,337 @@
...
+			</p>
+		</div>
+		<p>
+			Un feu d'artifice sera tiré depuis les jardins du Trocadéro.
+		</p>
+	</body>
+</html>
+```
+
+### Inapplicable
```
carlosapaduarte, line 270: More inapplicable examples are needed
- one with just a lang attribute on the html element
- one with an invalid language tag

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -285,13 +285,13 @@ This page has an undefined [default language][default page language] because it
...
 	<body>
-		<p>Paul put dire comment on tape</p>
+		<p>Paul put dire comment on tape.</p>
 	</body>
 </html>
 ```
 
 [valid language tag]: #valid-language-tag
-[default page language]: #default-page-language
+[default page language]: #default-element-language
```
carlosapaduarte, line 294: Why this change in this rule?

pages/glossary/default-element-language.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+The _default language of an element_ E is the _most common language_ used in the texts of its _candidate elements_. If there are no _words_ in the candidate elements, the default language is undefined.
+
+The _most common language_ is determined by counting the number of _words_ in the candidate elements that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. If the number of words is the same, the default language is undefined.
+
+The _candidates elements_ are all the elements for which all the following are true:
+
+- the element is an [inclusive descendant][] of E in the [flat tree][]; and
+- the element has no [inclusive ancestor][] in the [flat tree][] which is a descendant of E and has a non-empty (`""`) `lang` attribute (that is, no `lang` attribute "between" them).
```
carlosapaduarte, line 20: Should it just be non-empty or should it also be a valid language tag? I'm leaning towards the latter

pages/glossary/default-element-language.md
```diff
@@ -0,0 +1,43 @@
...
+
+The _words_ used in the following texts should be counted for the _most common language_:
+
+- **text nodes**: the value of any [text nodes][] that are [visible][] or [included in the accessibility tree][] and children of a candidate element;
+- **accessible text**: the [accessible name][] and [accessible description][] of any candidate element [included in the accessibility tree][];
+- **page title**: the value of the [document title][], only if E is a [document][] in a [top-level browsing context][].
```
carlosapaduarte, line 26: Why do we want to count the page title for determining the language of an element that is not the document itself?
This can lead to wrong language identification when a page as a long title and I'm just marking the language of a small passage.

carlosapaduarte (dismissed):

_rules/element-lang-matches-default-off6ek.md
```diff
@@ -0,0 +1,337 @@
+---
+id: off6ek
+name: HTML element language subtag matches default language
```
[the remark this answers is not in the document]
Jym77: (trying to keep the discussion here, there are several comments related to this)
Good point. I have been copying the existing rule without too much thinking. I now agree that 3.1.2 should fail on parts with mixed languages and that 
```html
<p lang="en">
  The Dutch phrase "Hij ging met de kippen op stok" literally translates into "He went to roost with the chickens", but it means that he went to bed early.
</p>
```
is probably failing it.

3.1.2 has a few exceptions for some words, but I guess it is OK to just copy them in the expectation since this is a manual rule anyway…

Then we can certainly require all words to be in the same language as the `lang` attribute.

A (small) consequence is that 
```html
<p lang="en">Paul put dire comment on tape</p>
<p lang="fr">Paul put dire comment on tape</p>
```
will become 2 Passed targets, which is actually rather good, I guess…

_rules/html-page-lang-matches-default-ucwvc8.md
Jym77: I've mostly renamed the definition +adapt it to make it work on any element (not just `html`), hence propagating the new name. This is actually the same definition (or so I hope 🤞).

Git doesn't see the rename anymore (sees a delete+create instead) because it ends up with a bit too many changes in its opinion…

pages/glossary/default-element-language.md
Jym77: Current definition is using "non-empty": https://act-rules.github.io/glossary/#default-page-language
I didn't want to change that (I haven't put much thoughts in it…)

I guess this is about how ATs are handling invalid language tags, whether they default to the previous language or do something else.
I'm summoning @WilcoFiers on that one. He wrote #1334 and hopefully can remember why he chose "non-empty" rather than "valid".

pages/glossary/default-element-language.md
Jym77: That's he other way around. We count the page title **only** for determining the language of a document. Not for other elements.

carlosapaduarte: You are correct. I must have been too tired already :-)

_rules/element-lang-matches-unique-language-off6ek.md
```diff
@@ -0,0 +1,340 @@
...
+
+- the element is an [inclusive descendant][] in the [flat tree][] of a `body` element; and
+- The element is in a [top-level browsing context][]; and
+- The element is in a [document][] with a [content type][] of `text/html`; and
+- The element has a `lang` [attribute value][] which is a [valid language tag][]; and
+- The element has some non-empty [text with the same programmatic language][].
+
+## Expectation
+
+For each test target, the [primary language][] of the [valid language tag][] is a [unique element language][].
```
carlosapaduarte, line 44: "a unique element language" is a strange concept in this sentence. Only after checking its definition I understood that you are checking that the primary language is the same language of the element.

```suggestion
For each test target, the [primary language][] of the [valid language tag][] is the [element language][].
```

_rules/element-lang-matches-unique-language-off6ek.md
```diff
@@ -0,0 +1,340 @@
...
+		</div>
+	</body>
+</html>
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+There is no HTML elements in this document.
```
carlosapaduarte, line 253:
```suggestion
There are no HTML elements in this document.
```

pages/glossary/element-language.md
```diff
@@ -0,0 +1,25 @@
...
+  - Language
+---
+
+The _default language of an element_ is the _most common language_ used in the [text with the same programmatic language][] as this element. If there are no _words_ in this text, the default language is undefined.
+
+The _most common language_ is determined by counting the number of _words_ in the text that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. If the number of words is the same, the default language is undefined.
+
+When the element is a [document][], its default language is also called the _default page language_.
+
+An element as a _unique language_ if all _words_ in the [text with the same programmatic language][] as an element are part of the same language from the [language subtag registry][], except for proper names, technical terms, words of indeterminate language, and words or phrases that have become part of the vernacular of the immediately surrounding text. Words can be part of other languages as long as they are also part of the unique language. In the rare case where all words are part of several languages, an element may have several _unique languages_…
```
carlosapaduarte, line 19:
```suggestion
An element has a _unique language_ if all _words_ in the [text with the same programmatic language][] of the element are part of the same language from the [language subtag registry][], except for proper names, technical terms, words of indeterminate language, and words or phrases that have become part of the vernacular of the immediately surrounding text. Words can be part of other languages as long as they are also part of the unique language. In the rare case where all words are part of several languages, an element may have several _unique languages_.
```

carlosapaduarte (changes requested):
Nearly there. I think the expectation could be better phrased.

Jym77:
@carlosapaduarte tried to rename the definition to improve things, but still not satisfied with it…

The problem is that a text can be part of several languages in some weird corner cases (such as the "Paul put dire comment on tape" example). So, it is incorrect to speak of **the** language of a text. It is OK to do so in example descriptions, but not in expectation where "**the** language of the text" is ambiguous and "**one of the** languages" must be used instead.

I also would like the definition's name to reflect that this has to be the language of all the words, not only of the vast majority of them. Intuitively, "the language" of `The Dutch word "gelukkig" has no equivalent in English.` is English because the single Dutch word doesn't really matter (especially given that the sentence itself says it is foreign), but this is not what the definition wants to capture…

I'm now trying to call it a "common language" of the element, using "common" in the sense of "shared" rather than "frequent". It's probably a bit better than "unique language", but doesn't feel perfect either…

If you have ideas for better name, I'm glad to heard them 😀

pages/glossary/element-language.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Element Language
```
carlosapaduarte, line 2: I'm not sure it's a good idea to have 3 definitions in 1
This definition of "element language" defines 3 concepts (and none is called element language):
- default language of an element
- default page element
- common language (of an element)

When I start reading a rule and find a reference to "common element language" pointing to this file, my first instinct is that it is a wrong reference because it takes me so long to find the definition.

_rules/element-lang-matches-common-language-off6ek.md
```diff
@@ -0,0 +1,340 @@
...
+
+- the element is an [inclusive descendant][] in the [flat tree][] of a `body` element; and
+- The element is in a [top-level browsing context][]; and
+- The element is in a [document][] with a [content type][] of `text/html`; and
+- The element has a `lang` [attribute value][] which is a [valid language tag][]; and
+- The element has some non-empty [text with the same programmatic language][].
+
+## Expectation
+
+For each test target, the [primary language][] of the [valid language tag][] is a [common element language][].
```
carlosapaduarte, line 44: Changing to "common element language" was a good move. I think it captures the need to have the words in the same language. 

But I still don't think the expectation is adequately phrased. I think it is too easy to read this with an incorrect meaning (i.e. the language tag must be one of common element languages, even if it is not used in the element - kind of checking if the language is valid). Phrasing it in a different way would make it easier to understand that we are talking about the target element.
```suggestion
For each test target, the [primary language][] of the [valid language tag][] is a [common language][] of the test target.
```

carlosapaduarte (dismissed):
Still not completely happy with the expectation. New suggestion. 
Plus a suggestion to split the element language definition in 3 definitions.

carlosapaduarte (dismissed):
Everything looks good to me now!

_rules/element-lang-matches-common-language-off6ek.md
```diff
@@ -0,0 +1,339 @@
...
+
+- the element is an [inclusive descendant][] in the [flat tree][] of a `body` element; and
+- The element is in a [top-level browsing context][]; and
+- The element is in a [document][] with a [content type][] of `text/html`; and
+- The element has a `lang` [attribute value][] which is a [valid language tag][]; and
+- The element has some non-empty [text with the same programmatic language][].
+
+## Expectation
+
+For each test target, the [primary language][] of the [valid language tag][] is a [common language][] of the test target.
```
WilcoFiers, line 44: It took me an hour reviewing this rule before I figured out that you want (almost) every word to match the `lang` attribute. I don't think that's as clear as it should be. But more importantly, I think this is far too strict. This SC is applicable to passages and phrases, not to individual words.

Putting a lang attribute on every word isn't just burdensome to content authors, it's distracting to screen reader users. Imagine having an English sentence with 2 French words in it. That results in 5 changes in voice, English -> French -> English -> French -> English, all in the span of a few seconds. That's not the intent of SC 3.1.2.

I think we may need to do something different in this rule. Maybe we need to figure out how to objectively define phrase and passage, so that we can make the rule applicable to those?

WilcoFiers (changes requested):

pages/glossary/default-element-language.md
```diff
@@ -0,0 +1,23 @@
...
+  - Accessibility tree
+  - CSS Styling
+  - Language
+---
+
+The _default language of an element_ is the _most common language_ used in the [text with the same programmatic language][] as this element. If there are no _words_ in this text, the default language is undefined.
+
+The _most common language_ is determined by counting the number of _words_ in the text that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. If the number of words is the same, the default language is undefined.
+
+When the element is a [document][], its default language is also called the _default page language_.
```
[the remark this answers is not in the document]
Jym77: This is a generalisation of the existing [Default Page Language](https://act-rules.github.io/glossary/#default-page-language)
[HTML page language subtag matches default language](https://act-rules.github.io/rules/ucwvc8) is using "default page language" all over the place. Without this precision, we need to change that other rule everywhere to use "default element language" which on top would seems weird with the rule being about page.

pages/glossary/text-same-language.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Text with the same Programmatic Language
```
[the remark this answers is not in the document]
Jym77: I'd rather keep the definition title as concepts rather than technology specific stuff (even if the definition is technology specific).

Among other, the definition is not restricted to elements that have a `lang` attribute in the first place. In `<div>foo</div>`, "foo" is text with the same programmatic language as the `div`, even though there is no `lang` attribute implied.

pages/glossary/text-same-language.md
```diff
@@ -0,0 +1,33 @@
...
+
+The _elements with the same programmatic language_ as an element E are all the elements for which all the following are true:
+
+- the element is an [inclusive descendant][] of E in the [flat tree][]; and
+- the element has no [inclusive ancestor][] in the [flat tree][] which is a descendant of E and has a non-empty (`""`) `lang` attribute (that is, no `lang` attribute "between" them).
+
+The _text with the same programmatic language_ as an element E is composed of all the following texts:
+
+- **text nodes**: the value of any [text nodes][] that are [visible][] or [included in the accessibility tree][] and children of an element with the same programmatic language as E;
+- **accessible text**: the [accessible name][] and [accessible description][] of any element with the same programmatic language as E, and [included in the accessibility tree][];
```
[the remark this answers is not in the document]
Jym77: I'd rather keeping visual cues for he defining instance of a term only. (actually, these should be `dfn` elements rather than just italics)

_rules/element-lang-matches-common-language-off6ek.md
```diff
@@ -0,0 +1,339 @@
...
+
+- This rule assumes that [grandfathered tags][] are not used as these will not be recognized as [valid language tags][].
+
+## Accessibility Support
+
+_There are no major accessibility support issues known for this rule._
+
+## Background
+
+This rule ignores elements that are nested inside `iframe`. `iframe` are intended to provide a layer of isolation, and it is not clear how language is inherited in nested browsing contexts. Incorrect `lang` attribute inside `iframe` are still likely to cause accessibility issues even though this rule won't flag them.
```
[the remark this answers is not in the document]
daniel-montalvo: I also think excluding iframes here is a bit odd. You do need to use the `lang` attribute within iframes just as you would need to use it in every other element.

Not feeling strongly, though. If we all agree that this is the way to go for now, happy to take this back.

daniel-montalvo (changes requested):
I also think excluding iframes here is a bit odd. You do need to use the `lang` attribute within iframes just as you would need to use it in every other element.

Not feeling strongly, though. If we all agree that this is the way to go for now, happy to take this back.

+ Some minor suggested corrections.

Jym77:
There seems to be a consensus that ignoring `iframe` was not a good move, so I've removed this restriction.
I don't think we need an `iframe` example, in the same way that, say, none of our "Foo has accessible name" rule has iframe example.

daniel-montalvo (dismissed):

_rules/element-lang-matches-common-language-off6ek.md
Jym77: The discussion seems to be raging on already in https://github.com/w3c/wcag/issues/1174 essentially with similar arguments, I'd say…

* Against the "single word" interpretation:
    * WCAG says "passage or phrase", not "word, passage or phrase".
    * It's disturbing for screen-reader users, so it's actually decreasing the accessibility.
* For the "single word" interpretation:
    * The Understanding document has a paragraph (4th of Intent) explaining when single words need to be considered, why if single words shouldn't be considered?
    * In many cases, single words need to be pronounced differently to be correctly understood.
    * Single words that do not really need language markup are falling under the "vernacular language" exception, so it doesn't mean "every single word" but leaves room for interpretation.
    
Still no strong opinion on that on my side 🤷

_rules/element-lang-matches-common-language-off6ek.md
```diff
@@ -0,0 +1,323 @@
...
+    - Jean-Yves Moyen
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # using aria-labelledby instead
+  - 'alt-require'
+---
+
+## Applicability
+
+This rule applies to any HTML element for which all the following are true:
```
WilcoFiers, line 34: I know this is a bit redundant, but I think the rule is easier to understand if we do this.

```suggestion
This rule applies to any HTML element with a `lang` attribute for which all the following are true:
```

_rules/element-lang-matches-common-language-off6ek.md
```diff
@@ -0,0 +1,323 @@
...
+## Background
+
+- Related rule:
+  - [_Element with `lang` Attribute Has Valid Language Tag_](https://act-rules.github.io/rules/de46e4)
+- [Understanding Success Criterion 3.1.2: Language of Page][usc312]
+- [H58: Using language attributes to identify changes in the human language](https://www.w3.org/WAI/WCAG21/Techniques/html/H58)
+- [BCP 47: Tags for Identifying Languages](https://www.ietf.org/rfc/bcp/bcp47.txt)
+- [The `lang` and `xml:lang` attributes](https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes)
+
+In all examples, the `html` element has itself a `lang` attribute in order to make sure that the examples do not fail [Success Criterion 3.1.1 Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page). These `html` elements are, however, never applicable because they are not descendants of a `body` element, and the example descriptions do not mention them further.
```
WilcoFiers, line 64:
```suggestion
In all examples, the `html` element has itself a `lang` attribute in order to make sure that the examples satisfy [Success Criterion 3.1.1 Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page). These `html` elements are, however, never applicable because they are not descendants of a `body` element, and the example descriptions do not mention them further.
```

pages/glossary/text-same-language.md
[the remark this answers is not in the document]
WilcoFiers: Right, but logically in the following example "foo" and "bar" are texts with the same programmatic language, but your definition says they wouldn't be. I don't think the name reflects what the definition actually does.

```html
<i lang="nl">foo</i>
<b lang="nl">bar</b>
```

WilcoFiers (dismissed):
Couple more minor points

pages/glossary/text-same-language.md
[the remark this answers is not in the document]
Jym77: Fair point. I was seeing "same" as physical rather than structural equality, but it doesn't make sense in common language…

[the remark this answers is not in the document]
Jym77: After more thoughts and work into this, I don't think the definition should be tied to a `lang` attribute as it ripples down further definitions.

This definition is essentially extracting the list from https://act-rules.github.io/rules/ucwvc8#default-page-language and is intended to define the default language of elements. If this definition is tied to a `lang` attribute, it means we only define "default language" of elements with a `lang` attribute which make the definition of "default language" extremely un-natural.

I'm struggling to find a better name for the definition, and I'm definitely welcoming ideas, but I'm very reluctant to tie it to the presence of a `lang` attribute.

pages/glossary/default-element-language.md
Jym77: Digging into this, an invalid language tag should be handled as its own language and carried as is (essentially assuming that something down the line (e.g. a CSS selector) knows what to do about it), so I think that "non-empty" is the correct stuff to use.

See example at https://html.spec.whatwg.org/multipage/dom.html#attr-lang and the paragraph before it.

Jym77:
@carlosapaduarte @daniel-montalvo @WilcoFiers I made updates after the CG call last week.
There is still an outstanding naming issue which I struggle to solve, see the discussion at https://github.com/act-rules/act-rules.github.io/pull/1521#discussion_r551878841

---
Essentially, I extracted the "where to look" part of the definition of [default language](https://act-rules.github.io/rules/ucwvc8#default-page-language) (the list and following exception). I think it's a good move both for separation of concerns and because it seems to be the way to solve #1525 and #1520 .

I initially went for a name of "text/element with the same programmatic language" but that has the unfortunate consequence that in
```html
<p lang="en">This is English text</p>
<p lang="en">This is also English text</p>
```html
the two `p` elements do not have the "same programmatic language" according to the definition, which is somewhat true (for a "physical equality" interpretation of "same"), but definitely counter-intuitive.

Wilco's suggesting "element affected by a `lang` attribute". My problem with that name is that it restricts the definition of elements having a `lang` attribute in the first place. This is not a problem per se, but the definition and the restriction then ripples to "most common" and "default language". So if we restrict it to element having a `lang` attribute, it means that
```html
<p>This is English text</p>
```
or
```html
<html>
  <p>This is English text</p>
</html>
```
do not have a "default language" (or a most common one), which is also extremely counter-intuitive (not a problem since the rules only apply to stuff with a `lang` attribute, but clearly not a good definition name…)

And I've been struggling with finding a better name for these concepts that would make the intuition and the definition match on all levels… Input is welcome 😄

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -0,0 +1,316 @@
...
+  # using aria-labelledby instead
+  - 'alt-require'
+---
+
+## Applicability
+
+This rule applies to any HTML element with a `lang` attribute for which all the following are true:
+
+- the element is an [inclusive descendant][] in the [flat tree][] of a `body` element; and
+- The element is in a [document][] with a [content type][] of `text/html`; and
```
carlosapaduarte, line 37:
```suggestion
- the element is in a [document][] with a [content type][] of `text/html`; and
```

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -0,0 +1,316 @@
...
+- The element's `lang` [attribute value][] is a [valid language tag][]; and
```
carlosapaduarte, line 38:
```suggestion
- the element's `lang` [attribute value][] is a [valid language tag][]; and
```

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -0,0 +1,316 @@
...
+- The element has some non-empty [text with the same programmatic language][].
```
carlosapaduarte, line 39:
```suggestion
- the element has some non-empty [text with the same programmatic language][].
```

carlosapaduarte, line 39:
```suggestion
- The element has non-empty [text with the same programmatic language][].
```

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -0,0 +1,316 @@
...
+			The Dutch phrase <span lang="nl">"Hij ging met de kippen op stok"</span> literally translates into "He went to
+			roost with the chickens", but it means that he went to bed early.
+		</p>
+	</body>
+</html>
+```
+
+#### Passed Example 2
+
+The second `p` element has `lang` attribute value of `nl` (Dutch), which matches its [most common language][]. The most common language is Dutch because all English words are in `span` elements with a `lang` attribute value of `en`. Both `span` elements also have a `lang` attribute matching their common language.
```
carlosapaduarte, line 91:
```suggestion
The second `p` element has `lang` attribute value of `nl` (Dutch), which matches its [most common language][]. The most common language is Dutch because all English words are in `span` elements with a `lang` attribute value of `en`. Both `span` elements also have a `lang` attribute matching their most common language.
```

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -0,0 +1,316 @@
...
+		<p lang="français">
+			I love ACT rules!
+		</p>
+	</body>
+</html>
+```
+
+#### Inapplicable Example 4
+
+The first `p` element has a no [text with the same programmatic language][] because it has no content.
```
carlosapaduarte, line 279:
```suggestion
The first `p` element has no [text with the same programmatic language][] because it has no content.
```

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -0,0 +1,316 @@
...
+	<body>
+		<p lang="fr"></p>
+		<p>I love ACT rules!</p>
+	</body>
+</html>
+```
+
+#### Inapplicable Example 5
+
+This `p` element has a no [text with the same programmatic language][] because it has no content that is either [visible][] or [included in the accessibility tree][].
```
carlosapaduarte, line 292:
```suggestion
This `p` element has no [text with the same programmatic language][] because it has no content that is either [visible][] or [included in the accessibility tree][].
```

pages/glossary/text-same-language.md
```diff
@@ -0,0 +1,34 @@
...
+- **page title**: the value of the [document title][], only if E is a [document][] in a [top-level browsing context][].
+
+An element F has the _same programmatic language_ as an element E if one of the following conditions is true (recursively):
+
+- F is E itself (an element always has the same programmatic language as itself); or
+- F does not have a non-empty `lang` attribute, and it the child in the [flat tree][] of an element with the same programmatic language as E; or
```
carlosapaduarte, line 21:
```suggestion
- F does not have a non-empty `lang` attribute, and is the child in the [flat tree][] of an element with the same programmatic language as E; or
```

pages/glossary/text-same-language.md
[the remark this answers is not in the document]
carlosapaduarte: But the definition ties it to the presence of a `lang` attribute, doesn't it?
I've been trying to come up with a proposal for a name, but I can't seem to improve on @WilcoFiers suggestion. Eventually "Text affected by the same `lang` attribute"?

carlosapaduarte (dismissed):

pages/glossary/text-same-language.md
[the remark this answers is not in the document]
Jym77: > But the definition ties it to the presence of a `lang` attribute, doesn't it?

No, not really (only as a descendants limit).

In 
```html
<div>
Some text in English.
<span lang="fr">Du texte français</span>
<span>More text in English.</span>
```
The first sentence, and the second `span` is "text with the same programmatic language" as the `div`. And that is true whatever the context in which the `div` is put.

If we move to "text affected by a `lang` attribute", then only the text in the first `span` will be affected by a `lang` attribute, but that changes with context (if the full snippet is inside a `<div lang="en">` then text becomes affected by that new `lang` attributes).

Now, the **rules** only use that on elements that do have a `lang` attribute, so it is not a big deal at that level.
**But** this is used in the definition of most common and default language. So it means these definitions would then only apply to elements that have a `lang` attribute. I don't like saying that in
```html
<div>
Some text in English.
<span lang="fr">Du texte français</span>
<span>More text in English.</span>
```
the `div` has no "default language" (just because it has no `lang` attribute). That makes the name "default language" counter-intuitive and we're just pushing the problem of poor definition name further…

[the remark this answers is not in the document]
carlosapaduarte: Weirder still, the below example does not have the same programmatic language (according to this definition).
```html
<div lang="en">Some text <span lang="en">in English</span>.<div>
```
Should we compare the value of the `lang` attributes when they exist to check if they have the same value? Then we could go all the way up to the `body` element which would address Wilco's example. And would make the name much more suitable.

carlosapaduarte (commented):
More inputs to the definition discussion

pages/glossary/text-same-language.md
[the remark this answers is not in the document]
Jym77: Yes, that's why "same programmatic language" is bad in the first place… 

I'm not too keen on looking at other element with the exact same `lang` attribute on the page because that is absolutely not what the rules need. In your example, when the rule checks whether `<div lang="en">` is correct, it really doesn't matter than there is a `<span lang="en">` inside, or another `<div lang="en">` on the page…

But maybe that leads to the way out 🎉 

What about "text/elements affected by the same `lang` attribute" (mixing both names).
That would work on stuff that has no `lang` attribute, and that would make it clear that "same" is about physical identity rather than structural one (`<div lang="en">` and `<div lang="en">`) may have the same programmatic language (English), but not the same `lang` attribute).

This should also work out that in
```html
<div>
Some text in English.
<span lang="fr">Du texte français</span>
<span>More text in English.</span>
```
the second `span` is "affected by the same `lang` attribute" than the `div`, even though there is none currently affecting them, whatever the context we put this where a `lang` attribute affects the `div`, that `lang` attribute would also affect the second `span` but not the first (even if it's also a `lang="fr"`, the first `span` doesn't care).

And if the definition can work on stuff without a `lang` attribute on them, then anything may have a default or most common language and that is cool for me 😎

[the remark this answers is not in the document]
carlosapaduarte: @Jym77 I have to admit I couldn't understand what you are proposing here :-(
Perhaps we can talk about this after the TF/CG meeting later today, if you have time?

Anyway, I've though about this a bit more and have another suggestion, looking at this from another perspective. What if instead of trying to identify text that has the same language, we focus on text that does not change the language? Something like "Inclusive text that does not change the programmatic language"?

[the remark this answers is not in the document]
Jym77: @carlosapaduarte I was (ultimately) suggesting to rename the def to "text/element affected by the same `lang` attribute". But I like your idea better 😄 ("inclusive text" is maybe a bit misleading…)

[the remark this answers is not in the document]
Jym77: @carlosapaduarte @WilcoFiers I end up with "text/element inheriting its programmatic language" from an element.

This is in the same idea of what Carlos was suggesting, but swapping the "does not change" by "receive from" which I feel is closer to the idea. I also felt that "inclusive text" was too close to "inclusive language" which has a totally different meaning (especially when talking about disabilities…)

I think that this is good. Especially, the "element inheriting the programmatic language" works well by using the same vocabulary as CSS inheritance since programmatic language is pretty much determined like an inherited CSS property… So the analogies that this could trigger will actually help understand the definition…

pages/glossary/text-inheriting-language.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Text with the same Programmatic Language
-key: text-same-language
+title: Text Inheriting its Programmatic Language from an Element
```
WilcoFiers, line 2: Bit verbose, but this works. Thanks.

WilcoFiers (approved):
LGTM

carlosapaduarte (approved):

daniel-montalvo (approved):

Jym77:
Final Call ends on February 10th.

Jym77:

PR #1522, opened 2020-12-16
New rule: Image contains no text (0va7u6)

carlosapaduarte:
New rule for 1.4.5 checking that images do not contain text unless it is essential

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,114 @@
...
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+---
+
+## Applicability
+
+This rule applies to any `img` element that is [visible][] except if the `src` [attribute value][] references an SVG document.
+
+## Expectation
+
+The image resource referenced by the `src` [attribute value][] of each target element either does not contain text expressing anything in a [human language][] or it is [essential][] that the text is rendered with that specific presentation.
```
Jym77, line 30: This should actually look at each element in the source set of the image.
Check https://act-rules.github.io/rules/9eb3f6 where I did that. It's fairly easy to do and will make the rule slightly more powerful. May need to change Applicability accordingly.

Jym77, line 30: I think I prefer forward loading the quantifier.
```suggestion
For each test target, the image resource referenced by its `src` [attribute value][] either does not contain text expressing anything in a [human language][] or it is [essential][] that the text is rendered with that specific presentation.
```

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Assumptions
+
+The text is the most significant content in the image. If there is text in the image, but it is not the most significant content, this rule might fail but the success criterion might still be satisfied.
```
Jym77, line 34: Not sure why this is an assumption rather than part of the Expectation. Do you feel it's ambiguous? (it certainly is subjective).
I guess this is to exclude, say, image of a street with street signs or shop signs showing text. I'm a bit afraid this is going to exclude a ton of images from the rule (i.e. that it is a very strong assumption that often won't hold).

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,114 @@
...
+#### Failed Example 1
+
+This `img` element references an image resource that contains text and the way the text is presented is not relevant.
+
+```html
+<img
+	src="/test-assets/0va7u6/textimage.jpg"
+	alt="The Accessibility Conformance Testing (ACT) Rules Format 1.0 defines a format for writing accessibility test rules."
+/>
+```
```
Jym77, line 76: My first reaction when seeing this example was "Wow, the SC does forbid screenshots?" Seeing the image and reading the SC, I think it doesn't since the presentation can be essential in a screenshot (but not in that image).

…which comes to the point of whether we should clear that upfront with a screenshot Passed Example. Typically instruction pages like https://www.howtogeek.com/school/using-windows-admin-tools-like-a-pro/lesson5/ (OK, not the first image in that page…) with screenshots to show precisely where to click. These seem to pass the SC, I think.

Jym77 (dismissed):

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,114 @@
...
+acknowledgments:
+  authors:
+    - Carlos Duarte
+  images:
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+---
+
+## Applicability
+
+This rule applies to any `img` element that is [visible][] except if the `src` [attribute value][] references an SVG document.
```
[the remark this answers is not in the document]
carlosapaduarte: It's because the user is capable of setting font family and size, color, and background of an SVG `text` element, which is what is required by the SC.
I'll document it in the background.

_rules/image-no-text-0va7u6.md
carlosapaduarte: Good suggestion. And it also makes sense to include image buttons.

_rules/image-no-text-0va7u6.md
carlosapaduarte: Yes, I do feel "most significant content" is ambiguous. That's why I moved this to an assumption.

It would be nice to have input on this from other reviewers...

_rules/image-no-text-0va7u6.md
carlosapaduarte: Interesting... that image is not a screenshot :-) I created an image of text, instead of capturing a screenshoot

A screenshot such as the ones in the page you mention seems to me to fall under the assumption. The SC applies to instances where images of text are used to convey information. Such images are the ones where text was rendered in a way to achieve a particular "visual effect" unless the picture contains other "significant visual content". 

The screenshot:
- Conveys information
- Has text rendered to achieve a particular visual effect, although the visual effect here is not about the styling of the text, but the relative placement of text within the user interface (and I have doubts about the SC applying to this)
- Has other significant visual content (or else it would be a text interface and not a graphical user interface and the tutorial would be about the command line)

What do you think @Jym77 ?

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,130 @@
...
+  images:
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][] for which one of the following is true:
+
+- the element is an `img` element where at least one of the [image sources][] in its [source set][] does not reference an SVG document; or
+- the element is an `input` element in the [Image Button][] state and its `src` [attribute value][] does not reference an SVG document.
```
Jym77, line 29:
```suggestion
- the element is an `input` element with a `type` [attribute value][] of `image` and its `src` [attribute value][] does not reference an SVG document.
```

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Expectation
+
+For each test target, the image resources referenced by its [image sources][] that do not reference an SVG document either do not contain text expressing anything in a [human language][] or it is [essential][] that the text is rendered with that specific presentation.
```
Jym77, line 33: I think I'd prefer the quantification on the image sources to be explicit.
```suggestion
For each test target, each of the image resources referenced by its [image sources][] that do not reference an SVG document either does not contain text expressing anything in a [human language][] or it is [essential][] that the text is rendered with that specific presentation.
```

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,130 @@
...
+acknowledgments:
+  authors:
+    - Carlos Duarte
+  images:
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][] for which one of the following is true:
```
Jym77, line 26: Should there be a mention somewhere that the image is not decorative? ("convey information")

Either by checking roles rather than element (can be in applicability), or adding a condition in expectation about "the image is [pure decoration][]".

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,130 @@
...
+
+- the element is an `img` element where at least one of the [image sources][] in its [source set][] does not reference an SVG document; or
+- the element is an `input` element in the [Image Button][] state and its `src` [attribute value][] does not reference an SVG document.
+
```
Jym77, line 30: Do we also want to add images inserted through `background-image` CSS? Maybe it's a bit hard to shoehorn that without too much twist on the language (not sure "image source" apply to these) and it would be better suited in a separate rule…

_rules/image-no-text-0va7u6.md
Jym77: Yes, I think I agree that the text in this kind on screenshot is not the most significant information 👍

Jym77 (changes requested):
Couple new questions…

_rules/image-no-text-0va7u6.md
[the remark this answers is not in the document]
Jym77: 👍
That does leave out cases where SVG shapes are used to "draw" letters without using `text`. But I guess this is rather unfrequent, usually needed, and anyway only passing the rule too often, so we're good.

_rules/image-no-text-0va7u6.md
carlosapaduarte: This was one of the the first thoughts that crossed my mind when I started thinking about this rule. I discussed this with @WilcoFiers and he easily convinced me that this is not relevant for this SC. If you are presenting text then you should use text and not images of text unless the presentation is essential. You can always mark text as decorative if it is not being used to convey information.

_rules/image-no-text-0va7u6.md
carlosapaduarte: I believe image source also applies here. An image source is a url, and the CSS image type can also be a url reference. I'll give it a go. Let me know if you think it holds up...

_rules/image-no-text-0va7u6.md
Jym77: 🤔 I have to disagree with the conclusion…
1.4.9, the corresponding AAA criterion, is explicitly allowing images of text that are pure decoration. It would be weird if they were failing 1.4.5…

_rules/image-no-text-0va7u6.md
```diff
@@ -26,11 +26,12 @@ acknowledgments:
 This rule applies to any HTML element that is [visible][] for which one of the following is true:
 
 - the element is an `img` element where at least one of the [image sources][] in its [source set][] does not reference an SVG document; or
-- the element is an `input` element in the [Image Button][] state and its `src` [attribute value][] does not reference an SVG document.
+- the element is an `input` element with a `type` [attribute value][] of `image` and its `src` [attribute value][] does not reference an SVG document; or
+- the element has a [`background-image`][background-image] CSS property with at least one [`image`][css-image] that is a [url reference][url-reference] that does not reference an SVG document.
```
Jym77, line 30: Which value of the property do you consider? I guess the computed value (after resolving cascade and such).

_rules/image-no-text-0va7u6.md
Jym77: Yes, that should be good 👍 
CSS is also calling it "image source" (it's the Section title in https://drafts.csswg.org/css-backgrounds-3/#background-image ) so it should be good.

_rules/image-no-text-0va7u6.md
```diff
@@ -121,10 +130,13 @@ There is no `img` element.
 ```
 
 [attribute value]: #attribute-value 'Definition of Attribute Value'
+[background-image]: https://drafts.csswg.org/css-backgrounds-3/#typedef-bg-image
```
Jym77, line 133:
```suggestion
[background-image]: https://drafts.csswg.org/css-backgrounds-3/#background-image
```
to link to the property rather than the type.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,142 @@
+---
+id: 0va7u6
+name: HTML graphics contain no text
+rule_type: atomic
+description: |
+  This rule checks that images of text are not used
+accessibility_requirements:
+  wcag20:1.4.5: # Images of Text (AA)
```
Jym77, line 8: It should probably also map to 1.4.9 (+ possibly need for a note in background saying that examples are designed for 1.4.5 and Failed Examples may not fail 1.4.9, see similar note in color contrast rule).

(+ we might have another "free" rule for 1.4.9 once this is done…)

Jym77 (changes requested):

_rules/image-no-text-0va7u6.md
carlosapaduarte: That line in 1.4.9 really made me change my mind. I've updated the applicability.

_rules/image-no-text-0va7u6.md
carlosapaduarte: Included 1.4.9. Good suggestion

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+
+This rule applies to any HTML element that is [visible][] and not [marked as decorative][] for which one of the following is true:
+
+- the element is an `img` element where at least one of the [image sources][] in its [source set][] does not reference an SVG document; or
+- the element is an `input` element with a `type` [attribute value][] of `image` and its `src` [attribute value][] does not reference an SVG document; or
+- the element has a [computed][] [`background-image`][background-image] CSS property with at least one [`image`][css-image] that is a [url reference][url-reference] that does not reference an SVG document.
+
+## Expectation
+
+For each test target, each of the image resources referenced by its [image sources][] that do not reference an SVG document either does not contain text expressing anything in a [human language][] or it is [essential][] that the text is rendered with that specific presentation.
```
nitedog, line 39: I think this expectation may be leaving out several uses cases in relation to 1.4.5? I think at least one of the following must be true to meet the expectation:

1. image does not contain text expressing anything in a human language (already covered)
2. the particular presentation of text is essential to the information being conveyed (already covered)
3. the font, size, color, and background of the essential text can be customized (not covered)
4. CSS cannot be used to achieve the visual presentation of the essential text (not covered)

Note: assumption is that CSS is part of "technologies being used"

_rules/image-no-text-0va7u6.md
nitedog: The WCAG definition for "image of text" already excludes "text that is part of a picture that contains significant other visual content", so I'm not sure we need to have this?

nitedog (dismissed):
I think the expectations section is inconsistent with SC 1.4.5?

_rules/image-no-text-0va7u6.md
carlosapaduarte: Regarding points 3 and 4, if you read the applicability, the rule leaves out SVG images, which could be customised through CSS. This is made clear in the first paragraph of the background section. I don't think that, using other image types, I can use CSS to change font-family, for instance.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Accessibility Support
+
+_No accessibility support issues known._
+
+## Background
+
+This rule does not consider SVG documents in its applicability because graphically inserted text using the SVG `text` element allows a user to set font family, size and color and the background color of the text, which is what [SC 1.4.5][sc1.4.5] requires.
+
+This rule is designed specifically for [SC 1.4.5 Images of Text][sc1.4.5], which allows images of text as long as the images are in technology that allows the user to customize its presentation. [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] does not even allow this exception, it is possible that this rule will consider some images inapplicable that are in fact applicable for [SC 1.4.9 Images of Text (No Exception)][sc1.4.9].
```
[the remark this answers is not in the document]
carlosapaduarte: I agree that this is not clear. I'll try rewriting.

_rules/image-no-text-0va7u6.md
carlosapaduarte: But we don't link to the WCAG definition of "image of text" so we need to have this here (or worked into the expectation).
And I don't think we could use it, because, at least for me, the part about "significant other visual content" is ambiguous. If you consider the example given in the definition, an image of "a person's name on a name tag in a photograph", assuming that the person is also on the photograph, I wouldn't consider it to be an image of text because there would be other significant visual content on the image.. So, I guess, this is ambiguous...

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Inapplicable Example 3
+
+This `img` element references an SVG document.
+
+```html
+<img src="/test-assets/shared/eu-logo.svg" alt="European Union flag" />
+```
+
+#### Inapplicable Example 4
```
WilcoFiers, line 138: This doesn't seem necessary. If inapplicable 1 and 2 aren't picked up, this one won't be either.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+This rule does not consider SVG documents in its applicability because graphically inserted text using the SVG `text` element allows a user to set font family, size and color and the background color of the text, which is what [SC 1.4.5][sc1.4.5] requires.
+
+This rule is designed specifically for [SC 1.4.5 Images of Text][sc1.4.5] which includes exceptions to the images it applies to. [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] does not consider any exceptions. Therefore, some images that are inapplicable for this rule can be applicable to [SC 1.4.9 Images of Text (No Exception)][sc1.4.9].
+
+- [Understanding Success Criterion 1.4.5: Images of Text][sc1.4.5]
+- [Understanding Success Criterion 1.4.9: Images of Text (No Exception)][sc1.4.9]
+
+## Test Cases
+
+### Passed
```
WilcoFiers, line 60: Just because it is such a common example, I think we should have an example of a logo in here. Would also be good to include different image types in passed examples, not just `img` elements.

WilcoFiers (changes requested):

_rules/image-no-text-0va7u6.md
nitedog: I don't think this requirement is limited to customizing text through CSS and SVG only. Here are some types of images of text, that I think would pass SC 1.4.5:

- Point 3: Image of text generated by script, where the user can set the font, size, color, and background (eg. by buttons/pull-down menus provided by the script, rather than browser/CSS settings)
- Point 4: Image of a mathematical formula used because CSS can't provide the styling and MathML is not a technology being used

I think the current rule would incorrectly fail these types of images/situations?

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+    - Carlos Duarte
+  images:
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][] and not [marked as decorative][] for which one of the following is true:
+
+- the element is an `img` element where at least one of the [image sources][] in its [source set][] does not reference an SVG document; or
```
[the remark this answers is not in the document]
carlosapaduarte: SVGs are considered in the update

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+acknowledgments:
+  authors:
+    - Carlos Duarte
+  images:
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+---
+
+## Applicability
+
+This rule applies to any HTML element that is [visible][] and not [marked as decorative][] for which one of the following is true:
```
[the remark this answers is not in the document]
carlosapaduarte: Good idea. Including definition of "embedded image" and considering the suggested elements.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,161 @@
...
+
+The text is the most significant content in the image. If there is text in the image, but it is not the most significant content, this rule might fail but the success criterion might still be satisfied.
+
+## Accessibility Support
+
+_No accessibility support issues known._
+
+## Background
+
+This rule does not consider SVG documents in its applicability because graphically inserted text using the SVG `text` element allows a user to set font family, size and color and the background color of the text, which is what [SC 1.4.5][sc1.4.5] requires.
```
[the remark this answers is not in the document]
carlosapaduarte: You are correct.
The updated version considers any SVG when it is embedded, but when using the `svg` element only images embedded through `image` svg elements are considered.

_rules/image-no-text-0va7u6.md
carlosapaduarte: My main concern would not be the "how much significant" but the "is this part of the image significant". As I said in my previous reply, the example of "image of text" in the WCAG definition is not something that I would consider an image of text (it's true that the example needs context that is not provided, so I had to assume the context), so I have to conclude that the "significant other visual content" is being ambiguously interpreted. 

Nevertheless, I seem to be the only one having this doubt, so I can give it a try and include this in the expectation.

_rules/image-no-text-0va7u6.md
carlosapaduarte: Multiple new examples have been included

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,205 @@
...
+
+For each of the image resources referenced by the [image sources][] of the test target, any of the following is true:
+
+- the image resource does not contain text expressing anything in a [human language][]; or
+- the text in the image resource is not the most significant visual content; or
+- it is [essential][] that the text in the image resource is rendered with that specific presentation.
+
+## Assumptions
+
+_There are currently no assumptions_
```
Jym77, line 44: I'm starting to think about a nasty corner case where 2 images are overlapping, the "bottom" one contains text, but the "top" one covers all the text, leaving a bit of non-text visible (e.g. a bit of the "background" of the bottom image).

Then, the bottom image would be applicable (it is visible because bit of its background is) and fail the rule (because of its text) but wouldn't fail the SC because the text is actually not visible…

=> Should there be an assumption that text in the image is visible? I don't think we can use our definition of "visible" since making part of a .jpg transparent is not really a thing. That should be corner case enough that it doesn't really happen often, so an assumption should be enough.



=> I'm also starting to freak out a bit that a similar situation happens with other rules that apply to "visible" stuff.

Jym77 (changes requested):
Good job!

_rules/image-no-text-0va7u6.md
```diff
@@ -24,32 +24,31 @@ acknowledgments:
...
+This rule applies to any [visible][] [embedded image][] that is not [marked as decorative][].
 
 ## Expectation
 
-For each test target, each of the image resources referenced by its [image sources][] that do not reference an SVG document either does not contain text expressing anything in a [human language][] or it is [essential][] that the text is rendered with that specific presentation.
+For each of the image resources referenced by the [image sources][] of the test target, any of the following is true:
+
+- the image resource does not contain text expressing anything in a [human language][]; or
+- the text in the image resource is not the most significant visual content; or
+- it is [essential][] that the text in the image resource is rendered with that specific presentation.
```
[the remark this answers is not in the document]
carlosapaduarte: I think this is a good suggestion so I'm using it, even if we haven't used in the past :-)

_rules/image-no-text-0va7u6.md
carlosapaduarte: Good catch. I'm adding the assumption, and later we can think about adding a similar assumption to other rules in need.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,205 @@
...
+## Applicability
+
+This rule applies to any [visible][] [embedded image][] that is not [marked as decorative][].
+
+## Expectation
+
+For each of the image resources referenced by the [image sources][] of the test target, at least one of the following is true:
+
+- the image resource does not contain text expressing anything in a [human language][]; or
+- the image resource contains text and the text is not the most significant visual content in the image; or
```
nitedog, line 39: I'm not sure about "most significant" as the criteria. If I have an arrow with the text "one way", is that text the most significant? Maybe better to say something like "not significant part of the visual content in the image" instead?

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,205 @@
...
+- the image resource contains text and it is [essential][] that the text is rendered with that specific presentation.
```
nitedog, line 40: I think we need to add the following situations as well:

- the image contains text, and the font, size, color, and background of the text in the image can be set;
- the image contains text, and the visual presentation of the text in the image cannot be achieved through CSS

nitedog, line 40: Do we need the "or" between every bullet? It seems the leading sentence, "at least one of the following is true", is sufficient.

carlosapaduarte: That's the current writing style for lists of items in rules :-)

carlosapaduarte: Regarding the first point: as far as I know, all the options in the "embedded image" definition do not allow setting font, size, color and background. Note that SVGs are only applicable if the image is embedded through the `image` tag. And SVGs being used as source for the other elements do not allow that as Wilco pointed out (and I confirmed).

Regarding the second point: what is the difference between that and the last point? If I cannot achieve the presentation with CSS and I chose to use an image of text then it must be because rendering the text with that specific presentation is essential. Otherwise, I wouldn't use an image of text and would render the text with a different presentation.

_rules/image-no-text-0va7u6.md
carlosapaduarte: Good suggestion.

_rules/image-no-text-0va7u6.md
nitedog: Regarding the first point, I think there are situations when you can set the font, size, color, etc. of text that is generated inside embedded images -- for example in some meme generators. The image itself is static but the script/app can regenerate the image according to some settings/preferences.

Regarding the second point, I think "essential" is like a logo or such -- even if that presentation can be done by CSS, it may be deemed necessary to keep the exact look or proportions, etc. I think 1.4.5 allows also non-essential text to be presented via images if that styling cannot be done by CSS (these situations have become rare with the steady improvements to CSS).

carlosapaduarte: I've been trying to find a meme generator that allows for such customisation (in run-time, not in generation-time) but I couldn't find one. I would really need to have a look at one so that I can understand if what is being used is something that falls under our definition of embedded image. 

Regarding that second point, what you're asking is a huge problem. Adding such a clause (it passes if there is no way to achieve that presentation via css - or other technology) would make it the burden of the evaluator to prove that there is no way to achieve such presentation with css. There would be situations where an evaluator would not be able to demonstrate that. But that does not mean it isn't possible, just that the evaluator does not know how to achieve it. So, we can't have that has an expectation. Would an assumption such as "the rule assumes that for the images of text it is not possible to achieve the desired presentation with text-based technology" work for you?

Jym77: > I've been trying to find a meme generator that allows for such customisation (in run-time, not in generation-time) but I couldn't find one. I would really need to have a look at one so that I can understand if what is being used is something that falls under our definition of embedded image.

I guess we can definitely imagine a page with an image of text and some scripting that allow resizing that text (e.g. trough a slider), maybe by having a client-side mash-up of the image + text or something. As often, there isn't anything we can't do with JS 😕 

In that case, the meme generator page itself is good (allow resize), but the result is bad (does not allow resize).

But that could probably go through an assumption that the page does not provide ways of resizing the text. Assumption because anything that is dynamic (JS) yield undecidable problems as soon as we try to look into it, so it can't really be in expectations.

carlosapaduarte: OK. Included two new assumptions to account for these two cases.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,207 @@
...
+- The page does not provide a mechanism to change the rendered text in the image resource. Otherwise, the rule might fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
+- The specific presentation of the text rendered in the image resource can be achieved through formatted text. Otherwise, the rule might fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
+
+## Accessibility Support
+
+_No accessibility support issues known._
+
+## Background
+
+This rule is designed specifically for [SC 1.4.5 Images of Text][sc1.4.5] which includes exceptions to the images it applies to. [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] does not consider any exceptions. Therefore, some images that are inapplicable for this rule can be applicable to [SC 1.4.9 Images of Text (No Exception)][sc1.4.9].
```
nitedog, line 54: Actually 1.4.9 **does** have an (implicit) exception, when the presentation is essential (eg. logo).

_rules/image-no-text-0va7u6.md
nitedog: I don't have sufficient knowledge to assess the inaccuracy that this approach introduces but I feel quite ambivalent about it. This is now a rule for 1.4.9 rather than for 1.4.5. Happy to discuss.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,207 @@
...
+
+For each of the image resources referenced by the [image sources][] of the test target, at least one of the following is true:
+
+- the image resource does not contain text expressing anything in a [human language][]; or
+- the image resource contains text and the text is not a significant part of the visual content in the image; or
+- the image resource contains text and it is [essential][] that the text is rendered with that specific presentation.
+
+## Assumptions
+
+- The text in the image resource is visible and not hidden by other resources rendered on the web page. If the text is not visible, even though parts of the image are, this rule may fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
```
nitedog, line 44: I don't understand the rationale for this assumption. Why would such a situation fail 1.4.5 but not 1.4.9?

Jym77: the **rule** would fail. Both SCs would be satisfied.

_rules/image-no-text-0va7u6.md
```diff
@@ -33,15 +33,17 @@ This rule applies to any [visible][] [embedded image][] that is not [marked as d
...
-- it is [essential][] that the text in the image resource is rendered with that specific presentation.
+- the image resource contains text and the text is not a significant part of the visual content in the image; or
+- the image resource contains text and it is [essential][] that the text is rendered with that specific presentation.
 
 ## Assumptions
 
-_There are currently no assumptions_
+- The text in the image resource is visible and not hidden by other resources rendered on the web page. If the text is not visible, even though parts of the image are, this rule may fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
+- The page does not provide a mechanism to change the rendered text in the image resource. Otherwise, the rule might fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
+- The specific presentation of the text rendered in the image resource can be achieved through formatted text. Otherwise, the rule might fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
```
Jym77, line 46: Wouldn't the photo of an old letter (used in examples) fail that assumption?
This looks like it is going to often overlap with the "essential" expectation. As in, if specific presentation is essential but can be done by CSS, it should; if specific presentation is essential but cannot be done by CSS, then OK, use an image.

Jym77 (changes requested):

_rules/image-no-text-0va7u6.md
carlosapaduarte: That is why we have this note in the background. Decorative images are inapplicable for this rule, but they are applicable (and conformant) for 1.4.9.

_rules/image-no-text-0va7u6.md
carlosapaduarte: I believe the second condition in the expectation makes this a rule for 1.4.5.
Why do you say that this is only about 1.4.9?

_rules/image-no-text-0va7u6.md
carlosapaduarte: This is really context dependant. If you argue the image is just showing cursive handwriting, then yes, it could be done with CSS. If you argue the image is showing human cursive handwriting, where the same letter is never written exactly in the same way, and the lines are not really straight, and so on, then no, it cannot be done using (text formatted) CSS.

It would be up for the evaluator to judge if the assumption is met or not and decide on whether to take into account the results of the rule. If we are not happy with that for this rule/assumption, then I see only one way forward: make this a 1.4.9 only rule, where this assumption would not be needed anymore (neither the other assumptions).

Jym77 (approved):

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,209 @@
...
+	</image>
+</svg>
+```
+
+#### Passed Example 4
+
+This `object` element references an image resource where text is not the most significant content.
+
+```html
+<object data="/test-assets/0va7u6/times_square.jpg"></object>
```
WilcoFiers, line 97:
```suggestion
<object data="/test-assets/0va7u6/times_square.jpg" title="Picture of Times Square, New York"></object>
```

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+#### Passed Example 5
+
+This `img` element references an image resource that contains text but where the presentation of the text is essential to convey the information.
+
+```html
+<p>The following image illustrates the use of cursive writing in the late nineteenth century.</p>
+
+<canvas id="canvas"> Your browser does not support the HTML5 canvas element. </canvas>
```
WilcoFiers, line 107: I may be contradicting my previous review here, but I think we should exclude canvas. It can be really difficult, if not practically impossible to determine the source of a canvas, and in addition if you do want customizable images of text, canvas is the way to do it.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,209 @@
...
+</script>
+```
+
+#### Passed Example 6
+
+This `div` element has a background image with a logo with text. Logotypes are considered an essential exception.
+
+```html
+<div
+	style="
```
WilcoFiers, line 131: Lets make this accessible. :)

```suggestion
	role="img"
	aria-label="W3C logo"
	style="
```

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,209 @@
...
+    height: 100px;
+    background-image: url(test-assets/shared/w3c-logo.png);
+    background-repeat: no-repeat;
+  "
+></div>
+```
+
+#### Passed Example 7
+
+This `img` element is [purely decorative][].
```
WilcoFiers, line 142: We should make this a failed example. An image doesn't become decorative when you put a caption under it. The information might be redundant, but it's still information. This is how we interpreted it in the decorative images rule. Specifically [inapplicable example 8](https://act-rules.github.io/rules/e88epe#inapplicable-example-8) has this case in it.

WilcoFiers (changes requested):

_rules/image-no-text-0va7u6.md
carlosapaduarte: In the "embedded image" definition I'm considering only canvas images that are rendered using the drawImage method. I can add an inapplicable example of using canvas with other rendering methods. Or would you still prefer to remove canvas from the applicability?

_rules/image-no-text-0va7u6.md
carlosapaduarte: I've updated the example (alt text and image used) so that this could be a passed example and moved this one to a failed example.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,219 @@
...
+  - DOM Tree
+  - CSS Styling
+  - Language
+acknowledgments:
+  authors:
+    - Carlos Duarte
+  images:
+    - Letter posted from Dulverton (Somerset, England) to Bristol (England) in 1894. Released into the public domain by Adrian Pingstone.
+    - Times Square image released into the public domain by (WT-shared) Ypsilonatshared at wts wikivoyage.
+---
```
Jym77, line 28: Should the words cloud have an entry here (not sure if you copied it or built it yourself…)

Jym77 (approved):

_rules/image-no-text-0va7u6.md
carlosapaduarte: I built it myself with a random words generator and word cloud generator :-)

_rules/image-no-text-0va7u6.md
```diff
@@ -29,20 +29,21 @@ acknowledgments:
...
 
-This rule applies to any [visible][] [embedded image][] that is not [marked as decorative][].
+This rule applies to any [visible][] [embedded image][].
 
 ## Expectation
 
-For each of the image resources referenced by the [image sources][] of the test target, at least one of the following is true:
+For each test target, at least one of the following is true:
 
+- the image resource contains text and it is [essential][] that the text is rendered with that specific presentation; or
```
WilcoFiers, line 38: We need to define what the image source is.

_rules/image-no-text-0va7u6.md
WilcoFiers: This description needs an update.

WilcoFiers (changes requested):

_rules/image-no-text-0va7u6.md
```diff
@@ -141,7 +141,7 @@ This `div` element has a background image with a logo with text. Logotypes are c
 
 #### Passed Example 7
 
-This `img` element is [purely decorative][].
+This `img` element is an image of text, but the text does not provide information (i.e., it could be rearranged and the meaning wouldn't change), therefore it is [purely decorative][].
 
 ```html
 <img src="/test-assets/0va7u6/wordcloud.png" alt="Word cloud of random words" />
```
Jym77, line 147: If it is purely decorative, shouldn't it have an empty `alt` (so it doesn't fail 1.1.1)?

_rules/image-no-text-0va7u6.md
```diff
@@ -209,10 +209,30 @@ This `svg` element does not have `image` element descendants.
...
+
+<script>
+	const canvas = document.getElementById('canvas')
+	canvas.width = 400
+	canvas.height = 400
+	const ctx = canvas.getContext('2d')
+	ctx.font = 'bold 32pt Times'
+	ctx.fillStyle = '#c2c'
+	ctx.fillText('Welcome!', 100, 100)
+</script>
```
Jym77, line 227: Non sure how "customizable" this text is. Given the SC, if the text is not customizable by user, we shouldn't have an Inapplicable example with it.
Can still be just drawing a circle or whatever without text.

Jym77 (changes requested):

_rules/image-no-text-0va7u6.md
carlosapaduarte: Replaced with another example, as discussed in the call

_rules/image-no-text-0va7u6.md
carlosapaduarte: Fair point. I could make the example customisable but I don't think we need a lot of scripting to convey the message that not all `canvas` elements are inapplicable

Jym77 (approved):

_rules/image-no-text-0va7u6.md
WilcoFiers: How is someone supposed to know if drawImage was used? There is no API you can call to figure that out. That doesn't seem possible.

_rules/image-no-text-0va7u6.md
WilcoFiers: This definition doesn't really work. I think I know what you're trying to say here, but the image source is a URL, not the actual image. I also don't think you can use canvas the way you're using it. What is the image source of a canvas? Even assuming someone can figure out whether or not it used drawImage to render; what if multiple drawImage calls were made?

pages/glossary/embedded-image.md
```diff
@@ -0,0 +1,28 @@
...
+  - DOM tree
+---
+
+An element presents an _embedded image_ when any of the following is true:
+
+- the element is an `img` element with a non-empty [source set][]; or
+- the element is an `input` element with a `type` [attribute value][] of `image` and its `src` [attribute value][] is not empty; or
+- the element is a `canvas` element with an [image][canvasImage] [drawn][drawImage]; or
+- the element is an `object` element with a `data` [attribute value][] referencing a resource with an [image MIME type](https://mimesniff.spec.whatwg.org/#image-mime-type); or
+- the element is an `svg` element having one or more `image` [descendants][] with a non-empty `href` [attribute value][]; or
```
WilcoFiers, line 16: What is the image source of an svg with multiple image elements? I think this need to target the individual image elements, not the svg for this rule to work.

pages/glossary/embedded-image.md
```diff
@@ -0,0 +1,28 @@
...
+- the element has a [computed][] [`background-image`][background-image] CSS property with at least one [`image`][css-image] that is a [url reference][url-reference].
```
WilcoFiers, line 17: Similar to the above, background can have multiple images. I think the rule needs to handle that.

WilcoFiers (changes requested):

_rules/image-no-text-0va7u6.md
carlosapaduarte: Would "rendered image resource" (as used in the first line of the expectation" solve this issue for you?

Regarding canvas, I'm going to drop it based on the difficulty that is finding out if drawImage is being used or not.

pages/glossary/embedded-image.md
carlosapaduarte: The expectation of the rule does target the individual rendered image sources, so it's in accordance to what you're asking.

pages/glossary/embedded-image.md
carlosapaduarte: The first assumption was added exactly because of such cases. Given that we are testing every rendered image source, some may overlap others, therefore we check each individually and we might fail some instances where text in one source is in fact obscured by other sources.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,205 @@
...
+
+- the rendered image resource contains text and it is [essential][] that the text is rendered with that specific presentation; or
+- the rendered image resource does not contain text expressing anything in a [human language][]; or
+- the rendered image resource contains text and the text is not a significant part of the visual content in the image; or
+- the test target is [purely decorative][].
+
+## Assumptions
+
+- The text in the image resource is visible and not obscured by other resources rendered on the web page. If the text is not visible, even though parts of the image are, this rule may fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
+- There is no mechanism to change the rendered text in the image resource. Otherwise, the rule might fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
```
nitedog, line 47: editorial: "might by satisfied" -> "might be satisfied"

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,205 @@
...
+- The specific presentation of the text rendered in the image resource can be achieved through formatted text. Otherwise, the rule might fail while [SC 1.4.5 Images of Text][sc1.4.5] and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9] might by satisfied.
```
nitedog, line 48: editorial: "might by satisfied" -> "might be satisfied"

_rules/image-no-text-0va7u6.md
nitedog: I agree with @Jym77 that this assumption can be tweaked a little. Maybe something like this:

"For images that present significant text, the specific presentation is either essential or can be presented as formatted text (eg. through CSS)."

_rules/image-no-text-0va7u6.md
nitedog: I think the middle sentence ("[SC 1.4.9 Images of Text (No Exception)][sc1.4.9] does not consider any exceptions." is wrong and can be safely removed. Consider this rewrite:

"This rule is designed specifically for [SC 1.4.5 Images of Text][sc1.4.5], which includes exceptions that are not part of [SC 1.4.9 Images of Text (No Exception)][sc1.4.9]. Therefore, some images that are inapplicable for this rule can be applicable to [SC 1.4.9 Images of Text (No Exception)][sc1.4.9].

carlosapaduarte:
@nitedog @WilcoFiers @Jym77 changes made in accordance to what we talked about. please review again

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,204 @@
...
+<object data="/test-assets/0va7u6/times_square.jpg" title="Picture of Times Square, New York"></object>
+```
+
+#### Passed Example 5
+
+This `img` element references an image resource that contains text but where the presentation of the text is essential to convey the information.
+
+```html
+<p>The following image illustrates the use of cursive writing in the late nineteenth century.</p>
+<img src="/test-assets/0va7u6/letter.jpg" alt="Nineteenth century handwritten letter" />
```
nitedog, line 106: This is independent of this particular SC but I would argue that the text alternative is probably insufficient -- if the hand writing is essential then the text alternative probably needs to describe that hand writing a little more. Maybe show only one character/letter and describe the glyph in just a little more detail?

nitedog: Also, the text alternative should contain all the text visible in the image! If it is an entire letter then it may need some form of longer description. If you show, again, just one letter then it may be a little easier.

_rules/image-no-text-0va7u6.md
```diff
@@ -36,24 +36,23 @@ This rule applies to any [visible][] [embedded image][].
 
 For the rendered image resource of the [image sources][] of each test target, at least one of the following is true:
 
-- the image resource contains text and it is [essential][] that the text is rendered with that specific presentation; or
-- the image resource does not contain text expressing anything in a [human language][]; or
-- the image resource contains text and the text is not a significant part of the visual content in the image; or
+- the [visible pixels][visible] of the rendered image resource contain text and it is [essential][] that the text is rendered with that specific presentation; or
+- the [visible pixels][visible] of the rendered image resource do not contain text expressing anything in a [human language][]; or
+- the [visible pixels][visible] of the rendered image resource contain text and the text is not a significant part of the visual content in the image; or
```
Jym77, line 41: I'm wondering if we should restrict also to "significant part of the visible pixels".
If everything from the image is hidden except the bit that contain text, then that text was maybe not the main part of the full image but is the main part of what's left and this is probably bad.

I'm not fully sure of this because there might be cases where this is OK and we don't want false positives, but every case where this is OK I can think of tend to have that below image with text showing actually become decorative (and pass the next item in the list)… So I  just leave it here…
(in any case, this only causes the rule to fail too often so it is OK).

Jym77 (approved):

_rules/image-no-text-0va7u6.md
nitedog: Coming to think of it, I think this would be better as an entire word rather than a single letter.

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+#### Passed Example 7
+
+This `img` element is an image of text (the book covers), but it is just meant to decorate the webpage of a book store, therefore it is [purely decorative][].
+
+```html
+<img src="/test-assets/0va7u6/books.jpg" alt="" />
+<p>Welcome to my book store</p>
+```
```
nitedog, line 133: I suggest another passed example: button image that has letters, such as large and small "T" for changing text size, or an "F" for changing font type, etc. -- buttons we commonly find in editors and word processors but do not actually qualify as images of text even though they include text.

_rules/image-no-text-0va7u6.md
carlosapaduarte: I agree, and will update accordingly.

_rules/image-no-text-0va7u6.md
carlosapaduarte: Done

_rules/image-no-text-0va7u6.md
carlosapaduarte: Added

carlosapaduarte:
@nitedog ready for your review again (changes made to the examples)

nitedog (approved):

_rules/image-no-text-0va7u6.md
```diff
@@ -36,24 +36,23 @@ This rule applies to any [visible][] [embedded image][].
 
 For the rendered image resource of the [image sources][] of each test target, at least one of the following is true:
 
-- the image resource contains text and it is [essential][] that the text is rendered with that specific presentation; or
-- the image resource does not contain text expressing anything in a [human language][]; or
-- the image resource contains text and the text is not a significant part of the visual content in the image; or
+- the [visible pixels][visible] of the rendered image resource contain text and it is [essential][] that the text is rendered with that specific presentation; or
```
WilcoFiers, line 39: This seems like it could be simplified:

```suggestion
- the [visible pixels][visible] of the rendered image resource contain text for which its presentation is [essential]; or
```

_rules/image-no-text-0va7u6.md
```diff
@@ -36,24 +36,23 @@ This rule applies to any [visible][] [embedded image][].
 
 For the rendered image resource of the [image sources][] of each test target, at least one of the following is true:
```
WilcoFiers, line 37: I think we should hoist the visible pixels thing into the setup to avoid repeating that phrase 3 times:

```suggestion
For the [visible pixels][] of the rendered image resource of the [image sources][] of each test target, at least one of the following is true:
```

pages/glossary/embedded-image.md
WilcoFiers: It is not clear to me how you intend for this to work. Perhaps the phrase rendered image resource is ambiguous. I'm not sure, but I don't see how this rule would work in the case of multiple images rendered by the same element. Be it repeating backgrounds, svg with multiple image elements, or canvas with multiple drawn parts.

pages/glossary/embedded-image.md
WilcoFiers: But you don't though. The applicability isn't testing visibly rendered image resources. It's testing elements, which may render multiple images at the same time.

WilcoFiers (changes requested):

_rules/image-no-text-0va7u6.md
carlosapaduarte: The issue is that there is a 4th condition...

_rules/image-no-text-0va7u6.md
```diff
@@ -0,0 +1,234 @@
...
+
+This image resource referenced by the `input` element does not contain text.
+
+```html
+<input type="image" src="test-assets/shared/file.svg" alt="New file" />
+```
+
+#### Passed Example 3
+
+This image resource displayed by the `svg` element does not contain text.
```
Jym77, line 82:
```suggestion
This image resource referenced by the `svg` element does not contain text.
```

Jym77 (approved):
I think the wordcloud asset is not used anymore and can be deleted.

_rules/image-no-text-0va7u6.md
```diff
@@ -23,23 +23,23 @@ acknowledgments:
...
+For each test target, at least one of the following is true:
 
-- the [visible pixels][visible] of the rendered image resource contain text and it is [essential][] that the text is rendered with that specific presentation; or
-- the [visible pixels][visible] of the rendered image resource do not contain text expressing anything in a [human language][]; or
-- the [visible pixels][visible] of the rendered image resource contain text and the text is not a significant part of the visual content in the image; or
-- the test target is [purely decorative][].
+- the [visible pixels][visible] of the test target contain text for which its presentation is [essential][]; or
+- the [visible pixels][visible] of the test target do not contain text expressing anything in a [human language][]; or
+- the [visible pixels][visible] of the test target contain text and the text is not a significant part of the [visible pixels][visible] of the test target; or
+- the [embedded image][] the test target belongs to is [purely decorative][].
```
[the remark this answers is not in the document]
carlosapaduarte: I would prefer to keep it as a list with all the conditions for the reasons you enumerate. If others prefer this solution, I won't oppose to the change.

WilcoFiers (approved):

nitedog (approved):
There are still some comments on the definition for Embedded Image but I understand that these are mostly editorial.

carlosapaduarte:
Final call until February 10

carlosapaduarte:
Final call ended. Merging

PR #1523, opened 2020-12-17
Take out unnecessary condition from hidden-state definition

WilcoFiers:
The "hidden" attribute gives an element the display:none style, so calling it out as a separate condition is unnecessary.

Need for Final Call: none, editorial

---

## How to Review And Approve

### DIFF
--- a/pages/glossary/hidden-state.md
+++ b/pages/glossary/hidden-state.md
@@ -10,7 +10,6 @@ input_aspects:
 
 An HTML element's _hidden state_ is "true" if at least one of the following is true for itself or any of its [ancestors][] in the [flat tree][]:
 
-- has a `hidden` attribute; or
 - has a [computed][] CSS property `display` of `none`; or
 - has a [computed][] CSS property `visibility` of `hidden`; or
 - has an `aria-hidden` attribute set to `true`


carlosapaduarte:
@WilcoFiers `hidden` being implemented as `display: none` is not normative, right? It is just the default rendering suggestion in the HTML spec.
So, do we really need to remove it from this list?

Jym77:
> @WilcoFiers `hidden` being implemented as `display: none` is not normative, right? It is just the default rendering suggestion in the HTML spec.
> So, do we really need to remove it from this list?

Correct https://html.spec.whatwg.org/multipage/interaction.html#attr-hidden says:
> User agents should not render elements that have the hidden attribute specified. This requirement may be implemented indirectly through the style layer. For example, an HTML+CSS user agent could implement these requirements using the rules suggested in the Rendering section.

But at the same time, there is a note saying that:
> Because this attribute is typically implemented using CSS, it's also possible to override it using CSS. For instance, a rule that applies 'display: block' to all elements will cancel the effects of the hidden attribute. Authors therefore have to take care when writing their style sheets to make sure that the attribute is still styled as expected.

which does means that our definition is somewhat incorrect on stuff like `<div hidden style:"display:block">` which has a "hidden state" of true because of its `hidden` attribute but is not really hidden… (maybe needs a slightly different styling, haven't checked the conflict resolution here).

I guess the easier fix is to add an assumption that `hidden` is implemented through `display: none` (on top of this PR), which should both correct the problem and make it explicit the reason why "hidden state" doesn't mention `hidden` (I see that easily creating confusion…) Main drawback is that the assumption needs to be copied in the many rules using the definition…

WilcoFiers:
@Jym77 I think `<div hidden style="display:block">` is actually a good reason to change it, because the hidden state of that content is `false` in the accessibility trees I checked.

Jym77:
> @Jym77 I think `<div hidden style="display:block">` is actually a good reason to change it, because the hidden state of that content is `false` in the accessibility trees I checked.

@WilcoFiers I agree.
But given that this is mostly serendipity (as in, it happens that UAs currently implement `hidden` via CSS, but this is not mandated by specs), I think this needs an assumption or a background note.

PR #1534, opened 2021-01-18
Autocomplete attribute has valid value- editorial changes in response to TF feedback

ajanec01:
A couple of changes made in response to the [Task Force feedback](https://github.com/w3c/wcag-act/issues/501)

- Edited paragraph two in the assumptions section to make it more explicit
- >Passed 2, 3, 5, 6, 7 examples need more visible label info to confirm that the optional tokens in the attribute values are correct.
-  Changed passed example 8 to a failed example- done in response to the following comments (don't think there's any other role that could be applied since input roles are a subclass of widgets): 
>Seems like Passed example 8 should be inapplicable since the widget role is no longer an input - it is overridden by ARIA but not as an input, select or textarea.
>I'm not too keen on passed example 8. Can we find a better role for this? 
- Added the disabled element definition to the applicability section in response to

> The rule does not account for different types of disabled elements. Only the aria-disabled attribute. Inapplicable example 6 seems like it should be a pass instead.

Need for Final Call: 

This will not require a Final Call- editorial changes


_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -24,7 +24,7 @@ acknowledgments:
 This rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` [attribute value][] that is neither empty (`""`) nor only [ASCII whitespace][], except if one of the following is true:
 
 - **hidden**: the element is not [visible][], and not [included in the accessibility tree][]; or
-- **disabled**: the element has an `aria-disabled` [attribute value][] of `true`; or
+- **disabled**: the element has the `disabled` attribute or the `aria-disabled` [attribute value][] of `true`; or
```
Jym77, line 27: This should probably use https://act-rules.github.io/glossary/#disabled-element instead.

Also, this question the need for the last assumption. I essentially understood the rule (+assumption) as looking only at `aria-disabled` because it assumed that if the element is disabled (common meaning), this is reflected in the `aria-disabled` attribute, notably, if `disabled` is set, then `aria-disabled` is also set (by default) (see https://www.w3.org/TR/html-aam-1.0/#html-attribute-state-and-property-mappings / disabled, and actually if trying to set both, then `disabled` seems to take precedence).

I think that the definition (and its assumption) is capturing this idea even better (it was written after this rule) and should be used here also.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -98,7 +97,7 @@ This `autocomplete` [attribute value][] only has the required token, and is vali
 This `autocomplete` [attribute value][] only has the required token, and is valid for a `textarea` element. Mixing upper and lower case letters is allowed for `autocomplete` attributes.
 
 ```html
-<textarea autocomplete="Street-Address"></textarea>
+<label for="sAddress">Street address</label> <textarea id="sAdress" autocomplete="Street-Address"></textarea>
```
Jym77, line 100:
```suggestion
<label for="sAddress">Street address</label> <textarea id="sAddress" autocomplete="Street-Address"></textarea>
```
(or rather use implicit label)

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -86,7 +84,8 @@ This `autocomplete` [attribute value][] only has the required token, and is vali
 
 ```html
 <form autocomplete="off">
-	<select autocomplete="bday-month">
+	<label for="bday">Birthday month</label>
```
Jym77, line 87: I think we should stick to implicit labels (`<label>Foo <widget>…</widget> </label>`) rather than explicit ones (`<label for="…">`) because this is not the core of the rule (this is not a rule about labels, so no need to show all ways to give a label), all the existing cases are using implicit labels, and it limits errors in IDs.

Jym77 (changes requested):

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: The disabled element definition, as far as I understand it, applies to elements such as `<button disabled>Submit</button>` or to elements that are in the shadow DOM. It does not seem to apply to elements such as `<button aria-disabled>Submit</button>`. I'm trying to cover both scenarios but I agree that we should use the definition of the disabled element instead of "disabled" attribute. 

Am I right in saying that you're referring to `<button disabled aria-disabled>Submit</button>`?

ajanec01: Sorry, I think that the penny dropped, we should use the disabled element definition because `disabled` is mapped to platform-specific accessibility API's as `aria-disabled="true"`, is that correct?

Jym77: The second point in the definition is about `aria-disabled` (and should read "inclusive ancestor", rather than "ancestor" 🙈 )

"shadow including" ancestor means that it indeed apply to elements in the shadow DOM, but not only. `<div><span></span></div>`, the `div` is both an ancestor and a shadow including ancestor (i.e., it's shadow including, not light excluding).

Copying DOM standard, emphasise mine:
> An object A is a shadow-including descendant of an object B, if **A is a descendant of B**, or A’s root is a shadow root and A’s root’s host is a shadow-including inclusive descendant of B.
>
> A shadow-including inclusive descendant is an object or one of its shadow-including descendants.
>
> An object A is a shadow-including ancestor of an object B, if and only if B is a shadow-including descendant of A.
>
> A shadow-including inclusive ancestor is an object or one of its shadow-including ancestors.

So, regular (inclusive) ancestors/descendants are also shadow-including ones.

ajanec01: Took a while, but got it in the end. Should be fixed now. Thanks!

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -24,7 +24,7 @@ acknowledgments:
 This rule applies to any HTML `input`, `select` and `textarea` element with an `autocomplete` [attribute value][] that is neither empty (`""`) nor only [ASCII whitespace][], except if one of the following is true:
 
 - **hidden**: the element is not [visible][], and not [included in the accessibility tree][]; or
-- **disabled**: the element has an `aria-disabled` [attribute value][] of `true`; or
+- **disabled**: the test target is the [disabled element]; or
```
Jym77, line 27:
```suggestion
- **disabled**: the element is a [disabled element]; or
```
all conditions starts with "the element" rather than "the test target".

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -86,10 +84,12 @@ This `autocomplete` [attribute value][] only has the required token, and is vali
 
 ```html
 <form autocomplete="off">
-	<select autocomplete="bday-month">
-		<option>January</option>
-		<option>...</option>
-	</select>
+	<label>Birthday month
+        <select id="bday" autocomplete="bday-month">
```
Jym77, line 88:
```suggestion
        <select autocomplete="bday-month">
```
No need for an `id` with an implicit label.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -98,7 +98,7 @@ This `autocomplete` [attribute value][] only has the required token, and is vali
 This `autocomplete` [attribute value][] only has the required token, and is valid for a `textarea` element. Mixing upper and lower case letters is allowed for `autocomplete` attributes.
 
 ```html
-<textarea autocomplete="Street-Address"></textarea>
+<label> Street address<textarea id="sAddress" autocomplete="Street-Address"></textarea></label>
```
Jym77, line 101:
```suggestion
<label> Street address<textarea autocomplete="Street-Address"></textarea></label> 
```

Jym77 (changes requested):
Small editorial polishing.

Jym77 (approved):

carlosapaduarte (approved):

WilcoFiers (approved):

PR #1537, opened 2021-01-20
"`role` attribute has valid value" (674b10): map to 1.3.1

### LINKED ISSUE #1496: What to do of "`role` attribute has valid value"?
# History
So, we have a rule [`role` attribute has valid value](https://act-rules.github.io/rules/674b10) it initially mapped to 4.1.2 Name Role Value, but the first TF review pointed out that this can happen on things that are not UI components and thus the mapping is incorrect (https://github.com/w3c/wcag-act/issues/467).

While addressing that concern, I couldn't really found a SC to map it to. Instead, the best I could find was the various ARIA specs and lists of roles: #1439. See notably the discussions with @carlosapaduarte and @WilcoFiers about the mapping issue.

The new TF review both pointed out that these mappings are sub-optimal and that without mapping to the SC the rule is not very good: https://github.com/w3c/wcag-act/issues/492

I'm now a bit at a lost as to what to do with the rule…

# Technical
Of course, the problem comes because a `role` attribute can be set on any element to any value. And the very nature of the rule prevents us from targeting "element with a widget role" or similar, because, well, the role is presumably invalid or the rule has no reason to be here…

The problem is made more difficult because the `role` attribute is **not an ARIA specific role**. `role` can be used for many reasons and has its own specs: https://www.w3.org/TR/role-attribute/#s_role_module_attributes Thus, we cannot simply map to the ARIA definition of `role` because it does not exist.
ARIA defines a list of roles, but the `role` attribute is not really restricted to these values. I would still stay that using a different violates ARIA specs, but this is not explicitly mentioned in ARIA specs.

# Possibilities
I think the rule can be kept at the very least as a good practice.
It is fairly easy to make a rule mapping to 4.1.2 and targeting elements in sequential focus navigation (tab order); these can be assumed to be UI components, thus the mapping would hold, see #1454.
But this leaves out many other cases and I'm not fully sure what to do with them.

Not sure also what to do very concretely with the latest TF review… Should we just drop that TF submission until we have other rules that map to SCs?

Jym77:
After CG call on 14/01/2021, we decided to map the "role attribute has valid value" rule to 1.3.1 with an assumption that a role attribute implies some styling to make the element look like it should.
See https://github.com/act-rules/act-rules.github.io/wiki/Minutes-2021-01-14#what-to-do-of-role-attribute-has-valid-value

I'm not super happy with the examples which are poor authoring practices 🤷 

Closes issue(s):

- closes #1496 

Need for Final Call:
This will require a 1 week Final Call (updating mappings and examples)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/role-attribute-valid-value-674b10.md
+++ b/_rules/role-attribute-valid-value-674b10.md
@@ -71,7 +71,7 @@ If an incorrect `role` attribute is specified on an element which is not [visibl
 This `role` attribute contains one token, and this token is a valid [WAI-ARIA role][].
 
 ```html
-<input type="text" role="searchbox" />
+<label>Search: <input type="text" role="searchbox" placeholder="Enter 3 or more characters"/></label>
 ```
 
 #### Passed Example 2
@@ -98,7 +98,7 @@ This `role` attribute contains two tokens, and these tokens are both valid [WAI-
 This `role` attribute contains two tokens, and one of these tokens (`searchbox`) is a valid [WAI-ARIA role][].
 
 ```html
-<input type="text" role="searchfield searchbox" />
+<label>Search: <input type="text" role="searchfield searchbox" placeholder="Enter 3 or more characters"/></label>
 ```
 
 ### Failed
@@ -108,7 +108,7 @@ This `role` attribute contains two tokens, and one of these tokens (`searchbox`)
 This `role` attribute contains one token, but this token is not a valid role in any of the [WAI-ARIA specifications][].
 
 ```html
-<input type="text" role="searchfield" />
+<label>Search: <input type="text" role="searchfield" placeholder="Enter 3 or more characters"/></label>
 ```
 
 #### Failed Example 2
--- a/_rules/role-attribute-valid-value-674b10.md
+++ b/_rules/role-attribute-valid-value-674b10.md
@@ -5,24 +5,11 @@ rule_type: atomic
 description: |
   This rule checks that each `role` attribute has a valid value.
 accessibility_requirements:
-  aria11:role_definitions:
-    title: ARIA 1.1, 5.4 Definition of Roles
+  wcag20:1.3.1: # Info and Relationship (A)
     forConformance: true
     failed: not satisfied
     passed: further testing needed
-    inapplicable: satisfied
-  graphics-aria:role_definitions:
-    title: ARIA Graphics module 1.0, 4.1 Definition of Roles
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: satisfied
-  dpub-aria:role_definitions:
-    title: ARIA Digital Publishing module 1.0, 4.1 Definition of Roles
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: satisfied
+    inapplicable: further testing needed
   wcag-technique:ARIA4: # Using a WAI-ARIA role to expose the role of a user interface component
     forConformance: false
     failed: not satisfied
@@ -54,7 +41,9 @@ Each test target has at least one token which is a valid value corresponding to 
 
 ## Assumptions
 
-_There are currently no assumptions_
+- This rule assumes that the `role` attribute is used to provide an ARIA [semantic role][] to the elements. If it is used for other purposes, this rule shouldn't be used.
+- This rule assumes that elements with a `role` attribute have their intended structure and relationship conveyed through some sort of presentation. If it is not the case, it is possible to fail this rule while still satisfying [Succes Criterion 1.3.1 Info and Relationship][sc131].
+- This rule assumes that the intended role of the element is not its [implicit role][]. If no token is valid, User Agents will default to the [implicit role][] for the element; if that role is the intended one, it is possible to fail this rule but still satisfy [Succes Criterion 1.3.1 Info and Relationship][sc131].
 
 ## Accessibility Support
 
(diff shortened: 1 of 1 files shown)


kasperisager (approved):
I'm on board 👍

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -80,23 +71,34 @@ Further reading:
 This `role` attribute contains one token, and this token is a valid [WAI-ARIA role][].
 
 ```html
-<input type="text" role="textbox" />
+<input type="text" role="searchbox" />
```
ajanec01, line 74:
```suggestion
<label for="srch">Search</label>
<input type="text" role="searchbox" />
```

Shouldn't we somehow imply in the examples what the visual presentation is?

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -64,9 +53,11 @@ Older browsers do not support more than one token in the value for a role attrib
 
 The `role` attribute is a set of [space separated tokens][]. Having a [whitespace](#whitespace) separated list of more than one token in the value of the role attribute is used for what is known as _fallback roles_. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role](#semantic-role) of the element, and so forth. Having the rule target attributes containing at least one non-[ASCII whitespace][] character ensures that there is at least one token in the set.
 
-Further reading:
+If an incorrect `role` attribute is specified on an element which is not [visible][], this is still likely to cause accessibility issues. However, this is not relevant to [Success Criterion 1.3.1 Info and Relationship][sc131] since nothing is conveyed through the presentation of that element.
```
carlosapaduarte, line 56: 1.3.1 also targets "information and relationships that are implied by [...] auditory formatting". So, an element that is not visible can still be presented.

_rules/role-attribute-valid-value-674b10.md
carlosapaduarte: Agreed, but the `input` needs an `id="srch"` attribute

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -80,23 +71,34 @@ Further reading:
...
 ```
 
 #### Passed Example 3
 
-This `role` attribute contains two tokens, and one of these tokens (`img`) is a valid [WAI-ARIA role][].
+This `role` attribute contains two tokens, and one of these tokens (`searchbox`) is a valid [WAI-ARIA role][].
 
 ```html
-<img role="image img" src="/test-assets/shared/w3c-logo.png" alt="W3C logo" />
+<input type="text" role="searchfield searchbox" />
```
carlosapaduarte, line 101: Do the same as is done in passed example 1

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -106,20 +108,26 @@ This `role` attribute contains two tokens, and one of these tokens (`img`) is a
 This `role` attribute contains one token, but this token is not a valid role in any of the [WAI-ARIA specifications][].
 
 ```html
-<img role="lnik" src="/test-assets/shared/w3c-logo.png" alt="W3C logo" onclick="location.href='https://www.w3.org/'" />
+<input type="text" role="searchfield" />
```
carlosapaduarte, line 111: Do the same as is done in passed example 1

carlosapaduarte (changes requested):

_rules/role-attribute-valid-value-674b10.md
ajanec01: yes, updated.

_rules/role-attribute-valid-value-674b10.md
ajanec01: That's truth, a whistling sound for the start of the main section and the footer section would need to be replicated programmatically.

ajanec01 (approved):

carlosapaduarte (approved):

Jym77:
Final Call ends on January 28th.

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -106,20 +106,26 @@ This `role` attribute contains two tokens, and one of these tokens (`img`) is a
 This `role` attribute contains one token, but this token is not a valid role in any of the [WAI-ARIA specifications][].
 
 ```html
-<img role="lnik" src="/test-assets/shared/w3c-logo.png" alt="W3C logo" onclick="location.href='https://www.w3.org/'" />
+<label>Search: <input type="text" role="searchfield" placeholder="Enter 3 or more characters"/></label>
```
WilcoFiers, line 109: I think this is an example of where that third assumption might be true. Without the role attribute, this would be a perfectly valid form field.

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -80,23 +69,34 @@ Further reading:
...
+<style>
+	.button {
+		border: 0.1em solid black;
+	}
+	.button:hover {
+		background-color: black;
+		color: white;
+	}
+</style>
+<span class="button" tabindex="0" role="button link" onclick="location.href='https://act-rules.github.io/'"
```
WilcoFiers, line 89: I don't much like this example. I don't think button is a particularly appropriate role for this. I think we should show an appropriate use of fallback roles if we want them demonstrated. Maybe use on of the DPUB roles, with `link` as a fallback?

WilcoFiers (changes requested):

Elli360:
Looks at this resource - https://zellwk.com/blog/style-hover-focus-active-states/

_rules/role-attribute-valid-value-674b10.md
Jym77: I would argue that the correct (and definitely the intended) role is here `searchbox`, so the third assumption still holds. (and this is indicated by the "Search" label).

_rules/role-attribute-valid-value-674b10.md
Jym77: I don't know what this example was supposed to show (I've only added the styling), but I definitely like the DPUB idea 😀

WilcoFiers (approved):

Jym77:

PR #1539, opened 2021-01-22
New rule: First focusable element is link to non-repeated content

Jym77:
Extracting the separate rule from #421 .

Closes issue(s):

- N/A

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


Jym77:
Summoning @carlosapaduarte and @WilcoFiers to continue the discussion on actionable element.

pages/glossary/keyboard-actionable-element.md
```diff
@@ -0,0 +1,23 @@
...
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An HTML element is _keyboard actionable_ if all the following are true:
+
+- the element is part of [sequential focus navigation][]; and
+- the element is [visible][] when [focused][]; and
+- it is possible to [fire a click event][] at the element by use of keyboard.
```
Jym77, line 16: Copying from https://github.com/act-rules/act-rules.github.io/pull/421/files#r562630136
> Under what circumstance would this not be true? This one is a little tricky. What do you mean with "by use of the keyboard"? If I have an on-screen cursor that I control using arrow keys, and use space to fire a click event from the cursor, does that count? Arguably it would, but I suspect that's not what you mean. Additionally, are you expecting the click event to be fired when the element is focused? Could event bubbling be involved?

Jym77: I don't think that a mouse-emulator should count. It's a separate assistive technology and I don't think that it should be taken into account.

If such an emulator would count, then we could argue that every thing satisfies 2.1.1 as soon as users have such an emulator. This is quite clearly not the intend of 2.1.1…

I think rules only test what is part of the test subject (usually a web page).

Now, if a web page implements a mouse emulation with keyboard, that is another question. I would say that such a page does satisfy 2.1.1 (maybe not in a very good way, however). In [Scrollable element is keyboard accessible](https://act-rules.github.io/rules/0ssw9k), we do have an assumption (maybe a bit weak) around this idea. (I would argue that currently a page with a mouse emulation implemented could fail the rule but satisfy 2.1.1 and it's not completely clear that this is caught by the assumption).

Anyway, here, if a page implements a mouse emulation, then yes, you can fire the click event by use of keyboard only. But that is not always the case.
```
<span tabindex="0" onclick="this.style.color='red'">Foo</span>
```
Is passing the first two conditions (it is part of sequential focus navigation, and visible when focused), but not the 3rd (that may be browser dependant, but on Chrome I do not manage to fire the click event with space or enter even when the element has focus). So I do not think that this condition is always true.

carlosapaduarte: I think we can take another approach. 
To meet the expectation of the rule, the element needs to be activated (so that a change of focus can be evaluated). So, for this rule, we don't need to check if it is possible to activate (fire a click event at) an element twice. 
I suggest we drop this definition and move being part of the sequential focus navigation and being visible when focused to the expectations of the rule. The third condition in this definition is already checked in the rule's expectation.

Jym77: The important point in this third condition is not that the element can be activated, but the fact that it can be done by using the keyboard. This is what fails Failed Example 7, and is not checked elsewhere in the rule.

Without this condition,
```html
<div role="link" onclick="location.assign('#main');" tabindex="0" id="skip-link">Skip to main content</div>
```
would pass the rule, which is a bit sad for a best practice rule.

carlosapaduarte: You are correct. I was focusing too much on the "how can we know if we can fire an event at a target" and leaving out the "using the keyboard" parte.

Still, what about changing the condition in the expectation to 
```
- when the element is [activated][] by use of keyboard, focus moves [just before][] a node of [non-repeated content after repeated content][]; and
```

Jym77: 🤔 Interesting…

I'm afraid that this condition is dangerously close to triggering a universal quantification on an empty set… The almost similar
> If the element is activated by use of keyboard, focus moves…

for example, would pass on elements that are not keyboard-actionable (or we can make this even more explicit with "every time the element is activated by use of keyboard, …")

Now, clearly, the current formulation also has that problem… It can be rephrased the same way as
> Every time the element is activated, …

and would pass on a non-activatable element (whatever that means…) But the current version does have a condition that the element is actionable (as part of the keyboard actionable requirement…)

---
Also, to continue discussion… If I open the devtools (Ctrl+Shift+I, usually), go to the console, and type in some `document.getElement(foo).sendEvent(click)`, is that keyboard activation? I guess we're safe-ish because that uses something (devtools) which is not part of the content (the web page), so if that gives extra power, we don't care (the same way that we wouldn't take into account a browser extension that allows to move the pointer with the k.eyboard).

_rules/first-focusable-element-is-link-to-non-repeated-content-8a213c.md
```diff
@@ -0,0 +1,658 @@
+---
+id: 8a213c
+name: First focusable element is link to non-repeated content
+rule_type: atomic
+description: |
+  This rule checks that the first focusable element is a link to non-repeated content in the page
+accessibility_requirements:
+  wcag-technique:G1: # Adding a link at the top of each page that goes directly to the main content area
```
ajanec01, line 8: This rule does not map to WCAG directly. We had this issue in #1496. Have no issues with reviewing but the TF may not publish it. Maybe it's worth tweaking it so that it could be part of a composite rule?

_rules/first-focusable-element-is-link-to-non-repeated-content-8a213c.md
```diff
@@ -0,0 +1,658 @@
...
+
+Each test target has a non-empty [sequential focus navigation order][sequential focus navigation].
+
+## Expectation 2
+
+For each test target, all the following are true for the first element in its [sequential focus navigation order][sequential focus navigation]:
+
+- the element is [keyboard actionable][]; and
+- the element is [included in the accessibility tree][]; and
+- the element is a [semantic link][]; and
```
WilcoFiers, line 40: Referencing back to our conversation of mechanism, I don't think these two are required. Actually if a skip mechanism was a button, I'm not entirely sure it'd even fail WCAG.

_rules/first-focusable-element-is-link-to-non-repeated-content-8a213c.md
```diff
@@ -0,0 +1,658 @@
...
+    - _The Three Kingdoms_ by Luo Guanzhong, translation by Yu Sumei (Tuttle publishing, 2014, ISBN 9780804843935)
+---
+
+## Applicability
+
+This rule applies to any [HTML web page][].
+
+## Expectation 1
+
+Each test target has a non-empty [sequential focus navigation order][sequential focus navigation].
```
WilcoFiers, line 32: Shouldn't this be part of the applicability? I don't quite see the logic of failing a page without focusable content.

_rules/first-focusable-element-is-link-to-non-repeated-content-8a213c.md
```diff
@@ -0,0 +1,658 @@
...
+
+## Expectation 2
+
+For each test target, all the following are true for the first element in its [sequential focus navigation order][sequential focus navigation]:
+
+- the element is [keyboard actionable][]; and
+- the element is [included in the accessibility tree][]; and
+- the element is a [semantic link][]; and
+- when the element is [activated][], focus moves [just before][] a node of [non-repeated content after repeated content][]; and
```
WilcoFiers, line 41: We should make sure there is repeated content in the page.

_rules/first-focusable-element-is-link-to-non-repeated-content-8a213c.md
```diff
@@ -0,0 +1,658 @@
...
+- the element has an [accessible name][] that communicates that it skips to the main content area.
```
WilcoFiers, line 42: This too seems a bit out of scope for the technique.

WilcoFiers (dismissed):

Jym77:
@WilcoFiers good catch on the presence of repeated content. That was lost in various iterations…
This lead me to make a composite rule with one atomic checking for repeated content and another for skip link.

colabottles (approved):

pages/glossary/keyboard-actionable-element.md
shunguoy: I am not sure if this can be programmatically checked.  I would like this to be a manual rule.

PR #1544, opened 2021-02-02
fix URL in meta-refresh rules

### LINKED ISSUE #1543: Change to w3c.org in tests leading to intermittent test failures
The change from example.com to w3c.org in https://github.com/act-rules/act-rules.github.io/commit/8e20cff8c510a64cb42678da55a8658fb8181a77 is leading to intermittent test failures

https://w3c.org redirects to https://w3.org but requests to https://w3c.org sometimes fail when running the test suite, and in Chrome. The screenshot shows what happens in Chrome after running the test suite. 

There may be rate limiting on w3c.org triggered by too many requests in a short time (which will happen when running a test test suite)

Screenshot 2021-01-30 at 08 05 29

Edit: after requests to https://w3c.org fail I can still access https://w3.org
Edit2: if we're all running the test suite and making requests to w3c.org the traffic pattern might look very similar to a DDOS

WilcoFiers:
w3c.org does not exists, and instead redirects to w3.org. Think we all know this was a mistake, not sure how this got overlooked in reviews though.

Closes #1543

Need for Final Call: **none**, fix is urgent and inconsequential to the rule's intent.

---

## How to Review And Approve

### DIFF
--- a/_rules/meta-refresh-no-delay-bc659a.md
+++ b/_rules/meta-refresh-no-delay-bc659a.md
@@ -95,8 +95,8 @@ First valid `<meta http-equiv="refresh">` redirects immediately.
 
 ```html
 <head>
-	<meta http-equiv="refresh" content="0; https://w3c.org" />
-	<meta http-equiv="refresh" content="5; https://w3c.org" />
+	<meta http-equiv="refresh" content="0; https://w3.org" />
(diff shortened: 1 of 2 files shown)


Jym77 (approved):
Ouch 🙈

ajanec01 (approved):

carlosapaduarte (approved):

Issue #1545, opened 2021-02-08
Should all table header cells have assigned data cells?


ajanec01:
We have a rule named [All table header cells have assigned data cells](https://act-rules.github.io/rules/d0f69e)

There are two points in the [Task Force feedback](https://www.w3.org/2002/09/wbs/93339/ACTTableHeaders/results) that I'm not sure we should act upon. The first one being:

> I don't think it is an accessibility problem if a header cell exists without an assigned data cell. I think it would be better if a rule checked that a table data cell has an assigned header cell that exists instead.

I think that the assumption section covers this quite well.

> This rule assumes that table header cells have a relationship conveyed through presentation with other cells within the same table. This excludes edge cases such as a table definition where there is only one header cell, or a table definition where there are multiple headers and no other cells. In such scenarios the rule fails, but success criterion 1.3.1 Info and Relationships could still be satisfied.

If we were to change the rule so that each data cell has assigned at least one header cell, the following pattern from the passed example 4 would be a failure even though it is a valid code
```
<table role="grid">
	<thead>
		<tr role="row">
			<td></td>
			<th scope="col" role="columnheader">Breakfast</th>
			<th scope="col" role="columnheader">Lunch</th>
			<th scope="col" role="columnheader">Dinner</th>
		</tr>
	</thead>
	<tbody>
		<tr role="row">
			<th scope="row" role="rowheader">Day 1</th>
			<td>8:00</td>
			<td>13:00</td>
			<td>18:00</td>
		</tr>
	</tbody>
</table>
```

Either case would need an assumption. Is it worth changing the rule altogether, or should we stay with it as is?

The second point in the feedback was:

> ... not sure what assigned means, there is no such requirement as i see.

I think that the suggestion is to change the "assigned" word to associated. Assigned is currently linking to [Forming relationships between data cells and header cells section](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics) of the HTML spec, which I think is valid.

@Jym77 has also raised an issue for this rule that would need to be addressed- #1517 

It would be good to get input from the community before we change the rule.

Jym77:
[...]

Issue #1517, opened 2020-12-11
"Table header cell has assigned data cells" (d0f69e): test tables and grids separately?


Jym77:
Currently, the rule targets a header which is descendant of "either `table` or `grid` and checks whether it has assigned `cell` or `gridcell`.

Potential problem is that it is possible to pass the rule with a `table` having only `gridcell` (or with a `grid` having only `cell`) which seems incorrect (not fully sure it is a SC failure, however).

Should we try to de-entangle that so that? (easiest way is to split rule, could be doable also by splitting Expectation)
Should we instead have a rule "`table` has no `gridcell` / `grid` has no `cell`"?

WilcoFiers:
[10:57] <Wilco> RESOLUTION: update rule title to "Header cell has assigned cells"

[10:58] <Wilco> RESOLUTION: Update the rule to only allow "cell" roles in "table" and "gridcell" roles in "grid"

ajanec01:
Following @Jym77 's [comment](https://github.com/act-rules/act-rules.github.io/pull/1548#issuecomment-1164497401) from #1548 I think we can close this issue as we are restricting the rule to HTML only. 

Some reasoning:
* We are unable to determine how table headers are assigned to grid cells in ARIA only grids as this is not covered by any related technical documentation,
* all cells within a `<table role="grid">` element should have the role of `gridcell` as per the definition of [`grid`](https://www.w3.org/TR/wai-aria-1.2/#grid), see below:
>if an author applies the grid role to an HTML table element, the author does not need to apply the [row](https://www.w3.org/TR/wai-aria-1.2/#row) and [gridcell](https://www.w3.org/TR/wai-aria-1.2/#gridcell) roles to the descendant HTML tr and td elements because the [user agent](https://www.w3.org/TR/wai-aria-1.2/#dfn-user-agent) will automatically make the appropriate translations.

Although it would be nice to have a rule that checks if all cells within a grid are grid cells it is likely that this happens most of the time because in practice, the authors tend to use the `<table role="grid">` element instead of going with `<div>` and `<span>` elements. The point is that I think we are already covering a high percentage of content out there. 

I'd like to know what others think though.

HelenBurge:
Closed as the PR is closed

PR #1548, opened 2021-02-17
Table header cell has assigned cells- changes in response to the Task Force feedback

ajanec01:
A couple of changes to [Table header cell has assigned cells](https://act-rules.github.io/rules/d0f69e). All in response to [the Task Force feedback](https://www.w3.org/2002/09/wbs/93339/ACTTableHeaders/results).

#### Key points

- Changed the assumptions section and Failed example 1 in response to 
> I'm not convinced that tables without td should be treated as failures of this rule.
> Failed Example 1 visually conveys a table header cell with no data cells in its column. I don't think this meets the assumption of this rule: This rule assumes that table header cells have a relationship conveyed through presentation with other cells within the same table. 
- Changed Passed example 5 in response to
> Per the guidline these are the 2 techniques to associate headers with data cells, some of the passed examples don't do that.
H63: Using the scope attribute to associate header cells and data cells in data tables
H43: Using id and headers attributes to associate data cells with header cells in data tables
- Changed Failed Example 3 as per
> The description of failed example 3 uses "table element" but it might be more appropriate to use "semantic role of grid" instead.
- Edited the applicability section in response to
> 1. The applicability does not establish that the header is part of a table. Just because there is a table ancestor, doesn't mean it is part of that table. There could be any number of ancestors between the header cell and the table.

> 2. The part about the table being included in the accessibility tree seems unnecessary. This may be my lack of imagination, but I can't come up with a way to take the table out of the accessibility tree, without either changing the role, or also taking the header cell out of the tree. That requirement seems unnecessary.

- Edited expectation as per #1517 

Closes issue #1545 and #1517 

Need for Final Call:
2 weeks


_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: d0f69e
-name: Table header cell has assigned cells
+name: Table header cell has assigned data cells
 rule_type: atomic
 description: |
-  This rule checks that each table header has assigned cells in a table element.
+  This rule checks that each table header has assigned data cells in a table element.
```
Jym77, line 6: "table element" is not really correct here.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,20 +27,19 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any HTML element with the [semantic role][] of [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any HTML element with the [semantic role][] of [rowheader][] or [columnheader][] within a [table][] or [grid][] element, where the [table][] or [grid][] element is [visible][].
```
Jym77, line 30: `[table]` and `[grid]` refer to ARIA roles (in the references list). I feel it's not super clear that "a table element" actually means "an element with a role of table".

Also, we normally backticks roles and element (``a [`table`][table] element``).

Jym77, line 30: I am not sure that "within" is unambiguous enough.
I understand that this is trying to address Wilco's first comment of https://www.w3.org/2002/09/wbs/93339/ACTTableHeaders/results#xQ8 but I think that "within" can easily be understood as "descendant of" (this is the meaning we use for ["inside a web page"](https://act-rules.github.io/glossary/#web-page-html), for example).

I agree with the initial comment, but I'm afraid we need to be more precise in explaining what it means for a cell to be part of a table. (and this is somewhat more complex due to the use of roles, so we cannot fully rely on HTML definitions from https://html.spec.whatwg.org/multipage/tables.html#table-processing-model such as "the element corresponds to a cell [=concept, not `td`/`th`] in the table [=concept, not the `table` element]" 😖

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,20 +27,19 @@ acknowledgments:
...
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][] is [included in the accessibility tree][].
+## Expectation 1
 
-## Expectation
+Each target element is [assigned][] to at least one element with a [semantic role][] of [cell][]. The test target and the [cell][] are within an element with the [semantic role][] of [table][] or [grid][].
```
Jym77, line 34: The second sentence looks like a different expectation.
Also, since the target is part of the table, maybe
```suggestion
Each target element is [assigned][] to at least one element with a [semantic role][] of [cell][]. which is within the same element with the [semantic role][] of [table][] or [grid][].
```
(+ same caveat about "within")

Jym77: Ah no, I get it… (reading the next expectation…)
So more like
```suggestion
For each test target, either it is not part of a [table][]; or it is [assigned][] to at least one element with a [semantic role][] of [cell][] which is within the sameelement with the [semantic role][] of [table][].
```

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,20 +27,19 @@ acknowledgments:
...
 
-Each target element is [assigned][] to at least one element with a [semantic role][] of [cell][] or inheriting from [cell][].
+## Expectation 2
+
+Each target element is [assigned][] to at least one element with a [semantic role][] of [gridcell][]. The test target and the [gridcell][] are within an element with the [semantic role][] of [grid][].
```
Jym77, line 38: Same concerns as for 1st expectation.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,20 +27,19 @@ acknowledgments:
...
 
 ## Assumptions
 
-This rule assumes that table header cells have a relationship conveyed through presentation with other cells within the same table. This excludes edge cases such as a table definition where there is only one header cell, or a table definition where there are multiple headers and no other cells. In such scenarios the rule fails, but [success criterion 1.3.1 Info and Relationships][sc1.3.1] could still be satisfied.
+This rule assumes that table header cells have a relationship conveyed through presentation with data cells within the same table. This excludes edge cases, where there are multiple headers and no data cells. In such scenarios the rule fails, but [success criterion 1.3.1 Info and Relationships][sc1.3.1] could still be satisfied.
```
Jym77, line 42: I agree with https://github.com/act-rules/act-rules.github.io/issues/1545#issuecomment-777604045 that we could move part of this Assumption to Applicability by requiring that the table/grid contains at least one non-empty cell/gridcell.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -240,10 +247,10 @@ This `div` with role of `columnheader` and text equal to "Occupant" does not hav
 		<div role="columnheader">Occupant</div>
 	</div>
 	<div role="row">
-		<div role="gridcell">1A</div>
+		<div role="gridcell"><button>1A</button></div>
```
Jym77, line 250: I'm not sure why we need to add `button` here (and in the next row).

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -327,27 +334,12 @@ This `th` element is not a descendant in the [flat tree][] of an element with a
 </div>
 ```
 
-#### Inapplicable Example 7
```
Jym77, line 330: I think we can keep the example but update the description to point out that the `table` element does not have a role of `table`, so the rule does not apply.

Jym77 (changes requested):
🙈 This seems to explode a bit. Sorry.

Also, can you add H63 to the background?

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: Not sure why "either it is not part of a [table][]"? Elements with the semantic role of `rowheader` or `columnheader` have required context roles that eventually end on an element with the semantic role of either `grid` or `table`.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: Something in the lines of "within an element with the semantic role of `table` or `grid` that is its closest ancestor in the accessibility tree"- the ancestor in the accessibility tree does not seem correct because it's used in the DOM standard so have a feeling it should not be used for accessibility tree, but please correct me on that.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: What if "table header has assigned at least one element with the semantic role of `cell` that has non-empty accessible name"? 
There might be a non-empty data cell but if it has `aria-hidden` than that would also be a failure under 1.3.1.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: This was not meant to be a failed example.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: If we apply this rule to either a `rowheader` or a `columnheader` and either of those has assigned at least one non-empty element with a semantic role of `cell` or inheriting from cell than that should work. What do you think?

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: d0f69e
-name: Table header cell has assigned cells
+name: Table header cell has assigned cell
 rule_type: atomic
 description: |
-  This rule checks that each table header has assigned cells in a table element.
+  This rule checks that each table header has assigned at least one non-empty cell.
```
Jym77, line 6:
```suggestion
  This rule checks that each table header is assigned to at least one non-empty cell.
```

_rules/table-header-cell-has-assigned-cells-d0f69e.md
Jym77: The problem with "closest ancestor" is that it is the current version and not really working…

Why I understand is that in
```
<table>
<div><div><tr><th>Foo</th></tr></tr>
</table>
```
the `th` is not really "part" of the `table` because the `div` are skipped. This actually gets more complex due to the fact that HTML parsing actually removes the `div` when building the DOM in that case… But we can reach a similar result with roles:
```html
<div role="table">
<div><div>
  <div role="tablerow"><div role="columnheader">Foo</div></div>
</div></div>
</div>
```
But then in that case it's not even clear what "being part of the table" means since the algorithm for forming a table only works on HTML tables, not on ARIA ones…

Using "within" is, I think, not good because of this:
```html
<table>
<tr><td>
  <table hidden>
  <tr><th>Foo</th><tr>
  </table>
</td></tr>
</table>
```
In which case, the `th` is arguably "within" both `table` elements (as in "within the subtree rooted at that element"), but we want to exclude it from the rule, so we need a definition of "within a table"… And copying the table building algorithm doesn't seem smart…

_rules/table-header-cell-has-assigned-cells-d0f69e.md
Jym77: You need the "either" bit because of the table/grid duality:
```html
<table role="grid">
<tr><th role=columnheader>Foo</th></tr>
<tr><td role="gridcell">Bar</td></tr>
</table>
```
Now, this `th` is not assigned to any element with a role of `cell` and thus fails the expectation. Yet, it is good because it is inside a `grid`.

When there are several expectations, a target needs to pass **all** of them in order to pass the rule.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
Jym77: What we need to avoid here is
```html
<table>
<tr><th>Foo</th></tr>
</table>
```
Where the `th` is not assigned to any `cell`, but this is OK because there is no relationship conveyed through presentation, so 1.3.1 doesn't care. But this does fail the rule because the `th` is assigned to no cell.
Rather than having it in an assumption (actually, in addition to the assumption which covers other corner cases), we could have that bit in the Applicability:
"the `table` contains at least one data cell".
This would remove from the start a large chunk of the corner cases covered by this assumption.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -146,19 +151,22 @@ Each of the 4 `th` elements has an assigned `td` element, within the same `table
...
 	<tr>
-		<td>Paris</td>
+		<th id="1" headers="objective">1</th>
+		<th id="2" headers="objective">2</th>
 	</tr>
 	<tr>
-		<td headers="col2">1</td>
+		<td headers="projects">40%</td>
+		<td headers="objective 1">20%</td>
+		<td headers="objective 2">25%</td>
```
Jym77, line 169: Is it on purpose that the percentages add up to only 95% 🤔

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -190,23 +198,49 @@ Each of the 5 `th` elements in this table has assigned cells, whether data or he
...
+	</div>
+</div>
+```
+
 ### Failed
 
 #### Failed Example 1
 
-The `th` element with text "Value" does not have an assigned cell within the same `table` element.
+The `th` elements do not have assigned data cells as per the [internal algorithm for scanning and assigning header cells](https://html.spec.whatwg.org/multipage/tables.html#internal-algorithm-for-scanning-and-assigning-header-cells). Their `scope` attribute is in the auto state and there is a non-empty table data slot in the same line.
```
Jym77, line 226:
```suggestion
The `th` elements do not have assigned data cells as per the [internal algorithm for scanning and assigning header cells](https://html.spec.whatwg.org/multipage/tables.html#internal-algorithm-for-scanning-and-assigning-header-cells). Their `scope` [attribute value][] is `auto` state and there is a non-empty table data slot in the same column or row.
```

Jym77 (changes requested):
See also comments in the other discussions.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,20 +32,19 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any HTML element with the [semantic role][] of [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any HTML element with the [semantic role][] of [`rowheader`][] or [`columnheader`][] that is within an element with the [semantic role][] of either [`table`][] or [`grid`][]. The [`table`][] or [`grid`][] is [visible][].
```
Jym77, line 35: The square brackets should be links.

Jym77 (dismissed):

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: I think that I've addressed that with the current format of the expectations section. The second expectation should account for the `table`/`grid` duality.

Jym77:
In #1552 Kasper is giving an example of mixed HTML/ARIA table where HTML parsing is throwing away the ARIA bit due to incorrect HTML 😕

Jym77 (dismissed):

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -23,36 +23,42 @@ acknowledgments:
...
 ## Applicability
 
-This rule applies to any HTML element with the [semantic role][] of [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any [HTML element][] with the [semantic role][] of [rowheader][] or [columnheader][] that is [included in the accessibility tree][] and for which all of the following are true:
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][] is [included in the accessibility tree][].
+- **in a row** the element is a child of an [accessible object][] with the [semantic role][] of [row][]; and
```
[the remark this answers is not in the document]
ajanec01: Agree with the definition of "owned by". As for the accessible object, I don't like the Conflict point for the definition of semantic role. As far as I understand it, `<row role="presentation">` would have a semantic role of row but that would breake the rule. An accessible object with the semantic role of row does not do it for me.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -23,36 +23,42 @@ acknowledgments:
...
+- **table** the element is an [owned element][] of an [accessible object][] with the [semantic role][] of either [table][] or [grid][] that is [visible][]; and
```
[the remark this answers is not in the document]
ajanec01: The definition of "owned by" does not fit very well in here. A `row` can be owned by a `table` or it can be owned by `rowgroup` that is owned by a `table`.  I wanted to catch both.

pages/glossary/assigned-cell.md
```diff
@@ -0,0 +1,31 @@
...
+
+In order for an [accessible object][] with the [semantic role][] of [rowheader][] to be assigned to an [accessible object][] with the [semantic role][] of either [cell][], [gridcell][], or inheriting from [cell][], at least one of the following is true:
+
+- there is at least one non-empty [accessible object][] with a [semantic role][] of either [cell][], or inheriting from [cell][], that has the `headers` attribute with the value that matches the `id` [attribute value][] of the [rowheader][] that is being evaluated, and it is an [owned element][] of the same [table][] or [grid][] as the [rowheader][]; or
+- there is at least one non-empty [accessible object][] with a [semantic role][] of either [cell][], or inheriting from [cell][], that is a child an [accessible object][] with the [semantic role][] of [row][] that is also the parent of the evaluated [rowheader][];
```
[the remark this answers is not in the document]
ajanec01: You're right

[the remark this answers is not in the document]
ajanec01: I've tested your example using NVDA and Firefox and NVDA with Chrome. Two different outcomes. NVDA/ FF- the data cell is announced as "Foo; row 2; 1". NVDA/ Chrome announces the same data cell as "row 2; Bar; column 1; 1" which is weired given that the "Bar cell" is announced as "row 1; Bar; column 2; Bar". Initially I wanted to go with `aria-colindex` and `aria-rowindex` to define assignment but those are not exposed consistently. Chrome does expose those two but Firefox doesn't. That may be what is causing the above issue. 

So, this example would be a failure, but the faiing test target would be different depending on the browser/ screen reader being used. I'm not sure what is the best way forward. I mean, those will be evaluated by the pass or fail of the example and not which test target is failing, right?

pages/glossary/assigned-cell.md
```diff
@@ -0,0 +1,31 @@
...
+
+- there is at least one non-empty [accessible object][] with a [semantic role][] of either [cell][], or inheriting from [cell][], that has the `headers` attribute with the value that matches the `id` [attribute value][] of the [columnheader][] that is being evaluated, and it is an [owned element][] of the same [table][] or [grid][] but it is not a child of the same [accessible object][] with the [semantic role][] of [row][] as the [columnheader][]; or
+- there is at least one non-empty [accessible object][] with a [semantic role][] of either [cell][], or inheriting from [cell][], that is a child of an [accessible object][] with the [semantic role][] of [row][], that is not the parent element of the [columnheader][] for which the assignment is evaluated; the [cell][]'s [row][] is an [owned element][] of the same [table][] or [grid][] to which the [columnheader][] belongs and has the same [child index][] as the [columnheader][], or it has a `colspan` attribute which together with the [cell][]'s [child index][] value spans over the [child index][] value of the [columnheader][];
```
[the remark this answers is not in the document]
ajanec01: As for the first example with a table cell that spans over two rows and columns, using NVDA on both Chrome and FF associates the `<td>Bar</td>` to `<th>H3</th>`.  Current definition and the accessibility tree that I was looking at it did not foresee such outcome.

In the second example with `aria-hidden`, the `<td>D3</td>` is assigned to H3 in FF but not in Chrome (in Chrome it's H2). I'm not sure at the moment what might be causing it.

As for the third example, that is currently a failure as the cells are not owned by a `row`.  Interestingly, FF and NVDA do not associate those cells at all whereas Chrome does.

In summary, there are examples that are clearly a failure but the test target that is failing is different depending on the browser. I'm not sure if that is a deal breaker or not since implemenations should fail such example anyway. I also need to take into consideration how `rowspan` and `colspan` affects association when the number of cells in a row is not consistent. I'd need to test that with Jaws and VO as well to get a better understanding.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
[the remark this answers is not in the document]
ajanec01: What about inclusive descendants?

colabottles (approved):

Jym77 (changes requested):
I'm still thinking that we try to do too much in one rule (or at least in one attempt) by
* checking both `table` and `grid`. This leads to heavy dealing of alternatives.
* checking both HTML and ARIA tables. This makes us unable to rely on the (already complex) table model in HTML and needing to manually handle all the shenanigans. The "assigned cell" definition is already quite a mouthful and feels far from being correct.

I still think it would be easier, at least as a first step, to restrict the rule to HTML tables (whose roles are not changed, to get the mapping), so we can focus on handling the other difficulties (such as the stuff around the meaning of "within" and all that). Maybe later, we can extend to other kinds of table.

I'm also wondering how often ARIA tables are used. I feel it's a lot of efforts for devs to use `<div role="row">` and `<span role="cell">` instead of `<tr>` and `<td>` (and since these have been here since the start of HTML it's not even a matter of backward compatibility and semantic HTML). I don't think I've ever encountered one.
If we are trying to build a very complex definition (assigned cell) just to handle a few corner cases of people building ARIA tables, I really think we should skip it for now, focus on the simpler and more frequent case, and revisit this once we are confident in the simple case.

ajanec01:
Hey @Jym77, I also said during the last TF call that I'm unhappy with this rule as is and would prefer to scope it to HTML tables. I've been trying to find any type of information of how the "assignment" process looks like but since there is no spec for it, that has not been possible. 

I'm not sure if this should be put forward to the CG in the next meeting before any commit or not?

HelenBurge (approved):

Jym77:
CG call: Restrict the rule to HTML tables

ajanec01:
This is a very old PR and some of the issues are no longer applicable. It will also be difficult to follow all comments and I rebased the branch so a lot of other files got added to the review. Closing this PR and will open a new one.

Issue #1541, opened 2021-01-28
With autocomplete, should type=tel be allowed on all numbers? [73f2c2]


WilcoFiers:
Common feedback we got on the autocomplete rule was people that wanted to use `type=tel` on number fields such as `bday-month`. The reason is that with type=tel IOS opens the number keypad, which it does not do if you use type=number. 

Axe-core has since built an exception in its autocomplete rule that allows type=tel. I wonder if this is something the autocomplete rule should allow. Or alternatively... if we should have a test case that shows axe-core's implementation is a partially consistent one, because of this exception.

ajanec01:
I think that we had a similar discussion. It is not a failure of sc 1.3.5. An incorrect control group may prevent auto-filling (doesn't mean it will), but the autocomplete property is still programmatically identifiable. The easiest fix is to remove Expectation 2.

ajanec01:
The `type` can be incorrect, see #1583

PR #1549, opened 2021-02-17
Autocomplete attribute has valid value- ACT-R meeting suggestions

### LINKED ISSUE #1540: autocomplete="off" fails "autocomplete attribute has valid value" (73f2c2)
The rule "autocomplete attribute has valid value" (73f2c2) has the assumption "The autocomplete attribute is used on form fields that correspond to Input Purposes for User Interface Components and collect information about the user."

This means that the use of autocomplete="off" is actually not applicable to the rule due to the expectation. 

However, I have twice this week had to debug automated testing results (from two different checker engines) that failed autocomplete="off". 
I think this is an indication that the rule incorrectly flags a common practice as an accessibility issue.

ajanec01:
Applying suggestions from the latest meeting.

- Changing applicability to make sure that `autocomplete="off"` is not a failure
- Provide more information on what is an appropriate autocomplete field.

Closes issue(s):

- closes #1540 and #1541


_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -26,7 +26,8 @@ This rule applies to any HTML `input`, `select` and `textarea` element with an `
 - **hidden**: the element is not [visible][], and not [included in the accessibility tree][]; or
 - **disabled**: the element is a [disabled element]; or
 - **fixed value**: the element is an `input` element with a `type` [attribute value][] of either `hidden`, `button`, `submit` or `reset`; or
-- **static**: the element is not part of [sequential focus navigation][] and has a [semantic role][] that is not a [widget role][].
+- **static**: the element is not part of [sequential focus navigation][] and has a [semantic role][] that is not a [widget role][]
+- the `autocomplete` attribute wears the [autofill expectation mantle](https://html.spec.whatwg.org/#autofill-expectation-mantle).
```
Jym77, line 30:
```suggestion
- **static**: the element is not part of [sequential focus navigation][] and has a [semantic role][] that is not a [widget role][]; or
- **autofill expectation** the `autocomplete` attribute wears the [autofill expectation mantle](https://html.spec.whatwg.org/#autofill-expectation-mantle).
```

Jym77, line 30: I'm wondering about that. As per the specs that are linked:
> On an input element whose type attribute is in the Hidden state, the autocomplete attribute wears the autofill anchor mantle. In all other cases, it wears the autofill expectation mantle.

So this just means that the `input` element is not in the Hidden state (i.e. that the type attribute value is not `hidden` in our words). Then this is already handled by the **fixed value** condition above.

Jym77, line 30: Also, this is a list of exception. Technically, this is currently saying that the rule does **not** apply when the autofill expectation mantle is on and thus **only** apply when the autofill anchor mantle is on. Combined with the **fixed value** condition that excludes the only case where the autofill anchor mantle is on, this means nothing is applicable (a negation is probably needed somewhere…)

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -39,15 +40,20 @@ Each test target's `autocomplete` [attribute value][] is a [space separated][] l
 
 ## Expectation 2
 
-Each test target's `autocomplete` [attribute value][] has a [correct autocomplete field][] that is [appropriate][appropriate field for the form control] for that test target.
+Each test target's `autocomplete` [attribute value][] has a [correct autocomplete field][] that is [appropriate][appropriate field for the form control] for that its control group. It means that, for example, an `input` element with `autocomplete="new-password"` must have the `type` attribute with one of the following values:
+
+- `hidden` (in this case the test target is inapplicable),
+- `text`,
+- `search`,
+- `password`.
```
Jym77, line 48: I think that the addition is more of a note than really part of the expectation. Expectations are not the best place to add example…
I think that could be moved to the background, maybe with a slightly lengthier explanation (not just an example).

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -39,15 +40,20 @@ Each test target's `autocomplete` [attribute value][] is a [space separated][] l
...
 
 ## Assumptions
 
 The `autocomplete` attribute is used on form fields that correspond to [Input Purposes for User Interface Components](https://www.w3.org/TR/WCAG21/#input-purposes) and collect information about the user.
 
-If the `autocomplete` attribute is used to describe "custom" taxonomy, for example `<input type="text" autocomplete="banner" />`, success Criterion [1.3.5 Identify Input Purpose][sc135] may be satisfied even if this rule failed.
+Controls with the `autocomplete` attribute that wears the [autofill expectation mantle](https://html.spec.whatwg.org/#autofill-expectation-mantle) (e.g. `<input autocomplete="off" />`) are assumed not to collect information about the user or do not correspond to [Input Purposes for User Interface Components](https://www.w3.org/TR/WCAG21/#input-purposes); therefore, they are inapplicable.
```
Jym77, line 54: That seems to directly contradict the specs:
> When wearing the autofill expectation mantle, the autocomplete attribute describes what input is expected from users.

_rules/autocomplete-valid-value-73f2c2.md
Jym77: Reading specs and the rest of the changes, I think we should rather explicitly exclude `autocomplete` which is either `on` or `off` (and nothing more).

Jym77 (changes requested):

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: Ohh, my 🤦 
The idea was that this rule should not apply to `autocomplete` attribute that wears the autofill expectation mantle that is a single token of either `off` or `on` and not an autofill detail token. Will get that sorted now. Thanks!

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: The suggestion was to explain what appropriate means without using the glossary. I'll have a think.

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: I made this assumption based on:

> The "off" keyword indicates either that the control's input data is particularly sensitive (for example the activation code for a nuclear weapon); or that it is a value that will never be reused (for example a one-time-key for a bank login) and the user will therefore have to explicitly enter the data each time, instead of being able to rely on the UA to prefill the value for them; or that the document provides its own autocomplete mechanism and does not want the user agent to provide autocompletion values.

To be honest, after reading it again, I'm not sure why I did it. Will be removed. Thanks!

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -26,7 +26,8 @@ This rule applies to any HTML `input`, `select` and `textarea` element with an `
 - **hidden**: the element is not [visible][], and not [included in the accessibility tree][]; or
 - **disabled**: the element is a [disabled element]; or
 - **fixed value**: the element is an `input` element with a `type` [attribute value][] of either `hidden`, `button`, `submit` or `reset`; or
-- **static**: the element is not part of [sequential focus navigation][] and has a [semantic role][] that is not a [widget role][].
+- **static**: the element is not part of [sequential focus navigation][] and has a [semantic role][] that is not a [widget role][]; or
+- **autofill expectation** the `autocomplete` attribute wears the [autofill expectation mantle](https://html.spec.whatwg.org/#autofill-expectation-mantle) that does not consist of a single token that is [ASCII case-insensitive][] match for string `off` or string `on`.
```
Jym77, line 30: I think we can get rid of the "mantle" bit (which is a bit tricky…)
The "anchor" mantle is only worn on `input` elements with a `type` attribute in the `Hidden` state, i.e a `type` attribute value of `hidden`. Hence, they are already excluded by the **fixed value** condition.
```suggestion
- **autofill expectation**: the `autocomplete` attribute consists of a single token that is [ASCII case-insensitive][] match for string `off` or string `on`.
```

Given that the Applicability starts with "with an autocomplete attribute value that is neither empty ("") nor only ASCII whitespace", we can probably refactor the full list as:
```
This rule applies to any HTML input, select and textarea element with an autocomplete attribute, except if one of the following is true:

- **autofill expectation**: the autocomplete attribute it either empty (""), only ASCI whitespace, or consists of a single token that is ASCII case-insensitive match for the string "off" or "on"; or
- **hidden*: the element is not visible, and not included in the accessibility tree; or
- **disabled**: the element is a disabled element; or
- **fixed value**: the element is an input element with a type attribute value of either hidden, button, submit or reset; or
- **static**: the element is not part of sequential focus navigation and has a semantic role that is not a widget role.
```

Jym77 (changes requested):

ajanec01:
Closed until #1562 is addressed.

PR #1554, opened 2021-03-11
"ARIA required context role" (ff89c9): resolve TF feedback

Jym77:
As per https://github.com/w3c/wcag-act/issues/507
This mostly inlines our definition of "owned by" into "child in the accessibility tree" to avoid the confusion with the ARIA definition. Notes are added to the background to explain a bit the reasoning behind that choice.

Note that the definition is contested inside ARIA: https://github.com/w3c/aria/issues/748 which is a good reason to not use it 😀 

Closes issue(s):

- N/A

Need for Final Call:
This will require a 1 week Final Call (technically only editorial, but changes structure quite a bit)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -86,7 +86,7 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 3
 
-These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not direct DOM children of it. The presentational nodes are not [included in the accessibility tree][], hence the two innermost `div` are children of the outermost in the [accessibility tree][] (but not in the DOM tree).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct children in DOM. The presentational node is not [included in the accessibility tree][].
 
 ```html
 <div role="list">
@@ -99,7 +99,7 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 4
 
-These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its DOM descendants. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
 
 ```html
 <div role="list" aria-owns="item1 item2"></div>
@@ -109,7 +109,7 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 5
 
-These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not direct DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct DOM descendants. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
 
 ```html
 <div role="list" aria-owns="item1 item2">
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -86,7 +86,7 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 3
 
-These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct children in DOM. The presentational node is not [included in the accessibility tree][].
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its children in DOM. The presentational node is not [included in the accessibility tree][].
 
 ```html
 <div role="list">
@@ -109,7 +109,7 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 5
 
-These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct DOM descendants. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its DOM children. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
 
 ```html
 <div role="list" aria-owns="item1 item2">
@@ -146,7 +146,7 @@ This element with an [explicit role][] of `listitem` is not a child in the [acce
 
 #### Failed Example 2
 
-These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][], but of the `tabpanel`.
+These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][], but of an element with the `tabpanel` role.
 
 ```html
 <div role="list">
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -29,7 +29,7 @@ This rule applies to any HTML or SVG element that is [included in the accessibil
 
 ## Expectation
 
-Each test target is [owned by][] an element that has a [semantic role][] that is one of the [required context roles][] of the target element.
+Each test target is the child in the [accessibility tree][] of an element that has a [semantic role][] that is one of the [required context roles][] of the target element.
 
 ## Assumptions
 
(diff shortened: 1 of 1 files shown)


_rules/aria-required-context-role-ff89c9.md
```diff
@@ -84,7 +86,7 @@ These elements with an [explicit role][] of `listitem` are [owned by][] an eleme
 
 #### Passed Example 3
 
-These elements with an [explicit role][] of `listitem` are [owned by][] an element with their [required context role][] even though they are not direct DOM children of it. ARIA [ownership][owned by] "skips" through presentational nodes because they are not [included in the accessibility tree][].
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not direct DOM children of it. The presentational nodes are not [included in the accessibility tree][], hence the two innermost `div` are children of the outermost in the [accessibility tree][] (but not in the DOM tree).
```
ajanec01, line 89:
```suggestion
These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] of `list`, even though they are not its direct children in DOM. The presentational node is not [included in the accessibility tree][].
```
It kind of read to me like `div`s are included in the accessibility tree and not the `listitem``s. I think it's well explained without the last sentence.

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -97,7 +99,7 @@ These elements with an [explicit role][] of `listitem` are [owned by][] an eleme
 
 #### Passed Example 4
 
-These elements with an [explicit role][] of `listitem` are [owned by][] an element with their [required context role][] even though they are not DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```
ajanec01, line 102:
```suggestion
These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its DOM descendants. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -107,7 +109,7 @@ These elements with an [explicit role][] of `listitem` are [owned by][] an eleme
 
 #### Passed Example 5
 
-These elements with an [explicit role][] of `listitem` are [owned by][] an element with their [required context role][] even though they are not direct DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not direct DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```
ajanec01, line 112:
```suggestion
These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct DOM descendants. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```

ajanec01 (dismissed):
Description for passed example 3 seems incorrect to me. Happy to approve otherwise.

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -47,7 +47,9 @@ The applicability of this rule is limited to the [WAI-ARIA 1.1 Recommendation][a
...
 
-The definition of [owned by][] used in this rule is different from the definition of ["owned element" in WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element). In this rule "owned by" includes only direct children of the owner element, whereas an "owned element" is any descendant of the owner.
+Being a child in the [accessibility tree][] is different from being a child in the DOM tree. Some DOM nodes have no corresponding node in the [accessibility tree][] (for example, because they are marked with `role="presentation"`). A child in the [accessibility tree][] can thus correspond to a descendant in the DOM tree. Additionally, the use of `aria-owns` attribute can change the tree structure to something which is not a subtree of the DOM tree.
+
+This rule is restricted to direct parent-child relation in the [accessibility tree][] which is more strict than the definition of ["owned element" in WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element). This rule mimics, on the roles level, the [content model](https://html.spec.whatwg.org/#concept-element-content-model) of HTML.
```
ajanec01, line 52: Shouldn't we delete the sentence where we are saying that this rule mimics the HTML content model? Asking because the definition of the content model is suggesting that it's both children and descendants of the element/ test target.

It's contradictory to the beginning of the paragraph where we insist that this rule is restricted to direct parent-child relation.

carlosapaduarte: I agree with Aron's suggestion

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -84,7 +86,7 @@ These elements with an [explicit role][] of `listitem` are [owned by][] an eleme
 
 #### Passed Example 3
 
-These elements with an [explicit role][] of `listitem` are [owned by][] an element with their [required context role][] even though they are not direct DOM children of it. ARIA [ownership][owned by] "skips" through presentational nodes because they are not [included in the accessibility tree][].
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct children in DOM. The presentational node is not [included in the accessibility tree][].
```
carlosapaduarte, line 89:
```suggestion
These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its children in DOM. The presentational node is not [included in the accessibility tree][].
```

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -107,7 +109,7 @@ These elements with an [explicit role][] of `listitem` are [owned by][] an eleme
 
 #### Passed Example 5
 
-These elements with an [explicit role][] of `listitem` are [owned by][] an element with their [required context role][] even though they are not direct DOM descendants of it. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its direct DOM descendants. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```
carlosapaduarte, line 112:
```suggestion
These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its DOM children. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -136,15 +138,15 @@ These elements with an [explicit role][] of `listitem` are [owned by][] an eleme
...
+This element with an [explicit role][] of `listitem` is not a child in the [accessibility tree][] of an element with its [required context role][].
 
 ```html
 <div role="listitem">List item 1</div>
 ```
 
 #### Failed Example 2
 
-These elements with an [explicit role][] of `listitem` are not [owned by][] an element with their [required context role][]. They are [owned by][] the `tabpanel` element, because it is the closest ancestor.
+These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][], but of the `tabpanel`.
```
carlosapaduarte, line 149:
```suggestion
These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][], but of an element with the `tabpanel` role.
```

carlosapaduarte (changes requested):

_rules/aria-required-context-role-ff89c9.md
WilcoFiers: I can go either way on this.

WilcoFiers (commented):
I agree with much of what Carlos has said.

_rules/aria-required-context-role-ff89c9.md
Jym77: Actually, it depends where we look 🤯 
The link used here indeed says
> A normative description of what content must be included as children and descendants of the element.

but a couple of line below we have [3.2.5 Content models](https://html.spec.whatwg.org/multipage/dom.html#content-models) stating
> Each element defined in this specification has a content model: a description of the element's expected contents. An HTML element must have contents that match the requirements described in the element's content model. The contents of an element are its children in the DOM.

🤷 

---
As far as I remember, HTML content model only refers to children, not further descendants (despite what the def says).

---
Additionally, the idea is not just to consider a single level in the tree (e.g., the rule allows the `table > rowgroup > row` chain which is 2 levels) but more than the chain have to be considered between parent/children and not ancestors/descendant.

Using the ARIA def of "owned element", something like `table > rowgroup > navigation > row` would be allowed because the `row` is "owned by" the `rowgroup` (as a descendant) but is not a "child in the accessiblity tree".

So the difference is not about only looking at children, but about looking one step at a time and not skipping any.

The reason why we do this stepwise checks (disregarding ARIA specs) is that the content model is doing likewise. That is `<table><tbody><nav><tr>` is not OK because `<nav>` is not allowed in a `<tbody>`. The content model is only looking one step at a time, and individual elements' content model only mention children, but builds up a model that restrict descendants further away.

Essentially, performing the same checks on implicit roles (only) should flag violations of the content model (not all of them, but some of them).

---
So, this is what I was trying to say there…
* We stray away from ARIA def.
* We do so because we take inspiration from content model.

I'm afraid that keeping only the first part (the action) without the second (its justification) could lead to much confusion ("why are you not following ARIA specs?")

Not sure how to rephrase it to convey that information… I'll let it simmer a bit in case any of you has a brillant idea 😁 
Easy fix might just be to switch the link and use the other one that does not mention descendants…

ajanec01 (approved):

WilcoFiers (approved):

_rules/aria-required-context-role-ff89c9.md
carlosapaduarte: I understand, and agre, with your reasoning.
Unfortunately, I don't have a better idea than what you suggested (i.e. change the link target).

carlosapaduarte (approved):

Jym77:
Call for review ends on June 17th.

colabottles (approved):

Jym77:

PR #1555, opened 2021-03-19
Update iframe-identical-name-equivalent-purpose-4b1c6c.md

tbostic32:
Deleted note in applicability and moved note in expectation to background.

Need for Final Call:

---

## How to Review And Approve

### DIFF
--- a/_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
+++ b/_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
@@ -29,13 +29,9 @@ This rule applies to any set of any two or more `iframe` elements which:
 - are [included in an accessibility tree][included in the accessibility tree]; and
 - that have [matching][] [accessible names][accessible name] that are not empty (`""`).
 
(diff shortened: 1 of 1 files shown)


_rules/iframe-identical-name-equivalent-purpose-4b1c6c.md
```diff
@@ -48,6 +44,8 @@ This rule assumes that assistive technologies are exposing all `iframe` elements
 
 ## Background
 
+When determining if target elements embed the same resource, resolving the embedded resource may include any redirects that are instant.
```
Jym77, line 47:
```suggestion
When determining if target elements embed the same resource, resolving the embedded resource includes any redirects that are instant.
```
I'd prefer a more direct formulation as "may include" can give the feeling that tester can choose whether they want to include it or not.

Jym77 (changes requested):

WilcoFiers (approved):

ajanec01 (approved):

daniel-montalvo (approved):

Jym77 (approved):

PR #1557, opened 2021-03-19
updated video failed example 1

zalmeraj:
Need for Final Call: None

## How to Review And Approve

### DIFF
--- a/_rules/video-alternative-for-visual-c5a4ea.md
+++ b/_rules/video-alternative-for-visual-c5a4ea.md
@@ -155,7 +155,7 @@ A video element that describes some of the text on the same page. The text on th
 
 #### Failed Example 1
 
-This 'video' element has an incorrect audio description. Testing.
+This 'video' element has an incorrect audio description.
 
 ```html
 <html lang="en">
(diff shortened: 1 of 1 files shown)


ajanec01 (approved):

_rules/video-alternative-for-visual-c5a4ea.md
```diff
@@ -155,7 +155,7 @@ A video element that describes some of the text on the same page. The text on th
 
 #### Failed Example 1
 
-A video element with an incorrect audio description.
+This 'video' element has an incorrect audio description.
```
Jym77, line 158:
```suggestion
This `video` element has an incorrect audio description.
```
Text enclosed in backticks is rendered as code, which is what we want here as we refer to the name of the HTML element.
('video' vs `video`)

Jym77 (changes requested):
If you feel confident, feel free to update the other examples description in the same way 👍

Jym77 (approved):
Github "pro tip":
After making changes requested by a review, you can
1. "dismiss" that review. This happens near the bottom of the page in the drop-down with the "1 change requested" label. => This ensures that other potential reviewers see that these concerns have been addressed; some reviewers are reluctant to review a PR which already has changes requested since it can sometimes feel "double work".
2. "re-request" a review from the same reviewer. This happens near the top of the page, by using the "recycling" arrows next to that reviewer's name. => This ensures that this reviewer gets notified that their concern have been handle and they need to get another look at it.

(no worries if you forget to do it)

WilcoFiers (approved):

WilcoFiers:
Looks like this is good to go.

PR #1560, opened 2021-03-19
Update 3 example descriptions to newer format

EmmaJP:
Minor updates to 3 of the example descriptions to the newer format requirements, in https://act-rules.github.io/rules/6a7281 - ARIA state or property has valid value. 

Need for Final Call:
- no need for final call.

---

## How to Review And Approve


EmmaJP:
@WilcoFiers Mira and I didn't go further because I didn't know if Passed Example 4 needed an element with the referenced id in the associated code snippet. Potentially a change that might need a final call, so not sure how you want to proceed.

Also, I don't know Mira's github id to assign the PR to her.

ajanec01 (dismissed):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -60,23 +60,23 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
...
+This `aria-expanded` attribute has a valid true/false/undefined value.
 
 ```html
 <div role="button" aria-expanded="undefined">A button</div>
 ```
 
 #### Passed Example 3
 
-`aria-pressed` state with valid tristate value
+This `aria-pressed` attribute has a valid true/false/mixed/undefined value.
```
Jym77, line 79:
```suggestion
This `aria-pressed` attribute has a valid tristate value.
```
"tristate" is the correct ARIA term for this attribute: https://www.w3.org/TR/wai-aria-1.2/#aria-pressed

Jym77 (dismissed):
I'm not sure about the change from "property/state" to "attribute". This difference does exist in ARIA, so the terms are a bit more precise than just "attribute". Not a blocker, however.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -66,39 +66,39 @@ For value type `URI`, this rule does not require that the destination URI exists
 
 #### Passed Example 1
 
-This `aria-required` attribute has a valid true/false value.
+This `aria-required` property has a valid true/false value.
```
Jym77, line 69:
```suggestion
This `aria-required` property has a valid [true/false](https://www.w3.org/TR/wai-aria-1.2/#valuetype_true-false) value.
```

Jym77 (approved):
We may want to link the value types to [their definitions](https://www.w3.org/TR/wai-aria-1.2/#propcharacteristic_value).

Not sure about it though (the same way we don't link element's or attribute's name to their HTML def).

WilcoFiers (approved):

colabottles (approved):

WilcoFiers:
This PR is completely out of date. All tests have since been rewritten.

Issue #1562, opened 2021-03-26
Does autocomplete attribute need appropriate `type`?


ajanec01:
This is a continuation of the conversation we had on whether or not the `autocomplete` attribute value needs to be used with a specific `type` attribute value (as per the HTML spec)? The question is raised in the context of sc 1.3.5 Identify input purpose.

This issue is closely related to #1541 and I don't think we can submit [`autocomplete` attribute has valid value](https://act-rules.github.io/rules/73f2c2) for a review until we resolve it.

I've created two [test pages for the `autocomplete` attribute]. One for [`autocomplete` with a correct control group](https://ajanec01.github.io/ajanec01/correct.html) and one for [`autocomplete` with an incorrect control group](https://ajanec01.github.io/ajanec01/). The findings are:
- Firefox is providing suggestions for both `bday-month` and `username` regardless of the `type`;
- Chrome is providing suggestions for `username` regardless of the `type` and does not provide any suggestions for `bday-month`
- Safari behaves just like Chrome

In summary, it looks like the browsers do not care about the `type` and do not provide suggestions/ autocompletion for all `autocomplete` attribute values.

The situation is slightly complicated given that sc 1.3.5 does not require autocompletion but for the property to be programmatically identifiable. If the rule was purely about autocompletion then I'm sure we should stick with the spec but because using `autocomplete` for 1.3.5 has somewhat become a secondary purpose of the `autocomplete`, I'm not so sure the rule should be so strict on the correct `type` attribute part, especially given that browsers do not seem to care.

It would be good to have a discussion about this during the next call.

WilcoFiers:
As discussed on the call, the CG agrees with the suggestion.

ajanec01:
## Results for `autocomplete` with the correct control group tests

As per our discussion in the last CG call, we decided to see how/ if browsers respect the suggestions from the HTML spec regarding the `type` attribute value being appropriate for an `autocomplete` attribute value. I’ve put together two pages that have form fields for all [types of input purposes](https://www.w3.org/TR/WCAG21/#input-purposes) listed under [sc 1.3.5](https://www.w3.org/TR/WCAG21/#identify-input-purpose). One page for form fields with the incorrect control group and the other for fields with the correct control group (links in the first comment). All tested on desktop machines in Firefox, Chrome, and Safari.

[...]

PR #1567, opened 2021-04-16
Autocomplete attribute has valid value updates

### LINKED ISSUE #1540: autocomplete="off" fails "autocomplete attribute has valid value" (73f2c2)

ajanec01:
Applying suggestions from the previous CG call.

- Making sure that `autocomplete="off"` is not a failure by changing the applicability
- Expanding assumption based on the findings from #1562 (Firefox, Safari, and Chrome do not care if `type` value is appropriate)

Closes issue(s):

- closes #1540 and #1541

No need for final call.

## How to Review And Approve


_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -37,21 +38,17 @@ Each test target's `autocomplete` [attribute value][] is a [space separated][] l
...
 
-The `aria-disabled` state is used on `input` elements which are not part of [sequential focus navigation][] and are not otherwise [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable). If this is not the case, this rule may be inapplicable on elements that are still [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable) and require a valid `autocomplete` attribute to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135].
+Mainstream user agents tend to provide autofill suggestions for `input` elements with or without an [appropriate field name for the form control][]. As such, the assistive technologies should be able recognize the purpose of a control even when the `autocomplete` [attribute value][] is inappropriate for the control's `type` [attribute value].
 
-## Accessibility Support
+## Accessibility Support
```
WilcoFiers, line 51:
```suggestion
## Accessibility Support
```

pages/glossary/embedded-image.md
```diff
@@ -15,7 +15,6 @@ An element presents an _embedded image_ when any of the following is true:
 - the element is an `svg` element having one or more `image` [descendants][] with a non-empty `href` [attribute value][]; or
 - the element has a [computed][] [`background-image`][background-image] CSS property with at least one [`image`][css-image] that is a [url reference][url-reference].
 
-
```
WilcoFiers, line 18: Please undo this. Don't want an autocomplete fix to be in the file history of embedded-image.md.

pages/glossary/rendered-image-resource.md
```diff
@@ -13,4 +13,4 @@ A _rendered image resource_ in a [web page][] is any resource with [visible pixe
 [embedded image]: #embedded-image 'Definition of Embedded Image'
 [image sources]: https://html.spec.whatwg.org/multipage/images.html#image-source
 [visible pixels]: #visible 'Definition of visible'
-[web page]: #web-page-html 'Definition of web page (HTML)'
\ No newline at end of file
+[web page]: #web-page-html 'Definition of web page (HTML)'
```
WilcoFiers, line 16: As above. Please undo this. Don't want an autocomplete fix to be in the file history.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -37,21 +38,17 @@ Each test target's `autocomplete` [attribute value][] is a [space separated][] l
 3. An optional token of either "home", "work", "mobile", "fax" or "pager", only if the last token is "email", "impp", "tel" or "tel-\*"; then
 4. A required token from the [correct autocomplete field][].
 
-## Expectation 2
```
WilcoFiers, line 40: You still have some failing examples that now pass the rule. Can you take those out and add in a passed example showing this should no longer be failed?

WilcoFiers (changes requested):

pages/glossary/rendered-image-resource.md
ajanec01: Sorry, not sure how that's happened.

ajanec01: @WilcoFiers I managed to get rid of the change in embedded-image file but I'm not sure what is happening here. There is literally no change at all. Do you have any suggestions?

_rules/autocomplete-valid-value-73f2c2.md
ajanec01: Will add an example with an incorrect type. Which failed example is now passing? I've looked through them again and I'm not sure which ones you mean. 
I've also checked the examples with optional tokens and if they are in the wrong order, querying the `autocomplete` returns an empty string, which is in line with what we currently have.

_rules/autocomplete-valid-value-73f2c2.md
```diff
@@ -37,19 +38,15 @@ Each test target's `autocomplete` [attribute value][] is a [space separated][] l
...
+The `aria-disabled` state is used on `input` elements which are not part of [sequential focus navigation][] and are not otherwise [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable). If this is not the case, this rule may be inapplicable on elements that are still operable and require a valid `autocomplete` attribute to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135].
 
-The `aria-disabled` state is used on `input` elements which are not part of [sequential focus navigation][] and are not otherwise [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable). If this is not the case, this rule may be inapplicable on elements that are still [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable) and require a valid `autocomplete` attribute to satisfy success Criterion [1.3.5 Identify Input Purpose][sc135].
+Mainstream user agents tend to provide autofill suggestions for `input` elements with or without an [appropriate field name for the form control][]. As such, the assistive technologies should be able recognize the purpose of a control even when the `autocomplete` [attribute value][] is inappropriate for the control's `type` [attribute value][].
```
Jym77, line 49: I'm not sure what is the assumption here…
"mainstream user agents tend to provide…" is not an assumption, it's a fact
"the AT should be able to…" is also not an assumption.

That should either be moved to the background or phrased in a way that mak the assumption obvious (even though I agree that the "This rule assumes that…" formulation is a bit heavy, I like it before it make it easy to spot the actual assumption and easier to write something which is actually an assumption).

Jym77 (changes requested):

pages/glossary/rendered-image-resource.md
ajanec01: I managed to reverse the updating with origin commit which has caused an issue but there was too much fiddling around with it so decided it would be clearer/ easier to cancel the pr and start from the beginning.

PR #1569, opened 2021-04-20
aria-required-context: Replace the assumption

WilcoFiers:
As far as I can figure, there is no way to override an explicit role. I don't think this assumption is right. On the other hand, we are making a bit of an assumption in saying that if something has the role of `cell`,

Need for Final Call: 1 week, because I'm removing an assumption.

## How to Review And Approve


_rules/aria-required-context-role-ff89c9.md
```diff
@@ -33,7 +33,7 @@ Each test target is [owned by][] an element that has a [semantic role][] that is
 
 ## Assumptions
 
-This rule assumes that the `role` attribute is used to give a [semantic role][] to the element according to ARIA specifications. If it is used for other purposes, and relationships between elements are already programmatically determinable by other means, it is possible to fail this rule but still satisfy [WCAG success criterion 1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).
+The rule assumes that the [explicit role][] of the applicable elements is appropriate for their element. I.e. A heading incorrectly marked up with `role="cell"` does not fail [success criterion 1.3.1 Info and Relationships][sc131] for not being in the context of a `row`. If the explicit role is both inappropriate and having that role does not cause other issues related to success criterion 1.3.1, the criterion can be satisfied even if this rule is `failed`.
```
ajanec01, line 36: It might be my lack of imagination but I can't think of an example where a `role` is inappropriate for an element, but it is contained inside its required context `role` and it passes 1.3.5. If the role is inappropriate. this is a failure of 1.3.5 regardless of the required context role/ owned elements being used.

Is there a particular example that you have in mind?

Jym77 (dismissed):
The reason for the current assumption is that the `role` attribute is not specific to ARIA and semantic roles. [Its specification](https://www.w3.org/TR/role-attribute/) says:
> Use cases include accessibility, device adaptation, server-side processing, and complex data description. 

I agree that as far as I know, the `role` attribute is only used to give a semantic role, but it is not an error to use it for some other purposes. In which case, a value which is not agreeing with ARIA specs is OK.
Maybe we need to rephrase this assumption, and/or move it to background, in the line of "whatever the purpose of the `role` attribute, AT are going to read it as giving a semantic role, so don't mess up with them".

_rules/aria-required-context-role-ff89c9.md
ajanec01: I was a bit confused when I first read it. It would be worth making it a bit more clear, as in `role` is not used to give semantics to elements as per @Jym77 comment.

WilcoFiers: @Jym77 @ajanec01 I'm not sure what you are suggesting to change. Can you maybe suggest what you are looking for?

WilcoFiers:
Dismissing the reviews, hoping to get a suggestion. Not sure what to do next with this one.

_rules/aria-required-context-role-ff89c9.md
Jym77: I feel there are two corner cases that we want (or not…) to address through Assumptions or Background notes.

---
One is the new assumption you're writing. If we have a
```html
<h1 role="row">Lorem Ipsum</h1>
```
out of the blue, the rule is going to fail but it is less clear than 1.3.1 is not satisfied. Typically, there is no info conveyed through presentation that this is a `row`, so 1.3.1 doesn't care that it's not in a `table`.
Of course, in that case one might argue that the info "this is a `heading`" is conveyed through presentation, so this might still violate 1.3.1, but for reasons other than what the rule looks at, so not clear we want the rule to flag it.

=> I agree we want something in that line. Could either be the way you write it or in a more tuned down version "the rule assume that if the role is not appropriate, this is in itself a violation of 1.3.1". Which is the complementary of your "if the role is inappropriate and doesn't violate 1.3.1" bit, but maybe looking a bit more nice on the rule (i.e. instead of "there are cases where the rule is plain wrong", it looks like "there are cases where the rule detects true errors but report them in a contrived way of 'missing context' rather than 'bad role'"). 

---
The other is how I've always read the old assumption. The `role` attribute is not restricted to specifying ARIA semantic roles. It is actually not even tied to ARIA. I agree that it is probably its only use currently (and it might have been better in the first place to have it as `aria-role` so we get all the benefits from non-interference and all, but 🤷 ).

Anyway, it is correct, according to specifications, to use the `role` attribute with an ontology that has nothing to do with ARIA roles. It is likely in the current landscape that this would result in no UA/AT being able to handle the page correctly, but it is not necessarily an error.

Of course, essentially all our rules (and definitions) actually assume that the `role` attribute is exclusively used with ARIA roles ontology. We probably don't want to mention it everywhere, but this rule is specifically about the `role` attribute. We do have a similar assumption in [role attribute has valid value](https://act-rules.github.io/rules/674b10).

So, I guess we could have, say,
```html
<div role="pier">
  <div id="dinghy" role="row">…</div>
<div>
```
where the rule would fail for having a `row` out of table, but 1.3.1 wouldn't fail since presentation doesn't show it as a `row`…

Maybe that bit is not needed 🤔 because it is so much on a corner case (and `role` is only used to give an ARIA role)…

Jym77 (approved):

ajanec01 (approved):

carlosapaduarte (approved):

colabottles (approved):

PR #1570, opened 2021-04-20
autoplay-3-seconds: editorial updates

WilcoFiers:
Need for Final Call: none

## How to Review And Approve

### DIFF
--- a/_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
+++ b/_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
@@ -25,10 +25,10 @@ acknowledgments:
 
 This rule applies to any `audio` or `video` element for which all the following are true:
 
-- (**autoplay**) the element has an `autoplay` [attribute value][] of `true`; and
-- (**not muted**) the element has a `muted` [attribute value][] of `false`; and
-- (**not paused**) the element has a `paused` [attribute value][] of `false`; and
-- (**duration**) the element has a [media resource][] lasting more than 3 seconds and that contains audio.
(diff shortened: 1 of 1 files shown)


Jym77 (approved):

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -25,10 +25,10 @@ acknowledgments:
...
 This rule applies to any `audio` or `video` element for which all the following are true:
 
-- (**autoplay**) the element has an `autoplay` [attribute value][] of `true`; and
-- (**not muted**) the element has a `muted` [attribute value][] of `false`; and
-- (**not paused**) the element has a `paused` [attribute value][] of `false`; and
-- (**duration**) the element has a [media resource][] lasting more than 3 seconds and that contains audio.
+- **autoplay**: the element has an `autoplay` [attribute value][] of `true`; and
+- **not muted**: the element has a `muted` [attribute value][] of `false`; and
+- **not paused**: the element has a `paused` [attribute value][] of `false`; and
+- **duration**: the element has a [media resource][] that contains audio lasting more than 3 seconds.
```
carlosapaduarte, line 31: I find this ambiguous. My first interpretation was that the "lasting more than 3 seconds" applied to the audio, not the media resource. Only after reading the expectation I realised that it must apply to the media resource, otherwise the expectation would always fail. What about?
```suggestion
- **duration**: the element has a [media resource][] lasting more than 3 seconds that contains audio.
```

carlosapaduarte (changes requested):

daniel-montalvo (approved):

carlosapaduarte (approved):

PR #1610, opened 2021-05-13
link has name: Fix missing tabindex on passed example 2

WilcoFiers:
Examples need to show good practice. If you're going to use role=link on a div, it needs to have tabindex.

Need for Final Call: 1 week

---

## How to Review And Approve


susanatx (approved):

_rules/link-non-empty-accessible-name-c487ae.md
```diff
@@ -85,7 +85,7 @@ This `a` element has an [accessible name][] from its content.
 This `div` element has an [explicit semantic role](#explicit-role) of `link` and an [accessible name][] from its content.
 
 ```html
-<div role="link" onclick="window.location.href='https://www.w3.org/WAI/'">Web Accessibility Initiative (WAI)</div>
+<div role="link" onclick="window.location.href='https://www.w3.org/WAI/'" tabindex="0">Web Accessibility Initiative (WAI)</div>
```
tbostic32, line 88: We should add a keyboard event here as well, otherwise even with the tabindex you cannot activate the link.

tbostic32 (changes requested):

kengdoj (approved):

_rules/link-non-empty-accessible-name-c487ae.md
Jym77: https://github.com/act-rules/act-rules.github.io/blob/develop/test-assets/bypass-blocks-cf77f2/click-on-enter.js (or a stripped down version) should do the trick.

WilcoFiers: Huh! I did not know that! Thank you.

Jym77 (approved):

carlosapaduarte (approved):

colabottles (approved):

Issue #1573, opened 2021-04-20
"ARIA state or property is permitted" (5c01ea) should allow more attributes, following "ARIA in HTML"


Jym77:
[ARIA state or property is permitted](https://act-rules.github.io/rules/5c01ea) only allows for global ARIA attributes and the ones explicitly mentioned in ARIA.
However, [ARIA in HTML](https://w3c.github.io/html-aria/#docconformance) is having an extended list of attributes (and roles) that are allowed depending on the HTML element. There are more of them than the stuff mentioned in ARIA.

Among other, `<input type="password">` have no role which, according to our rule, only allow for global `aria-*` attribute, but [ARIA in HTML of password](https://w3c.github.io/html-aria/#el-input-password) explicitly allows the attribute allowed on `textbox`.

Thus, our rule is failing `<input type="password" aria-required>` since `aria-required` is not a global attribute, but it shouldn't fail since `aria-required` is allowed on `textbox` and therefore on `input type="password"`.

See https://github.com/w3c/aria/issues/1461

Not entirely sure how to salvage the rule without being too verbose…

WilcoFiers:
The group is in agreement that this needs to be accounted for in the expectation. Feel free to suggest a change to this.

WilcoFiers:
As part of resolving this, we should also check if other rules are impacted by this.

ajanec01:
Fixed in #1612

PR #1612, opened 2021-05-16
Update aria-state-or-property-permitted-5c01ea.md

ajanec01:
Trying to resolve the issue raised in #1573. 

Not exactly sure if this is the right way so would appreciate the input.

Closes issue(s): #1573

Need for Final Call: 
1 week

## How to Review And Approve


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -32,9 +32,7 @@ This rule applies to any [WAI-ARIA state or property][] that is specified on an
 
 ## Expectation
 
-Each test target is either an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the attribute is specified. If the element has no [semantic role][], the attribute must be a [global state or property][global].
-
-**Note:** Assessing the value of the attribute is out of scope for this rule.
+Each test target is either an [inherited][], [supported][], or [required][] [state][] or [property][] for the element's [semantic role][] or the [WAI-ARIA state or property][] may be used on the language feature. If the element has no [semantic role][], the attribute must be a [global state or property][global].
```
Jym77, line 35: I'm a bit uneasy in leaving "language feature" as it and only explain it in a background note (versus trying to have a definition).
The rule is only targeting HTML and SVG, and I don't think there is a similar "ARIA in SVG" (or other feature document) out there. So maybe we can be a bit more prescriptive here (to make sure and remove the ambiguity). Pointing here to ARIA in HTML and saying that if the target is an HTML element, it may also have stuff described there.

ajanec01: Yes, I thought about that too. In the end, I left the "language feature" so that everybody else can say what they think about it. It's unlikely, I think, there will be an SVG accessibility API mappings spec so we could tighten it.

ajanec01: Yes, I thought about that too. In the end, I decided to leave the "language feature" for everybody else to decide on it. Yeah, I'm not sure if there ever will be an equivalent concept for SVG so we can tighten it.

WilcoFiers: I think you're trying to pack too much information into one sentence. It might be better to break this up into a list:

```suggestion
One of the following is true for each test target:

- **Global**: the test target is a [global state or property][global]; or

- **Semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or

- **language feature**: the test target is allowed as a language feature on the element on which it is specified. Which ARIA state or property may be used on which element is describe in [ARIA in HTML][].
```

This will also make it easier to add prohibited attributes when ARIA 1.2 makes it to rec.

WilcoFiers (dismissed):

carlosapaduarte (approved):

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -32,9 +32,11 @@ This rule applies to any [WAI-ARIA state or property][] that is specified on an
 
 ## Expectation
 
-Each test target is either an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the attribute is specified. If the element has no [semantic role][], the attribute must be a [global state or property][global].
+One of the following is true for each test target:
```
WilcoFiers, line 35: Maybe just:

```suggestion
For each test target, one of the following is true:
```

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -46,12 +48,17 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
...
+
+Assessing the value of the attribute is out of scope for this rule.
+
 - [ARIA state or property has valid value](https://act-rules.github.io/rules/6a7281)
 - [Understanding Success Criterion 4.1.1: Parsing](https://www.w3.org/WAI/WCAG21/Understanding/parsing.html)
 - [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
 - [WAI-ARIA 1.1, Supported States and Properties](https://www.w3.org/TR/wai-aria-1.1/#states_and_properties)
 - [WAI-ARIA 1.1, Global States and Properties](https://www.w3.org/TR/wai-aria-1.1/#global_states)
 - [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA5)
+- [Document conformance requirements for use of ARIA attributes in HTML](https://w3c.github.io/html-aria/#docconformance)
```
Jym77, line 61: All links to ARIA in HTML should be https://www.w3.org/TR/html-aria/ rather than github.io

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -46,12 +48,17 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 ## Background
 
+In HTML, there are language features that do not have corresponding implicit WAI-ARIA semantics. As per [ARIA in HTML Editor's Draft](https://w3c.github.io/html-aria/), those elements can have [global states or properties][global]. Some of those elements can also have [inherited][], [supported][], or [required][] [states][state] or [properties][property] that correspond to a [WAI-ARIA role](https://www.w3.org/TR/wai-aria-1.1/#usage_intro). For example, `audio` has no corresponding ARIA semantics but it can have [inherited][], [supported][], or [required][] [states][state] or [properties][property] of the [`application` role](https://www.w3.org/TR/wai-aria-1.1/#application).
```
Jym77, line 51:
```suggestion
In HTML, there are language features that do not have corresponding implicit WAI-ARIA semantics. As per [ARIA in HTML Editor's Draft](https://w3c.github.io/html-aria/), those elements can have [global states or properties][global]. Some of those elements can also have [inherited][], [supported][], or [required][] [states][state] or [properties][property] that correspond to a [WAI-ARIA role](https://www.w3.org/TR/wai-aria-1.1/#usage_intro). For example, the `audio` element has no corresponding ARIA semantics but it can have [inherited][], [supported][], or [required][] [states][state] or [properties][property] of the [`application` role](https://www.w3.org/TR/wai-aria-1.1/#application).
```

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -139,6 +146,14 @@ This `button` element has an [explicit role][] of `none`. However, it is [focusa
 <button role="none" aria-pressed="false">ACT rules are cool!</button>
 ```
 
+#### Passed Example 11
+
+This `input` element does not have an [explicit role][] of `textbox`, but the `aria-required` property may be used on such language feature.
```
Jym77, line 151:
```suggestion
This `input` element does not have an [explicit role][] of `textbox`, but the `aria-required` property may be used `input` element with a `type` [attribute value][] of `password`.
```
+ probably linking to the precise section of ARIA in HTML that says so.

Jym77 (dismissed):
Looks good 👍

_rules/aria-state-or-property-permitted-5c01ea.md
ajanec01: Done.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -32,9 +32,11 @@ This rule applies to any [WAI-ARIA state or property][] that is specified on an
 
 ## Expectation
 
-Each test target is either an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the attribute is specified. If the element has no [semantic role][], the attribute must be a [global state or property][global].
+One of the following is true for each test target:
 
-**Note:** Assessing the value of the attribute is out of scope for this rule.
+- **Global**: the test target is a [global state or property][global]; or
+- **Semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
+- **language feature**: for an HTML element, the test target is allowed on the element on which it is specified. Which ARIA state or property may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
```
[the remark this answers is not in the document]
ajanec01: Lowercase it is.

_rules/aria-state-or-property-permitted-5c01ea.md
ajanec01: Done.

Jym77 (approved):

daniel-montalvo (approved):

colabottles (approved):

PR #1630, opened 2021-05-27
table-header-has-assigned-cells: Remove H43 from mapping

### LINKED ISSUE #1626: [Quick Review] Table header cell has assigned cells
Rule Link: [Table header cell has assigned cells](https://act-rules.github.io/rules/d0f69e)

I am finding it hard to understand why technique [H43 Using id and headers attributes to associate data cells with header cells in data tables ](https://www.w3.org/WAI/WCAG21/Techniques/html/H43) is included in the accessibility requirements since it seems only applicable to passed example 5. In other words, if Passed Example 1 fails it seems like a bit of a stretch to me to consider H43 also failed as is stated in the requirements mapping. 

If we do follow the same logic, I might also expect [H51: Using table markup to present tabular information](https://www.w3.org/WAI/WCAG21/Techniques/html/H51) and [H63: Using the scope attribute to associate header cells and data cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H63) to be included since they follow a similar format of having specific examples very relevant for them.

WilcoFiers:
Closes #1626

Need for Final Call: 2 weeks

----

eview And Approve


daniel-montalvo (approved):

carlosapaduarte (changes requested):
Should remove H43 from the Background section also.

carlosapaduarte:
And shouldn't changing accessibility requirements be a 2 week final call?

carlosapaduarte:
I believe you didn't push your changes. At least I'm still seeing the reference to H43 in the background section.

ajanec01 (approved):

ajanec01:
Isn't that ok to keep the technique in the background section? Seems fine with the [description of the background section](https://act-rules.github.io/pages/design/rule-design/#background) on the rule design page.

Jym77 (approved):

annethyme (approved):

Issue #1617, opened 2021-05-24
ARIA Valid Value (6a7281) Pass Example 4 idref missing


tombrunet:
Rule: https://act-rules.github.io/rules/6a7281
Passed example 4: https://act-rules.github.io/testcases/6a7281/23f69af205f3c0032d60d64288fe96c59e2b0fa2.html
```
<div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div>
```
Simply says 
> aria-errormessage property with valid ID reference value

But, `#my-error` doesn't exist in the testcase, so the ID reference isn't valid...

tombrunet:
I saw a note in another issue about 'maybe' ignoring idrefs that aren't required. If that's the situation, would be nice to update the passing explanation. Otherwise, seems like the my-error reference needs to be added. I'm not sure which of the two is intended (feels like a valid id ref should be added to the testcase). Allowing for an invalid id-ref simply because that field isn't required feels like it's asking for trouble - devs won't be alerted to all sorts of bad refs.

tombrunet:
I can, but I'm not familiar with the testcase update process, and I don't see them in the repo. Do testcase ids change on update?

tombrunet:
A possible update would be:
```
<div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div>
```

Jym77:
> I can, but I'm not familiar with the testcase update process, and I don't see them in the repo.

That's why I'm suggesting you do it so you can learn the process 😉 

The test cases are stored as part of the rules. So you can edit the rule file itself (https://github.com/act-rules/act-rules.github.io/blob/develop/_rules/aria-state-or-property-valid-value-6a7281.md and click on the pen in the top right area).
When you save your changes, you'll be offered to create a new branch (probably a "fork" branch, not 100% sure) and a Pull Request.

PR #1633, opened 2021-05-28
#1617: Add missing reference to 6a7281 Pass4

tombrunet:
6a7281 testcase Pass4 was:
```html
```

This includes a reference to a non-existent `my-error`, which results in a failure. This reference should be added.

Closes issue(s):

- closes https://github.com/act-rules/act-rules.github.io/issues/1617

Need for Final Call:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-state-or-property-valid-value-6a7281.md
+++ b/_rules/aria-state-or-property-valid-value-6a7281.md
@@ -88,6 +88,7 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
 
 ```html
 <div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div>
+<div id="my-error">My error message</div>
 ```
 
 #### Passed Example 5


_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -88,6 +88,7 @@ Some user agents treat the value of `aria-*` attribute as case-sensitive (even w
 
 ```html
+<div id="my-error">My error message</div>
```
WilcoFiers, line 91: This is not required by the rule, see note 1 in the expectation:

> Only for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement.

An example of what might happen with aria-errormessage is that the aria-errormessage attribute is static, and including or removing the element with the ID is part of setting the error. Something like this:

```html
<template>
  <input aria-errormessage="err1" :aria-invalid="error ? true : false">
  <p v-if="error" id="err1">{{ error }}<p>
</template>
```

I've seen this in practice too. There is no difference between this, and actually removing the aria-errormessage attribute. The way this is written using a note isn't very clear, and will need to be improved.

WilcoFiers (changes requested):
Thank you for the PR, but this is very much intentional. I don't believe this should be changed, but if there is something we missed, please elaborate. Happy to bring it back to the group if it needs to be.

_rules/aria-state-or-property-valid-value-6a7281.md
tombrunet: I opened #1617 because the testcase does not indicate that as a reasoning. It says:
> aria-errormessage property with valid ID reference value

That's misleading if that's the reason it's passing. So, I was asked to add a PR to add the reference rather than change the message.

On our end, we'll probably still end up flagging it anyway since allowing missing references will also mask mistaken references. If you have aria-errormessage pointing at er1 and the actual id is err1, the developer would never notice it, even though they think they did the right thing. In your example, it's just as easy to do something like `error ? "messageId": undefined`.

WilcoFiers: Yeah, I agree with you, this isn't worded as well as it could be. It also shouldn't be in a note, it needs to be part of the expectation text itself. It's too easy to miss.

Would it help your case if we added aria-invalid=false on this example, to have a more explicit clue about what is going on? If so, feel free to make that change.

tombrunet: Honestly, if that's the intent, I would just change the description from:
> aria-errormessage property with valid ID reference value

to:

> aria-errormessage property with a non-existent ID reference value, but aria-errormessage is not required. See expectation note.

And not mess with changing the actual testcase. Does that make sense?

Jym77: Hi @tombrunet, we've discussed that on today call and indeed agreed that updating the description of the example is what makes sense 👍 
The example (and Failed Example 4) is more or less here to point out that "valid ID" doesn't mean "existing" but "OK according to ID parsing rules".

We can also add a link (from the expectation) to the HTML spec explaining that. https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute
> When specified on HTML elements, the id attribute value must be unique amongst all the IDs in the element's tree and must contain at least one character. The value must not contain any ASCII whitespace.

Can you handle that?

Jym77:
This is superceded by #1649 (after last week discussion in the CG call).

tombrunet:
Sorry - I had been out for a funeral. Looks like this has been taken care of separately, so I'll close this.

Issue #1593, opened 2021-05-05
Change "element with a semantic role of X" to "semantic X"?


WilcoFiers:
I was rereading some old rules, and we have quite a lot of places where the phrase "with a [semantic role] of X" appears. I wonder if we should simplify that phrase to just be "a [semantic] heading", where the word "semantic" is linked to the semantic role definition. If we wanted to make that a little clearer, we could consider be explicit in the semantic role definition about this. 

We may need to make the semantic role definition slightly more specific, to clarify that when in a phrase like "all semantic headings" refers to all elements with a semantic role of `heading`, but I think that can be done. We're essentially already doing this with the "semantic link" definition - the difference there being that this is anything that inherits from link too. 

This change hides some of the jargon behind definitions, making the rules a little easier to read. I really liked what we did with the "attribute value" definition, that helped quite a bit. I think this could do the same.

WilcoFiers:
Decision from the group:

- Use "semantic X" for just X as role
- Use  "inheriting semantic X" for X + anything inheriting from X

We need a new definition "inheriting semantic role" which replaces the "semantic link" definition.

Issue #459, opened 2019-03-26
Define SVG elements and HTML elements


WilcoFiers:
This question came up discussing the SVG image rule. It is not clear whether or not svg[role=img] is already covered by the image-has-name rule. If we take "any HTML element" to mean an element that is listed in the HTML spec, it does. However, while the HTML 5.2 spec includes the SVG element, it does **not** include a semantic description of the SVG element, which is part of the SVG spec. We need to distinguish between these two interpretations.

https://www.w3.org/TR/html52/semantics-embedded-content.html#svg

The solution to this is that we need to define what an SVG element is, and what an HTML element is. Basically the answer depends on the namespace. Anything with the SVG namespace is an SVG element, and anything with an HTML namespace is an HTML element.

PR #1636, opened 2021-05-31
define namespaced elements

WilcoFiers:
First version of this. It needs a to be used in a bunch more places, but before I do that work I'd like to hear what people think of the approach. It's a little different from what was discussed on last week's call.

Need for Final Call: 1 week

---

## How to Review And Approve


pages/glossary/namespaced-element.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Namespaced Element
+key: namespaced-element
+unambiguous: true
+objective: true
+---
+
+An [element][] with a specific [namespaceURI][] value from [HTML namespaces][]. For example an "SVG element" is an element with the "SVG namespace", which is `http://www.w3.org/2000/svg`.
+
+Namespaced elements are not limited to elements described in a specification. They also include custom elements. Elements such as `a`, `title`, and `svg` have a different namespace depending on where they are used. For example a `title` in an HTML page usually has the HTML namespace. When used in an `svg` element, a `title` elements has the SVG namespace instead.
```
carlosapaduarte, line 10:
```suggestion
Namespaced elements are not limited to elements described in a specification. They also include custom elements. 
Elements such as `a`, `title`, and `svg` have a different namespace depending on where they are used. For example a `title` in an HTML page usually has the HTML namespace. When used in an `svg` element, a `title` element has the SVG namespace instead.
```

carlosapaduarte (changes requested):
I'm okay with this. But I'm failing to understand how does this address what we discussed in the call?

Jym77:
> I'm okay with this. But I'm failing to understand how does this address what we discussed in the call?

That bit: https://www.w3.org/2021/05/27-act-rules-minutes.html#t03 / #459, 
Not that bit: https://www.w3.org/2021/05/27-act-rules-minutes.html#t04 / #1593

carlosapaduarte:
> > I'm okay with this. But I'm failing to understand how does this address what we discussed in the call?
> 
> That bit: https://www.w3.org/2021/05/27-act-rules-minutes.html#t03 / #459,
> Not that bit: https://www.w3.org/2021/05/27-act-rules-minutes.html#t04 / #1593

Thanks for pointing it out! I was looking in the wrong place 👍

pages/glossary/namespaced-element.md
Jym77, line 10: I actually think that a `<svg>` element inside an HTML document is in the SVG namespace…

[`svg` element in HTML specs](https://html.spec.whatwg.org/multipage/embedded-content-other.html#svg-0) links to [SVG `foreignObject`](https://svgwg.org/svg2-draft/embedded.html#ForeignObjectElement) which explains it's here for inclusion in other XML docs and has a note saying:
> The HTML parser treats elements inside the ‘`foreignObject`’ equivalent to elements inside an HTML document fragment. Any `svg` or `math` element, and their descendents, will be parsed as being in the SVG or MathML namespace, respectively;

Jym77 (changes requested):

pages/glossary/namespaced-element.md
WilcoFiers: I prefer to keep this as is.

WilcoFiers: Good catch! You are right, `svg` has the SVG namespace in HTML.

WilcoFiers, line 10:
```suggestion
Namespaced elements are not limited to elements described in a specification. They also include custom elements. Elements such as `a` and `title` have a different namespace depending on where they are used. For example a `title` in an HTML page usually has the HTML namespace. When used in an `svg` element, a `title` elements has the SVG namespace instead.
```

pages/glossary/namespaced-element.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Namespaced Element
+key: namespaced-element
+unambiguous: true
+objective: true
+---
+
+An [element][] with a specific [namespaceURI][] value from [HTML namespaces][]. For example an "SVG element" is an element with the "SVG namespace", which is `http://www.w3.org/2000/svg`.
+
+Namespaced elements are not limited to elements described in a specification. They also include custom elements. Elements such as `a` and `title` have a different namespace depending on where they are used. For example a `title` in an HTML page usually has the HTML namespace. When used in an `svg` element, a `title` elements has the SVG namespace instead.
```
carlosapaduarte, line 10:
```suggestion
Namespaced elements are not limited to elements described in a specification. They also include custom elements. Elements such as `a` and `title` have a different namespace depending on where they are used. For example a `title` in an HTML page usually has the HTML namespace. When used in an `svg` element, a `title` element has the SVG namespace instead.
```

carlosapaduarte (dismissed):
typo

Jym77 (approved):

colabottles (approved):

ajanec01 (approved):

tbostic32 (approved):

WilcoFiers:
After this one's merged, I'm going to update other uses of "HTML" and "SVG". I'll do that in a separate PR.

colabottles (approved):

PR #1649, opened 2021-06-16
Update aria-state-or-property-valid-value-6a7281.md

### LINKED ISSUE #1617: ARIA Valid Value (6a7281) Pass Example 4 idref missing

### LINKED ISSUE #1642: ARIA state or property has valid value- shadow DOM [6a7281]
In [ARIA state or property has valid value](https://act-rules.github.io/rules/6a7281), the second line of the expectation says:

> For value types ID Reference and ID Reference List for WAI-ARIA required properties at least one of the elements with the given ids exists in the same document tree or shadow tree as the element that specifies the target attribute.

I was not able to to find how to break shadow DOM boundaries when using ID reference or ID reference list value. Can we come up with an example before this rule gets published?

ajanec01:
Editorial changes to [ARIA state or property has valid value](https://act-rules.github.io/rules/6a7281)

Closes issue(s):

- closes #1642, closes #1617, closes #1633

Need for Final Call: none

## How to Review And Approve


Jym77 (approved):

colabottles (approved):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -28,15 +28,10 @@ This rule applies to any [WAI-ARIA 1.1 state or property](https://www.w3.org/TR/
 
 Each test target has a valid value according to its [WAI-ARIA 1.1 value type](https://www.w3.org/TR/wai-aria-1.1/#propcharacteristic_value).
 
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
+For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or in the same [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
```
WilcoFiers, line 31: I don't think this is accurate. This says the reference must exist, and then in the background it says, yes but only if the prop is required. We should update this text to only require the element to exist if the attribute is required.

WilcoFiers (dismissed):

_rules/aria-state-or-property-valid-value-6a7281.md
ajanec01: It says "for WAI-ARIA required properties". Isn't that enough?

carlosapaduarte (approved):

WilcoFiers:
I think this PR will also close https://github.com/act-rules/act-rules.github.io/issues/1432.

WilcoFiers (approved):

PR #1653, opened 2021-06-18
autocomplete: remove invalid example

WilcoFiers:
This example should have been removed when we changed the rule to no longer require the autocomplete be on an appropriate input type.

Need for Final Call: 1 week

## How to Review And Approve


carlosapaduarte (approved):

ajanec01 (changes requested):
I think the original idea was to catch an input with incorrect programmatic purpose. Yes, the example is wrong but  I feel that we should change it to:
```
<label>Email<input autocomplete="impp" /><label>
```

ajanec01 (approved):

Jym77 (approved):

PR #1654, opened 2021-06-18
Add assumption about no changes to default CSS styling 

### LINKED ISSUE #1445: Assumption about default styling for rules with "CSS Styling" as input aspect
As discussed in today's ACT CG meeting, test cases can have different outcomes if user stylesheets are used or if the user agent's default settings are modified by the user. 
Consequently, an assumption needs to be added to rules that have "CSS Styling" as an input aspect reflecting the test cases reliance on defaults.

carlosapaduarte:
Added the following assumption to all the rules that include CSS styling as input aspect:

- No [user stylesheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes of the ones presented here. 

Closes issue(s):
- closes #1445 

Need for Final Call:
- This will require a 1 week Final Call 

---

## How to Review And Approve


_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -41,7 +41,7 @@ None of the target elements are part of [sequential focus navigation](https://ht
 
 ## Assumptions
 
-_There are currently no assumptions_
+No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes of the ones presented here.
```
ajanec01, line 44:
```suggestion
No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are made. Otherwise the test cases might have different outcomes of the ones presented here. 
```

Wouldn't that sound better?

WilcoFiers (approved):
LGTM

_rules/aria-hidden-no-focusable-content-6cfa84.md
carlosapaduarte: I do prefer "are in place" over "are made". Are in place sounds more like a state (i.e. the user agent is in a specific state) while "are made" sounds more like an action (i.e. change the user agent to be in a specific state).

But this is not a strong preference, so if more people prefers the change I'll implement it.

ajanec01 (dismissed):

_rules/aria-hidden-no-focusable-content-6cfa84.md
Jym77: I am a bit uncomfortable having this as an **assumption**. It is not an assumption made by the rules, but one made by the examples. The rules work fine in presence of user agent styles.

I think it would be better maybe under the "Examples" heading (and before the "Passed Examples" one) as a note that the **examples** assume default styles.

Maybe also add a link to [Rendering](https://html.spec.whatwg.org/multipage/rendering.html#rendering) which is more or less what we consider to be the "default style".

Jym77 (changes requested):

carlosapaduarte:
@Jym77 I agree with your concern. Moved the assumption to a note in the examples section
@WilcoFiers @ajanec01 can you guys check if you agree with the change?

ajanec01 (approved):

WilcoFiers (changes requested):
I'm kind of reluctant to add this text to 70 rules, and lengthen already very long pages. Would it make sense to add something about this to the CSS Styling input aspect description instead?
https://www.w3.org/TR/act-rules-aspects/#input-aspects-css

We control this document, so we can add to it what we want.

Jym77:
I like @WilcoFiers idea of putting it in the "Input aspects" document 👍 

(and in the same line, we did add some times ago an assumption on language rule that "language is understood" which maybe could also move there for the same reason 🤔 )

carlosapaduarte:
@WilcoFiers I reverted the changes, but I don't think I'm allowed to edit this resource: https://www.w3.org/TR/act-rules-aspects/ 
Or am I wrong?

carlosapaduarte:
I created https://github.com/w3c/wcag-act/pull/520 to address this

Jym77:
Should we just close this one and instead link #1445 directly to https://github.com/w3c/wcag-act/pull/520 ?

carlosapaduarte:
The update to the input aspects made this PR void. Closing.

PR #1632, opened 2021-05-27
Instrument accessibility note

Jym77:
Add a note in the definition of instrument, and the rules using it, to indicate that instruments should be accessible.

Closes issue(s):

- closes #1524 

Need for Final Call:
This will not require a Final Call (editorial changes)

---

### **When creating PR:**

- [X] Make sure you're requesting to **pull a branch** (right side) to the `develop` branch (left side).
- [X] Make sure you do not remove the "How to Review and Approve" section in your pull request description

### **After creating PR:**

- [X] Add yourself (and co-authors) as "Assignees" for PR.
- [X] Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- [X] [Link the PR to any issue it solves](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue). This will be done automatically by [referencing the issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) at the top of this comment in the indicated place.
- [X] Optionally request feedback from anyone in particular by assigning them as "Reviewers".

### **When merging a PR:**

- [ ] Close any issue that the PR resolves. This will happen automatically upon merging if the PR was correctly linked to the issue, e.g. by referencing the issue at the top of this comment.

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Final Call period. In case of disagreement, the longer period wins.

PR #1655, opened 2021-06-18
Update 80f0bf - audio or video avoids automatically playing audio

### LINKED ISSUE #1502: How to update `audio` or `video` has no audio that plays automatically (80f0bf)
Early feedback from the TF reveals that the "3 seconds" condition on the applicability and in one specific atomic rule is confusing.

There are, at least, 2 ways to address this:

Rewrite the applicability in order to include what is now the expectation of the atomic rule Audio or video that plays automatically has no audio that lasts more than 3 seconds, that is, the rules would be applicable only when the audio lasts more than 3 seconds. This would mean this atomic rule would no longer be needed, and we would end up with just a single atomic rule: audio or video that plays automatically has a control mechanism

Remove the 3 seconds condition from the applicability and keep the composite with the 2 atomic rules.

carlosapaduarte:
- Change composite rule to atomic
- Update applicability to only consider media with more than 3 seconds of audio
- Update expectations and test cases to match the ones in the previous atomic "audio or video that plays automatically has a control mechanism"

Closes issue(s):
- closes #1502 

Need for Final Call:
- This will require a 2 weeks Final Call

---

## How to Review And Approve


_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -51,20 +53,23 @@ This rule applies to any `audio` or `video` element for which all the following
...
+## Expectation 1
 
-For each test target, the [outcome](#outcome) of at least one of the following rules is passed:
+For each test target, there is at least one [instrument][] in the same [web page][] to pause or stop the audio, or turn the audio volume off independently from the overall system volume control.
 
-- [Audio Or Video That Plays Automatically Has A Control Mechanism](https://act-rules.github.io/rules/4c31df)
-- [Audio Or Video That Plays Automatically Has No Audio That Lasts More Than 3 Seconds](https://act-rules.github.io/rules/aaa1bf)
+## Expectation 2
+
+The [instrument][] to pause or stop or turn the audio volume off is [visible](#visible), has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace), and is [included in the accessibility tree](#included-in-the-accessibility-tree).
```
ajanec01, line 64:
```suggestion
The [instrument][] to pause or stop or turn the audio volume off is [visible](#visible), [included in the accessibility tree](#included-in-the-accessibility-tree), and it has an [accessible name](#accessible-name) that is not only [whitespace](#whitespace).
```
My preference, feel free to ignore.

ajanec01 (dismissed):

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
Jym77: I actually think this expectation should go away in favour of #1632 (where I somehow seem to have forgotten to remove it 😖 )

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -89,12 +96,12 @@ This `audio` element has an [instrument][] to pause, stop, or turn the audio vol
 
 #### Passed Example 2
 
-This `video` element does not play for longer than 3 seconds.
+This `video` element has an [instrument][] to pause, stop, or turn the audio volume off.
```
Jym77, line 99:
```suggestion
This `video` element has (in its default controls) an [instrument][] to pause it, an [instrument][] to stop it, and an [instrument][] to turn its audio volume off.
```

_rules/auto-play-audio-does-not-exceed-3-seconds-aaa1bf.md
```diff
@@ -1,131 +0,0 @@
----
```
Jym77, line 1: I  am not sure we should bluntly delete the file, or rather first deprecate the rule.
(see [DEPRECATED — Image filename is accessible name for image](https://act-rules.github.io/rules/9eb3f6) for another deprecated rule)

Jym77 (dismissed):

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -51,29 +53,32 @@ This rule applies to any `audio` or `video` element for which all the following
...
+
+This rule assumes that the [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) to control the sound must be visible and accessible in order to be effective and usable by all kinds of users. If the mechanism is hidden to some users, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes of the ones presented here.
```
ajanec01, line 68: Shouldn't that part go under Test Cases heading to be in line with #1654 ?

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -51,29 +53,32 @@ This rule applies to any `audio` or `video` element for which all the following
 - **autoplay**: the element has an `autoplay` [attribute value][] of `true`; and
 - **not muted**: the element has a `muted` [attribute value][] of `false`; and
 - **not paused**: the element has a `paused` [attribute value][] of `false`; and
-- **duration**: the element has a [media resource][] lasting more than 3 seconds and that contains audio.
+- **audio duration**: the element has a [media resource][] for which the audio output lasts more than 3 seconds.
```
WilcoFiers, line 56: This seems slightly ambiguous to me. Should we do this based on the duration property?

WilcoFiers (dismissed):

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
carlosapaduarte: It does not work based on the duration because the media resource could have a duration longer than 3 seconds, while the audio in the media resource is shorter than 3 seconds. 

What exactly is ambiguous in the current description?

carlosapaduarte:
Following @WilcoFiers suggestion, I deprecated the 3 rules and created a new one with the changes that have been made to transform the prior composite into an atomic.
Ready for reviewing again

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,316 @@
...
+
+This rule assumes that the [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) to control the sound must be visible and accessible in order to be effective and usable by all kinds of users. If the mechanism is hidden to some users, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes of the ones presented here.
```
Jym77, line 70:
```suggestion
No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes from the ones presented here. 
```

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,316 @@
...
+
+This `audio` element has an [instrument][] to pause, stop, or turn the audio volume off.
+
+```html
+<audio src="/test-assets/moon-audio/moon-speech.mp3" autoplay controls></audio>
+```
+
+#### Passed Example 2
+
+This `video` element has (in its default controls) an [instrument][] to pause it, an [instrument][] to stop it, and an [instrument][] to turn its audio volume off.
```
Jym77, line 99: The "(in its default controls)" bit should probably be present either on all (relevant) cases or none.
Here, it seems that only that example has default control and needs the precision.
(OK, seems like only Passed Example 1 and this have default control, so probaby add the same precision on PE1)

Jym77 (dismissed):

carlosapaduarte:
@carlosapaduarte  make sure to update frontmatter also

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: 80f0bf
-name: '`audio` or `video` avoids automatically playing audio'
+name: DEPRECATED - '`audio` or `video` avoids automatically playing audio'
```
Jym77, line 3: Need a `deprecated` field in frontmatter too 😁

Jym77 (dismissed):

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,317 @@
...
+- **not paused**: the element has a `paused` [attribute value][] of `false`; and
+- **audio duration**: the element has a [media resource][] for which the audio output lasts more than 3 seconds.
+
+## Expectation
+
+For each test target, there is at least one [instrument][] in the same [web page][] to pause or stop the audio, or turn the audio volume off independently from the overall system volume control.
+
+## Assumptions
+
+This rule assumes that the [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) to control the sound must be located in the same [web page][]. Mechanisms located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen readers users) since the autoplaying sound will interfere with their ability to find and activate the mechanism. If a [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) external to the [web page][] is provided, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
```
WilcoFiers, line 66: Should this use the word "instrument" instead of "mechanism"?

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,317 @@
...
+
+This rule assumes that the [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) to control the sound must be visible and accessible in order to be effective and usable by all kinds of users. If the mechanism is hidden to some users, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes from the ones presented here.
+
+## Accessibility Support
+
+The native `video` and `audio` controls in several browser and assistive technology combinations are not keyboard accessible and the `video` or `audio` element itself may not be announced. Authors are recommended to use custom controls for keyboard navigation and cross browser accessibility support in general.
```
WilcoFiers, line 74: From https://github.com/act-rules/act-rules.github.io/pull/1787/files

```suggestion
The native `video` and `audio` controls in several browser and assistive technology combinations are not keyboard accessible and the `video` or `audio` element itself may not be announced. Authors are recommended to use custom controls for keyboard navigation and cross browser accessibility support in general. Some major browsers no longer automatically play the 'video' unless the 'video' is muted.
```

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Background
+
+The [instruments][instrument] used to pass this rule (if any), must meet all level A Success Criteria in order to fully satisfy [Success Criterion 1.4.2 Audio Control][sc142]. These extra requirements are left out of this rule, and should be tested separately.
+
+- [Understanding Success Criterion 1.4.2: Audio Control](https://www.w3.org/WAI/WCAG21/Understanding/audio-control.html)
```
WilcoFiers, line 80: Can you use the new headings?

WilcoFiers (dismissed):
Couple minor points.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Indeed it should. "Instrument" is already used in the expectation and the examples' descriptions.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Thanks for the reminder!

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Expectation
+
+For each test target, there is at least one [instrument][] in the same [web page][] to pause or stop the audio, or turn the audio volume off independently from the overall system volume control.
+
+## Assumptions
+
+This rule assumes that the [instrument][] to control the sound must be located in the same [web page][]. [Instruments][instrument] located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen readers users) since the autoplaying sound will interfere with their ability to find and activate the [instrument][]. If a [instrument][] external to the [web page][] is provided, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+This rule assumes that the [instrument][] to control the sound must be visible and accessible in order to be effective and usable by all kinds of users. If the [instrument][] is hidden to some users, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
```
WilcoFiers, line 68: There's nothing about the instrument being visible or accessible. I think the assumption is slightly different:

```suggestion
This rule assumes that the [instrument][] to control the sound is visible and accessible in order to be effective and usable by all kinds of users. If the [instrument][] is hidden to some users, it is possible to pass this rule but still not satisfy [Success Criterion 1.4.2 Audio Control][sc142].
```

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+
+No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes from the ones presented here.
+
```
WilcoFiers, line 71: This is generally true. I don't think there is any reason to put this assumption into the rule.

```suggestion
```

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+## Accessibility Support
+
+The native `video` and `audio` controls in several browser and assistive technology combinations are not keyboard accessible and the `video` or `audio` element itself may not be announced. Authors are recommended to use custom controls for keyboard navigation and cross browser accessibility support in general. Some major browsers no longer automatically play the 'video' unless the 'video' is muted.
```
WilcoFiers, line 74: Is this still true? Which browsers do not have keyboard accessible controls?

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+- [Failure of Success Criterion 1.4.2 for absence of a way to pause or stop an HTML5 media element that autoplays](https://www.w3.org/WAI/WCAG21/Techniques/failures/F93)
+- [G170: Providing a control near the beginning of the Web page that turns off sounds that play automatically](https://www.w3.org/WAI/WCAG21/Techniques/general/G170)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `audio` element has (in its default controls) an [instrument][] to pause, stop, or turn the audio volume off.
```
WilcoFiers, line 93:
```suggestion
This `audio` element has its default controls as an [instrument][] to pause, stop, and mute the audio.
```

I don't think we should do the brackets thing. That this comes through default controls is important information. Can you apply these changes to the other examples too?

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Assumptions
+
+This rule assumes that the [instrument][] to control the sound must be located in the same [web page][]. [Instruments][instrument] located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen readers users) since the autoplaying sound will interfere with their ability to find and activate the [instrument][]. If a [instrument][] external to the [web page][] is provided, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+This rule assumes that the [instrument][] to control the sound must be visible and accessible in order to be effective and usable by all kinds of users. If the [instrument][] is hidden to some users, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+No [user style sheets](https://drafts.csswg.org/css-cascade/#cascade-origin-user) are used and no changes to the [user agent default style sheet](https://drafts.csswg.org/css-cascade/#cascade-origin-ua) are in place, otherwise the test cases might have different outcomes from the ones presented here.
+
+## Accessibility Support
```
WilcoFiers, line 72: In Chrome you can no longer use autoplay on video elements, unless they are muted. I think we should have something in acc support about this.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+```html
+<video autoplay>
+	<source src="/test-assets/rabbit-video/video.mp4" type="video/mp4" />
+	<source src="/test-assets/rabbit-video/video.webm" type="video/webm" />
+</video>
+```
+
+#### Failed Example 3
+
+This `video` element has an [instrument][] to pause, stop, or turn the audio volume off but the [instrument][] is not visible.
```
WilcoFiers, line 168: The expectation doesn't ask that the control be visible. Unlike the following examples, I can see an argument for why this would fail 1.4.2.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+			<button type="button" id="mute">Mute</button>
+		</div>
+	</div>
+	<script src="/test-assets/80f0bf/no-autoplay.js"></script>
+</body>
+```
+
+#### Failed Example 4
+
+This `video` element has an [instrument][] to pause, stop, or turn the audio volume off but its `button` elements do not have accessible names.
```
WilcoFiers, line 204: The expectation doesn't ask that the control has an accessible name. That also seems out of scope for this SC. If there was a button there with no accessible name, I'd fail that under 4.1.2, not under 1.4.2.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+			<button type="button" id="mute"></button>
+		</div>
+	</div>
+	<script src="/test-assets/80f0bf/no-autoplay.js"></script>
+</body>
+```
+
+#### Failed Example 5
+
+This `video` element has an [instrument][] to pause, stop, or turn the audio volume off but the [instrument][] is not included in the accessibility tree.
```
WilcoFiers, line 239: As above. I think that's a 4.1.2 failure.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+
+```html
+<head>
+	<style>
```
WilcoFiers, line 243: Do you need this style elm?

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+```html
+<audio src="/test-assets/moon-audio/moon-speech.mp3" autoplay></audio>
+```
+
+#### Failed Example 2
+
+This `video` element's audio autoplays for longer than 3 seconds, and does not have an [instrument][] to pause, stop, or mute the audio
+
+```html
+<video autoplay>
```
WilcoFiers, line 160: Should we add a note to these that this will not autoplay in several major browsers, or would it be sufficient to mention that in accessibility support?

WilcoFiers (dismissed):
Nice work on this. I have a few thoughts though.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: You are correct. And in the definition of instrument we have a note stating that rules using the definition should not check for issues related to 2.1.1 and 4.1.2 so that we don't report one element in two different rules for the same reason.

And I'm curious... what is your argument for failing 1.4.2 in this instance?

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Removed the example

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Removed the example

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Not really. Removing it.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: We already had 😃

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: I've made the accessibility support note more explicit. Let me know if you feel that the note is enough or that it should be added to all examples with autoplay.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: I guess you're correct. Did some tests and couldn't find one. I'll remove the note.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,319 @@
...
+- **not paused**: the element has a `paused` [attribute value][] of `false`; and
+- **audio duration**: the element has a [media resource][] for which the audio output lasts more than 3 seconds.
+
+## Expectation
+
+For each test target, there is at least one [instrument][] in the same [web page][] to pause or stop the audio, or turn the audio volume off independently from the overall system volume control.
+
+## Assumptions
+
+This rule assumes that the [instrument][] to control the sound must be located in the same [web page][]. [Instruments][instrument] located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen readers users) since the autoplaying sound will interfere with their ability to find and activate the [instrument][]. If a [instrument][] external to the [web page][] is provided, it is possible to fail this rule but still satisfy [Success Criterion 1.4.2 Audio Control][sc142].
```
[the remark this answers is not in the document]
carlosapaduarte: Rephrased

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
Jym77: Is the "not playing" achieved by ignoring the `autoplay` attribute, or by overwritting it? That is on a browser that doesn't autoplay unmuted videos, does `video.getAttribute("autoplay")` still returns `true` (as set in the HTML), or does it returns `false` (ignoring the attribute)?
In the second case, the Applicability already makes the video inapplicable on such browsers… (still need an AS note, but likely not one on each example).

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,198 @@
...
+---
+
+## Applicability
+
+This rule applies to any `audio` or `video` element for which all the following are true:
+
+- **autoplay**: the element has an `autoplay` [attribute value][] of `true`; and
+- **not muted**: the element has a `muted` [attribute value][] of `false`; and
+- **not paused**: the element has a `paused` [attribute value][] of `false`; and
+- **audio duration**: the element has a [media resource][] for which the audio output lasts longer than 3 seconds.
```
Jym77, line 58: We can use `dfn` elements for the condition list (and possibly link to the correct condition from corresponding examples).

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,198 @@
...
+This `video` element's audio autoplays for longer than 3 seconds, and does not have an [instrument][] to pause, stop, or mute the audio.
+
+```html
+<video autoplay>
+	<source src="/test-assets/rabbit-video/video.mp4" type="video/mp4" />
+	<source src="/test-assets/rabbit-video/video.webm" type="video/webm" />
+</video>
+```
+
+### Inapplicable
```
Jym77, line 151: Should we add an example with a long video but short audio? I understand this should pass 1.4.2, and it should be inapplicable for the rule due to the **audio duration** condition. But this also gets to the definition of "audio output" (Does John Cage's *4'33* have an "audio output"?)

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Added `dfn` elements and updated inapplicable examples accordingly

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Chrome autoplays, Firefox and Safari don't. 
All of them return the empty string, since `autoplay` does not really have a value. Even if it is a boolean attribute, `autoplay='false'` [does not work](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-autoplay). It is present and it autoplays, or it is not present and it does not autoplay.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Added inapplicable example 5

carlosapaduarte:
Changes made

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Assumptions
+
+This rule assumes that it is not sufficient for the [instrument][] to control the sound to be located on a different [web page][], or a different state of the same [web page][] to pass the rule. [Instruments][instrument] located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen reader users) since the autoplaying sound will interfere with their ability to find and activate the [instrument][]. If an [instrument][] external to the [web page][] is provided, this rule will fail but it is still possible to satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+This rule assumes that the [instrument][] to control the sound is visible and accessible in order to be effective and usable by all kinds of users. If the [instrument][] is hidden to some users, it is possible to pass this rule but still not satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+## Accessibility Support
+
+Some major browsers no longer automatically play the 'video' unless the 'video' is muted.
```
Jym77:
```suggestion
Some major browsers do not automatically play the 'video' unless the 'video' is muted.
```
(more timeless wording).
Also, should we expand a bit like "for these UA, the rule is pointless/never applicable; but for UA that do autoplay unmuted video, the rule still makes sense" (notably as a note for future TF to remember and check whether some browser still autoplay videos in the future when they review the rule yearly).

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Expectation
+
+For each test target, there is at least one [instrument][] in the same [web page][] to pause, stop, or mute the audio.
+
+## Assumptions
+
+This rule assumes that it is not sufficient for the [instrument][] to control the sound to be located on a different [web page][], or a different state of the same [web page][] to pass the rule. [Instruments][instrument] located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen reader users) since the autoplaying sound will interfere with their ability to find and activate the [instrument][]. If an [instrument][] external to the [web page][] is provided, this rule will fail but it is still possible to satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+This rule assumes that the [instrument][] to control the sound is visible and accessible in order to be effective and usable by all kinds of users. If the [instrument][] is hidden to some users, it is possible to pass this rule but still not satisfy [Success Criterion 1.4.2 Audio Control][sc142].
```
Jym77: Should that be merged with the Background note on the same topic?
I'm not sure this needs to be an Assumption (since we do have a "further testing needed" mapping already).

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,208 @@
...
+Some major browsers no longer automatically play the 'video' unless the 'video' is muted.
+
+## Background
+
+The [instruments][instrument] used to pass this rule (if any), must meet all level A Success Criteria in order to fully satisfy [Success Criterion 1.4.2 Audio Control][sc142]. These extra requirements are left out of this rule, and should be tested separately.
+
+### Bibliography
+
+- [Understanding Success Criterion 1.4.2: Audio Control](https://www.w3.org/WAI/WCAG21/Understanding/audio-control.html)
+- [Accessible Multimedia](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/Multimedia)
```
Jym77: Do we want to link to MDN from a W3C site?
Is there some similar W3C document we could link to instead?

Jym77 (approved):
Minor comments, approving already.

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Done

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
carlosapaduarte: Good suggestion!

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,206 @@
...
+
+For each test target, there is at least one [instrument][] in the same [web page][] to pause, stop, or mute the audio.
+
+## Assumptions
+
+This rule assumes that it is not sufficient for the [instrument][] to control the sound to be located on a different [web page][], or a different state of the same [web page][] to pass the rule. [Instruments][instrument] located on other pages can still create accessibility issues for users relying on sound to navigate (e.g. screen reader users) since the autoplaying sound will interfere with their ability to find and activate the [instrument][]. If an [instrument][] external to the [web page][] is provided, this rule will fail but it is still possible to satisfy [Success Criterion 1.4.2 Audio Control][sc142].
+
+## Accessibility Support
+
+Some major browsers do not automatically play the 'video' unless the 'video' is muted. For these, the rule is never applicable. But, for UA that autoplay not muted video, the rule is still applicable.
```
WilcoFiers: Let's not abbreviate.

```suggestion
Some major browsers do not automatically play the 'video' unless the 'video' is muted. For these, the rule is never applicable. But for browsers that autoplay not muted video, the rule is still applicable.
```

_rules/audio-or-video-avoids-automatically-playing-audio-x0paj4.md
```diff
@@ -0,0 +1,206 @@
...
+
+#### Failed Example 1
+
+This `audio` element autoplays, lasts for more than 3 seconds, and does not have an [instrument][] to pause, stop, or mute the audio.
+
+```html
+<audio src="/test-assets/moon-audio/moon-speech.mp3" autoplay></audio>
+```
+
+#### Failed Example 2
```
WilcoFiers, line 138: This example somewhat contradicts the idea that whether a failed example fails shouldn't depend on accessibility support. I'm a little worried this example could block implementations that run in Chromium. I know that could happen in axe Pro (although we have no tools testing this today). I'm pretty sure Trusted Tester would be impacted though, as it is tested in Edge.

So from that angle I'm not sure I would want this example in. On the other hand, not having a failed video example in it feels like a problem too. We could possibly wait with this rule until we have this "optional examples" idea worked out. A different option could be to modify the rule so that it isn't just about autocomplay, but more generally about media elements that start playing without user interaction within the first 3 second of the page loading? That feels like a reasonable improvement on this rule anyway, and we could then add videos that get turned on through a script during onload.

WilcoFiers (changes requested):

PR #1659, opened 2021-06-23
Clarify 'one of the following' phrases

### LINKED ISSUE #1624: Use "at least one of the following" instead of "one of the following" in conditionals
Our recommended text for conditional lists is to use "for which one of the following is true", followed by a list of conditions. This could lead someone to think that if two conditions are true, the statement would be false. So instead I think these need to be changed to say "for which at least one of the following is true".

WilcoFiers:
I used "at least one of the following" in phrases where the outcome is something desirable. It seemed a little odd to say "except if at least one... " so I used "one ore more" in a few cases where "at least" seemed out of place.

- closes #1624

Need for Final Call: 2 weeks. This change touches a large number of rules.

## How to Review And Approve

### DIFF
--- a/pages/design/rule-design.md
+++ b/pages/design/rule-design.md
@@ -163,7 +163,7 @@ Applicability and expectations, often have multiple conditions that need to be m
 
 This phrasing is designed to be easily readable, but may not work in every situation. In all cases prefer readability over prescriptive formatting. Additionally to this, keep the following in mind when listing conditions:
 
-- The phrasing "for which (all | one of) the following is" can be modified based on use, such as "to achieve one of the following objectives".
+- The phrasing "for which (all | at least one of) the following is" can be modified based on use, such as "to achieve at least one of the following objectives".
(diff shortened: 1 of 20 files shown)


pages/design/rule-design.md
```diff
@@ -163,7 +163,7 @@ Applicability and expectations, often have multiple conditions that need to be m
 
 This phrasing is designed to be easily readable, but may not work in every situation. In all cases prefer readability over prescriptive formatting. Additionally to this, keep the following in mind when listing conditions:
 
-- The phrasing "for which (all | one of) the following is" can be modified based on use, such as "to achieve one of the following objectives".
+- The phrasing "for which (all | at least one of) the following is" can be modified based on use, such as "to achieve at least one of the following objectives".
```
Jym77, line 166: Should we mention here the "except if one or more of the following" construction?

Jym77 (approved):

carlosapaduarte (approved):

ajanec01 (approved):

Issue #1446, opened 2020-09-14
All audio and video are "non streaming" (afb423, e7aa44, 2eb176, eac66b, c5a4ea, ab4d13, 1ea59c, f51b46, f196ce, c3232f, fd26cf, d7ba54, ac7dc6, ee13b5, 1ec09b, 1a02b0)


Jym77:
Our definition of [non-streaming](https://act-rules.github.io/glossary/#non-streaming-media-element) is 
> A non-streaming media element is an HTML Media Element for which the `duration` property is not 0.

However, the HTML specs for [`duration`](https://html.spec.whatwg.org/multipage/media.html#dom-media-duration) states:
> If the media resource is not known to be bounded (e.g. streaming radio, or a live event with no announced end time), then the attribute must return the positive Infinity value.

Therefore, the `duration` is never 0 (actually, it might be for an empty media), and according to our definition, every media element is "non streaming"…

I am afraid that changing to "`duration` is not positive infinity" is not good enough because I understand that streams may have a non-infinite `duration` (e.g. live event with an announced end time). And it is bad if we qualify an actual streaming event as "non-streaming" because that makes it applicable for all the rules requiring captions or transcript and breaks the fail-on-fail relationship.

carlosapaduarte:
Thinking about how to address this:
- the SCs apply to pre-recorded media, and define that as non live media
- (technically) streaming is different from live media, i.e. we can stream media that has been pre-recorded 

So, I don't think we should use "non-streaming" in these rules. But, I'm not sure we can come up with a working objective definition of "live" 😕
We could say that the media resource is available in its entirety prior to the media player starts its playback. That is objective. But how can we test this?  Any ideas?

Jym77:
Good points…
"Available prior to start playing" feels like it's never going to be the case. Because indeed, most video platforms these days effectively stream or in any case start playing asap while downloading still happens (and stop downloading if playing stops), so unless we manage to say "available on the server" (which is even less testable…) that's bad (rule would be OK as a rule but if it never applies…)

What about adding to the composition: rule(s) applies to all videos and we add another atomic "video is non-live". Then we only need an unambiguous definition, but it can be subjective. And live video pass the rule by passing the new atomic while non-live video have to pass one of the old atomic rules…

We might add "known duration" to the applicability. That will not catch all live video, but still some. And thus makes the rule a bit more focused.

carlosapaduarte:
[...]

PR #1661, opened 2021-06-25
Replacing non-streaming audio with live audio

carlosapaduarte:
Updating '`audio` element content has text alternative' (#e7aa44) so that it no longer uses the non-streaming definition which was found to be inaccurate

I'm looking for comments on:
- The appropriateness of the expectation in the new atomic rule that checks if the audio content is live
- Suggestions on resources to use for the passed example

After getting this change approved I'll open a new PR changing the other rules that also use the non-streaming definition

Need for Final Call:

---

## How to Review And Approve


ajanec01:
I must've missed that part, what's inaccurate with "non-streaming"?

carlosapaduarte:
This: https://github.com/act-rules/act-rules.github.io/issues/1446#issue-701040347

_rules/audio-is-live-213x3x.md
```diff
@@ -0,0 +1,91 @@
...
+
+_No assumptions._
+
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [Understanding SC 1.2.1:Audio-only and Video-only (Prerecorded)](https://www.w3.org/WAI/WCAG21/Understanding/audio-only-and-video-only-prerecorded)
```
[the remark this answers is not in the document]
carlosapaduarte: I think I'll address this one after sorting out this rule.

_rules/audio-is-live-213x3x.md
```diff
@@ -0,0 +1,91 @@
...
+This rule applies to every `audio` element that is:
+
+- playing; or,
+- has a "play button" that is [visible][] and [included in the accessibility tree][].
+
+**Note:** A play button is an interactive element that when activated, plays the audio.
+
+## Expectation 
+
+The target element's [media resource](https://html.spec.whatwg.org/multipage/media.html#media-resource) is being broadcasted in real time without being first recorded and stored on the origin server.
```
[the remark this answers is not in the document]
carlosapaduarte: Changing the expectation.

_rules/audio-is-live-213x3x.md
```diff
@@ -0,0 +1,91 @@
...
+  authors:
+    - Carlos Duarte
+---
+
+## Applicability
+
+This rule applies to every `audio` element that is:
+
+- playing; or,
+- has a "play button" that is [visible][] and [included in the accessibility tree][].
```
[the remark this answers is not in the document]
carlosapaduarte: It sounds reasonable to have an atomic for all media, but I'll also address this after sorting this rule first.

_rules/audio-is-live-213x3x.md
```diff
@@ -19,13 +19,13 @@ acknowledgments:
...
-- has a "play button" that is [visible][] and [included in the accessibility tree][].
-
-**Note:** A play button is an interactive element that when activated, plays the audio.
+- has a [play button][] that is [visible][] and [included in the accessibility tree][].
 
 ## Expectation 
 
-The target element's [media resource](https://html.spec.whatwg.org/multipage/media.html#media-resource) is being broadcasted in real time without being first recorded and stored on the origin server.
+For each test target one of the following is true:
+- it is not possible to [seek][] the end of the target element's [media resource][]; or
```
Jym77, line 27: I like that 👍 
OTOH, I'm not sure it's possible to "not be possible to seek". The algorithm seems to always return (possibly after doing nothing). Would that rather be part of the second bullet (and so the first can be removed)?

Jym77 (dismissed):

_rules/audio-is-live-213x3x.md
carlosapaduarte: I believe you are correct. I've updated the expectation accordingly.

carlosapaduarte:
Updated rule, now with a working example.

_rules/audio-is-live-213x3x.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `audio` element plays a live media resource by streaming the audio from the device's microphone.
+
+This example requires a browser supporting the experimental MediaStreamTrack API. The browser must be authorized to access a microphone. You can successfully test this example on Google Chrome with the flag --enable-blink-features=WebCodecs,MediaStreamInsertableStreams
```
Jym77, line 48: I think we should (also) put that text inside the HTML page of the example itself.
That way, (manual) testers that navigate to the example's page (without necessarily going through the ACT rules page) would be able to see that text and act accordingly.

_rules/audio-is-live-213x3x.md
```diff
@@ -0,0 +1,123 @@
...
+
+- [Understanding SC 1.2.1:Audio-only and Video-only (Prerecorded)](https://www.w3.org/WAI/WCAG21/Understanding/audio-only-and-video-only-prerecorded)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `audio` element plays a live media resource by streaming the audio from the device's microphone.
```
Jym77, line 46: Also, have you tested that by "seeking" the end, to make sure this matches the expectation?
Wondering if the "seek" bit should be used also in examples descriptions 🤔

Jym77 (changes requested):

_rules/audio-is-live-213x3x.md
Jym77: This now seems related to #1679 since "give access to your mic" sounds a bit like a test instruction 🤔

carlosapaduarte: * Moved the note to the example so that manual testers are aware of the instructions
* Added a test instructions section after the description of the example

_rules/audio-is-live-213x3x.md
carlosapaduarte: I just found that MediaStream is not seekable! https://w3c.github.io/mediacapture-main/#mediastreams-in-media-elements 
I will add this to the expectation and the example's description.

carlosapaduarte: And I suppose this lands us again exactly where we were before... needing to find a live stream that we can use in the example 😞

_rules/audio-is-live-213x3x.md
```diff
@@ -0,0 +1,140 @@
...
+					.catch(function(err) {
+						log.innerHTML += err.name + ': ' + err.message
+					})
+			},
+			false
+		)
+	</script>
+</html>
+```
+
```
daniel-montalvo, line 99: I am assuming we are expanding these passed examples with real streaming audios/videos of broadcasting orgs, right?

carlosapaduarte: That was the main idea. 

As was also mentioned in TPAC, we might want to explore the possibility of W3C hosting a streaming server where we could serve the examples (we would just loop one of the videos we already use).

PR #1663, opened 2021-07-01
Correctly handle visibility in Hidden State

### LINKED ISSUE #1656: "Hidden state" doesn't handle reverting the `visibility`
[Hidden state](https://act-rules.github.io/glossary/#hidden-state) of an element is `true` if any of its ancestor has `visibility: hidden`.

This means that in
```html
<div style="visibility: hidden">
  Hello <span style="visibility: visible"><span>World</span></span>
</div>
```
Both `span` have a "hidden state" of `true` even though they are actually visible (and included in the accessibility tree (well not really as `span` are only containers, but that's the idea)).
Since `visibility` is inherited, that condition should only look at the element, not on its ancestors.

Also, the def should handle `visibility: collapse`.

Jym77:
Update "hidden state" to only look at `visibility` on the element itself, not its ancestors, because `visibility: hidden` can be overwritten by a later `visibility: visible` (opposite to `display: none` or `aria-hidden=true`).

Also rephrase to handle `visibility: collapse`.

Closes issue(s):

- closes #1656 

Need for Final Call:
This will require a 1 week Final Call (small bugfix in a definition)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/pages/glossary/hidden-state.md
+++ b/pages/glossary/hidden-state.md
@@ -8,15 +8,16 @@ input_aspects:
   - DOM tree
 ---
 
-An HTML element's _hidden state_ is "true" if at least one of the following is true for itself or any of its [ancestors][] in the [flat tree][]:
+An HTML element's _hidden state_ is "true" if either it has a [computed][] CSS property `visibility` whose value is not `visible`; or at least one of the following is true for any of its [inclusive ancestors][] in the [flat tree][]:
 
 - has a `hidden` attribute; or
 - has a [computed][] CSS property `display` of `none`; or
-- has a [computed][] CSS property `visibility` of `hidden`; or
 - has an `aria-hidden` attribute set to `true`
 
 In any other case, the element's _hidden state_ is "false".
 
-[ancestors]: https://dom.spec.whatwg.org/#concept-tree-ancestor 'Definition Ancestor'
+> **Note**: Contrarily to the other conditions, the `visibility` CSS property may be reverted by descendants.
+
+[inclusive ancestors]: https://dom.spec.whatwg.org/#concept-tree-inclusive-ancestor 'DOM Definition of Inclusive Ancestor'
 [computed]: https://www.w3.org/TR/css-cascade/#computed-value 'CSS definition of computed value'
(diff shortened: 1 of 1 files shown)


ajanec01 (approved):

pages/glossary/hidden-state.md
```diff
@@ -8,15 +8,16 @@ input_aspects:
...
 
 - has a `hidden` attribute; or
 - has a [computed][] CSS property `display` of `none`; or
-- has a [computed][] CSS property `visibility` of `hidden`; or
 - has an `aria-hidden` attribute set to `true`
 
 In any other case, the element's _hidden state_ is "false".
 
-[ancestors]: https://dom.spec.whatwg.org/#concept-tree-ancestor 'Definition Ancestor'
+> **Note**: Contrarily to the other conditions, the `visibility` CSS property may be reverted by descendants.
```
WilcoFiers, line 19:
```suggestion
**Note**: Contrarily to the other conditions, the `visibility` CSS property may be reverted by descendants.
```

Otherwise this will end up in a blockquote.

WilcoFiers (changes requested):
Editorial

daniel-montalvo (approved):

WilcoFiers (approved):

Jym77:
Call for Review ends on August 26, 2021.

dd8:
The `hidden` attribute is implemented as a user agent CSS rule like this in all modern browsers:
https://html.spec.whatwg.org/multipage/rendering.html#hiddenCSS

That means CSS can override `hidden` as mentioned here:
https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute

For example these buttons with a `hidden`  attribute are visible because author rules win cascade over user agent rules:

```
<style>
    button.big { display: block; }
</style>

<button hidden style="display:block" >Buy Now</button>

<button hidden class="big" >Buy Now</button>
```

These are visible in Chrome, Edge, Firefox, Safari and Internet Explorer.

Jym77:
@dd8 Yes, this has already been discussed and somehow seemed to have been dropped without us recording a clear conclusion 😢 #1523 
In any case, this is likely a topic for a separate PR.

PR #1669, opened 2021-07-20
Replace semantic widget/link definitions

WilcoFiers:
Part of #1593 was to add a single definition that makes it more explicit when a we mean the exact role, or any role that inherits. This adds a new definition "inheriting semantic role" which replaces "semantic widget" and "semantic link". I'll open a separate PR to change the definition of `semantic role` to allow phrases like "semantic button".

Need for Final Call: 1 week.

## How to Review And Approve


pages/glossary/inheriting-semantic-role.md
```diff
@@ -0,0 +1,17 @@
...
+title: Inheriting Semantic Role
+key: inheriting-semantic
+unambiguous: true
+objective: true
+input_aspects:
+  - Accessibility tree
+  - DOM tree
+---
+
+An element with a _inheriting semantic role_ of X is any element which has the [semantic role][] of X or a [semantic role][] that inherits from the X.
```
WilcoFiers, line 11: I'm considering excluding abstract roles from this.... What do people think?

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,16 +27,16 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any HTML element with the [semantic role][] of [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any HTML element with is a [semantic][semantic role] [rowheader][] or [columnheader][] for which all of the following is true:
```
Jym77, line 30: I feel this requires a precision in the def of "Semantic Role" that we can say "an element is a semantic foo" to mean "has a semantic role of foo".

Jym77 (changes requested):

tbostic32 (approved):

_rules/table-header-cell-has-assigned-cells-d0f69e.md
kherr-a11y, line 30: I don't have edit privileges yet, but "is a" - please remove the "is"

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -27,16 +27,16 @@ acknowledgments:
...
 - the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] with a [semantic role][] of either [table][] or [grid][] is [included in the accessibility tree][].
+- the element has at least one ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][]; and
+- the element's closest ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][] is [included in the accessibility tree][].
 
 ## Expectation
 
-Each target element is [assigned][] to at least one element with a [semantic role][] of [cell][] or inheriting from [cell][].
+Each target element is [assigned][] to at least one element this is an [inheriting semantic][] [cell][].
```
kherr-a11y, line 39: "this is" should be "with"

kherr-a11y (commented):
two small editorial changes suggested.  I do not have edit privileges.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
WilcoFiers, line 39:
```suggestion
Each target element is [assigned][] to at least one element with an [inheriting semantic][] [cell][].
```

_rules/table-header-cell-has-assigned-cells-d0f69e.md
daniel-montalvo:
```suggestion
This rule applies to any HTML element with a [semantic][semantic role] [rowheader][] or [columnheader][] for which all of the following is true:
```

Is that what you mean, @kherr-a11yb

Jym77: Solved in https://github.com/act-rules/act-rules.github.io/pull/1671/files#diff-a0b8349df87c854df34f5481b6c6a983ca594f8b4ea62347239579f80d6d4e0fR16

Jym77 (approved):

daniel-montalvo (approved):

PR #1671, opened 2021-07-20
Define that "semantic X" means "semantic role of X"

WilcoFiers:
Part of #1593, this describes that "semantic `button`" means an element with a semantic role of `button`.

Need for Final Call: 1 week

## How to Review And Approve

### DIFF
--- a/pages/glossary/semantic-role.md
+++ b/pages/glossary/semantic-role.md
@@ -13,7 +13,7 @@ The _semantic role_ of an element is determined by the first of these cases that
 1. **Explicit** If the element has an [explicit role][], then its _semantic role_ is its [explicit role][].
 1. **Implicit** The _semantic role_ of the element is its [implicit role][].
 
-This definition can be used in expressions such as "semantic `button`" meaning any element has a semantic role of `button`.
+This definition can be used in expressions such as "semantic `button`" meaning any element with a semantic role of `button`.
 
 #### Accessibility Support for Definition of Semantic Role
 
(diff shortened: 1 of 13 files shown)


tbostic32 (approved):

colabottles (approved):

kherr-a11y (approved):

WilcoFiers:
I'm going to hold off sending out a call for review until we have #1669 ready as well. I don't think this should be merged separately.

pages/glossary/semantic-role.md
```diff
@@ -13,6 +13,8 @@ The _semantic role_ of an element is determined by the first of these cases that
 1. **Explicit** If the element has an [explicit role][], then its _semantic role_ is its [explicit role][].
 1. **Implicit** The _semantic role_ of the element is its [implicit role][].
 
+This definition can be used in expressions such as "semantic `button`" meaning any element has a _semantic role_ of `button`.
```
Jym77, line 16:
```suggestion
This definition can be used in expressions such as "semantic `button`" meaning "any element with a _semantic role_ of `button`".
```

Jym77: I'm also not too keen on the italic on the non-defining instance of the term (`_semantic role_`) (because I'm reading the italic with the semantics "defining instance", maybe I'm reading too much into it…)
```suggestion
This definition can be used in expressions such as "semantic `button`" meaning any element has a semantic role of `button`.
```

Jym77 (changes requested):
Small editorial.

Jym77 (approved):

PR #1674, opened 2021-07-22
Update examples of included-in-the-accessibility-tree.md

### LINKED ISSUE #1581: Fix examples of "included in the accessibility tree"
[Examples of included in the accessibility tree](https://act-rules.github.io/pages/examples/included-in-the-accessibility-tree/) are not handling containers correctly.

Especially the first one, `<span>ACT rules</span>` incorrectly claims that the `span` element is included in the accessibility tree. The `span` element is not included because it's a container. Only the contained text node is included.

We should probably update all the examples to use elements with an implicit role.

kengdoj:
- Replaced `<span>` in most of the examples. 
- Changed the description of the last example since I still found the interactive element in the accessibility tree.

Closes issue(s):

- closes #1581

Need for Final Call: 
<< check https://act-rules.github.io/pages/design/process/#final-call-aka-call-for-consensus-cfc >>
This will not require a Final Call << choose reason(s): editorial changes (including to the applicability, expectation or examples section), changes to assumptions, background, accessibility support, change to website/test code (not rule), other (explain). >>
This will require a 1 week Final Call << small changes affecting a small number of test cases, if in doubt do not use this. >>
This will require a 2 weeks Final Call << new rule, or substantial changes affecting a large number of test cases, if in doubt, use this. >>

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


pages/examples/included-in-the-accessibility-tree.md
```diff
@@ -6,53 +6,53 @@ These are examples of the [Included in the accessibility tree][] definition. The
...
 ```
 
 ## Removed but not ignored
 
-Although the `span` element with an `id` of "label" is not itself included in the accessibility tree, it still provides an [accessible name][] to the other `span`, via the `aria-labelledby` attribute. Thus, it is still indirectly exposed to users of assistive technologies. Removing an element from the accessibility tree is not enough to remove all accessibility concerns from it since it can still be indirectly exposed.
+Although the `h3` element with an `id` of "label" is not itself included in the accessibility tree, it still provides an [accessible name][] to the `span`, via the `aria-labelledby` attribute. Thus, it is still indirectly exposed to users of assistive technologies. Removing an element from the accessibility tree is not enough to remove all accessibility concerns from it since it can still be indirectly exposed.
 
 ```html
-<span id="label" style="display:none">ACT rules</span>
+<h3 id="label" style="display:none">ACT rules</h3>
```
ajanec01, line 44:
```suggestion
<h2 id="label" style="display:none">ACT rules</h2>
```

pages/examples/included-in-the-accessibility-tree.md
```diff
@@ -6,53 +6,53 @@ These are examples of the [Included in the accessibility tree][] definition. The
...
 <span aria-labelledby="label">Accessibility Conformance Testing rules</span>
```
ajanec01, line 45:
```suggestion
<h3 aria-labelledby="label">Accessibility Conformance Testing rules</h3>
```
This produces weird interactions. VO announces "ACT Rules" and the "Accessibility Conformance Testing rules" as a grouping whereas NVDA and Narrator ignore completely the "ACT rules" text. Using an `h3` fixes that.

pages/examples/included-in-the-accessibility-tree.md
```diff
@@ -6,53 +6,53 @@ These are examples of the [Included in the accessibility tree][] definition. The
...
 
 ```html
-<span style="position: absolute; top:-9999em">ACT rules</span>
+<h3 style="position: absolute; top:-9999em">ACT rules</h3>
 ```
 
 ## Removed but not ignored
 
-Although the `span` element with an `id` of "label" is not itself included in the accessibility tree, it still provides an [accessible name][] to the other `span`, via the `aria-labelledby` attribute. Thus, it is still indirectly exposed to users of assistive technologies. Removing an element from the accessibility tree is not enough to remove all accessibility concerns from it since it can still be indirectly exposed.
+Although the `h3` element with an `id` of "label" is not itself included in the accessibility tree, it still provides an [accessible name][] to the `span`, via the `aria-labelledby` attribute. Thus, it is still indirectly exposed to users of assistive technologies. Removing an element from the accessibility tree is not enough to remove all accessibility concerns from it since it can still be indirectly exposed.
```
ajanec01, line 41:
```suggestion
Although the `h2` element with an `id` of "label" is not itself included in the accessibility tree, it still provides an [accessible name][] to the `h3`, via the `aria-labelledby` attribute. Thus, it is still indirectly exposed to users of assistive technologies. Removing an element from the accessibility tree is not enough to remove all accessibility concerns from it since it can still be indirectly exposed.
```

pages/examples/included-in-the-accessibility-tree.md
```diff
@@ -6,53 +6,53 @@ These are examples of the [Included in the accessibility tree][] definition. The
...
-<span id="label" style="display:none">ACT rules</span>
+<h3 id="label" style="display:none">ACT rules</h3>
 <span aria-labelledby="label">Accessibility Conformance Testing rules</span>
 ```
 
-## Removed but focusable
+## Focusable elements
 
-Although this `input` element is not included in the accessibility tree, it is still [focusable][], hence users of assistive technologies can still interact with it by sequential keyboard navigation. This may result in confusing situations for such users (and is in direct violation of [the fourth rule of ARIA (working draft)](https://www.w3.org/TR/using-aria/#fourth)).
+Using `role="presentation"` or `aria-hidden="true"` on a focusable element (or its ancestor) will not remove the element from the accessibility tree and may result in confusion for users of assistive technology. The `input` is still [focusable][], hence users of assistive technologies can still interact with it by sequential keyboard navigation. This is in direct violation of [the fourth rule of ARIA (working draft)](https://www.w3.org/TR/using-aria/#fourth).
```
ajanec01, line 50: Something like `<input role="presentation" />` would be in the accessibility tree as per the [presentational conflict resolution](https://www.w3.org/TR/wai-aria-1.2/#presentational-roles-conflict-resolution). Something like `<div role="presentation"><input/></div>` would also be included. We should keep the original text.

ajanec01 (changes requested):

pages/examples/included-in-the-accessibility-tree.md
kengdoj: The original text says 

> Although this `input` element is not included in the accessibility tree,..

but it would be included as you say. Why keep the original text?

ajanec01: The original example does not have `role="presentation"` in the description. It only mentions `aria-hidden`. They behave differently when it comes to focusable elements. As per the spec, elements with `role="presentation"` are included in the accessibility tree if they are focusable whereas elements with `aria-hidden` are not (although VO tends include those element). Essentially, a screen reader would not announce anything when element with `aria-hidden` receives focus but it would announce the role and name of an element with `role="presentation"`.

kengdoj: I am using the Edge inspection tool to view the Accessibility tree. Where I have `<h2 aria-hidden="true">`, it is shown as "ignored" in the acc tree. That is how I expected to see `<input aria-hidden="true">`, but it is shown as "textbox".

I see all of the following in the Acc tree as "textbox":  
- **`<input aria-hidden="true">`**
- **`<input>` as a child of `<span aria-hidden="true">`**
- `<input role="presentation">`
- `<input>` as a child of `<span role="presentation">`

acctree

ajanec01: Yes, Edge and Chrome expose focusable inputs with `aria-hidden="true" in the accessibility tree. Firefox doesn't (follows the spec). I think this description should stay as is. It's a common pitfall that can cause issues so people should know that adding `aria-hidden`  to a parent element can cause issues.

WilcoFiers: Yikes! So I did some testing on this, and it seems the accessibility tree Chrome Devtools shows is different from the accessibility tree I get from Puppeteer. My guess is that aria-hidden elements aren't really in the accessibility tree, but Devtools shows them anyway to make this easier for developers. But it might as well be the other way around and those elements are hidden from the Puppeteer AXTree snapshot.

I'm not entirely sure how to figure out which is which, or if it even matters.

kengdoj: Should we keep the original text and add a note like: 

Note: some browsers may include interactive elements with `aria-hidden="true"` in the accessibility tree.

pages/examples/included-in-the-accessibility-tree.md
```diff
@@ -6,53 +6,53 @@ These are examples of the [Included in the accessibility tree][] definition. The
...
+This `h3` element is not included in the accessibility tree because it is hidden to everybody by the CSS property.
 
 ```html
-<span style="display:none">ACT rules</span>
+<h3 style="display:none">ACT rules</h3>
 ```
 
 ## Removed with aria-hidden
 
 This `span` element is not included in the accessibility tree because it is explicitly removed by the `aria-hidden` attribute.
```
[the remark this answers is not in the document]
kengdoj: change made.

pages/examples/included-in-the-accessibility-tree.md
kengdoj: change made.

pages/examples/included-in-the-accessibility-tree.md
kengdoj: Kept the original example as discussed in today's ACT TF meeting.

We also discussed adding a note to the definition but looks like it is already in the [definition for included in accessibility tree](https://act-rules.github.io/glossary/#included-in-the-accessibility-tree):

> Note: Users of assistive technologies might still be able to interact with elements that are not included in the accessibility tree. An example of this is a focusable element with an aria-hidden attribute with a value of true. Such an element could still be interacted using sequential keyboard navigation regardless of the assistive technologies used, even though the element would not be included in the accessibility tree. [examples of included in the accessibility tree]: https://act-rules.github.io/pages/examples/included-in-the-accessibility-tree/

ajanec01 (approved):

WilcoFiers (approved):

Jym77 (approved):

PR #1679, opened 2021-08-03
Definition of test instructions

ajanec01:
Defining "Test instructions" for rules that require the tester to get a test case into a specific state.

Need for Final Call: none

## How to Review And Approve

### DIFF
--- a/pages/design/rule-design.md
+++ b/pages/design/rule-design.md
@@ -149,6 +149,8 @@ The description should:
 
 > _For example: "This page has a `title` element with content."_
 
+At times, the aplicability and the expectation of the rule may require analysing the before and after state of a web page. It happens when the rule applies to an element that changes some of its properties and the pass or failure condition can be evaluated only after the change of the state. 
+
 For a detailed description on what to write test cases for see [test case design](../test-cases). For more details, see [ACT Rules Format: Test Cases](https://www.w3.org/TR/act-rules-format/#test-cases).
 
 ## Listed conditions
--- a/pages/design/rule-design.md
+++ b/pages/design/rule-design.md
@@ -149,9 +149,20 @@ The description should:
 
 > _For example: "This page has a `title` element with content."_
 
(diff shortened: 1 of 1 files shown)


colabottles (approved):

colabottles (approved):

Jym77:
I'm also considering whether the test instructions should (also) be included in the example itself. So that a tester that navigates to the example's page (without going through the ACT rules page) can still see them and act accordingly.

See also https://github.com/act-rules/act-rules.github.io/pull/1661#discussion_r705136203 where giving access to the computer's microphone seems fairly similar to a test instruction.

pages/design/rule-design.md
```diff
@@ -151,6 +151,19 @@ The description should:
 
 For a detailed description on what to write test cases for see [test case design](../test-cases). For more details, see [ACT Rules Format: Test Cases](https://www.w3.org/TR/act-rules-format/#test-cases).
 
+### Test instructions
+
+At times, it may be necessary for the user to follow a set of instructions to get a test case into a state in which it can be evaluated against the expectation. It happens when the pass, failure, or inapplicability of a test case can only be determined after a change of the state of a test case that is triggered by the user's action and not the content itself. 
+
+Such instructions should be explicit to make sure that the desired state of a test case is obtained. The steps necessary to get the test case into the desired state need to be listed out. The instructions for evaluating the rule should not be included in the list of instructions to improve readability. For example, a test case description with a set of instructions could look like the following:
```
daniel-montalvo, line 158:
```suggestion
Explicit, step-by-step instructions should be provided to ensure the desired state of a test case is obtained. For example, a test case description with a set of instructions could look like the following:
```

> The instructions for evaluating the rule should not be included in the list of instructions to improve readability.

I can't relate this sentence to any of these, other than "Don't use these instructions to improve rule readability". This seems like good advice, but probably not that relevant. Maybe I am missing something...

daniel-montalvo (commented):
Hope these edits help clarify the message a bit.

I think the message is conveyed through my edits, but happy to further discuss if I missed something. Not feeling strongly on this.

HelenBurge (approved):

PR #1683, opened 2021-08-12
Update line-height-not-important-78fd32.md

tbostic32:
Removed extra "1.5" and changed failed example descriptions from using recommend -> required.

Need for Final Call:

This will require 2 week Final Call - Significant changes to the applicability, expectation, background, and some to the test cases. 

---

## How to Review And Approve


ajanec01 (dismissed):

_rules/line-height-not-important-78fd32.md
```diff
@@ -28,7 +28,7 @@ This rule applies to any HTML element that is [visible][] and for which the `sty
 For each test target, at least one of the following is true:
 
 - **not important**: the [computed][] value of its [line-height][] property is not [important][]; or
-- **large enough**: the [computed][] value of its [line-height][] property is not `normal`, and is at least `1.5` or 1.5 times the [computed][] value of its [font-size][] property; or
+- **large enough**: the [computed][] value of its [line-height][] property is not `normal`, and is at least 1.5 times the [computed][] value of its [font-size][] property; or
```
Jym77, line 31: It's actually not duplicated…
Computed `line-height` can be a unitless number, with the meaning "X times the `font-size`", or a length (with unit).

```
<p style="font-size: 16px; line-height: 10 !important">Foo</p>
```
has a **computed** `line-height` of `10`, which is not 1.5 times the `font-size` (`16px`) (notably because a unitless number is not comparable to a unit), yet it is nonetheless OK.

Maybe saying that the **used** value is 1.5 times the `font-size` is good (https://drafts.csswg.org/css-inline/#line-height-property seems down right now, so I can't really check that). But I remember we spent a lot of time chiselling the conditions in that rule, so I'd be very careful before making such a change.

Jym77 (changes requested):

_rules/line-height-not-important-78fd32.md
tbostic32: I see your point, but I think that the phrasing was confusing. I will checkout the link when it is back up, but in the meantime, could we say something like "at least the unitless constant `1.5` or 1.5 times...", something that disambiguates the first 1.5 from the second.

Jym77: Yes, I'm not saying it shouldn't be changed 😄
Improving the formulation is definitely possible, but we do need to keep something to handle the unitless numbers.

WilcoFiers (dismissed):

_rules/line-height-not-important-78fd32.md
```diff
@@ -21,32 +21,32 @@ acknowledgments:
...
 For each test target, at least one of the following is true:
 
 - **not important**: the [computed][] value of its [line-height][] property is not [important][]; or
-- **large enough**: the [computed][] value of its [line-height][] property is not `normal`, and is at least `1.5` or 1.5 times the [computed][] value of its [font-size][] property; or
+- **large enough**: the [used](https://www.w3.org/TR/css-cascade-4/#used) value of its [line-height][] property is at least 1.5 times the [computed][] value of its [font-size][] property; or
 - **cascade**: the [cascaded][] value of its [line-height][] property is not a value [declared][] in its `style` attribute.
 
 ## Assumptions
 
 - There is no mechanism available on the page to adjust [line-height][]. If there is such a mechanism, it is possible to fail this rule while [Success Criterion 1.4.12 Text Spacing][sc1412] is still satisfied.
```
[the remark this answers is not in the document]
WilcoFiers: I would suggest we do those two fixes separately.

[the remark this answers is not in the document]
tbostic32: I'm a bit confused by your wording, should I add an assumption for now or just disregard the issue for the moment and do it in another MR?

_rules/line-height-not-important-78fd32.md
```diff
@@ -21,32 +21,34 @@ acknowledgments:
...
 
-- This rule assumes that when the [computed][] value of the [line-height][] property is `normal`, user agents chose a [used][] value below 1.5. [CSS recommendation][line-height normal] is to have a [used][] value between 1.0 and 1.2, thus too small to satisfy [Success Criterion 1.4.12 Text Spacing][sc1412].
+- The font size is constant for all text in the element. If font-size changes (e.g., through use of the `::first-line` pseudo-element) then the required line height would also change throughout the element. This is untested by the current rule.
```
tbostic32, line 38: This is definitely a weird assumption. The problem is that with an HTML element being the applicable "unit" we don't really want to say that the line height must be the 1.5 times the size of the largest font-size in the element. This would be really weird if the first line is very big, but all of the following lines are small.

WilcoFiers: Yeah true. I'm not sure it makes much of a difference though.

_rules/line-height-not-important-78fd32.md
```diff
@@ -45,7 +47,7 @@ When a style is [declared][] in the `style` attribute with an [important][] decl
...
 
-This rule does not consider other style attributes that could be used to increase the distance between lines of text. For example, style attributes such as `position`, `padding`, and `margin` could be used to increase the distance between lines of text to meet `1.4.12`. Oppositely, those style attributes could also be used to reduce the distance between lines of text. Thus, it is possible to pass this rule, but still fail `1.4.12` due to other styling choices.
+This rule does not consider other style attributes that could be used to increase the distance between lines of text. For example, style attributes such as `position`, `padding`, and `margin` could be used to increase the distance between lines of text to meet [Success Criterion 1.4.12 Text Spacing][sc1412]. Oppositely, those style attributes could also be used to reduce the distance between lines of text. Thus, it is possible to pass this rule, but still fail [Success Criterion 1.4.12 Text Spacing][sc1412] due to other styling choices.
```
Jym77, line 50: I'm not sure this is (or was) really best located in "Accessibility Support". This doesn't sound like this concern UA/AT behaving differently 🤔 Should it be a normal Assumption instead?

_rules/line-height-not-important-78fd32.md
```diff
@@ -21,20 +21,22 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any HTML element with multi-line text that is [visible][] and for which the `style` attribute [declares][declared] the [line-height][] CSS property.
+This rule applies to any HTML element with [visible][] text that includes a [line-break](https://www.w3.org/TR/css-text-3/#line-breaking) and for which the `style` attribute [declares][declared] the [line-height][] CSS property.
```
Jym77, line 24: I'm a bit unsure about the "line break" (CSS) definition:
> When inline-level content is laid out into lines, it is broken across line boxes. Such a break is called a line break. When a line is broken due to explicit line-breaking controls (such as a preserved newline character), or due to the start or end of a block, it is a forced line break. When a line is broken due to content wrapping (i.e. when the UA creates unforced line breaks in order to fit the content within the measure), it is a soft wrap break. The process of breaking inline-level content into lines is called line breaking.

On one hand, it seems to only concern inline-level content, but on the other hand it mentions start/end of block as being one kind of line breaks. That puzzles me a bit 🤔 🧩 

Where it gets is to the following:
```html
<div style="line-height: 1px !important">
  <p>Hello</p>
  <p>World</p>
</div>
```
Does the `div` contain a line break or not? On one hand it has no (direct) inline content, but on the other hand the start/end of `p` is a start/end of block, hence a "force line break". Point is that this is not failing the SC (due do the paragraph spacing), and therefore shouldn't fail the rule; but if this contains a line break, then it fails the rule.

---
Maybe just using
```suggestion
This rule applies to any HTML element with [visible][] text that includes a [soft wrap break](https://www.w3.org/TR/css-text-3/#soft-wrap-break) and for which the `style` attribute [declares][declared] the [line-height][] CSS property.
```
will at least make sure we restrict the rule to line breaks created by UA for display reason, (not considering author line break). This is maybe not enough, but should ensure we don't have false positives.

Jym77 (changes requested):

_rules/line-height-not-important-78fd32.md
tbostic32: Understood, I was initially intending the example that you had above not to be applicable, but given the forced line break you pointed how I see how it could cause problems. I am fine going using soft-wrap as well.

_rules/line-height-not-important-78fd32.md
tbostic32: Moved this from background to assumption.

_rules/line-height-not-important-78fd32.md
```diff
@@ -181,17 +184,17 @@ The [computed][] value of the `line-height` property of this `p` element is **no
 
 #### Failed Example 1
 
-This `p` element has a [computed][] `line-height` equal to the font size, which is below the recommended minimum.
+This `p` element has a [computed][] `line-height` equal to the font size, which is below the required minimum.
```
Jym77, line 187: Since the rule now considers the used value and not the computed one, the descriptions should be updated accordingly.
Not sure all examples are in the PR, can you double check that no other description uses "computed" value?
```suggestion
This `p` element has a [used][] `line-height` equal to the font size, which is below the required minimum.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -21,32 +21,35 @@ acknowledgments:
...
 
 ## Background
 
 When a style is [declared][] in the `style` attribute with an [important][] declaration, it "wins" the [cascade sort] over any other style from [author origin][], i.e. it cannot be overridden by any of these. On the other hand, if such a style is [declared][] in a style sheet, it can still "lose" the [cascade sort][] to declarations with higher [specificity][] or simply coming from a later style sheet (such as ones injected by assistive technologies). This rule ensures that the element is not in the first case and that the style can be overridden by users, unless it is already at least the minimum recommended threshold. [Important][] styles that are declared with the [user][user origin] or [user agent][user agent origin] origin can win the [cascade sort][] over styles with the [author origin][].
 
 CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
 
+
```
Jym77, line 52: I'm wondering whether we should add a background note explaining why this rule considers the "used" value and not the "computed" one.
This is a slightly tricky thing (just consider the amount of discussion we had on it internally) and may deserve some explanation…

_rules/line-height-not-important-78fd32.md
```diff
@@ -112,7 +115,7 @@ This `p` element has a [computed][] `line-height` of `25.6px` (160% of `16px`) w
 This `p` element has a [computed][] `line-height` of `1.6` which is **large enough**.
```
Jym77, line 115:
```suggestion
This `p` element has a [used][] `line-height` of `1.6` which is **large enough**.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -181,17 +184,17 @@ The [computed][] value of the `line-height` property of this `p` element is **no
...
-<p style="line-height: 1em !important">
+<p style="line-height: 1em !important; max-width: 200px;">
 	The toy brought back fond memories of being lost in the rain forest.
 </p>
 ```
 
 #### Failed Example 2
 
-This `p` element has a [computed][] `line-height` of `20px`, which is below the recommended minimum given the specified font size is 20 pixels.
+This `p` element has a [computed][] `line-height` of `20px`, which is below the required minimum given the specified font size is 20 pixels.
```
Jym77, line 197:
```suggestion
This `p` element has a [used][] `line-height` of `20px`, which is below the required minimum given the specified font size is 20 pixels.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -216,37 +219,37 @@ This `p` element has a [computed][] `line-height` of `19.2px` (120% of `16px`) w
...
-<p style="line-height: 120% !important">
+<p style="line-height: 120% !important; max-width: 200px;">
 	The toy brought back fond memories of being lost in the rain forest.
 </p>
 ```
 
 #### Failed Example 4
 
-This `p` element has a [computed][] `line-height` of `1.2` which is below the recommended minimum.
+This `p` element has a [computed][] `line-height` of `1.2` which is below the required minimum.
```
Jym77, line 229:
```suggestion
This `p` element has a [used][] `line-height` of `1.2` which is below the required minimum.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -216,37 +219,37 @@ This `p` element has a [computed][] `line-height` of `19.2px` (120% of `16px`) w
...
-<p style="line-height: 1.2 !important">
+<p style="line-height: 1.2 !important; max-width: 200px;">
 	The toy brought back fond memories of being lost in the rain forest.
 </p>
 ```
 
 #### Failed Example 5
 
-This `p` element has a [computed][] `line-height` of `normal` which is below the recommended minimum ([used][] value is generally around 1.2).
+This `p` element has a [computed][] `line-height` of `normal` which is below the required minimum ([used][] value is generally around 1.2).
```
Jym77, line 239:
```suggestion
This `p` element has a [used][] `line-height` of `normal` which is below the required minimum ([used][] value is generally around 1.2).
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -216,37 +219,37 @@ This `p` element has a [computed][] `line-height` of `19.2px` (120% of `16px`) w
...
-<p style="line-height: normal !important">
+<p style="line-height: normal !important; max-width: 200px;">
 	The toy brought back fond memories of being lost in the rain forest.
 </p>
 ```
 
 #### Failed Example 6
 
-This `p` element has a [computed][] `line-height` of `normal` which is below the recommended minimum ([used][] value is generally around 1.2).
+This `p` element has a [computed][] `line-height` of `normal` which is below the required minimum ([used][] value is generally around 1.2).
```
Jym77, line 249:
```suggestion
This `p` element has a [used][] `line-height` of `normal` which is below the required minimum ([used][] value is generally around 1.2).
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -21,32 +21,35 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any HTML element that is [visible][] and for which the `style` attribute [declares][declared] the [line-height][] CSS property.
+This rule applies to any HTML element with [visible][] text that includes a [soft wrap break](https://www.w3.org/TR/css-text-3/#soft-wrap-break) and for which the `style` attribute [declares][declared] the [line-height][] CSS property.
```
Jym77, line 24: Should we have viewport width to test the line break?
(maybe extracting "multi-line text" into a separate definition)

Notably, Inapplicable example 3 does contain a line break if the viewport is small enough…

Jym77 (changes requested):

_rules/line-height-not-important-78fd32.md
tbostic32: I agree, given the amount of discussion we should document it for the future.

_rules/line-height-not-important-78fd32.md
tbostic32: I think I got most of the others, but I don't think this should be changed. Wouldn't `normal` be the `computed` value and 1.2 be the `used` value as indicated by the comment at the end "([used][] value is generally around 1.2)".

Jym77: I agree 👍

_rules/line-height-not-important-78fd32.md
tbostic32: Are you suggesting that we specify a view port size that should be used when testing with the rule? Is that something that has been done before, I tried to cheat through this by specifying widths for specific elements so it would automatically have a line break.

Agreed its trick with the inapplicable example, we can't really make an example like that without a line break because the viewport can always get smaller. I could add an `overflow: scroll` container around it, though thats basically an anti-pattern. E.g.,
```html
    <div style="overflow-x: scroll;">
      <p style="line-height: 1em !important; width: 1000px;">
      </p>
    </div>
```

tbostic32:
Looks like tests are failing because it doesn't like the word unitless (thinks its a misspelling) and the definition of `html element` is declared but not used.

Jym77:
> Looks like tests are failing because it doesn't like the word unitless (thinks its a misspelling) and the definition of `html element` is declared but not used.

The unused definitions can simply be removed from the definitions list at the bottom of the file (lines 307 and 315 it seems).
The spelling exception can be added to [this file](https://github.com/act-rules/act-rules.github.io/blob/develop/__tests__/spelling-ignore.yml).

_rules/line-height-not-important-78fd32.md
tbostic32: I did the above modification so that we have an inapplicable example where changing the view port size does not cause a line break to occur.

Jym77: I was more thinking something like 
"This rule applies to any element with text that includes a line break at viewport size of X by Y"
with a note, likely in Accessibility Support or Assumption, that "X by Y" is arbitrarily but based on some measurements used elsewhere in WCAG and stuff like that (and assumed that WCAG needs good line spacing at least at that viewport size).

We do have a similar trick in [Zoomed text node is not clipped with CSS overflow)(https://act-rules.github.io/rules/59br37).

_rules/line-height-not-important-78fd32.md
```diff
@@ -23,32 +23,36 @@ acknowledgments:
...
 
 CSS specifications define each declaration as being either [important][] (if is as the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
 
+In the Expectation of this rule, we evaluate the [used](https://www.w3.org/TR/css-cascade-4/#used) value of the [line-height][] property instead of the [computed][] value since the [used](https://www.w3.org/TR/css-cascade-4/#used) value is guaranteed to use absolute units (i.e., pixels). This allows us to easily determine if the 1.5 times threshold is met versus a [computed][] which may be a unitless number that is difficult to evaluate.
```
Jym77, line 54:
```suggestion
This rule evaluates the [used][] value of the [line-height][] property instead of its [computed][] value because the [used][] value is guaranteed to use absolute units (i.e., pixels). This streamlines comparison with the [computed][] [font-size][] which is also absolute. The [computed][] [line-height][] may be a unitless number that is harder to compare.
```

Jym77 (approved):
Wondering whether we should have a viewport specified in the Applicability, but I guess I'd like to see what other think about that.

The rest are polishing. So I'm approving now to let other review and give their opinion on the viewport question 👍

_rules/line-height-not-important-78fd32.md
tbostic32: I was hesitant to include an arbitrary port side which is why I decided to fix the inapplicable example. It makes me feel a bit better that there is precedent for using the viewport size in the CSS overflow rule. That rule hasn't been viewed by TF yet and I am expecting that this will raise a discussion there, which I plan to bring up this week to get ahead of the problem.

kherr-a11y (approved):

kengdoj (approved):

wcreedle (approved):

wcreedle (approved):

colabottles (approved):

WilcoFiers:
Call for review e-mail here: https://lists.w3.org/Archives/Public/public-act-r/2022Jan/0006.html

We forgot to label it. Apologies.

PR #1684, opened 2021-08-12
Update definition of included-in-the-accessibility-tree.md

kengdoj:
Moved focusable element with `aria-hidden="true"` out of note. Related to #1674. 

Need for Final Call: This will not require a Final Call - editorial

---

## How to Review And Approve


pages/glossary/included-in-the-accessibility-tree.md
```diff
@@ -11,7 +11,7 @@ Elements included in the accessibility tree of platform specific accessibility A
 
 The general rules for when elements are included in the accessibility tree are defined in the [core accessibility API mappings](https://www.w3.org/TR/core-aam/). For native markup languages, such as HTML and SVG, additional rules for when elements are included in the accessibility tree can be found in the [HTML accessibility API mappings (working draft)](https://www.w3.org/TR/html-aam/) and the [SVG accessibility API mappings (working draft)](https://www.w3.org/TR/svg-aam/).
 
-For more details, see [examples of included in the accessibility tree][].
+For more details, see [examples of included in the accessibility tree](https://act-rules.github.io/pages/examples/included-in-the-accessibility-tree/).
+
+Elements with `aria-hidden="true"` should not be included in the accessibility tree. However, some browsers include [focusable](#focusable) elements with `aria-hidden="true"`in the accessibility tree. In such a browser, a user of assistive technology could still interact with the element by using sequential keyboard navigation, even though the element should not be included in the accessibility tree.
```
ajanec01, line 16: My preference would be to Remove the "**Note**" word, leave the text as is and then on the bottom of that text to add a "**Note**" that would say that some browsers tend to expose focusable elements with `aria-hidden` but this behaviour is not consistent; therefore we consider them not exposed.

It's not a strong preference though, so it would be good to know what others think about it.

Jym77: I think it might be well in a "Background" section, like we have for [Block of repeated content](https://act-rules.github.io/glossary/#block-of-repeated-content) or [Instrument to achieve an objective](https://act-rules.github.io/glossary/#instrument-to-achieve-an-objective).

WilcoFiers, line 16: Maybe like this?

```suggestion
Some browser will leave [focusable][] elements with a [hidden state][] of `true` in the accessibility tree. Because they are hidden, these elements are considered **not** included in the accessibility tree.

**Note:** Users of assistive technologies might still be able to interact with elements that are not included in the accessibility tree. An example of this is a [focusable][] element with an `aria-hidden` attribute with a value of `true`. Such an element could still be interacted using sequential keyboard navigation regardless of the assistive technologies used, even though the element would not be included in the accessibility tree.

For more details, see [examples of included in the accessibility tree][].

[focusable]: #focusable
[hidden-state]: #hidden-state
[examples of included in the accessibility tree]: https://act-rules.github.io/pages/examples/included-in-the-accessibility-tree/
```

WilcoFiers (changes requested):

WilcoFiers (dismissed):

ajanec01 (dismissed):

carlosapaduarte:
Let's wait for the merging of https://github.com/act-rules/act-rules.github.io/pull/1699 which is already approved. Then we should replace "[hidden state](#hidden-state) of `true`" with "[programmatically hidden](#programmatically-hidden)" because the earlier definition will no longer be supported.

Jym77:
> Let's wait for the merging of #1699 which is already approved. Then we should replace "[hidden state](#hidden-state) of `true`" with "[programmatically hidden](#programmatically-hidden)" because the earlier definition will no longer be supported.

It has been merged, so we can go on with the changes here 😄

pages/glossary/included-in-the-accessibility-tree.md
```diff
@@ -7,11 +7,13 @@ input_aspects:
...
 
 For more details, see [examples of included in the accessibility tree][].
 
-**Note:** Users of assistive technologies might still be able to interact with elements that are not included in the accessibility tree. An example of this is a [focusable](#focusable) element with an `aria-hidden` attribute with a value of `true`. Such an element could still be interacted using sequential keyboard navigation regardless of the assistive technologies used, even though the element would not be included in the accessibility tree.
+Elements with a [hidden state](#hidden-state) of `true` are removed from the accessibility tree. However, some browsers will leave [focusable](#focusable) elements with an `aria-hidden` attribute set to `true` in the accessibility tree. Because they are hidden, these elements are considered **not** included in the accessibility tree. This may cause confusion for users of assistive technologies because they may still be able to interact with these focusable elements using sequential keyboard navigation, even though the element should not be included in the accessibility tree.
```
carlosapaduarte, line 16:
```suggestion
[Programmatically hidden](#programmatically-hidden) elements are removed from the accessibility tree. However, some browsers will leave [focusable](#focusable) elements with an `aria-hidden` attribute set to `true` in the accessibility tree. Because they are hidden, these elements are considered **not** included in the accessibility tree. This may cause confusion for users of assistive technologies because they may still be able to interact with these focusable elements using sequential keyboard navigation, even though the element should not be included in the accessibility tree. 
```

carlosapaduarte (dismissed):
Update to use the new "programmatically hidden" definition

carlosapaduarte (approved):

WilcoFiers (approved):

ajanec01 (approved):

PR #1694, opened 2021-08-23
Add "funding" to acknowledgments

WilcoFiers:
This PR adds a "funding" field to all rules developed under the WAI-Tools project. This way we can indicate rules that were not created as part of WAI-Tools, and may have other sources of funding. This PR is part of migrating the ACT website content to the WAI website.

---

## How to Review And Approve


Jym77 (approved):

carlosapaduarte (approved):
Wondering if we should have a link somewhere to each funding entry in the list (similar to the tool links from the implementation reports)

WilcoFiers:
@carlosapaduarte We'll do so when these are rendered. Each of them takes some customisation anyway, so I'm not putting that into each rule. Same way we don't include links to author profiles into the rule but pull them from somewhere. Just makes updating easier.

PR #1700, opened 2021-08-31
Context role- updating to aria 1.2 and removing PE 5

ajanec01:
Making sure that the [required context rule](https://act-rules.github.io/rules/ff89c9) is ready when ARIA 1.2 becomes a recommendation.

Also, As I've been going through the examples, I've noticed that in [Passed Example 5](https://act-rules.github.io/rules/ff89c9#passed-example-5) the `<div role="navigation">` is included in the accessibility tree, well, sort of. It looks like the accessibility tree is malformed by this example. The element does not have an accessible role and it is not a parent of the two list items but it is a child of the list element (see screenshot below). Technically a failure of the [required owned elements rule](https://act-rules.github.io/rules/bc4a75) but not a failure of ARIA. I don't think such an example should be included as a passed example.
Malformed-tree

Also, closing issue #1703 

Need for Final Call: 1 week.

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -163,7 +163,7 @@ These elements with an [explicit role][] of `listitem` are not children in the [
 
 ```html
 <div role="list">
-	<div aria-label="busy">
+	<div aria-busy="true">
 		<div role="listitem">List item 1</div>
 		<div role="listitem">List item 2</div>
 	</div>
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -159,11 +159,11 @@ These elements with an [explicit role][] of `listitem` are not children in the [
 
 #### Failed Example 3
 
-These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][]. They are instead children in the [accessibility tree][] of the `div` with an `aria-label` attribute; even though this `div` has no role, it has a global ARIA attribute and is thus [included in the accessibility tree][].
+These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][]. They are instead children in the [accessibility tree][] of the `div` with an `aria-busy` attribute; even though this `div` has no role, it has a global ARIA attribute and is thus [included in the accessibility tree][].
 
 ```html
 <div role="list">
-	<div aria-label="menu">
+	<div aria-label="busy">
 		<div role="listitem">List item 1</div>
 		<div role="listitem">List item 2</div>
 	</div>
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -109,6 +109,19 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 5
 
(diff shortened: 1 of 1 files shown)


colabottles (approved):

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -109,19 +109,6 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
...
-These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its DOM children. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
-
-```html
-<div role="list" aria-owns="item1 item2">
-	<div role="navigation">
-		<div id="item1" role="listitem">List item 1</div>
-		<div id="item2" role="listitem">List item 2</div>
-	</div>
-</div>
-```
```
Jym77, line 121: I think this example is important to show how `aria-owns` twists the accessibility tree and that the rule cares about this (rather than the DOM tree). OTOH, I do agree that the resulting accessibility tree is "bad" due to the `<div role="navigation">` being a child of the list. maybe replace the example with:
```html
<div role="list" aria-owns="item1 item2">
</div>
<div role="navigation">
	<div id="item1" role="listitem">List item 1</div>
	<div id="item2" role="listitem">List item 2</div>
</div>
```
Which is arguably more artificial, but would pass the other rule 🤔

Jym77 (changes requested):

_rules/aria-required-context-role-ff89c9.md
ajanec01: I see your point. I think that the alternative you gave would be better since it's a passed example.

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -159,11 +159,11 @@ These elements with an [explicit role][] of `listitem` are not children in the [
 
 #### Failed Example 3
 
-These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][]. They are instead children in the [accessibility tree][] of the `div` with an `aria-label` attribute; even though this `div` has no role, it has a global ARIA attribute and is thus [included in the accessibility tree][].
+These elements with an [explicit role][] of `listitem` are not children in the [accessibility tree][] of an element with their [required context role][]. They are instead children in the [accessibility tree][] of the `div` with an `aria-busy` attribute; even though this `div` has no role, it has a global ARIA attribute and is thus [included in the accessibility tree][].
 
 ```html
 <div role="list">
-	<div aria-label="menu">
+	<div aria-label="busy">
```
tombrunet, line 166: If this is for #1703, I believe the intent was `<div aria-busy="true">` or `<div aria-busy="false">` rather than `<div aria-label="busy">`

ajanec01: yes, it was :face_palm: Thanks!

tombrunet (approved):

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -109,6 +109,19 @@ These elements with an [explicit role][] of `listitem` are children in the [acce
 
 #### Passed Example 5
 
+These elements with an [explicit role][] of `listitem` are children in the [accessibility tree][] of an element with their [required context role][] even though they are not its DOM children. The `aria-owns` attribute is used to alter the accessibility tree and place the target elements in their [required context role](https://www.w3.org/TR/wai-aria-1.1/#scope).
```
Jym77: Link should point to ARIA 1.2 😀

Jym77 (changes requested):
Typo.

colabottles (approved):

carlosapaduarte (approved):

HelenBurge (approved):

PR #1715, opened 2021-09-21
Add namespaced element references

### LINKED ISSUE #459: Define SVG elements and HTML elements

WilcoFiers:
Add more links to the namespaced element definition. 

- closes #459

---

## How to Review And Approve

### DIFF
--- a/_rules/menuitem-non-empty-name-m6b1q3.md
+++ b/_rules/menuitem-non-empty-name-m6b1q3.md
@@ -25,7 +25,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to HTML elements that is a [semantic][semantic role] `menuitem` [included in the accessibility tree][].
+This rule applies to [HTML elements][] that is a [semantic][semantic role] `menuitem` [included in the accessibility tree][].
 
 ## Expectation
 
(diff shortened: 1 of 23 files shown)


HelenBurge (approved):

_rules/menuitem-non-empty-name-m6b1q3.md
```diff
@@ -163,3 +163,4 @@ This element with the `menuitem` role does not need an [accessible name][] becau
 [included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of included in the accessibility tree'
 [presentational roles conflict resolution]: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none 'Presentational Roles Conflict Resolution'
 [semantic role]: #semantic-role 'Definition of Semantic Role'
+[html element]: #namespaced-element
```
WilcoFiers, line 166:
```suggestion
[html elements]: #namespaced-element
```

HelenBurge (approved):

Jym77 (approved):

carlosapaduarte (approved):

PR #1717, opened 2021-09-23
"Image button has non-empty accessible name" [59796f]: Fail default name ("Submit Query")

### LINKED ISSUE #1457: "Image button has non-empty accessible name" (59796f) does not consider default names and all Failed Examples are actually passing…
[Image button has non-empty accessible name](https://act-rules.github.io/rules/59796f) ignores that `<input type="image">` have a default accessible name of "Submit Query" according to the HTML AAM (https://www.w3.org/TR/html-aam-1.0/#input-type-image step 4).

Thus, the 3 Failed examples actually trigger this default name and result in image buttons with a name of "Submit Query", passing the rule.
(I am actually not sure it is possible to fail the rule in its current state, due to the default name).

In a similar case, for [Button has non-empty accessible name (Passed Example 7)](https://act-rules.github.io/rules/97a4e1#passed-example-7) we did take the default name into account (PR #566 after issue #565).

Jym77:
Explicitly fails the default name of image buttons. Explicitly accept default names of reset and submit buttons.

Closes issue(s):

- closes #1457

Need for Call for Review:
This will require a 1 week Call for Review (small change to expectation)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -59,6 +59,8 @@ _There are currently no assumptions._
 
 ## Background
 
+Some elements have a role of `button` and a default accessible name defined by the [HTML Accessibility API Mapping][html aam input button], for example `input` elements whose `type` [attribute value][] is either `submit` or `reset`. This rule consider that these default names can be descriptive and therefore does not fail them.
```
tbostic32, line 62: I think the statement - `This rule consider that the default names can be descriptive and therefore does not fail them.`  I think should be included in the `Assumptions` section as a counter part to the assumption in the `image-button` rule.

tbostic32:
```suggestion
Some elements have a role of `button` and a default accessible name defined by the [HTML Accessibility API Mapping][html aam input button], for example `input` elements whose `type` [attribute value][] is either `submit` or `reset`. This rule considers that these default names can be descriptive and therefore does not fail them.
```

tbostic32, line 62: I'm confused why this change is being made to `image-non-empty-accessible-name` instead of being made to `Button has non-empty accessible name`. I don't think this rule references any buttons. 

I think the statement `This rule considers that these default names can be descriptive and therefore does not fail them.` should be added as an assumption to `button has non-empty accessible name` as a counter part to the assumption in `image-button has non-empty accessible name`.

tbostic32 (dismissed):
Small editorial and I think the second change may have been made to the wrong rule.

_rules/button-non-empty-accessible-name-97a4e1.md
```diff
@@ -42,6 +40,8 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Background
 
+Some elements have a role of `button` and a default accessible name defined by the [HTML Accessibility API Mapping][html aam input button], for example `input` elements whose `type` [attribute value][] is either `submit` or `reset`. This rule considers that these default names can be descriptive and therefore does not fail them.
```
tbostic32, line 43: Only thought is that this statement could still be moved into the assumption, i.e., we assume that if the default name is used then it is intended and descriptive for the situation.

tbostic32 (approved):

WilcoFiers (approved):
Slight preference for Trevor's suggestion, but no objection to how you did this.

daniel-montalvo (approved):
I can live with this as is, but I also think this is more an assumption than a background comment really.

Jym77:
Call for Review ends on November 11th.

colabottles (approved):

JenniferChadwick (approved):

Jym77:

PR #1728, opened 2021-09-30
Update to cross-reference Element with aria-hidden has no focusable content

kherr-a11y:

---

## How to Review And Approve


carlosapaduarte (commented):
Hi @kherr-a11y 
What is the reasoning supporting this change?

WilcoFiers:
@carlosapaduarte This was an action on the ACT TF. We felt these two rules are closely related, so that we should cross-link between them.

WilcoFiers (approved):

kherr-a11y:
Incorrect edit made.  Will create new pull request with correct edit.

PR #1739, opened 2021-10-28
xml:lang matches, remove incorrect inapplicable examples

WilcoFiers:
Inapplicable example 5 shouldn't be inapplicable. Inapplicable example 9 and 10 seem to me like they should fail the success criterion.

---

## How to Review And Approve


kengdoj (commented):
These look ok to me. The examples have invalid lang tags so should be inapplicable. The rule's [Applicability](https://act-rules.github.io/rules/5b7ae0#applicability) is to html element that has a lang attribute that has a valid language tag;

wcreedle (approved):

kherr-a11y (approved):

Jym77 (approved):
Agreed. These 3 examples fail [HTML page lang attribute has valid language tag](https://act-rules.github.io/rules/bf051a) and therefore 3.1.1 and shouldn't be here.

PR #1743, opened 2021-11-17
scrollable-element-keyboard: inapplicable example 5 started failing

WilcoFiers:
Chrome 96 changed how it padding-right interacts with scrollbar. In previous versions, and still in all other major browser it would pretend right padding does not exist for the purpose of scrolling. Effectively just making the box wider by the amount of padding-right that it has, and pretending that padding-right is 0. That's no longer the case, and so this inapplicable example was suddenly a failure.

I've updated this example so that it functions consistently again in all major browsers.

**Edge 95**

Inapplicable example 5, text can not be scrolled off screen in Edge 95

**Chrome 96**

Inapplicable example 5, text scrolled off screen in Chrome 96

Here's a link to the example so you can see for yourself:
https://act-rules.github.io/testcases/0ssw9k/9646f97144271b3e95b6648f5814085ddaa8ee08.html

Need for Call for Review: 1 week, **but will be merged first due to the urgency of it**

---

## How to Review And Approve

### DIFF
--- a/_rules/scrollable-element-keyboard-accessible-0ssw9k.md
+++ b/_rules/scrollable-element-keyboard-accessible-0ssw9k.md
@@ -49,6 +49,8 @@ This rule assumes that all [scrollable elements][scrollable] with visible conten
 
 Some browsers will automatically make any [scrollable element][scrollable] focusable to ensure keyboard accessibility. However, the browser does not include these elements in [sequential focus navigation][] when it has a negative number as a tabindex [attribute value][].
 
+There are also differences between browsers in if they make right padding scrollable. While some browsers allow scrolling the full border-width of the content of a scrollable region, others constrain scrolling to the content box instead.
+
 ## Background
 
 To ensure there is some element from which arrow keys can be used to control the scroll position, focus must be on or in a scrollable region. If scripts are used to prevent the keyboard events from reaching the scrollable region, this could still cause a keyboard accessibility issue. This must be tested separately.
@@ -226,7 +228,7 @@ This [scrollable][] `section` element has no [visible][] content.
 This `section` element has a [horizontal scroll distance][scrollable] that is less than its horizontal [padding][], and [vertical scroll distance][scrollable] that is less than its vertical [padding][].
 
 ```html
-<section style="height: 210px; width: 500px; overflow: scroll; padding: 30px;">
+<section style="height: 210px; width: 530px; overflow: scroll; padding: 30px 0 30px 30px;">
 	<div role="heading" aria-level="1">WCAG 2.1 Abstract</div>
 	<div style="width: 520px">
 		Web Content Accessibility Guidelines (WCAG) 2.1 covers a wide range of recommendations for making Web content more
--- a/_rules/scrollable-element-keyboard-accessible-0ssw9k.md
+++ b/_rules/scrollable-element-keyboard-accessible-0ssw9k.md
@@ -43,7 +43,7 @@ Each test target is either included in [sequential focus navigation][] or has a 
 
 ## Assumptions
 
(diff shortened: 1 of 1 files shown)


_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -49,6 +49,8 @@ This rule assumes that all [scrollable elements][scrollable] with visible conten
 
 Some browsers will automatically make any [scrollable element][scrollable] focusable to ensure keyboard accessibility. However, the browser does not include these elements in [sequential focus navigation][] when it has a negative number as a tabindex [attribute value][].
 
+There are also differences between browsers in if they make right padding scrollable. While some browsers allow scrolling the full border-width of the content of a scrollable region, others constrain scrolling to the content box instead.
```
Jym77, line 52:
```suggestion
Some browsers restrict scrolling to the [content box](https://drafts.csswg.org/css-box-4/#content-box) of elements; while others allow to scroll the full [border box](https://drafts.csswg.org/css-box-4/#border-box), hence including the element's padding. This results in some elements being scrollable with a browser but not with another.
```

carlosapaduarte (approved):

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -226,7 +228,7 @@ This [scrollable][] `section` element has no [visible][] content.
 This `section` element has a [horizontal scroll distance][scrollable] that is less than its horizontal [padding][], and [vertical scroll distance][scrollable] that is less than its vertical [padding][].
 
 ```html
-<section style="height: 210px; width: 500px; overflow: scroll; padding: 30px;">
+<section style="height: 210px; width: 530px; overflow: scroll; padding: 30px 0 30px 30px;">
 	<div role="heading" aria-level="1">WCAG 2.1 Abstract</div>
 	<div style="width: 520px">
```
Jym77, line 233: I'm not sure the example illustrates the same thing now.
The old example is great at seeing if tools take `padding` into consideration: without `padding` it looks like a `520px` element inside a `500px` one and requires scrolling, with padding the inner element is less wide. No, the inner element is always less wide than the outer, and the example doesn't differentiate tools on whether they use `padding` or not.

Jym77 (changes requested):
I'm afraid the rabbit hole goes deeper 😕 

* It seems that it is the definition of [Scrollable elements](https://act-rules.github.io/rules/0ssw9k#scrollable-element) that is now browser dependent, not just this rule. I assume because `scrollWidth` (or `clientWidth`?) is now browser dependent. We should record the information at the correct place. Note that [`scrollWidth`](https://www.w3.org/TR/cssom-view/#dom-element-scrollwidth) goes to 
   > Return the width of the element’s [scrolling area](https://www.w3.org/TR/cssom-view/#scrolling-area). 
   
   in turn leading to
   > The right-most edge of the element’s right padding edge 

   So I would argue that Chrome is actually correct in scrolling through all the padding 🤔 (in any case, it is worth checking why they decided to change the value of either `scrollWidth` or `clientWidth`: either there has been some recent change in CSS and we can expect everybody to follow suit shortly, or something weird is happening).

* The old example is now code that is Inapplicable on some browsers, but Failed on others. But then, as I understand when the page loads (on Chrome), all the text is nonetheless visible (it is only some padding area that is "hidden"). So, even without any possibility to scroll it, nothing would be missing and I am not sure that it would fail 2.1.1 Keyboard. 

   In this case, it means that this code is actually a false positive on Chrome (failing the rule, not the SC), which is bad.

Jym77 (approved):

HelenBurge (approved):

PR #1745, opened 2021-11-18
Add note on designing inapplicable test cases

### LINKED ISSUE #1689: Add explanation on designing inapplicable test cases in test case document
When reviewing #1675, it took me some time to remember why we want inapplicable examples to be "bad but don't matter" 🤔 

Essentially, Inapplicable and Passed are almost interchangeable in mapping. In order to have Inapplicable examples pertinent to test implementation (and detect incorrect tools), we want Inapplicable examples that "if you don't handle the reason why this is inapplicable you'll incorrectly fail that case", aka "this looks bad but actually is in one of the case the rule doesn't look at".

I think it might be good to add some guidance in [ACT Rules Test Case Design](https://act-rules.github.io/pages/design/test-cases/) about that. Especially since it seems counter-intuitive with the [Consistency With Accessibility Requirements](https://act-rules.github.io/pages/design/test-cases/#consistency-with-accessibility-requirements) requirement that Inapplicable cases satisfy the SC (`<input autocomplete="badname" disabled/>` looks bad at first until you realise it is disabled, that's why it's a good Inapplicable example, but also not obvious it had to be done that way).

Jym77:
Add a paragraph to designing document explaining why Inapplicable test cases should "look bad" to be good…

Closes issue(s):

- closes #1689 

Need for Call for Review:
This will not require a Call for Review (change to process document)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


WilcoFiers (approved):

colabottles (approved):

pages/design/test-cases.md
```diff
@@ -41,6 +41,8 @@ In order for the mapping of implementations to the rules to work, test cases nee
...
 
+Because mapping of implementations makes little difference between inapplicable and passed results, it is often better to have an inapplicable test case that "looks bad" but is actually satisfying the accessibility requirements. That way, tools that don't implement the Applicability correctly would fail this test case. For example, when testing [1.4.3 Contrast (minimum)](https://www.w3.org/TR/WCAG21/#contrast-minimum), a rule may be inapplicable to disabled button (since this matches the second exception in the Success Criterion); an inapplicable test case for this condition is better if it includes a disabled button whose text has a 2.0:1 contrast ratio with its background, than if it includes a disabled button with a 5.0:1 contrast ratio (in this second case, a tool that doesn't ignore disabled buttons would pass the test case and can still claim a correct implementation of the rule). Depending on the part of Applicability which a case tests, it is not always possible to have an inapplicable example that "looks bad".
```
carlosapaduarte, line 44:
```suggestion
Because mapping of implementations makes little difference between inapplicable and passed results, it is often better to have an inapplicable test case that "looks bad" but is actually satisfying the accessibility requirements. That way, tools that don't implement the Applicability correctly would fail this test case. For example, when testing [1.4.3 Contrast (minimum)](https://www.w3.org/TR/WCAG21/#contrast-minimum), a rule may be inapplicable to disabled buttons (since this matches the second exception in the Success Criterion); an inapplicable test case for this condition is better if it includes a disabled button whose text has a 2.0:1 contrast ratio with its background, than if it includes a disabled button with a 5.0:1 contrast ratio (in this second case, a tool that doesn't ignore disabled buttons would pass the test case and can still claim a correct implementation of the rule). Depending on the part of Applicability which a case tests, it is not always possible to have an inapplicable example that "looks bad".
```

carlosapaduarte (changes requested):
Small editorial change

carlosapaduarte (approved):

wcreedle (approved):

kherr-a11y (approved):

PR #1746, opened 2021-11-18
"Element language" rules: handle whitespace

### LINKED ISSUE #1693: "Element with lang attribute has valid language tag" and "HTML element language subtag matches language" [de46e4, off6ek] should ignore whitespace only text
```
<body><div lang="invalid"> <span lang="en">Hello</span> </div></body>
```
[Element with lang attribute has valid language tag](https://act-rules.github.io/rules/de46e4) fails this because there are two non-empty text nodes within the `div`, but these are whitespace and don't create problem.

[HTML element language subtag matches language](https://act-rules.github.io/rules/off6ek) is a bit mixed up there, these whitespace text nodes do not have any [most common language](https://act-rules.github.io/rules/off6ek#most-common-element-language) due to having no words, which make the expectation fails (`invalid` is not a most common language).

Suggestions (choose one):
1. In both rules, replace the last Applicability bullet from "non-empty text" to "non-whitespace text"
    **Problem**: we're still running into the same issue with corner cases of non-text "text" (e.g. smiley, punctuation, …); OTOH I'm not sure how this is announced… But at least `<span lang="en">😀</span>` doesn't have a most common language and therefore fails [HTML element language subtag matches language](https://act-rules.github.io/rules/off6ek). (yes, it's an Unicode smiley inside that text node, also works with an ASCII art one like `:-)`)
2. In both rules, update Expectation to say something like "either the element [on which the attribute is declared] has no most common language, or [current check]"
    **Problem**: this makes [Element with lang attribute has valid language tag](https://act-rules.github.io/rules/de46e4) much harder to fully automate since it needs to know the most common language. I sorta assume that tools will default to heuristic like "non-whitespace text". Depending on how strict we build our tests cases, this may result in having less implementations for the rule 😢 
3. Update Applicability of [Element with lang attribute has valid language tag](https://act-rules.github.io/rules/de46e4) to reject whitespace text nodes; and Expectation of [HTML element language subtag matches language](https://act-rules.github.io/rules/off6ek) to explicitly pass if there are no most common language.
    **Problem?**: [Element with lang attribute has valid language tag](https://act-rules.github.io/rules/de46e4) would still fail `<span lang="invalid">😀</span>` (or `<span lang="invalid">:-)</span>`). But maybe that is OK because the emojis (or symbols) are actually announced and therefore screen readers need a language to pick (i.e. "smiling face" is not the same as "visage souriant", so having an invalid `lang` attribute on such a `span` is indeed a real problem) 🤔

Jym77:
On both rules about elements' `lang`:
* make whitespace only text inapplicable
* add an assumption that the text contain human language (clears up corner cases of text which would not have a "most common language" and similar stuff).

Closes issue(s):

- closes #1693 

Need for Call for Review:
This will require a 1 week Call for Review (updating Applicability)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -325,11 +327,25 @@ The `lang` [attribute value][] of this `p` element is a [grandfathered tag][gran
 </html>
 ```
 
+#### Inapplicable Example 8
+
+The [text inheriting its programmatic language][] from this `div` element it only [whitespace][].
+
+```html
+<html>
```
WilcoFiers, line 335: I don't think SC 3.1.1 allows this TBH. There is no "except if the page is empty" or "except if the page has no default language" exception in that SC. The other rule I agree with, but I don't think we should do the same to this rule.

WilcoFiers (changes requested):

_rules/element-lang-matches-default-language-off6ek.md
WilcoFiers, line 335: Needs a lang attribute :)

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -325,11 +327,25 @@ The `lang` [attribute value][] of this `p` element is a [grandfathered tag][gran
 </html>
 ```
 
+#### Inapplicable Example 8
+
+The [text inheriting its programmatic language][] from this `div` element it only [whitespace][].
```
carlosapaduarte, line 332:
```suggestion
The [text inheriting its programmatic language][] from this `div` element is only [whitespace][].
```

_rules/element-lang-valid-de46e4.md
```diff
@@ -322,15 +324,30 @@ There is no [text inheriting its programmatic language][] from this `div` elemen
 </html>
 ```
 
+#### Inapplicable Example 5
+
+The [text inheriting its programmatic language][] from this `div` element it only [whitespace][].
```
carlosapaduarte, line 329:
```suggestion
The [text inheriting its programmatic language][] from this `div` element is only [whitespace][].
```

carlosapaduarte (changes requested):
Found a typo :-)

carlosapaduarte (approved):

_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -325,11 +327,25 @@ The `lang` [attribute value][] of this `p` element is a [grandfathered tag][gran
 </html>
 ```
 
+#### Inapplicable Example 8
+
+The [text inheriting its programmatic language][] from this `div` element is only [whitespace][].
```
WilcoFiers, line 332: I'm not following... the div is empty. It has no whitespace. Did something go wrong with Prettier or something?

WilcoFiers (changes requested):

_rules/element-lang-matches-default-language-off6ek.md
Jym77: Yes, Prettier seems to remove them 😢 
Do you have an idea how to circumvent that?

(my current best would be to add some JS to do `foo.innerText = "    "` but that seems a bit too crazy…)

WilcoFiers (approved):

kherr-a11y (approved):

Jym77:
Call for Review ends on February 10.

colabottles (approved):

Jym77:
Call for Review has ended, merging.

PR #1748, opened 2021-11-21
Remove 2 WCAG references:  1.3.1 and 4.1.2

HelenBurge:
Removed 1.3.1 and 4.1.2 as feel this applies to 2.4.3 more (a single checkpoint)

Asked to make sure there is only 1 SC dependency and this rule fits more to 2.4.3 than 1.3.1 and 4.1.2.

Closes issue(s):

N/A

Need for Call for Review:

### DIFF
--- a/_rules/aria-hidden-no-focusable-content-6cfa84.md
+++ b/_rules/aria-hidden-no-focusable-content-6cfa84.md
@@ -5,12 +5,7 @@ rule_type: atomic
 description: |
   This rule checks that elements with an `aria-hidden` attribute do not contain focusable elements.
 accessibility_requirements:
-  wcag20:1.3.1: # Info and Relationships (A)
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
-  wcag20:4.1.2: # Name, Role, Value (A)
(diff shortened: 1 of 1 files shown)


carlosapaduarte:
I don't agree this impacts 2.4.3. What this is checking is that something that is in the sequential focus navigation is not aria-hidden. It won't have an impact on the focus order (because it is not taken out of the sequential focus navigation), so it should not impact 2.4.3.

On the other hand, I maintain that it fails 4.1.2 (you won't be able to programatically determine the role of the element if it is aria-hidden) and 1.3.1 (because it will prevent programatic access to information of focusable elements).

carlosapaduarte (changes requested):
https://github.com/act-rules/act-rules.github.io/pull/1748#issuecomment-1000158326

HelenBurge:
I am in 2 minds, I think 2.4.3 should fail here as it is about the focusable order, not the DOM order, and this rule is testing that child elements of a hidden attribute are not in the focus order, or that aria-hidden is not used on a parent item of something that should be in the focus order.

However, the state that relates to 4.1.2 can also be valid here. So I will leave this for more thoughts before taking any action.

HelenBurge:
Carlos is right

PR #1749, opened 2021-11-21
Remove 2nd WCAG reference: 1.4.9

HelenBurge:
Removed the 2nd WCAG criteria that is AAA as the next level of compliance for this rule (no exceptions for the logo etc. This might be valid to stay in though.

Closes issue(s):

N/A

Need for Call for Review:


tbostic32:
I think we discussed in the meeting we should probably keep these, though there is an ongoing discussion around having passed examples that fail AAA criterion.

HelenBurge:
Agree with @tbostic32

PR #1751, opened 2021-11-21
Remove 2 WCAG references: 2.4.4 and 2.4.9

HelenBurge:
Removed the link specific checkpoints 2.4.4 and 2.4.9 as this is a name issue identified in 4.1.2.

Closes issue(s):

N/A

Need for Call for Review:


tbostic32:
Could you elaborate a bit more? I'm leaning towards keeping these due to [F89](https://www.w3.org/WAI/WCAG21/Techniques/failures/F89)

HelenBurge:
I think this one can be rejected, as was only trying to whittle it down.

PR #1754, opened 2021-11-21
Remove 2 WCAG references: 2.2.4 and 3.2.5

HelenBurge:
Removed the AAA checkpoints as this was designed for 2.2.1

Closes issue(s):

N/A

Need for Call for Review:

### DIFF
--- a/_rules/meta-refresh-no-delay-bc659a.md
+++ b/_rules/meta-refresh-no-delay-bc659a.md
@@ -10,16 +10,6 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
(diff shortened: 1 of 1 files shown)


tbostic32:
Was this one of the cases where you were trying to pick the best ones to remove? I think that this rule is basically testing [F41](https://www.w3.org/TR/WCAG-TECHS/F41.html) directly, so I think we could keep 2.2.4 and 3.2.5.

Seeing the second assumption in here as opposed to the other rule without the exception, I am wondering if a second assumption that the `refresh was essential` is needed in order for 2.2.1 to not apply.

HelenBurge:
Yep - closing this now

PR #1755, opened 2021-11-21
Remove 2nd WCAG reference: 3.2.5

HelenBurge:
Removed 3.2.5 as the no exceptions part is more for the 2.2.4 but it is a grey area/hard to decide here.

Closes issue(s):

N/A

Need for Call for Review:


tbostic32:
I would disagree with removing 3.2.5. In [3.2.5s understanding doc](https://www.w3.org/WAI/WCAG21/Understanding/change-on-request.html) it has a specific example about the page refreshing. I do agree with keeping 2.2.4.

HelenBurge:
Yes - closing this now.

PR #1759, opened 2021-11-21
Remove 2 WCAG references: 1.2.5 and 1.2.8

HelenBurge:
Removed the AAA checkpoints for the dependencies as the A covers this test well.

Closes issue(s):

N/A

Need for Call for Review:


WilcoFiers:
TF decided not to make this change.

PR #1764, opened 2021-11-24
Update programmatically-hidden.md

### LINKED ISSUE #1698: Hidden state / Programmatically hidden doesn't handle collision between `hidden` and `display: none`
Our definition of [Programmatically Hidden](https://act-rules.github.io/glossary/#programmatically-hidden) contains (roughly):
> has a `hidden` attribute or `display: none`

However, most (all?) browsers implement `hidden` as `display: none` which means that it can actually be overriden (`<p hidden style="display: block">Foo</p>` is actually not hidden.

How can we handle that in the definition?

1. Remove the bit about `hidden` attribute (keep only `display: none`), add an assumption that `hidden` is handled by `display: none`.
    **Problem**: that's an assumption we need to copy to every rule using this definition.
    **Problem**: if we end up with UA that don't do that, the def is broken for them…
2. Leave as is, add an Accessibility Support that "if `hidden` is handled as `display: none`, the `hidden` condition should not be used.
    **Problem**: that sounds pretty wonky and doesn't really help.
3. Replace `hidden` by something like "`hidden` and not visible".
    **Problem**: something like `<p hidden style="display: block; position: absolute; left: -9999em">` would match but would actually be exposed to ATs, so precisely the thing this definition tries to catch…

See also:
#1523 and https://github.com/act-rules/act-rules.github.io/pull/1663#issuecomment-903672856

carlosapaduarte:
Update the [programmatically hidden](https://act-rules.github.io/glossary/#programmatically-hidden) definition following the [CG discussion](https://www.w3.org/2021/11/11-act-r-minutes.html#item06)

Closes issue(s): 

- closes #1698 

Need for Call for Review:

---

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


pages/glossary/programmatically-hidden.md
```diff
@@ -10,12 +10,13 @@ input_aspects:
 
 An HTML element is _programmatically hidden_ if either it has a [computed][] CSS property `visibility` whose value is not `visible`; or at least one of the following is true for any of its [inclusive ancestors][] in the [flat tree][]:
 
-- has a `hidden` attribute; or
 - has a [computed][] CSS property `display` of `none`; or
 - has an `aria-hidden` attribute set to `true`
 
 **Note**: Contrarily to the other conditions, the `visibility` CSS property may be reverted by descendants.
 
+**Note**: The [HTML standard suggests](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements) rendering elements with the `hidden` attribute with a CSS rule that applies the value `none` to the CSS property `display` of the element. Although the suggestion is not normative, know user agents render it according to the suggestion. If a user agent does not follow the suggestion, this definition will not apply in the user agent's context.
```
Jym77, line 18:
```suggestion
**Note**: The [HTML standard suggests](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements) rendering elements with the `hidden` attribute with a CSS rule that applies the value `none` to the CSS property `display` of the element. Although the suggestion is not normative, know user agents render it according to the suggestion. If a user agent does not follow the suggestion, this definition may produce incorrect results for this user agent.
```
making it a bit less dramatic.

Jym77 (approved):
Small tweaking, feel free to ignore.

pages/glossary/programmatically-hidden.md
WilcoFiers, line 18: I would like to know what AT does with `hidden style="display:block"`. I think we should test this scenario, just to be safe.

carlosapaduarte: A `style` attribute setting the value of `display` overrules the `hidden` attribute. Irrespective of the CSS being declared as external, internal or inline. I've clarified that in the note.

WilcoFiers (approved):

kengdoj (approved):

colabottles (approved):

carlosapaduarte:
Call for review ended

PR #1765, opened 2021-11-30
Include ancestor generating a block container in the programatically determined link context definition

### LINKED ISSUE #1572: "Programmatically Determined Link Context" is too restrictive, causing problem in "Links with identical accessible names and context serve equivalent purpose" (fd3a94)
Consider:
```html
<div>Lorem ipsum <a href="lorem">Read more</a></div>
<div>Dolor sit amet<a href="dolor">Read more</a></div>
```
The `<div>` are ignored in the [Programmatically Determined Link Context](https://act-rules.github.io/glossary/#programmatically-determined-link-context). Therefore, both links have the same (empty) context but lead to different resources, hence the rule fails. This is not correct since the content of the `div` can reasonably be considered context.

Proposed fix: consider elements with text node children as part of the context. This should remove the "structural" `<div>` but keep the meaningful ones. Might need a bit more tweak to avoid
```html
<div>Lorem ipsum <a href="lorem">Read more</a></div>
<div>Lorem <span>ipsum</span> <a href="ipsum">Read more</a></div>
```
being two contexts because of the `<span>`.

carlosapaduarte:
Update the programatically determined link context definition to consider situations similar to the described in #1572 
Change was to also consider the parent node if it contains text nodes. Only the parent node, not ancestor nodes, is considered because of the impossibility to determine what node would enclose the "sentence" the link is in.

Closes issue(s):

- closes #1572 

Need for Call for Review:

---

### **When merging a PR:**

## How to Review And Approve


Jym77:
I've seen quite a bit of case where links were styled by an enclosing element 🙈 :
```html
<style>a { text-decoration: none }</style>
<div>Lorem <b><a href="foo">link</a></b></div>
<div>Ipsum <b><a href="bar">link</a></b></div>
```
These would still be considered as same context by the rule (because the `div` is not the parent and thus not part of the context) and thus incorrectly fail the rule…

The other way around,
```html
<div>Lorem <span>Ipsum</span><a href="foo">link</a></div>
<div>Lorem Ipsum<a href="bar">link</a></div>
```
Would be considered as different context and pass the rule even though they should arguably fail it. (a lesser problem).

---
Maybe there is something to dig in the direction of [inner text](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute), the idea would be that the context would be a block of inner text without newline, so maybe in the line of "furthest ancestor whose inner text does not contain a newline (except at start or end)" Or maybe simply "closest ancestor whose `display` is `block`" (or the correct version of it)?

---
There is also a tweak with hidden text that should or not be part of the context… (text that is transparent, or absolutely positioned out of viewport is still part of the inner text).

---
Sorry, not sure I'm helping much 😅

WilcoFiers (dismissed):

pages/glossary/programmatically-determined-link-context.md
```diff
@@ -12,6 +12,7 @@ The _programmatically determined context_ of a link (or _programmatically determ
 
 - being an [ancestor][] of the link in the [flat tree][] with a [semantic role][] of `listitem`; or
 - being the closest [ancestor][] of the link in the [flat tree][] that is a `p` element; or
+- being the [parent][] of the link in the [flat tree][] if it contains [text nodes][] that are not contained by the link; or
```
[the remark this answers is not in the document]
Jym77: Currently, the rule passes
```html
<p><a href="#lorem">Read more</a> about Lorem</p>
<p><a href="#ipsum">Read more</a> about Ipsum</p>
```
but fails
```html
<div><a href="#lorem">Read more</a> about Lorem</div>
<div><a href="#ipsum">Read more</a> about Ipsum</div>
```

I'm having a hard time to believe that the later violate 2.4.4 (especially if we say that the former satisfies it).

It is true that H77, H78, H79, H81 list some elements (and there is also a [WCAG def of programatically determined link context](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determined-link-context), but they are also all englobed into [G53: Identifying the purpose of a link using link text combined with the text of the enclosing sentence](https://www.w3.org/WAI/WCAG21/Techniques/general/G53). I would argue that the later example passes G53.

Jym77:
@carlosapaduarte Also noticing:
* Technique [H80](https://www.w3.org/WAI/WCAG21/Techniques/html/H80) accepts preceding heading to clarify link name, which we don't. It seems the rule fails the first example of H80 and shouldn't…
* When moving to ARIA 1.2, we should probably replace "`p` element" by "role of `paragraph`" in the definition.

Note sure whether we want to take these in this PR or separate ones.

carlosapaduarte:
> * Technique [H80](https://www.w3.org/WAI/WCAG21/Techniques/html/H80) accepts preceding heading to clarify link name, which we don't. It seems the rule fails the first example of H80 and shouldn't…

Indeed... I'll add the preceding heading to the programmatically determined context definition.

> * When moving to ARIA 1.2, we should probably replace "`p` element" by "role of `paragraph`" in the definition.

This one I agree it's best to tackle in a different PR (when we start addressing ARIA 1.2)

pages/glossary/programmatically-determined-link-context.md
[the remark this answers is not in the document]
carlosapaduarte: Using @Jym77's suggestion of updating this to use display:block
I believe this to be a better match for the semantics of paragraph in the WCAG definition of programatically determined link context.

Jym77 (changes requested):
We should add test cases with the new kinds of context.

carlosapaduarte:
> We should add test cases with the new kinds of context.

Test cases added

pages/glossary/programmatically-determined-link-context.md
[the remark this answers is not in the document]
WilcoFiers: Again I should say, why is it any block? I think this is somewhat speculative, and may be going outside of the intent of the SC. I think it would be good to see if AG could weigh in on this. I've certainly never tested it how this is being suggested, and would fail things that this rule would now pass.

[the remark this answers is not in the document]
carlosapaduarte: That would be false negatives for the rule, which we are sort of okay with, right? 
But considering the examples from Jean-Yves, wouldn't you pass some examples that the rule is currently failing?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -376,11 +376,44 @@ These two HTML `a` elements have the same [accessible name][] and link to the [s
...
+```
+
+#### Inapplicable Example 7
+
+These two HTML `a` elements have the same [accessible name][] but different [programmatically determined link contexts][programmatically determined link context] because the `div` elements place them in different display blocks.
+
+```html
+<div><a href="https://www.w3.org/WAI/">Read more</a> about the W3C WAI</div>
+<div><a href="https://www.w3.org/International/">Read more</a> about the W3C internationalization</div>
+```
```
Jym77, line 415: A bit sidetracking, but I feel like this ties to #1784 🤔 
Notably, if we restrict the rule to "same" (which is implied by Inapplicable Example 7 description), then the headings are unnecessarily in Inapplicable Example 6.
Which also sort of implies (backtracking my previous review 🙈 ) that the rule might not need to consider headings since I'm not sure that links can be in the **same** context but with different heading (except maybe with some heavy CSS  wrangling to avoid creating new blocks before and after headings).

pages/glossary/programmatically-determined-link-context.md
```diff
@@ -12,16 +12,20 @@ The _programmatically determined context_ of a link (or _programmatically determ
 
 - being an [ancestor][] of the link in the [flat tree][] with a [semantic role][] of `listitem`; or
 - being the closest [ancestor][] of the link in the [flat tree][] that is a `p` element; or
```
Jym77, line 14: That one might be removable in favour of the next one.
Even if `p` are styled with `display: line`, I'd say they might not be enough to  be a context by themselves and need to be wrapped in the closest block 🤔

Jym77 (dismissed):

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
Jym77: But see also https://github.com/act-rules/act-rules.github.io/issues/1784#issuecomment-1029046782
I'm not sure any more where I stand on this issue 🙃

carlosapaduarte: @jym I'm going to remove headings from the programatically determined context. We can't say that WCAG supports it. 
From WCAG 2.0 to WCAG 2.1, [headings were removed from F63](https://github.com/w3c/wcag/pull/832). 
WCAG2.0 - https://www.w3.org/TR/WCAG20-TECHS/F63.html
WCAG2.1 - https://www.w3.org/WAI/WCAG21/Techniques/failures/F63
Further, H80 moved from sufficient to advisory. So, all the evidence points against headings being part of programatically determined context.

pages/glossary/programmatically-determined-link-context.md
carlosapaduarte: Agreed.

Jym77 (approved):

HelenBurge (approved):

tbostic32 (approved):

colabottles (approved):

carlosapaduarte:
Call for review ended. Merging.

PR #1775, opened 2021-12-27
Updated examples to ensure tests are case insensitive in valid lang tag rules

### LINKED ISSUE #1499: Add test case with different capitalization in valid language tag rules (de46e4 and bf051a)
I have found a bug in an implementation of the rule "Element with lang attribute has valid language tag" (de46e4) that has to do with the implementation only accepting a certain capitalization in language tags. 
According to the implementation report, this tool had a correct implementation, so I suggest adding a new test case with different capitalization (e.g. lang="EN") to ensure consistency in implementations in this area.

The same applies to the rule "HTML page lang attribute has valid language tag" (bf051a).

carlosapaduarte:
…subtag are done in a case insensitive manner

Closes issue(s):

- closes #1499 

Need for Call for Review:

---

## How to Review And Approve


Jym77 (dismissed):
Not fully sure whether we should instead add new passed example (with appropriate description pointing on case insensitivity).
Mostly because these examples probably already exist for a reason (at least I hope) and that they now have 2 reasons: the old one and the case insensitivity; which we normally frown upon.

carlosapaduarte:
The definition of valid language tag already includes this:
> Subtags, notably the primary language subtag, are case insensitive. Hence comparison with the language subtag registry must be done in a case insensitive way.

Do you really think that we need to have tests exclusively checking this?

Jym77 (approved):

WilcoFiers (approved):
@carlosapaduarte can you change the PR title of this, so we end up with a complete message in the changelog?

daniel-montalvo (approved):

carlosapaduarte:
Call for review ends on February 18

colabottles (approved):

carlosapaduarte:
Call for review ended. Merging

PR #1744, opened 2021-11-18
Use `dfn` elements for defining terms

Jym77:
Showcasing how we could use `dfn` elements for defining terms in definitions and conditional lists.
Some of the changes are not really needed but used to illustrate some possibilities 😄 

In order to ensure name uniqueness (the definitions are included inside the rules), I namespace the IDs in definition with the definition key, as far as I understand the `:` in an ID shouldn't be a problem. (otherwise we can just namespace with `_`). IDs in rules are namespaced by the rule ID.

Pros:
* Semantically correct 😁 
* Can link to the precise part of the definition that is of interest.
* Can backlink to a precise condition in a conditional list.

Closes issue(s):

- closes #1664 

Need for Call for Review:
This will require a 1 week Call for Review (should be purely cosmetic, but could be touching a lot of stuff).

---

### **When creating PR:**

- [X] Make sure you're requesting to **pull a branch** (right side) to the `develop` branch (left side).
- [X] Make sure you do not remove the "How to Review and Approve" section in your pull request description

### **After creating PR:**

- [X] Add yourself (and co-authors) as "Assignees" for PR.
- [X] Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- [X] [Link the PR to any issue it solves](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue). This will be done automatically by [referencing the issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) at the top of this comment in the indicated place.
- [x] Optionally request feedback from anyone in particular by assigning them as "Reviewers".

### **When merging a PR:**

- [ ] Close any issue that the PR resolves. This will happen automatically upon merging if the PR was correctly linked to the issue, e.g. by referencing the issue at the top of this comment.

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Call for Review period. In case of disagreement, the longer period wins.

PR #1779, opened 2022-01-12
Survey amendments iframe-not-focusable-has-no-interactive-content-akn7bn

kherr-a11y:
Removed the reference to tabindex="-`" from Assumptions, per ACT meeting survey discussion.
Removed second sentence from Expectation, per ACT meeting survey discussion
Added new pass example where iframe is hidden (not visible)

## How to Review And Approve


kherr-a11y:
@tbostic32 can you give this a first review?

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,11 +28,11 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+This rule applies to any non-focusable `iframe` element that has focusable content.
```
tbostic32, line 31: Are you changing to non-focusable because its more clear or is there a scenario that the negative tab-index doesn't catch? Asking because I do somewhat like how specific the negative tab-index is. 

Also, you will probably need to include a reference to [focusable](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/glossary/focusable.md) here. You should be able to add it by doing `[focusable][]`

tbostic32: You will probably also need to change the description of the rule to mirror this.

tbostic32 (changes requested):

colabottles (approved):

HelenBurge (approved):
Agree with Trevor's feedback but nothing to add myself

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77, line 31: I'm also not sure about adding "that has focusable content" in the Applicability. Skipping over the (lack of) definition of "has", this means that most Passed Example (e.g. Passed Example 1) are now Inapplicable, since the `iframe` has no focusable content 🤔

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,11 +28,11 @@ acknowledgments:
...
 
 ## Expectation
 
-For each test target, the [nested browsing context][] does not include elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
+For each test target, the [nested browsing context][] does not contain elements that are [visible][] and part of the [sequential focus navigation][].
```
Jym77, line 35: Getting into technical nerd mode here also 🤓 
A [browsing context container](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-container) (i.e. `iframe`) "instantiate" a "browsing context" which is called "nested" because the `iframe` itself is already inside another browsing context (typically the browser window).

Now, the (nested) [browsing context](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context) only has a [container document](https://html.spec.whatwg.org/multipage/browsers.html#bc-container-document). In turn, this document contains elements (or rather "own" them). 

As much as I agree that the simplified text proposed here makes sense at the human level, it is not technically correct and I'm not sure it is sufficiently unambiguous to be used in an ACT rule (first draft of the rule only said "include" without the precision, but that was deemed unclear at the time: https://github.com/act-rules/act-rules.github.io/pull/1515#discussion_r542392991)

I have no problem replacing "include" by "contain", but I'd rather keep the precision in some form or other. To make the Expectation look lighter, we could move the second sentence to a new line/paragraph (possibly with a link after #1744  has landed):
```markdown
For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].

An element is <dfn id="#akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
```
or fully move it to a new definition.

Jym77 (changes requested):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,11 +28,13 @@ acknowledgments:
...
 
-This rule applies to any `iframe` element that has a negative number as a `tabindex` [attribute value][].
+This rule applies to any non-focusable `iframe` element that has [focusable][] content.
 
 ## Expectation
 
-For each test target, the [nested browsing context][] does not include elements that are [visible][] and part of the [sequential focus navigation][]. An element is "included" in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
+For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
+
+An element is <dfn id="#akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
```
Jym77, line 37:
```suggestion
An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
```

Jym77 (changes requested):

Jym77 (approved):

tbostic32 (approved):

PR #1785, opened 2022-01-21
Hide content using left rather than top positioning

### LINKED ISSUE #1776: Consider not hiding content with `top`
We have a few rules that are making content invisible with `position: absolute; top: -9999px`.

As far as I'm reading, using `top` is not a very good practice because (i) if used on an element high enough, or whose offset ancestor is low enough, it won't hide it fully;  and (ii) if used on focusable content, it may causes the page to jump back and forth when trying to scroll to the element.

None of these problems occurs in our examples because they are so small. Nonetheless, we might want to not show this poor practice and instead hide stuff with `left: -9999px` which is safer. Notably because people might consider what we write as examples to follow.

Jym77:
Use `position: absolute; left: -9999px` to visually hide content, rather than `top: …` which is not so good practice.

Closes issue(s):

- closes #1776

Need for Call for Review:
This will require a 1 week Call for Review (superficially changing some test cases)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


tbostic32 (approved):

carlosapaduarte (approved):

WilcoFiers (changes requested):
I'm not quite understanding why this is necessary. An element that's off screen is off screen. If it's off the top, the implementor needs to check that the content is far enough off the top that it's actually hidden. If it's off to the left, same thing.

Even if left is a better technique in CSS for content authors to use, shouldn't implementors know to consider both of them? Not immediately against this, but I'd like to understand it better before it goes forward.

Jym77:
> Not immediately against this, but I'd like to understand it better before it goes forward.

It is a matter of showcasing better practice since our code might be used as example of "things to do".

WilcoFiers:
@Jym77 Should we put this on the agenda?

Jym77:
[CG call summary](https://www.w3.org/2022/03/24-act-r-minutes.html#item05): we do not care about having best practices example only; our examples are meant to exercice corner cases. Therefore we close this PR (and issue).
We should add some more (in)visibility examples; possibly sprinkling some over all rules.

PR #1790, opened 2022-02-04
New rule: 'Widget has non empty accessible name' (rdzs6q)

### LINKED ISSUE #1374: Add menuitem and button to form-control-acc-name? [e086e5]
Question from [@jnurthen](https://github.com/act-rules/act-rules.github.io/pull/1357/files#r450379608):

> I question why menuitemcheckbox and menuitemradio appear here and menuitem does not.

And a related question from myself:

> Fair Similarly I think it's worth asking why this include spinbutton but not button.

It makes logical sense to say a menuitem is not a form control, but a checkbox in a menu is, but is there any particular reason that that rule be limited to form controls, e.g. controls that have a value.

carlosapaduarte:
Replace

- [Button has non-empty accessible name](https://act-rules.github.io/rules/97a4e1)
- [Form field has non-empty accessible name](https://act-rules.github.io/rules/e086e5)
- [Image button has non-empty accessible name](https://act-rules.github.io/rules/59796f)
- [Link has non-empty accessible name](https://act-rules.github.io/rules/c487ae)
- [Menuitem has non-empty accessible name](https://act-rules.github.io/rules/m6b1q3)

with a single widget rule, following the CG recommendation (https://github.com/act-rules/act-rules.github.io/issues/1374#issuecomment-916170574)

This will also help to streamline the 4.1.2 Understanding Document by having just a single rule checking widgets

Closes issue(s):

- closes #1374 

--- 

## Need for Call for Review:

## How to Review And Approve


_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Assumptions
+
+_There are currently no assumptions_
+
+## Accessibility Support
+
+- Certain assistive technologies can be set up to ignore the title attribute, which means that to some users the title attribute will not act as an [accessible name][].
+- Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have one of the applicable [semantic roles][] and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+
```
Jym77, line 64: We should probably keep the second point of https://www.w3.org/WAI/standards-guidelines/act/rules/button-non-empty-accessible-name-97a4e1/#accessibilitysupport explaining why we accept "submit" and "reset", especially since that rule is deprecated.

Jym77 (changes requested):

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,377 @@
...
+    - Image used in passed example 14 is courtesy of NASA/JPL-Caltech.
+htmlHintIgnore:
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'alt-require'
+---
+
+## Applicability
+
+This rule applies to any element that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has a [semantic role][] that is one of the following [semantic roles][]:
```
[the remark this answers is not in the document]
carlosapaduarte: Nice catch. I'm also adding the required accessibility support note about `doc-*` attributes.

_rules/widget-non-empty-accessible-name-rdzs6q.md
carlosapaduarte: I'm wandering... given that we are not deprecating the "image button" rule, do we need to fail image buttons that have the default name here? Or is it enough to fail them in the "image button" rule?
This rule would just check the "non-emptyness" of the accessible name, without judging whether the default names are appropriate or not...

Jym77: That's a fair point.
`<input type="image" />` will have an accessible name of "Submit Query" and fails both 1.1.1 and 4.1.2 via "Image button has non-empty accessible name". Having it pass this rule creates a false negative for 4.1.2 but makes the rule so much simpler that it is totally worth it (vs "accessible name is not empty nor the default name of image button" which would be awkward).

Then I guess just a background note explaining this (and why one rule accepts "Submit Query" while the other rejects it).

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -61,6 +61,8 @@ _There are currently no assumptions_
 
 - Certain assistive technologies can be set up to ignore the title attribute, which means that to some users the title attribute will not act as an [accessible name][].
 - Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have one of the applicable [semantic roles][] and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+- Accessibility support for some elements inheriting the semantic role of `link` (e.g. elements with `doc-*` attributes) may vary depending on the assistive technology in use.
```
Jym77, line 64:
```suggestion
- Accessibility support for some elements inheriting the semantic role of `link` (e.g. elements with `doc-*` roles) may vary depending on the assistive technology in use.
```
maybe mention/link to the DPUB ARIA spec?

Jym77 (changes requested):
Looks good, small tweak plus maybe adding note to sum up the image button discussion.

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,382 @@
...
+
+This element with a `button` role has an [accessible name][] because of its `aria-label` attribute.
+
+```html
+<span role="button" aria-label="My button"></span>
+```
+
+#### Passed Example 11
+
+This `input` element has an [accessible name][] because of the default accessible name for an `input` element with a `type` attribute set to `reset`.
```
Jym77, line 190:
```suggestion
This `input` element has an [accessible name][] because of the default accessible name for an `input` element with a `type` [attribute value][] of `reset`.
```

Jym77 (approved):

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,382 @@
...
+
+```html
+<div id="country">Country</div>
+<textarea aria-labelledby="country"></textarea>
+```
+
+#### Passed Example 5
+
+This `input` element has an [accessible name][] because of its `placeholder` attribute.
+
```
kengdoj, line 142: Can we keep the note from https://act-rules.github.io/rules/e086e5#passed-example-5?

Note: While the placeholder attribute is sufficient to provide an [accessible name](https://act-rules.github.io/rules/e086e5#accessible-name), a [visible](https://act-rules.github.io/rules/e086e5#visible) [label](https://www.w3.org/TR/WCAG21/#dfn-labels) that does not disappear when a user starts to enter data is still required for [success criterion 3.3.2 Labels or Instructions](https://www.w3.org/TR/WCAG21/#labels-or-instructions).

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,382 @@
...
+- do not have a [required context](https://www.w3.org/TR/wai-aria/#scope) role that itself inherits from one of those roles.
+
+Elements with the `option` role are not tested in this rule because they do not meet the definition of a [User interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components). If these elements are presented as user interface components, these need to be tested separately from this rule.
+
+Unlike the [Image button has non-empty accessible name](https://act-rules.github.io/rules/59796f) rule, this rule does not check whether an element's accessible name is different from the default name. While that rule assumes that, for image buttons, the default name is never descriptive, this rule relaxes that assumption to ensure the simplicity of this rule.
+
+### Related rules
+
+- [Image button has non-empty accessible name](https://act-rules.github.io/rules/59796f)
+- [Link has non-empty accessible name](https://act-rules.github.io/rules/c487ae)
```
kengdoj, line 81: if this c487ae rule isn't getting deprecated, should link be removed from this widget rule? Otherwise, we'll have 2 rules to test link has non-empty accessible name.

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,382 @@
...
+  # https://www.npmjs.com/package/htmlhint
+  # (used with `npm test` to ensure validity of code snippets)
+  - 'alt-require'
+---
+
+## Applicability
+
+This rule applies to any element that is [included in the accessibility tree](#included-in-the-accessibility-tree) and has a [semantic role][] that is or inherits from one of the following [semantic roles][]:
+
+- `button`; or
```
WilcoFiers, line 37: I think we're going to have problems with image buttons if we do this. If we added an image button example to this rule, an implementation that would report that as failing 1.1.1 would not be considered a consistent implementation of this rule, per the new definitions we're working on for consistency (see https://github.com/w3c/wcag-act/pull/522).

I'm not sure how to handle this case yet. But I would suggest we discuss this on an upcoming call. I'd like us to carefully work through the implications of not having an image button exception here.

_rules/widget-non-empty-accessible-name-rdzs6q.md
```diff
@@ -0,0 +1,382 @@
...
+- `checkbox`; or
+- `combobox`; or
+- `link`; or
```
WilcoFiers, line 40: Related to the above, but worse since all elements with link roles should fail 2.4.4 and 2.4.9.

WilcoFiers (changes requested):

HelenBurge (approved):

carlosapaduarte:
The CG discussed this in a [meeting](https://www.w3.org/2022/03/24-act-r-minutes.html#item04) and decided to close this PR due to the consistency issues that it would raise.

PR #1797, opened 2022-02-16
Update front matter of multiple rules to acknowledge source of used assets

### LINKED ISSUE #1093: Cite source of rabbit video (and other resources?)
Many video rules are using the video of a rabbit, seems to be taken from https://www.w3schools.com/html/html5_video.asp itsel/ff coming from https://peach.blender.org/

I think we should cite our source. Probably a global note at the start of the Examples section in the line of "The video used in these examples come from…"

We should also check other sources and give credit accordingly if needed.

carlosapaduarte:
Update the front matter of rules that included
- moon-speech.mp3
- rabbit video
- web accessibility perspective video

to acknowledge the sources

### Closes issue(s):

- closes #1093 

### Need for Call for Review:

This will not require a Call for Review

---

## How to Review And Approve


JenniferChadwick (approved):

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -26,6 +26,8 @@ acknowledgments:
     - Jean-Yves Moyen
   funding:
     - WAI-Tools
+  assets:
+    - JFK's We Choose the Moon speech excerpt is courtesy of NASA.
```
Jym77, line 30:
```suggestion
    - JFK's "We Choose the Moon" speech excerpt is courtesy of NASA.
```
Or in italics? Not sure which applies best: https://www.scribbr.com/mla/titles/
(needs to be done on all of them)

Jym77 (dismissed):
Typographic nerd nitpicking… 🤓

_rules/aria-state-or-property-permitted-5c01ea.md
carlosapaduarte: I'm going with the quotation marks... according to your source, that's what seems most appropriate for an excerpt

WilcoFiers (changes requested):
Could you remove the .DS_Store file?

Jym77 (approved):

Jym77:
There's still a `.DS_store` file in the PR 🙈 
(you can set up a global `.gitignore` file on your system)

WilcoFiers (approved):

carlosapaduarte:
> There's still a `.DS_store` file in the PR 🙈

Yes, but I guess it was in the repo beforehand. This PR initially modified it and then deletes it, so it should go away.

> (you can set up a global `.gitignore` file on your system)
Done!

carlosapaduarte:
Call for review not needed. Merging.

PR #1802, opened 2022-02-24
Tweaks to expectations to "Block of repeated content is collapsible"

### LINKED ISSUE #1579: [Rejected] 3e12e1 -- Block of repeated content is collapsible
Ref:
https://act-rules.github.io/rules/3e12e1

First paragraph of “expectation” section is really difficult to understand.

>   For each [block of repeated content](https://act-rules.github.io/rules/3e12e1#block-of-repeated-content) in each test target, which is before (in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree)) at least one node of [non-repeated content after repeated content](https://act-rules.github.io/rules/3e12e1#non-repeated-content), all the following are true:

Before what?

---

The background section is difficult to understand.

There are two different things we are trying to communicate here I think, but it is not clear to me where the second starts and the first ends.

To me we want to say:

1. Technique does not require a specific location for the instrument. hence, instruments placed after the repeated content in the reading order would pass the rule, but they would not make sense as they would not really solve the issue
1. We don't use the `main` landmark because that would pass the SC itself.

---

An editorial pass should be taken to passed test cases descriptions. Instead of

>   In this document, the visibility and inclusion in the accessibility tree of the navigational block of repeated content can be toggled on and off by the link at the start of the document.

It could be

>   This document has a link at the start of the document to toggle the visibility and inclusion in the accessibility tree of the navigational block of repeated content

And so for the others.

WIP in #1582

daniel-montalvo:
<< Describe the changes >>

Closes issue(s):

- closes #1579 (ADD ISSUE NUMBER HERE)

Need for Call for Review:
<< check [Process Document on Call for Review](https://act-rules.github.io/pages/design/process/#call-for-review) >>
This will require a 1 week Call for Review << small changes affecting a small number of test cases, if in doubt do not use this. >>

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/block-collapsible-3e12e1.md
```diff
@@ -29,7 +29,7 @@ This rule applies to any [HTML web page][].
 
 ## Expectation
 
-For each [block of repeated content][] in each test target, which is before (in the [flat tree][]) at least one node of [non-repeated content after repeated content][], all the following are true:
+For each [block of repeated content][] which is before at least one node of [non-repeated content after repeated content][] in the [flat tree][], all the following are true:
```
Jym77, line 32: I think we need to keep the "each test target" quantification to scope the "each block" one (otherwise, it could be "for each block in the entire web").
```suggestion
For each test target, and each of its [block of repeated content][] which is before at least one node of [non-repeated content after repeated content][] in the [flat tree][], all the following are true:
```

Otherwise, we might be able to play around the idea of:
```markdown
For each [applicable block](#3e12e1:applicable-block), all the following are true:
- (condition 1)
…
- (condition n)

For each test target, the <dfn id="3e12e1:applicable-block">Applicable Blocks</dfn> are the blocks of repeated content, …
```
that can split the concerns a bit and make the full easier to swallow.

Jym77 (changes requested):

Jym77 (approved):

colabottles (approved):

_rules/block-collapsible-3e12e1.md
```diff
@@ -29,7 +29,7 @@ This rule applies to any [HTML web page][].
 
 ## Expectation
 
-For each [block of repeated content][] in each test target, which is before (in the [flat tree][]) at least one node of [non-repeated content after repeated content][], all the following are true:
+For each test target, and each of its [block of repeated content][] which is before at least one node of [non-repeated content after repeated content][] in the [flat tree][], all the following are true:
```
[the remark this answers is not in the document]
daniel-montalvo: > As this is currently phrased I would want it to say `each of its blocks of repeated content` (make blocks plural) which I realize would break the definition tool. Could we say something like `each descendant block of repeated content which is ...`.

Making it plural and adding a new difinition in [block of repeated content]: #block-of-repeated-content 'Definition of Block of Repeated Content'G[blocks of repeated content]: #block-of-repeated-content 'Definition of Block of Repeated Content'

_rules/block-collapsible-3e12e1.md
```diff
@@ -46,7 +46,9 @@ _There are no major accessibility support issues known for this rule._
...
+[Technique SCR28: Using an expandable and collapsible menu to bypass block of content][tech scr28] does not have any requirements concerning the location of the [instruments][instrument] in relation to the [block of repeated content][] they control, hence this rule doesn't. It is likely a good idea to either keep each [instrument][] close to the start of the [block of repeated content][] it controls; or to group them all in one place near the start of the document.
+
+[Instruments][instrument] located after the block they collapse in reading order are likely not satisfying [Success Criterion 2.4.1 Bypass blocks][sc241], which this rule is designed for. Thus, it is possible to pass this rule without satisfying [Success Criterion 2.4.1 Bypass blocks][sc241].
```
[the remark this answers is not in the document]
daniel-montalvo: > This seems like it could possibly go in the assumptions section. E.g.,
> 
> `We assume the instrument is located before the block it collapses. If the instrument is located later in the reading order, it is likely not satisfying [Success Criterion 2.4.1 Bypass blocks][sc241], Thus, it is possible to pass this rule without satisfying [Success Criterion 2.4.1 Bypass blocks][sc241].`
> 
> But maybe the "likely not" makes the statement too weak. I think it could be up for discussion but would be fine leaving it as is.

@Jym77  Would you be comfortable with me trying to make this an assumption?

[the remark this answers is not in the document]
daniel-montalvo: > I do not think this really is an assumption **made by the rule**. The rule does not assume anything about the position of the instrument. It's quite the contrary, the rule accepts instruments located anywhere on the page, which is likely too lax.
> 
> This is more of a known case where the rule can pass content which is bad (false negative). Maybe it can be rephrased to make this more explicit:
> 
> > The rule accepts instruments located anywhere in the page. Instruments located after the block they collapse …
> 
> Maybe we can also drop the "in reading order" precision which makes the sentence hard to parse. Given that this is in a non-normative note, it is OK to just say "after the block" even if it is a bit ambiguous.

Changed to "sequentially after the block". I think that is what we want to underline here, as this is likely to affect tab order mostly.

_rules/block-collapsible-3e12e1.md
```diff
@@ -46,7 +46,9 @@ _There are no major accessibility support issues known for this rule._
...
+[Technique SCR28: Using an expandable and collapsible menu to bypass block of content][tech scr28] does not have any requirements concerning the location of the [instruments][instrument] in relation to the [block of repeated content][] they control, hence this rule doesn't. It is likely a good idea to either keep each [instrument][] close to the start of the [block of repeated content][] it controls; or to group them all in one place near the start of the document.
+
+[Instruments][instrument] located sequentially after the block they collapse are likely not satisfying [Success Criterion 2.4.1 Bypass blocks][sc241], which this rule is designed for. Thus, it is possible to pass this rule without satisfying [Success Criterion 2.4.1 Bypass blocks][sc241].
```
kengdoj, line 51: Should the instrument being before the block they collapse be in the expectation?
- there exists an [instrument][] before the [block][] to make all nodes in this [block][] not [visible][]; and
- there exists an [instrument][] before the [block][] to remove all nodes in this [block][] from the [accessibility tree][included in the accessibility tree].

daniel-montalvo:
Thanks @kengdoj

That's added now in 4e618b012d48214038e09803ae6b54b0c98c055f

kengdoj (approved):

_rules/block-collapsible-3e12e1.md
```diff
@@ -29,10 +29,10 @@ This rule applies to any [HTML web page][].
 
 ## Expectation
 
-For each [block of repeated content][] in each test target, which is before (in the [flat tree][]) at least one node of [non-repeated content after repeated content][], all the following are true:
+For each test target, and each of its [blocks of repeated content][block of repeated content] which are before at least one node of [non-repeated content after repeated content][] in the [flat tree][], all the following are true:
 
-- there exists an [instrument][] to make all nodes in this [block][] not [visible][]; and
-- there exists an [instrument][] to remove all nodes in this [block][] from the [accessibility tree][included in the accessibility tree].
+- there exists an [instrument][] before the [block][] to make all nodes in this [block][] not [visible][]; and
+- there exists an [instrument][] before the [block][] to remove all nodes in this [block][] from the [accessibility tree][included in the accessibility tree].
```
WilcoFiers, line 35: "before" is a bit ambiguous here. Is that visually above, is that the relation in the flat tree? Is that the accessibility tree (since with aria-owns that can differ from the flat tree).

WilcoFiers, line 35: If we don't allow the instrument to come after the block, I think we need a failed example for this.

_rules/block-collapsible-3e12e1.md
Jym77: I am not sure we should reject instruments that are after the block they collapse. Because I am not sure that WCAG would fail in these case. The SC says:
> A [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism) is available to bypass blocks of content that are repeated on multiple [Web pages](https://www.w3.org/TR/WCAG21/#dfn-web-page-s).

The question is a bit whether "bypass" implies that the mechanism is located before the corresponding block of content or not. But it doesn't says so explicitly. Similarly, [Document has an instrument to move focus to non-repeated content](https://act-rules.github.io/rules/ye5d6e) (aka glorified skip link) does not require the instrument to be before its destination.

There is also a can of worm with a "collapse" button at the end of a block of text with no tabbable inside the block of text. Then it is possible to collapse the content by tabbing into the button without being hampered by the text. Is that good or not? I'm not sure.

So, I'd rather keep the rule a bit lax, allow the instrument anywhere on the page, with a note that this is a known weakness of the rule. I'm seeing high risks of false positives otherwise…
Happy to discuss that further, of course.

Jym77 (changes requested):

_rules/block-collapsible-3e12e1.md
kengdoj: True, SC doesn't say it explicitly. But the mechanism should work. If the mechanism is after the block of content, can it achieve the result of bypassing the block? 

> definition of [mechanism](https://www.w3.org/TR/WCAG21/#dfn-mechanism): [process](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html#dfn-process) or technique for achieving a result

This is interesting:

> There is also a can of worm with a "collapse" button at the end of a block of text with no tabbable inside the block of text. Then it is possible to collapse the content by tabbing into the button without being hampered by the text. Is that good or not? 

One could argue that putting any link after the block of text is a way to bypass. No additional skip or collapse function is necessary.

Jym77: > True, SC doesn't say it explicitly. But the mechanism should work. If the mechanism is after the block of content, can it achieve the result of bypassing the block?

I do agree that a mechanism located after the block it collapses is most likely not going to be good for 2.4.1.
What I'm having more trouble with is to convince me that it would always violate 2.4.1 in all cases.

Notably, "before/after" can relate to reading order, to tab-order, to DOM order, to accessibility tree order… (as pointed in https://github.com/act-rules/act-rules.github.io/pull/1802#discussion_r854940717)
```html
<div style="position: relative">
  <button style="position: absolute; bottom: 0">Collapse paragraph</button>
  <p> style="position: absolute; top: 0">Lorem Ipsum</p>
</div>
```
(plus some `height` to make sure stuff don't collide)
This has the "collapse" mechanism before the block in DOM order and tab-order, but not in reading order. Is it bad?
What if the `<button>` is after the `<p>` (with some `tabindex` shenanigans), thus after in DOM order (or flat tree order).
What if we use `aria-owns` to tweak even more the accessibility tree.

That's why I feel uncomfortable rejecting everything where the collapsing mechanism is "after" the block and feel it's easier to just not look where it is… At least, we keep the fail-on-fail relationship (no false positives) which is central in ACT rules design. (and keeping the non-normative note that this is a known limitation of the rule)

If we want to keep this, we do need a precise definition of "before/after" in that case to make sure all testers test the same thing (would be bad if some look at tab-order and others look at reading order).

---
> This is interesting:
> 
> > There is also a can of worm with a "collapse" button at the end of a block of text with no tabbable inside the block of text. Then it is possible to collapse the content by tabbing into the button without being hampered by the text. Is that good or not?
> 
> One could argue that putting any link after the block of text is a way to bypass. No additional skip or collapse function is necessary.

That would depend on whether the block has tab-stops or not.
But that's indeed an interesting case. We could add an atomic rule (to the Bypass Blocks composite one) which would pass if blocks of repeated content have no tab-stop and there is a tab-stop in non-repeated content after them.

That's however make things trickier because we could mix and match bypass methods (e.g. have one collapsible block, and one no-tab-stop block). The current rules are built in a way that methods cannot mix and match, which does create fairly big limitations (e.g. the ["Skip link"](https://act-rules.github.io/rules/ye5d6e) rule accepts any link to non-repeated content. This consciously avoids digging into G123 (one "local skip link" per block as in its [Passed Example 3](https://act-rules.github.io/rules/ye5d6e#passed-example-3))). Putting more requirements to try and detect "local skip links" becomes problematic since each block of content could be skipable by different mean (and building the rule becomes fairly difficult, notably we'd need to target individual blocks of content to keep in line with the ACT rules format, or make a very complex rule of "for each block of repeated content, either: local skip link; collapsible; no tab-stop").

It might, however, be worth adding an assumption to the composite rule to pre-emptively discard blocks with no tab-stop. Or maybe this could be part of [Document has no repeated content followed by non-repeated content](https://github.com/act-rules/act-rules.github.io/pull/1539/files#diff-b48514b7b1bb4c6ed432698bc7cab3872d5d83402055a13bea1fb5dada166e32R3) that was never finished (and that might also be added to Bypass Blocks" 🤔 )

(sorry, I'm not sure if I'm making things clearer or more confusing 😕 )

Jym77:
Cancelling the Call for Review due to the heavy discussions going on.

_rules/block-collapsible-3e12e1.md
kengdoj: Can of worms busted open! 😆 

For a mechanism to achieve the result of bypassing the block, I think "before" means in tab order and reading order (at a minimum). It wouldn't make sense to have a user tab/read through the block and then get to the option to bypass the block. Can we use the same phrasing from the [rule](https://act-rules.github.io/rules/3e12e1#expectation): 

> Expectation: For each [block of repeated content](https://act-rules.github.io/rules/3e12e1#block-of-repeated-content) in each test target, which is before (in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree)) at least one node...

- there exists an instrument _before (in the flat tree) the block_ to...

For the scenario where a link after a block of text allegedly serves as a bypass mechanism, I wouldn't advocate for this approach. It might pass a rule for 2.4.1 Bypass Blocks, but I would want to fail it for something (like 2.4.4 Link Purpose).

daniel-montalvo: I would try to further define "before" in this context, even if the SC does not explicitly get into this. If the instrument is after either the tab order or the accessibility tree, the whole purpose of the instrument would then be defeated. The users would have already bypassed the block of repeated content themselves before getting to the instrument, as @kengdoj has pointed out.

Not really sure if this pr continues to make sense as the discussions here seem to have gone far beyond what this PR was originally about.

WilcoFiers:
Closing, as agreed on by Daniel.

PR #1804, opened 2022-03-03
Text spacing not !important [78fd32, 24afc2, 9e45ec]: Consider actual font size and ignore empty text

### LINKED ISSUE #1687: Text spacing not !important rules should check text is actually affected [78fd32, 24afc2, 9e45ec]
This issue applies to the following three rules:
- [Letter spacing in style attributes is not !important](https://act-rules.github.io/rules/24afc2)
- [Line height in style attributes is not !important](https://act-rules.github.io/rules/78fd32)
- [Word spacing in style attributes is not !important](https://act-rules.github.io/rules/9e45ec)

These ACT rules should check that there are actual text nodes affected by the style attribute. For example, the following scenarios probably shouldn't be failing the rule:

```html
<div style="letter-spacing: 2px !important; border-top: 1px solid black;">
  <!-- empty div, border to make it "visible" -->
</div>
```

```html
<div style="letter-spacing: 2px !important">
  <p style="letter-spacing: 3px !important;">
  </p>
</div>
```

### LINKED ISSUE #1688: Text spacing not !important does not account for actual font-size [78fd32, 24afc2, 9e45ec]
This issue applies to the following three rules:
- [Letter spacing in style attributes is not !important](https://act-rules.github.io/rules/24afc2)
- [Line height in style attributes is not !important](https://act-rules.github.io/rules/78fd32)
- [Word spacing in style attributes is not !important](https://act-rules.github.io/rules/9e45ec)

This is a false positive. The rule fails the `div` for not having a letter-spacing that's less than 0.15 times the font-size, but then the `p` element reduces the font-size, and so in actuality the text ends up with letter spacing that is sufficient.

```html
<div style="font-size: 16px; letter-spacing: 2px !important">
  <p style="font-size: 10px;">
  </p>
</div>
```

Jym77:
Superceded by #1923 

Currently updating only "Line height is not `!important`" to check whether that solution works or not.

The second and third condition where about the test target, which is incorrect since both the `font-size` or the text spacing property can be updated before the actual text. This replaces them with a condition that look at these properties on the parent of each text node inside the test target.

Closes issue(s):

- closes #1687 
- closes #1688

Need for Call for Review:
This will require a 2 weeks Call for Review (big change to expectation reworking the logic significantly)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -29,9 +29,13 @@ This rule applies to any [HTML element][] that is [visible][] and for which the
 
 For each test target, at least one of the following is true:
 
-- **not important**: the [computed][] value of its [letter-spacing][] property is not [important][]; or
-- **wide enough**: the [computed][] value of its [letter-spacing][] property is at least 0.12 times the [computed][] value of its [font-size][] property; or
-- **cascade**: the [cascaded][] value of its [letter-spacing][] property is not a value [declared][] in its `style` attribute.
+- <dfn id="24afc2:not-important">not important</dfn>: the [computed][] value of its [letter-spacing][] property is not [important][]; or
+- <dfn id="24afc2:spaced-text">spaced text</dfn>: for each text node descendant of this target, one of the following is true:
```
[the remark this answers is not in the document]
Jym77: > Would it be easier to change the applicability for this like the line height rule? E.g., the applicability would become
> 
> `This rule applies to any HTML element with [visible](https://act-rules.github.io/rules/78fd32#visible) text for which the style attribute [declares](https://www.w3.org/TR/css-cascade-4/#declared) the [letter-spacing](https://www.w3.org/TR/css-text-3/#propdef-letter-spacing) CSS property.`

It is the case already: [Applicability](https://github.com/act-rules/act-rules.github.io/pull/1804/files#diff-3ca6de72ad2dd5a4af825a1785d687778e21647d8311d90cf0696a3734626393R26)

> I'm not sure I completely understand what you are trying to fix here. Specifically I don't quite follow what you mean when you say `which is incorrect since both the font-size or the text spacing property can be updated before the actual text. ` in your description above.

That are the examples in issues #1687 and #1688 which this PR fixes.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -29,9 +29,13 @@ This rule applies to any [HTML element][] that is [visible][] and for which the
...
+  - <dfn id="24afc2:empty">empty</dfn>: the text node is only [whitespace][]; or
+  - <dfn id="24afc2:wide-enough-cond">wide enough</dfn>: the parent of the text node has [wide enough letter spacing](#24afc2:wide-enough-def); or
+  - <dfn id="24afc2:cascade">cascade</dfn>: the parent of the text node has a [cascaded][] [letter-spacing][] property which is not the one [declared][] in the `style` attribute of the target.
+
+An element has a <dfn id="24afc2:wide-enough-def">wide enough letter spacing</dfn> if the computed value of its [letter-spacing][] property is at least 0.12 times the computed value of its [font-size][] property.
```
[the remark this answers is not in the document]
Jym77: Fair point. It is not really needed for `letter-spacing` and `word-spacing` since the used value and computed value should be equal. But alignment is a good idea…

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -29,9 +29,13 @@ This rule applies to any [HTML element][] that is [visible][] and for which the
...
 For each test target, at least one of the following is true:
 
-- **not important**: the [computed][] value of its [letter-spacing][] property is not [important][]; or
-- **wide enough**: the [computed][] value of its [letter-spacing][] property is at least 0.12 times the [computed][] value of its [font-size][] property; or
-- **cascade**: the [cascaded][] value of its [letter-spacing][] property is not a value [declared][] in its `style` attribute.
+- <dfn id="24afc2:not-important">not important</dfn>: the [computed][] value of its [letter-spacing][] property is not [important][]; or
+- <dfn id="24afc2:spaced-text">spaced text</dfn>: for each text node descendant of this target, one of the following is true:
+  - <dfn id="24afc2:empty">empty</dfn>: the text node is only [whitespace][]; or
+  - <dfn id="24afc2:wide-enough-cond">wide enough</dfn>: the parent of the text node has [wide enough letter spacing](#24afc2:wide-enough-def); or
+  - <dfn id="24afc2:cascade">cascade</dfn>: the parent of the text node has a [cascaded][] [letter-spacing][] property which is not the one [declared][] in the `style` attribute of the target.
```
tbostic32, line 36: I've been wrestling with this definition a bit. I finally understand it, but it seems weird since it feels like it is almost defining an inapplicable case and then calling it a pass (e.g., the text node doesn't use the target elements style which results in a pass). I'm not sure I have a suggestion for this, but figured I'd leave my thoughts in case it helps to figure out how to clarify this language somehow.

tbostic32 (dismissed):

_rules/letter-spacing-not-important-24afc2.md
Jym77: Yes, the problem is that a single target can have all three cases (or more), something like:
```html
<div style="letter-spacing: 2.4 px !important; font-size: 30px">
  <span>&nbsp;</span> <!-- OK because too small but only whitespace -->
  <span style="font-size: 20px">Lorem</span> <!-- OK because font-size has changed -->
  <span style="letter-spacing: 3.6px !important">ipsum</span> <!-- OK because letter-spacing has changed -->
  <span>dolor sit amet</span> <!-- bad -->
</div>
```

The 4th case totally only fails because of the `div`, so we do want the `div` as a target.

We could update applicability to say something like "…and has at least one text node descendant which is not only whitespace and whose parent's cascaded `letter-spacing` is not the one declared here". That would make cases with only the first or third case Inapplicable, which is quite good (these are cases where the `!important` declaration does not affect anything. But we would still need to keep the "empty" and "cascade" exceptions in the Expectation due to the possibility of mixing things up, as here. Hence, I feel that doing so would result in complicating the Applicability without simplifying the Expectation, and duplicating some condition (making maintenance harder).

Another solution could be to target the text nodes (which are not empty and whose parent's cascaded `letter-spacing` is `!important` and declared in a style attribute), but that would result in failing the text nodes rather than the (element with the) bad declaration, so it is not that good for reporting. This would result in only targeting "Lorem" (pass) and "dolor sit amet" (fail).

__tests__/spelling.js
```diff
@@ -91,16 +91,28 @@ const checkSpelling = text => {
...
 	}
 
+	if (stripRefNamesAndURLs) {
+		// Start of line, Opening [, no ] (repeated), closing ], /[[][^\]]*]: [^ ]*/ => matches "[name]: url"
+		// => the leftover bit is the title, which should be spellchecked.
+		// The name is only spellchecked if it appears in text, which is OK.
+		// 'm' flag checks in multiline mode, aka each line is matched separately.
+		const refListRegex = /^\[[^\]]*]: [^ ]*/gm
+		body = body.replace(refListRegex, "")
+	}
```
Jym77, line 114: Adding code to ignore the names and URL parts of reference list items when spellchecking, i.e. in
> `[name]: url 'title'

only the title needs to be spellchecked (the name being spellchecked if it appears in the text). This became useful here with all the `#xxxxxx:anchor` links in the reference list.

_rules/letter-spacing-not-important-24afc2.md
```diff
@@ -157,6 +161,67 @@ The [computed][] value of the `letter-spacing` property of this `p` element is *
...
+<p style="font-size: 24px; letter-spacing: 2px !important">
+	<span style="letter-spacing: 3px !important;">
+		The toy brought back fond memories of being lost in the rain forest.
+	</span>
+</p>
+```
+
+#### Passed Example 13
+
+This `p` element has three text node descendants. The first one is [empty](#24afc2:empty); the second one is [wide enough][] because of the smaller `font-size` on its parent; and the last one is [wide enough][] because of the `letter-spacing` specified on its parent. The third `span` element has a [wide enough][] text node descendant.
```
carlosapaduarte: I feel like it would be good to specifically highlight there are two test targets in this example. Also, added the reference to the cascade condition.

```suggestion
This `p` element has three text node descendants. The first one is [empty](#24afc2:empty); the second one is [wide enough][] because of the smaller `font-size` on its parent; and the last one is [wide enough][] because of the `letter-spacing` specified on its parent, meeting the [cascade](#24afc2:cascade) condition. The third `span` element, which is also applicable, has a [wide enough][] text node descendant.
```

_rules/line-height-not-important-78fd32.md
```diff
@@ -185,6 +189,61 @@ The [computed][] value of the `line-height` property of this `p` element is **no
...
+	<span>&nbsp;</span>
+	<span style="font-size: 16px;">
+		The toy brought back fond memories of being lost in the rain forest.
+	</span>
+</p>
+```
+
+#### Passed Example 13
+
+Both this `p` and `span` elements have a single text node descendant, whose parent (the `span` element) is [large enough][].
```
carlosapaduarte: Shouldn't the description here mention that the `p` passes because of the cascade condition? The large enough condition is only relevant for the `span`

_rules/line-height-not-important-78fd32.md
```diff
@@ -185,6 +189,61 @@ The [computed][] value of the `line-height` property of this `p` element is **no
...
+<p style="font-size: 16px; line-height: 16px !important; max-width: 200px">
+	<span style="line-height: 24px !important;">
+		The toy brought back fond memories of being lost in the rain forest.
+	</span>
+</p>
+```
+
+#### Passed Example 14
+
+This `div` element has three text node descendants. The first one is [empty](#78fd32:empty); the second one is [large enough][] because of the smaller `font-size` on its parent; and the last one is [large enough][] because of the `letter-spacing` specified on its parent. The third `span` element has a [large enough][] text node descendant.
```
carlosapaduarte:
```suggestion
This `div` element has three text node descendants. The first one is [empty](#78fd32:empty); the second one is [large enough][] because of the smaller `font-size` on its parent; and the last one matches the [cascade](#78fd32:cascade) condition because of the `letter-spacing` specified on its parent. The third `span` element has a [large enough][] text node descendant.
```

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -29,9 +29,13 @@ This rule applies to any [HTML element][] that is [visible][] and for which the
 
 For each test target, at least one of the following is true:
 
-- **not important**: the [computed][] value of its [word-spacing][] property is not [important][]; or
-- **wide enough**: the [computed][] value of its [word-spacing][] property is at least 0.16 times the [computed][] value of its [font-size][] property; or
-- **cascade**: the [cascaded][] value of its [word-spacing][] property is not a value [declared][] in its `style` attribute.
+- <df id="9e45ec:not-important">not important</dfn>: the [computed][] value of its [word-spacing][] property is not [important][]; or
+- - <dfn id="9e45ec:spaced-text">spaced text</dfn>: for each text node descendant of this target, one of the following is true:
```
carlosapaduarte:
```suggestion
- <dfn id="9e45ec:spaced-text">spaced text</dfn>: for each text node descendant of this target, one of the following is true:
```

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -157,6 +161,65 @@ The [computed][] value of the `word-spacing` property of this `p` element is **n
...
+
+```html
+<p style="word-spacing: 1px !important; border-top: 1px solid black;">
+	&nbsp:<!-- empty div, border to make it "visible" -->
+</p>
+```
+
+#### Passed Example 12
+
+Both this `p` and `span` elements have a single text node descendant, whose parent (the `span` element) is [wide enough][].
```
carlosapaduarte: Same here... for the `p` element, what is relevant is the cascade condition, right?

_rules/word-spacing-not-important-9e45ec.md
```diff
@@ -157,6 +161,65 @@ The [computed][] value of the `word-spacing` property of this `p` element is **n
...
+<p style="font-size: 24px; word-spacing: 2px !important">
+	<span style="word-spacing: 4px !important;">
+		The toy brought back fond memories of being lost in the rain forest.
+	</span>
+</p>
+```
+
+#### Passed Example 13
+
+This `p` element has three text node descendants. The first one is [empty](#9e45ec:empty); the second one is [wide enough][] because of the smaller `font-size` on its parent; and the last one is [wide enough][] because of the `word-spacing` specified on its parent. The third `span` element has a [wide enough][] text node descendant.
```
carlosapaduarte:
```suggestion
This `p` element has three text node descendants. The first one is [empty](#9e45ec:empty); the second one is [wide enough][] because of the smaller `font-size` on its parent; and the last one matches the [cascade](#9e45ec:cascade) condition because of the `word-spacing` specified on its parent. The third `span` element has a [wide enough][] text node descendant.
```

carlosapaduarte (dismissed):
Good job with the updated expectations. I've made some changes towards clarifying example descriptions.

_rules/letter-spacing-not-important-24afc2.md
Jym77: For the `p` target, the text inside the third `span` is actually meeting both the "cascade" condition **and** the "wide enough" one (because the parent of this text node, the `span`, has a used `letter-spacing` of 3.6px, and a computed `font-size` of 30px; and 30×1.2 ⩽ 3.6).
I felt that "wide enough" is easier to understand 😅 But I guess we can keep both…

_rules/line-height-not-important-78fd32.md
Jym77: Here also, the large enough condition is relevant for both 🤷

carlosapaduarte (dismissed):

colabottles (approved):

WilcoFiers (dismissed):
This is getting really very complicated. I'm not sure this is actually necessary. To be honest I'd much rather we do this by just adding an assumption. Previously the rule of thumb has been that if we have no real-world examples of a problem, we're can use those. WDYT?

Jym77:
> This is getting really very complicated. I'm not sure this is actually necessary. To be honest I'd much rather we do this by just adding an assumption. Previously the rule of thumb has been that if we have no real-world examples of a problem, we're can use those. WDYT?

I'm not sure how much of an actual problem the full rule detects in the first place 😅

Thus said, I feel the way to simplicity would be to move focus of the rule from the element to the text node (so from where the problem is created to where it actually occurs):

Applicability: any (parent of a) (non-whitespace) text node.

Expectation: either (i) specified `letter-spacing` is not declared in a `style` attribute; or (ii) specified `letter-spacing` is not `!important`; or (iii) computed `letter-spacing` is larger than X times computed `font-size`.

This immediately rules out the corner cases where a definition is not affecting any text, and freely handle the cases where the properties get changed. Maybe that would need to deprecate these rules and create new ones since it is a significant change 🤔 

@WilcoFiers what do you think about this?

WilcoFiers:
@Jym77 Just for how much that changes these rules, I'd rather not. If we wanted to fix this (and I'm not convinced we need to) perhaps the way to do it would be by doing this:

```md
- **wide enough**: the [computed][] [letter-spacing][] is at least 0.12 times the [computed][] [font-size][] of all descending text node in the flat tree; or
```

Jym77:
> @Jym77 Just for how much that changes these rules, I'd rather not. If we wanted to fix this (and I'm not convinced we need to) perhaps the way to do it would be by doing this:
> 
> ```lisp
> - **wide enough**: the [computed][] [letter-spacing][] is at least 0.12 times the [computed][] [font-size][] of all descending text node in the flat tree; or
> ```

That fixes #1688 (`font-size` changes), but not #1687 (`letter-spacing` change) where in the second example the `div` would still incorrectly fail the rule.

dd8:
How about wording like this:

> ### Applicability
> Elements where the computed value of the `letter-spacing:` property was assigned from a `letter-spacing:` declared as `!important` in a `style` attribute.
> ### Expectation
> For each test target, at least one of the following is true:
> - only whitespace: the element contains only whitespace text nodes; or
> - wide enough: the element has wide enough letter spacing

The avoids having to embed details of how the CSS the cascade works in the rule, and lets implementors choose how they implement this (looking at cascaded value, looking at specificity, calling an API, etc)

The definition above seems to work for all these cases:

```
<div style="letter-spacing: 2.4px !important; font-size: 30px">
  <span>&nbsp;</span> <!-- OK because too small but only whitespace -->
  <span style="font-size: 20px">Lorem</span> <!-- OK letter-spacing assigned from div, but font-size is smaller -->
  <span style="letter-spacing: 3.6px !important">ipsum <!-- OK letter-spacing assigned from span --></span> 
  <span style="letter-spacing: 4.6px">ipsum <!-- OK letter-spacing assigned from span, and doesn't inherit from div --></span> 
  <span>dolor sit amet </span> <!-- Bad: letter-spacing assigned from div via implicit inheritance -->
  <span style="letter-spacing: inherit">ipsum <!-- Bad: letter-spacing assigned from div via explicit inheritance --></span> 
  
 Also bad <!-- Bad: letter-spacing assigned from div -->
</div>
```

Jym77:
> How about wording like this:
> 
> > ### Applicability
> > Elements where the computed value of the `letter-spacing:` property was assigned from a `letter-spacing:` declared as `!important` in a `style` attribute.

I'm not sure what "the computed value was **assigned from**" means. I guess i will boil down to something like "the specified value was declared in" (an `!important` `style` attribute) 🤔 In any case we'll need to define it since it doesn't refer to any existing term as far as I can tell.

dd8:
**assign** comes from here:

>Once a user agent has parsed a document and constructed a document tree, it must **assign**, to every element in the tree, and correspondingly to every box in the formatting structure, a value to every property that applies to the target media type

https://www.w3.org/TR/css-cascade-3/#value-stages

Here's a tweaked version with a link to the definition:

> ## Applicability
> Elements where the computed value of the letter-spacing: property was [assigned](https://www.w3.org/TR/css-cascade-3/#value-stages) from a `letter-spacing:` declaration marked as as `!important` in an HTML `style` attribute.

Jym77:
Closed by #1923

PR #1809, opened 2022-03-10
form-field descriptive label: Improve the examples

WilcoFiers:
- Ensure inapplicable examples don't fail the SC
- Make it clearer why some of these failed examples are failing.

---

## How to Review And Approve


_rules/form-field-label-descriptive-cc0f0a.md
```diff
@@ -198,29 +198,32 @@ The `span` element is a [programmatic label][] of the `input` element but does n
 
 #### Failed Example 4
 
-The `label` elements are [programmatic labels][programmatic label] of their respective `input` elements. The `label` elements, are not descriptive enough (because they are repeated over several fields). The headings are not [visible][]. Therefore, they do not provide [visual context][].
+The `label` elements are [programmatic labels][programmatic label] of their respective `input` elements. The `label` elements, are not descriptive enough because they are reused on multiple fields. The headings are not [visible][]. Therefore, they do not provide [visual context][].
```
daniel-montalvo, line 201:
```suggestion
These `label` elements are [programmatic labels][programmatic label] of their respective `input` elements. They are not descriptive enough because they are reused on multiple fields. The headings are not [visible][]. Therefore, they do not provide [visual context][].
```

_rules/form-field-label-descriptive-cc0f0a.md
```diff
@@ -198,29 +198,32 @@ The `span` element is a [programmatic label][] of the `input` element but does n
...
+		<label>Name: <input type="text" name="billing-name" /></label>
+		<label>Street: <input type="text" name="billing-street" /></label>
+	</fieldset>
 </html>
 ```
 
 #### Failed Example 5
 
-Both the `div` and the `span` elements are [programmatic labels][programmatic label] of the `input` element, but only the `div` is [visible][]. It has no [visual context][], and is not descriptive.
+Both the `button` and the `span` elements are [programmatic labels][programmatic label] of the `input` element, but only the `button` is [visible][]. It has no [visual context][], and is not descriptive.
```
daniel-montalvo, line 220:
```suggestion
These `button` and `span` elements are both [programmatic labels][programmatic label] of the `input` element, but only the `button` is [visible][]. It has no [visual context][], and is not descriptive.
```

daniel-montalvo (changes requested):

JenniferChadwick:
I am good with the edits.

wcreedle (approved):
These look good

_rules/form-field-label-descriptive-cc0f0a.md
```diff
@@ -243,8 +246,9 @@ The `label` element is not [visible][].
 
 ```html
 <html lang="en">
-	<label for="fname" style="display:none;">First name:</label>
-	<input id="fname" type="text" name="fname" />
+	<label style="display:none;">
+		First name: <input type="text" />
+	</label>
```
Jym77, line 251: This makes the `input` not visible. The Applicability is that both the label and the form field are visibles, so I'd rather keep an Inapplicable test case where one is visible but not the other (so we can check that implementers do indeed test both visibility and not just the form field visibility).

Jym77 (dismissed):

_rules/form-field-label-descriptive-cc0f0a.md
WilcoFiers: We have the visible label, hidden control case covered by inapplicable example 3. But I don't think we can have a hidden label, visible control case without it being a failure of SC 3.3.2. I don't really think we should have things like this as inapplicable examples.

Jym77: The rule does not map to 3.3.2, so it's not a problem (with respect to the rule format and our design document) to have such an Inapplicable example.

---
We can also make the example something like
```html
<label for="foo" class="sr-only">Foo</label>
<span>Foo:</span><input id="foo"/>
```
This would pass 3.3.2 since the `span` does serve as a WCAG "label", but would have a visible form field with an invisible programmatic label and thus be Inapplicable for the rule.

Jym77 (dismissed):

_rules/form-field-label-descriptive-cc0f0a.md
WilcoFiers, line 251:
```suggestion
	<label for="fname" style="position: absolute; left: -9999px;">First name:</label>
	<label>First name:</label>
	<input id="fname" type="text" name="fname" />
```

daniel-montalvo (approved):

colabottles (approved):

PR #1813, opened 2022-03-23
Fix issues with automated deploy to WAI website

WilcoFiers:
I overlooked a few things.... like permissions. This PR solves:

- Node version was too old
- Git wasn't set up with the right permissions
- Script errored if it had nothing new to commit

Need for Call for Review: None

---

## How to Review And Approve

### DIFF
--- a/.github/scripts/wai-build.mjs
+++ b/.github/scripts/wai-build.mjs
@@ -44,10 +44,19 @@ async function generateTestCases({ tmpDir, rulesDir, testAssetsDir }) {
 }
 
 async function commitAndPush({ tmpDir }) {
-  const commitMessage = (await $`git log -1 --pretty=%B`).stdout
   cd(tmpDir);
-  await $`git add .`;
-  await $`git commit -m ${commitMessage}`;
-  await $`git push`;
-  cd(`../`);
+  try {
+    const diff = (await $`git diff --name-status`).stdout;
+    if (diff.trim().length === 0) {
+      console.log('No changes detected, skipping git commit')
+      return;
+    }
+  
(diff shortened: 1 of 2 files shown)


.github/scripts/wai-build.mjs
```diff
@@ -19,7 +19,7 @@ async function cloneWcagActRules({ tmpDir }) {
   await $`git clone \
     --single-branch \
     --branch master \
-    git@github.com:w3c/wcag-act-rules.git ${tmpDir} \
+    https://github.com/w3c/wcag-act-rules.git ${tmpDir} \
```
WilcoFiers, line 22: Use HTTP to make the git URL thing work.

.github/workflows/wai-deploy.yml
```diff
@@ -10,8 +10,21 @@ jobs:
...
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - name: Configure git
+        run: |
+          git config --global url."https://${{ secrets.WAI_GIT_NAME }}:${{ secrets.WAI_GIT_ACCESS_TOKEN }}@github.com".insteadOf "https://github.com"
+          git config --global user.name "${{ secrets.WAI_GIT_NAME }}"
+          git config --global user.email "${{ secrets.WAI_GIT_EMAIL }}"
```
WilcoFiers, line 26: These are new secrets. They're currently set up with my personal credentials. I'll ask W3C what their policy on credentials is. I'd prefer to get an account set up by the W3C for this.

carlosapaduarte (approved):

Issue #1811, opened 2022-03-14
More clarity around passed case 4 for Element with aria-hidden has no focusable content [6cfa84]


straker:
Axe-core recently tried to implement the [passed example 4](https://act-rules.github.io/rules/6cfa84#passed-example-4) for Element with aria-hidden has no focusable content. I was hoping for some clarity around the example, namely if you might have any real-world scenarios around the `onfocus` attribute.

Currently we don't have a way to determine what the `onfocus` attribute does on an element, so are left with just not failing any element with the attribute. This seems a bit too broad of a stroke though so were hoping to gather some more real-world examples of the `onfocus` attribute. I tried to find some real-world examples, but was unsuccessful (all I found were mostly just tutorials and explanations about the attribute).

Jym77:
Hi @straker ,
would the case discussed in #643 answer your question? (it's not `onFocus`, but a similar use of redirecting focus with Javascript).
I do agree that static tools essentially cannot guess what Javascript is doing 😢

Jym77:
@straker #1819 is improving the example to a more realistic one (elements behind a modal redirect focus while the modal is open).

straker:
Yep, that's very helpful. Thanks for the updated example.

PR #1819, opened 2022-03-24
6cfa84: Modify pass example 4 to be more realistic, using a dialog

tombrunet:
Per todays ACT Task Force call, modifying Pass 4 to use a dialog to better highlight why a focus sentinel might be used.

Need for Call for Review:
This will require a 1 week Call for Review (Testcase and description are essentially the same, but using more of a real world example as discussed on the call)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


tombrunet:
I'll open a separate PR regarding the focusable vs tabbable since that feels like an independent discussion.

tombrunet:
See also https://github.com/act-rules/act-rules.github.io/pull/1820

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -100,10 +100,18 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
...
+    <label>Some input <input id="dialogFirst"></label><br />
+    <button>Close button</button>
+</div>
 <div aria-hidden="true">
-	<a href="/" style="position:absolute; top:-999em" onfocus="document.querySelector('input').focus()">First link</a>
+    <a href="#" id="sentinelAfter" style="position:absolute; top:-999em">Focus sentinel</a>
 </div>
-<input />
+<script>
+    sentinelAfter.addEventListener("focus", () => {
```
WilcoFiers, line 111: Oh wow! There's DOM quirkiness you don't see every day.

I know this works, but for ease of understanding, would it be better to replace this with:

```suggestion
    document.getElementById("sentinelAfter").addEventListener("focus", () => {
```

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -100,10 +100,18 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
 This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus.
 
 ```html
+<div role="dialog" aria-label="Sample Modal" aria-modal="true" style="border: solid black 1px; padding: 1rem;">
+    <label>Some input <input id="dialogFirst"></label><br />
+    <button>Close button</button>
+</div>
 <div aria-hidden="true">
-	<a href="/" style="position:absolute; top:-999em" onfocus="document.querySelector('input').focus()">First link</a>
+    <a href="#" id="sentinelAfter" style="position:absolute; top:-999em">Focus sentinel</a>
```
WilcoFiers, line 108: I don't think we should leave the off screen style here. That's generally not what happens when a modal pops up. Could we instead turn down the opacity perhaps?

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -100,10 +100,18 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
 This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus.
 
 ```html
+<div role="dialog" aria-label="Sample Modal" aria-modal="true" style="border: solid black 1px; padding: 1rem;">
+    <label>Some input <input id="dialogFirst"></label><br />
+    <button>Close button</button>
```
WilcoFiers, line 105: Should this be a functional button? Otherwise we end up with half a keyboard trap on this page.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -100,10 +100,18 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
 This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus.
```
WilcoFiers, line 100:
```suggestion
This `a` element is not [focusable][] because it moves focus to the `button` element whenever it receives focus.
```

WilcoFiers (dismissed):
Can you add a failed modal example as well that doesn't have the focus trap script?

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: Added script so that the close makes the dialog display:none

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: Focus moves to the input (top of the dialog) and not the button.

tombrunet:
@WilcoFiers Added Fail 7 in Contrast to Pass 4. I couldn't think of a real world scenario for this other than the developer simply has a bug.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -100,10 +100,21 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
...
 
 ```html
+<div id="sampleModal" role="dialog" aria-label="Sample Modal" aria-modal="true" style="border: solid black 1px; padding: 1rem;">
+    <label>Some input <input id="dialogFirst"></label><br />
+    <button id="closeButton">Close button</button>
```
Jym77, line 105: Should we have a more real example, e.g. "input your name" or something like that. "Some input" doesn't sound like a good label name (which this rule doesn't check, but it really add no value to the example to have placeholder text).

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -169,6 +180,25 @@ This `summary` element is [focusable][].
 </details>
 ```
 
+#### Failed Example 7
+
+This `a` element is [focusable][] because it fails to move focus when it receives focus. This is in contrast to Passed Example 4.
```
Jym77, line 185: We usually do not refer from one example to an other.
Notably because that would easily become a nightmare to maintain if some example number changes and we have to remember another example description 😖

_rules/aria-hidden-no-focusable-content-6cfa84.md
Jym77: Technically, the link does not need to be invisible. It only needs to be `aria-hidden` to trigger the rule 🤔

Jym77 (dismissed):

Jym77:
Looks like this resolves #1811. Can you add the link to the initial description?

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: @Jym77 The intent of this was to be more of a real-world dialog example. This is how we normally see dialog components built where they add this invisible, yet focusable sentinel that upon receiving focus wraps focus. I skipped the top sentinel in the interest of simplicity.

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: Will change to "First and last name"

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: Will change to "This is in contrast to a focus sentinel that immediately jumps focus to a valid location." Is that better?

_rules/aria-hidden-no-focusable-content-6cfa84.md
tbostic32: While I understand the appeal of realism, I'm wondering if having the sentinel be visible after the dialog box and have the words like `When focus is placed here it will wrap to the dialog box` to make it even more obvious and self-documenting.

tbostic32 (approved):
I think we could have discussions about the overall styling choices, but I think this is already an improvement as is.

_rules/aria-hidden-no-focusable-content-6cfa84.md
WilcoFiers: Conclusion from TF call: needs more of an example description.

tombrunet:
@WilcoFiers @Jym77 Adjusted Pass 4 / Fail 7 to be more descriptive. Let me know if that helps.

WilcoFiers (approved):

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -169,6 +184,27 @@ This `summary` element is [focusable][].
 </details>
 ```
 
+#### Failed Example 7
+
+This `a` element is [focusable][] because it fails to move focus when it receives focus. This is in contrast to a focus sentinel that 
+immediately jumps focus to a valid location. Focus sentinels are typically used before and after modal dialogs in order to contain
```
WilcoFiers, line 190: I don't think "dialogs" is a word. If you want to use it, add it to `__tests__/spelling-ignore.yml` so the test doesn't fail it.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -97,13 +97,28 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
 
 #### Passed Example 4
 
-This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus.
+This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus. These elements
+are sometimes referred to as 'focus sentinel' or 'bumper'. They are typically found before and after a modal / dialog in 
+order to contain focus within the modal. Page authors do not want the sentinel to be visible, nor do they want them to be read by
+a screen reader. But, they do want the element to be part of the [sequential focus navigation][]. This allows the page author
```
HelenBurge, line 103: The following definition needs to be defined at the bottom of the file to stop the failing checks
> [sequential focus navigation][]

HelenBurge (changes requested):

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: Done

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: Changed to "after a modal dialog"

colabottles (approved):

Issue #1807, opened 2022-03-07
Should negative tabindex fail in aria-hidden?


WilcoFiers:
This example was recently added to the `Element with aria-hidden has no focusable content` rule:

https://act-rules.github.io/rules/6cfa84#failed-example-5

```html
<div aria-hidden="true">
	<button tabindex="-1">Some button</button>
</div>
```

I didn't realise this when it went in, but looking at implementing this I can't quite figure out why this would fail SC 4.1.2. This is not a button in any sort of meaningful way. The only way I can see to get to it is with VoiceOver in reading mode. But it then announces the text node, not the button. It's explicitly called "some button text".

I think the expectation needs to be adjusted to only apply to elements in sequential focus navigation that do not redirect focus.

straker:
The pattern to put `aria-hidden` on a parent element and `tabindex=-1` on all focusable descendant elements is a common pattern for hiding main content when a modal is open. The `inert` polyfill relies on this pattern. If the test case shown is truly a fail in terms of accessibility, is there an alternative accessible pattern to doing this?

PR #175, opened 2018-07-10
Rule: SC4-1-2-aria-hidden-focus

WilcoFiers:
# Checklist for Pull Request
🚨Please review the [guidelines for contributing](https://auto-wcag.github.io/auto-wcag/pages/contribute.html) to this repository.

- [ ] Make sure you are requesting to **pull a issue/feature/bugfix branch** (right side). Don't request your master branch!
- [ ] Make sure you are making a pull request against the **master branch** (left side). Also you should start *your branch* off *master branch*.
- [ ] Kindly prefix the pull request name with `DRAFT` or `FINAL` accordingly. For more details, read [Rule Design Guidelines](https://auto-wcag.github.io/auto-wcag/pages/design/review.html)
- [ ] Provide a reference to the issue being addressed.

#Reviewing Guidelines
- For `DRAFT` PR's, leave any comments you have and approve if you’ve reviewed the entire pull request. 
- Once there have been three such reviews the pull request can be merged. 
- Any comments left will be taken up in an updated pull request, which will be necessary before a FINAL release”.
- For `FINAL` PR's, review and approve changes to be merged.

# How to Review And Approve
- Go to the “files changed” tab, there you will have the option to leave comments on different lines. 
- Once the review is completed, find the “Review changes” button in the top right, select “approve” and click “Submit review”

### Description
Please describe your pull request.

💔Thank you!

PR #1820, opened 2022-03-24
6cfa84: Change aria-hidden rule to focus on tabbable rather than focusable

tombrunet:
Per Task Force discussions:
focusable, by spec, includes elements with `tabindex="-1"` because they can be programmatically focused, but also browsers allow them to be clickable and focused. The more significant issue is items that are tabbable within content that is `aria-hidden`.

Proposing a change in language that would focus on items that are tabbable.

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-hidden-no-focusable-content-6cfa84.md
+++ b/_rules/aria-hidden-no-focusable-content-6cfa84.md
@@ -1,9 +1,9 @@
 ---
 id: 6cfa84
-name: Element with `aria-hidden` has no focusable content
+name: Element with `aria-hidden` has no content in sequential focus navigation
 rule_type: atomic
 description: |
-  This rule checks that elements with an `aria-hidden` attribute do not contain focusable elements.
+  This rule checks that elements with an `aria-hidden` attribute do not contain elements that are part of the sequential focus navigation.
 accessibility_requirements:
   wcag20:4.1.2: # Name, Role, Value (A)
     forConformance: true
@@ -32,11 +32,11 @@ This rule applies to any element with an `aria-hidden` [attribute value][] of `t
 
 ## Expectation
 
-None of the target elements are [focusable][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [focusable][].
+None of the target elements are part of the [sequential focus navigation][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are part of the [sequential focus navigation][], unless, upon receiving focus, focus is moved to an element that does not fail this rule.
 
 ## Assumptions
 
-Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and not [focusable][] elements become [focusable][].
+Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and elements not part of the [sequential focus navigation][] become part of the [sequential focus navigation][].
 
 ## Accessibility Support
 
@@ -48,7 +48,7 @@ Using `aria-hidden="false"` on a descendant of an element with `aria-hidden="tru
 
 By adding `aria-hidden="true"` to an element, content authors ensure that assistive technologies will ignore the element. This can be used to hide parts of a web page that are [pure decoration](https://www.w3.org/TR/WCAG21/#dfn-pure-decoration), such as icon fonts - that are not meant to be read by assistive technologies.
 
-A [focusable][] element with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making its state of [visible](#visible) or hidden unclear.
+An element that is part of the [sequential focus navigation][] with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making its state of [visible](#visible) or hidden unclear.
 
 The 1 second time span introduced in the exception of the definition of [focusable][] is an arbitrary limit which is not included in WCAG. Given that scripts can manage the focus state of elements, testing the focused state of an element consistently would be impractical without a time limit.
 
@@ -71,7 +71,7 @@ The 1 second time span introduced in the exception of the definition of [focusab
 
 #### Passed Example 1
 
-This `p` element is not [focusable][].
+This `p` element is not part of the [sequential focus navigation][].
 
 ```html
 <p aria-hidden="true">Some text</p>
@@ -79,7 +79,7 @@ This `p` element is not [focusable][].
 
 #### Passed Example 2
 
-This `a` element is not [focusable][] because it is hidden through CSS.
+This `a` element is not part of the [sequential focus navigation][] because it is hidden through CSS.
 
 ```html
 <div aria-hidden="true">
@@ -89,7 +89,7 @@ This `a` element is not [focusable][] because it is hidden through CSS.
 
 #### Passed Example 3
 
-This `input` element is not [focusable][] because of the `disabled` attribute.
+This `input` element is not part of the [sequential focus navigation][] because of the `disabled` attribute.
 
 ```html
 <input disabled aria-hidden="true" />
@@ -97,7 +97,7 @@ This `input` element is not [focusable][] because of the `disabled` attribute.
 
 #### Passed Example 4
 
-This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus.
+This `a` element is part of the [sequential focus navigation][], but moves focus to the `input` element whenever it receives focus.
 
 ```html
 <div aria-hidden="true">
@@ -110,7 +110,7 @@ This `a` element is not [focusable][] because it moves focus to the `input` elem
 
 #### Failed Example 1
 
-This `a` element positioned off screen is [focusable][] using the keyboard.
+This `a` element positioned off screen is part of the [sequential focus navigation][] using the keyboard.
 
 ```html
 <div aria-hidden="true">
@@ -120,7 +120,7 @@ This `a` element positioned off screen is [focusable][] using the keyboard.
 
 #### Failed Example 2
 
-This `input` element is [focusable][] because it was incorrectly disabled.
+This `input` element is part of the [sequential focus navigation][] because it was incorrectly disabled.
 
 ```html
 <div aria-hidden="true">
@@ -130,7 +130,7 @@ This `input` element is [focusable][] because it was incorrectly disabled.
 
 #### Failed Example 3
 
-This `button` element is [focusable][] and a descendant of an element with an `aria-hidden` [attribute value][] of `true` because `aria-hidden` can't be reset once set to true on an ancestor.
+This `button` element is part of the [sequential focus navigation][] and a descendant of an element with an `aria-hidden` [attribute value][] of `true` because `aria-hidden` can't be reset once set to true on an ancestor.
 
 ```html
 <div aria-hidden="true">
@@ -142,7 +142,7 @@ This `button` element is [focusable][] and a descendant of an element with an `a
 
 #### Failed Example 4
 
-This `p` element is [focusable][] because of the `tabindex` attribute.
+This `p` element is part of the [sequential focus navigation][] because of the `tabindex` attribute.
 
 ```html
 <p tabindex="0" aria-hidden="true">Some text</p>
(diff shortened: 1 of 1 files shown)


tombrunet:
Update to Pass 4 is handled in https://github.com/act-rules/act-rules.github.io/pull/1819

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -32,11 +32,11 @@ This rule applies to any element with an `aria-hidden` [attribute value][] of `t
 
 ## Expectation
 
-None of the target elements are [focusable][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [focusable][].
+None of the target elements are part of the [sequential focus navigation][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are part of the [sequential focus navigation][], unless, upon receiving focus, focus is moved to an element that does not fail this rule.
```
Jym77, line 35: We might be able to rephrase like
```suggestion
None of the target elements or their [descendants][] in the [flat tree][] are part of the [sequential focus navigation][], unless, upon receiving focus, focus is moved to an element that does not fail this rule.
```
or maybe
```suggestion
None of the target elements has [inclusive descendants][] in the [flat tree][] that are part of the [sequential focus navigation][],  upon receiving focus, focus is moved to an element that does not fail this rule.
```
which does keep the text more "focused" on the targets.

Jym77, line 35: I'm not sure about how the redirect exception is written currently.
```html
<div aria-hidden="true">
  <a href="…" onfocus='document.getElementById("next").focus()'>Redirect</a>
  <a href="…" id="next" tabindex="-1">Not tabbable?</a>
</div>
```
would pass the rule since the tabbable element redirect focus to an element that "does not fail the rule" (since it's not tabbable, hence ignored by the rule)

Even worse, the rule targets the `div` (the element with `aria-hidden`), not the links. So even if the second `a` was tabbable, it would "not fail the rule" since only the `div` can fail the rule.

What about "focus is moved to an element which is [included in the accessibility tree][]"?

Jym77, line 35: There is also a full can of worms around what "upon receiving focus" actually means 🙈 
Is it good if the redirection happens after 5 seconds? 0.5s? 0.5ms?
See also [this discussion](https://github.com/act-rules/act-rules.github.io/pull/1444#discussion_r761084251).

Since we somewhat have resolved this point in earlier definition, I'd suggest something like
```suggestion
None of the target elements are part of the [sequential focus navigation][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are part of the [sequential focus navigation][], unless, they are not [focusable][].
```
This is not perfect  since it doesn't look at where the focus is redirect. OTOH it would also accept elements that self-blur (without focusing anything else).

Jym77, line 35: To sum up all in a single change proposal:
```suggestion
None of the target elements has [inclusive descendants][] in the [flat tree][] that are [focusable][] and part of the [sequential focus navigation][].
```
Should catch almost the same thing and be unambiguous.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -32,11 +32,11 @@ This rule applies to any element with an `aria-hidden` [attribute value][] of `t
...
 
 ## Assumptions
 
-Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and not [focusable][] elements become [focusable][].
+Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and elements not part of the [sequential focus navigation][] become part of the [sequential focus navigation][].
```
Jym77, line 39: We can probably keep this as is if using "focusable and part of sequential focus navigation" in expectation.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -48,7 +48,7 @@ Using `aria-hidden="false"` on a descendant of an element with `aria-hidden="tru
 
 By adding `aria-hidden="true"` to an element, content authors ensure that assistive technologies will ignore the element. This can be used to hide parts of a web page that are [pure decoration](https://www.w3.org/TR/WCAG21/#dfn-pure-decoration), such as icon fonts - that are not meant to be read by assistive technologies.
 
-A [focusable][] element with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making its state of [visible](#visible) or hidden unclear.
+An element that is part of the [sequential focus navigation][] with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making its state of [visible](#visible) or hidden unclear.
```
Jym77, line 51: I'm actually not sure what this AS note meant in the first place 🤔 Notably what does visibility has to do with `aria-hidden`. Was it supposed to referred to [programatically hidden](https://act-rules.github.io/glossary/#programmatically-hidden) rather than visibility?

Seems to have been here [from the start of the rule](https://github.com/act-rules/act-rules.github.io/pull/175/files#diff-695b07cb3f307e2d42f30d9af89d1ba529ab49d749f451d4f8738edc0ee767d6R41), though… So not related to this PR at all.

Jym77 (dismissed):

Jym77:
Also, I think this resolves #1807.

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: I agree. Will fix.

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: My reading of this is that it would be reachable by a screen reader user when tabbing, but they wouldn't find it when using other reading modes (item navigation). visible is probably the wrong word. "discoverable" maybe?

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -32,11 +32,11 @@ This rule applies to any element with an `aria-hidden` [attribute value][] of `t
 
 ## Expectation
 
-None of the target elements are [focusable][], nor do they have [descendants](https://dom.spec.whatwg.org/#concept-tree-descendant) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) that are [focusable][].
+None of the target elements has [inclusive descendants][] in the [flat tree][] that are [focusable][] and part of the [sequential focus navigation][].
 
 ## Assumptions
 
-Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and not [focusable][] elements become [focusable][].
+Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden and elements not [focusable][] and part of the [sequential focus navigation][] become [focusable][] and part of the [sequential focus navigation][].
```
tbostic32, line 39:
```suggestion
Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden elements  that were not [focusable][] or part of the [sequential focus navigation][] become [focusable][] and part of the [sequential focus navigation][].
```

tbostic32: The example here might need some work, it seems a little wordy right now with how `focusable` and `sequential focus navigation` are currently included in my opinion.

tbostic32 (dismissed):
We need a little wording change in the assumptions section to clean things up, otherwise I think this PR looks good. The issue @Jym77 brought up is something we should look at and probably deal with in a separate issue.

_rules/aria-hidden-no-focusable-content-6cfa84.md
WilcoFiers: Lets try to elaborate on this a bit. It can certainly be clearer.

tombrunet: Part of this was already said in the prior sentence. I reworded this to tie those two together better and elaborate slightly. Let me know if that makes more sense.

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -48,7 +48,7 @@ Using `aria-hidden="false"` on a descendant of an element with `aria-hidden="tru
 
 By adding `aria-hidden="true"` to an element, content authors ensure that assistive technologies will ignore the element. This can be used to hide parts of a web page that are [pure decoration](https://www.w3.org/TR/WCAG21/#dfn-pure-decoration), such as icon fonts - that are not meant to be read by assistive technologies.
 
-A [focusable][] element with `aria-hidden="true"` is ignored as part of the reading order, but still part of the focus order, making its state of [visible](#visible) or hidden unclear.
+However, if the element is also part of the [sequential focus navigation][], the assistive technology will want to announce the element where focus has landed. If focus lands on an element with `aria-hidden="true"`, the element's state of decorative vs non-decorative and [visible](#visible) vs hidden unclear.
```
kengdoj, line 51: Is decorative/non-decorative a state? 

Also, end of last sentence should be "is unclear" - a word is missing.

kengdoj (changes requested):
Just a question for line 51.

_rules/aria-hidden-no-focusable-content-6cfa84.md
HelenBurge: Also need to define 

> [sequential focus navigation][] 

at the bottom of the file

HelenBurge (dismissed):
as per my comment - add a definition to [sequential focus navigation][]

_rules/aria-hidden-no-focusable-content-6cfa84.md
tombrunet: @HelenBurge I see that already defined in this PR - it was missing in the other PR. I copied the definition from the other definitions - did I do something wrong there?

tombrunet: @kengdoj "state" may be the wrong word, but I'm not sure of a better one offhand. `aria-hidden` is intended to indicate that the content is decorative. If it's focusable, it's interactive and therefore implied that it's not decorative. Would it be better to say "If an element with `aria-hidden="true"` can be tabbed to, it is unclear if that content is intended to be decorative or not"?

tombrunet: @HelenBurge Disregard. I think you meant the other two definitions I missed. I'll get those.

kengdoj: How about:
An element with an aria-hidden attribute set to true that is also part of the [sequential focus navigation][] may cause confusion for users of assistive technologies because the element should be hidden and not [included in the accessibility tree][].

This is the [definition of included in the accessibility tree](https://act-rules.github.io/glossary/#included-in-the-accessibility-tree): 

> [Programmatically hidden](https://act-rules.github.io/glossary/#programmatically-hidden) elements are removed from the accessibility tree. However, some browsers will leave [focusable](https://act-rules.github.io/glossary/#focusable) elements with an aria-hidden attribute set to true in the accessibility tree. Because they are hidden, these elements are considered not included in the accessibility tree. This may cause confusion for users of assistive technologies because they may still be able to interact with these focusable elements using sequential keyboard navigation, even though the element should not be included in the accessibility tree.

You can see the discussion about the wording of this in https://github.com/act-rules/act-rules.github.io/pull/1684

tombrunet: @kengdoj Does this work?
> An element with an `aria-hidden` attribute set to `true` that is also part of the [sequential focus navigation][] may cause confusion for users of assistive technologies because the element can be reached via [sequential focus navigation][], but it should be hidden and not [included in the accessibility tree][].

kengdoj: yes!

kengdoj (approved):
I also agree with @tbostic32 suggested change (on Apr 6).

HelenBurge (approved):

_rules/aria-hidden-no-focusable-content-6cfa84.md
tbostic32:
```suggestion
Interacting with the page does not result in changing the `aria-hidden` [attribute value][] of target elements. An example of such a situation would be when closing a modal dialog makes previously hidden elements not [focusable][] and part of the [sequential focus navigation][] become [focusable][] and part of the [sequential focus navigation][].
```

colabottles (approved):

PR #1821, opened 2022-03-25
Add note to button has accessible name rule to explain the exception #97a4e1

carlosapaduarte:
The [CG decided](https://www.w3.org/2022/03/24-act-r-minutes.html#item04) to add a note explaining the reason for the image buttons exception in the [button has non-empty accessible name ](https://act-rules.github.io/rules/97a4e1)rule.

Need for Call for Review:

---

## How to Review And Approve

### DIFF
--- a/_rules/button-non-empty-accessible-name-97a4e1.md
+++ b/_rules/button-non-empty-accessible-name-97a4e1.md
@@ -42,6 +42,8 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


Jym77 (approved):

HelenBurge (approved):

WilcoFiers (approved):

carlosapaduarte:
Call for review ends May 23

colabottles (approved):

PR #1822, opened 2022-03-31
Editorial updates to the meta-viewport rule

WilcoFiers:
These changes are intended to be editorial. It does break up the expectation and removes the "property/value pairs in a user-agent specific manner" sentence. That seemed unnecessary to me.

---

## How to Review And Approve

### DIFF
--- a/_rules/meta-viewport-b4f0c3.md
+++ b/_rules/meta-viewport-b4f0c3.md
@@ -261,7 +261,6 @@ This viewport `meta` element does not have a `content` attribute.
 </html>
 ```
 
-[case-insensitive]: https://infra.spec.whatwg.org/#ascii-case-insensitive 'ASCII case-insensitive'
 [content]: https://www.w3.org/TR/WCAG21/#dfn-content 'content (Web content)'
 [maximum-scale]: https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale 'The initial-scale, minimum-scale, and maximum-scale properties'
 [meta]: https://html.spec.whatwg.org/#the-meta-element 'The meta element'
--- a/_rules/meta-viewport-b4f0c3.md
+++ b/_rules/meta-viewport-b4f0c3.md
@@ -22,22 +22,23 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to each `meta` element with a `name` attribute whose value is a [case-insensitive][] match for `viewport` and has a `content` attribute.
+This rule applies to each `content` attribute on a `meta` element with a `name` [attribute value][] of `viewport`.
 
-## Expectation
+## Expectation 1
 
-For each test target, the `content` attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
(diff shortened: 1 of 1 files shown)


_rules/meta-viewport-b4f0c3.md
```diff
@@ -22,22 +22,23 @@ acknowledgments:
...
+## Expectation 1
 
-For each test target, the `content` attribute, whose value is mapped to a list of property/value pairs in a user-agent specific manner, does not:
+For each test target, the [attribute value][] does hot have a `user-scalable` property with a value of `no`.
 
-- specify the property `user-scalable` with a value of `no`; nor
-- specify the property `maximum-scale` with a value of less than 2.
+## Expectation 2
+
+For each test target, the [attribute value][] does hot have a `maximum-scale` property with a value less than 2.
```
kengdoj, line 33:
```suggestion
For each test target, the [attribute value][] has a `maximum-scale` property with a value of 2 or more.
```

JenniferChadwick (approved):

_rules/meta-viewport-b4f0c3.md
Jym77: There's a bit of a weird trick happening here, according to [this bit of non-normative CSS specs](https://drafts.csswg.org/css-device-adapt/#min-scale-max-scale).
Namely, `maximum-scale` can be a number, or the keywords `yes`, `device-width`, `device-height`, `no` (or some unknown value). These values are then translated into a number for the [`max-zoom` descriptor](https://drafts.csswg.org/css-device-adapt/#max-zoom-desc) of `@viewport` CSS rules.

[`user-scalable`](https://drafts.csswg.org/css-device-adapt/#user-scalable) has similar tricks.

On `user-scalable`, the rule explicitly mention that `no` is a possible value (rather than using the keywords of the `user-zoom` descriptor). So I'd rather have `maximum-scale` also acknowledge that `yes` is good.
This is not really a problem with this PR, given that it was already present in the previous version, so I'm accepting this.

---
Thus said, this also hints that we should probably rewrite (or replace) the rule to look into `@viewport` descriptors, **including the `meta viewport` mappings** rather than only look at the `meta viewport`. This would make the rule a bit more generic at relatively low cost 🤞

Jym77 (approved):

HelenBurge (approved):

_rules/meta-viewport-b4f0c3.md
WilcoFiers: @kengdoj The use of negative language is intentional. This expectation is true if there is no maximum-scale too. In your rewrite that wouldn't be the case.

patrickhlauke (approved):

carlosapaduarte (approved):

dd8:
Typo:

> does hot have

should be

> does not have

colabottles (approved):

_rules/meta-viewport-b4f0c3.md
WilcoFiers, line 33:
```suggestion
For each test target, the [attribute value][] does not have a `maximum-scale` property with a value less than 2.
```

PR #1826, opened 2022-04-07
form label descriptive: Tweak the examples

WilcoFiers:
Particularly failed example 5 I think is a little more difficult than it needs to be. If I encounter a form field that has a label above it saying "shipping address" I think the logical conclusion there is that it's an address field. That there's a hidden "name" field also makes this a little more confusing. By just using "shipping", it's clearer that information is missing.

Alternatively, or maybe additionally, we could add a second field with a hidden label to failed example 5?

---

## How to Review And Approve

### DIFF
--- a/_rules/form-field-label-descriptive-cc0f0a.md
+++ b/_rules/form-field-label-descriptive-cc0f0a.md
@@ -140,11 +140,11 @@ The `label` elements are [programmatic labels][programmatic label] of their resp
 
 ```html
 <html lang="en">
-	<h2>Shipping address</h2>
+	<h2>Shipping</h2>
 	<label>Name<input id="shipping-name" type="text" name="name"/></label>
 	<label>Street<input id="shipping-street" type="text" name="street"/></label>
 
-	<h2>Billing address</h2>
+	<h2>Billing</h2>
 	<label>Name<input id="billing-name" type="text" name="name"/></label>
 	<label>Street<input id="billing-street" type="text" name="street"/></label>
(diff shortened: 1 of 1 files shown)


_rules/form-field-label-descriptive-cc0f0a.md
```diff
@@ -218,7 +218,7 @@ Both the `div` and the `span` elements are [programmatic labels][programmatic la
 
 ```html
 <html lang="en">
-	<div id="shipping">Shipping address</div>
+	<div id="shipping">Shipping</div>
 	<span id="name" style="display: none">Name</span>
 	<input id="shipping-name" type="text" name="name" aria-labelledby="shipping name" />
 </html>
```
Jym77, line 224: Maybe the way to make this example better is by adding another field to create the confusion:
```html
<html lang="en">
	<div id="shipping">Shipping</div>
	<span id="name" style="display: none">Name</span>
	<input id="shipping-name" type="text" name="name" aria-labelledby="shipping name" />
	<span id="address" style="display: none">Address</span>
	<input id="shipping-address" type="text" name="address" aria-labelledby="shipping address" />
</html>
```
Now, it is clear that both input have the same visible context ("Shipping") which is not descriptive enough, even though the programmatic context (accessible names) are good enough.

Jym77 (approved):
Looks good. Possible improvement.

HelenBurge (approved):

wcreedle (approved):

PR #1827, opened 2022-04-07
Heading descriptive: Fix some of the examples

WilcoFiers:
- There's a typo in one of the attributes which caused the image not to load
- I don't think there's a good reason for us to hide that paragraph in example 6. It that test more complicated to understand / interpret than it needs to be

---

## How to Review And Approve

### DIFF
--- a/_rules/heading-descriptive-b49b2e.md
+++ b/_rules/heading-descriptive-b49b2e.md
@@ -108,7 +108,7 @@ Heading marked up with `h1` element with an image that describes the topic or pu
 ```html
 <html lang="en">
 	<h1>
-		<img scr="/test-assets/descriptive-heading-b49b2e/opening_hours_icon.png" alt="Opening hours" />
+		<img src="/test-assets/descriptive-heading-b49b2e/opening_hours_icon.png" alt="Opening hours" />
 	</h1>
 	<p>We are open Monday through Friday from 10 to 16</p>
 </html>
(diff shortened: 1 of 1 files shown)


_rules/heading-descriptive-b49b2e.md
```diff
@@ -136,12 +136,12 @@ Heading marked up with `h1` element that is a single character that describes th
...
 #### Passed Example 6
 
-Heading marked up with `role="heading"` that describes the topic or purpose of the following [palpable content][]. The heading is positioned off screen and is [included in the accessibility tree][].
+Heading marked up with `role="heading"` that describes the topic or purpose of the following [palpable content][]. The heading is positioned off screen but is [included in the accessibility tree][].
 
 ```html
 <html lang="en">
 	<span role="heading" aria-level="1" style="position: absolute; top: -9999px; left: -9999px;">Opening Hours</span>
-	<p style="position: absolute; top: -9999px; left: -9999px;">
+	<p>
```
Jym77, line 144: I'm not fully sure either why this (and FE3) were having the paragraph off-screen. I guess the idea was that both the heading and its "section" were `sr-only` 🤔 Might be a bit borderline for 2.4.10 otherwise (since the visual content does not have headings to organise content, it's OK for a 2 lines toy examples but would be bad if there were 5 or 6 of these in a page).

OTOH, if that's the intention, it might be cleaner and clearer to wrap both the heading and the paragraph into the sami `div` and move that `div` off screen instead…

Anyway, no real opposition to this change.

Jym77 (approved):

daniel-montalvo (approved):

wcreedle (approved):

carlosapaduarte (approved):

PR #1829, opened 2022-04-08
Element lang correct: remove inapplicable examples

WilcoFiers:
This PR removes inapplicable examples from the `HTML element language subtag matches language` rule, which according to `Element with lang attribute has valid language tag` should be considered failures of 3.1.2.

Two of those examples are literal failed examples of the lang attribute valid rule.

AGWG has agreed that those examples are failures of SC 3.1.2, so we can't use them as inapplicable examples here. I'm open to suggestions for alternative examples that would not fail other rules.

---

## How to Review And Approve

### DIFF
--- a/_rules/element-lang-matches-default-language-off6ek.md
+++ b/_rules/element-lang-matches-default-language-off6ek.md
@@ -271,20 +271,6 @@ There is no [descendant of a `body`](#off6ek:in-body) element with a `lang` attr
 
 #### Inapplicable Example 3
 
-This `p` element does not have a [valid language tag](#off6ek:valid-lang).
-
-```html
-<html lang="en">
-	<body>
-		<p lang="français">
-			I love ACT rules!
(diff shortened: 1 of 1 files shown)


Jym77:
> Two of those examples are literal failed examples of the lang attribute valid rule.
> 
> AGWG has agreed that those examples are failures of SC 3.1.2, so we can't use them as inapplicable examples here. I'm open to suggestions for alternative examples that would not fail other rules.

Maybe we should just add "Related rules" links between these two rules, the same way we did for the three "document lang" rules 🤔

daniel-montalvo (approved):
Tests complain that we are defining "language subtag registry" but we are not using it as we have removed the examples where we used it.

In any case, approving.

tombrunet (approved):

carlosapaduarte (approved):

PR #1756, opened 2021-11-21
Remove 2nd WCAG reference: 1.4.10

HelenBurge:
Removed 1.4.10 as zoom restrictions are easier to note with 1.4.4

Closes issue(s):

N/A

Need for Call for Review:
This will require a 1 week Call for Review

PR #1834, opened 2022-04-14
Revert "Remove 2nd WCAG reference: 1.4.10"

HelenBurge:
Reverts act-rules/act-rules.github.io#1756

### DIFF
--- a/_rules/meta-viewport-b4f0c3.md
+++ b/_rules/meta-viewport-b4f0c3.md
@@ -10,6 +10,11 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:1.4.10: # Reflow (AA)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
 input_aspects:
   - DOM Tree
 acknowledgments:


WilcoFiers (approved):

JenniferChadwick (approved):

wcreedle (approved):

kengdoj (approved):

Jym77 (approved):

patrickhlauke:
Most (all?) browsers on mobile/tablet don't do reflow when zooming, so not sure how these are related?

WilcoFiers:
@patrickhlauke That's an interesting point. Are you saying that because mobile browsers don't reflow when you zoom anyway, preventing zoom isn't a violation of the reflow SC? That's not an angle I had considered.

What does that mean for 1.4.10 reflow in relation to mobile sites? Do those just fail because mobile browsers don't reflow when you zoom, or is it simply not appropriate to test 1.4.10 if you're only targeting browsers that can't reflow in the first place?

patrickhlauke:
1.4.10 is certainly related/dependent on the ability to zoom or resize the browser/viewport (to reach the 320 CSS px mark for vertical scrolling / 256 CSS px for horizontal scrolling content, which is normatively the only size at which reflow needs to pass - which is a problem in its own right, see https://github.com/w3c/wcag/issues/698). *but* if you can't resize, it doesn't fail 1.4.10 because normatively that SC doesn't say "you must be able to resize to that viewport size". that scenario fails 1.4.4 Resize Text, but not 1.4.10 Reflow.

now on the separate point, 1.4.10 itself can't really be tested on most (practically all, with a few exotic outliers) mobile/tablet OS browsers, since they simply don't reflow. See https://www.w3.org/WAI/WCAG21/Understanding/reflow.html#browsers-on-mobile-operating-systems

WilcoFiers:
@patrickhlauke Thanks for your thoughts. I'll take this back to the ACT Task Force next week. I think you've convinced me. Let me see what the rest thinks.

HelenBurge:
As no longer viable

PR #1846, opened 2022-05-16
Document how to use WCAG SC numbers in EARL

WilcoFiers:
I wanted to provide a way that implementors could map to WCAG criteria without needing to look up WCAG IDs. This describes how to do it. The code for processing them has already been pushed to act-tools.

Need for Call for Review: None, editorial

---

## How to Review And Approve

### DIFF
--- a/pages/implementations/earl-reports.md
+++ b/pages/implementations/earl-reports.md
@@ -111,4 +111,32 @@ For each result, we'll need to know which rule or procedure in the tool reported
 
 **Note**: The IDs of WCAG success criteria can be found in [sc-urls.json](https://github.com/act-rules/act-tools/blob/main/src/data/sc-urls.json), on the `scId` property. These are the IDs introduced in WCAG 2.1, and continued in WCAG 2.2. Use these IDs, even when reporting for WCAG 2.0.
 
+## isPartOf
+
+Alternatively, it is possible to specify WCAG 2 success criteria by number instead of by ID. This can be done in the following format:
+
+```json
+{
+	"@context": "https://act-rules.github.io/earl-context.json",
+	"@graph": [
+		{
+			"@type": "TestSubject",
+			"source": "https://act-rules.github.io/testcases/a1b64e/6c3ac31577c3cb2d968fc26c4075dd533b5513fc.html",
+			"assertions": [
+				{
+					"@type": "Assertion",
+					"result": { "outcome": "earl:passed" },
+					"test": {
+						"title": "image-button-has-name",
+						"isPartOf": [{ "title": "WCAG 2: 1.1.1" }, { "title": "WCAG 2: 4.1.2" }]
+					}
+				}
+			]
+		}
+	]
+}
+```
+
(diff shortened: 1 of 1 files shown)


pages/implementations/earl-reports.md
```diff
@@ -111,4 +111,32 @@ For each result, we'll need to know which rule or procedure in the tool reported
 
 **Note**: The IDs of WCAG success criteria can be found in [sc-urls.json](https://github.com/act-rules/act-tools/blob/main/src/data/sc-urls.json), on the `scId` property. These are the IDs introduced in WCAG 2.1, and continued in WCAG 2.2. Use these IDs, even when reporting for WCAG 2.0.
 
+## isPartOf
```
carlosapaduarte, line 114: Shouldn't this be an `<h4>` instead of an `<h2>`?

carlosapaduarte (approved):

pages/implementations/earl-reports.md
```diff
@@ -111,4 +111,32 @@ For each result, we'll need to know which rule or procedure in the tool reported
...
+						"isPartOf": [{ "title": "WCAG 2: 1.1.1" }, { "title": "WCAG 2: 4.1.2" }]
+					}
+				}
+			]
+		}
+	]
+}
+```
+
+The `title` format is fairly permissive. To be recognized as a success criterion, the string `WCAG 2` or `WCAG2`, and the success criterion number (separated with dots) must be included. For example `WCAG 2.1 criterion 1.1.1 Non-text content` would be recognized. To explicitly add the success criterion ID the `@id` property can be used.
```
Jym77, line 140: I was a bit confused at first because there is both a `test.title` and a `test.isPartOf.title`. I initially thought that "`title`" in this sentence was referring to the former, not the later (because the later is a bit "buried" and I've only seen the former while quickly scanning the code searching for a `title`), and the rest of the sentence made no sense. 

```suggestion
The format of the `title` property in the `isPartOf` array is fairly permissive. To be recognized as a success criterion, the string `WCAG 2` or `WCAG2`, and the success criterion number (separated with dots) must be included. For example `WCAG 2.1 criterion 1.1.1 Non-text content` would be recognized. To explicitly add the success criterion ID the `@id` property can be used.
```

Jym77 (changes requested):
Small clarification, feel free to ignore.

pages/implementations/earl-reports.md
```diff
@@ -111,4 +111,32 @@ For each result, we'll need to know which rule or procedure in the tool reported
 
 **Note**: The IDs of WCAG success criteria can be found in [sc-urls.json](https://github.com/act-rules/act-tools/blob/main/src/data/sc-urls.json), on the `scId` property. These are the IDs introduced in WCAG 2.1, and continued in WCAG 2.2. Use these IDs, even when reporting for WCAG 2.0.
 
+### isPartOf
+
+Alternatively, it is possible to specify WCAG 2 success criteria by number instead of by ID. This can be done in the following format:
+
+```json
+{
+	"@context": "https://act-rules.github.io/earl-context.json",
```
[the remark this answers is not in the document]
WilcoFiers: Good idea. I don't think I can get rid of this one any time soon, but I could probably stop updating it and encourage folks to move to a new one. I'll put it on my todo list, but if you don't mind I'll leave that for a different PR.

Jym77 (approved):

daniel-montalvo (approved):

PR #1848, opened 2022-05-18
presentational children not focusable; fix issue in failed example 1

WilcoFiers:
Because it is missing a role, there are a bunch of things wrong with failed example 1 on this rule. This should have a role, so that presentational children is the only problem on this example.

---

## How to Review And Approve

### DIFF
--- a/_rules/presentational-children-no-focusable-content-307n5z.md
+++ b/_rules/presentational-children-no-focusable-content-307n5z.md
@@ -102,7 +102,7 @@ This `button` element has a [child][] `span` element. Because the `span` element
 ```html
 <button>
 	Save
-	<span aria-label="save options" aria-expanded="false" tabindex="0">▼</span>
+	<span role="button" aria-label="save options" aria-expanded="false" tabindex="0">▼</span>
 </button>
 ```
 


carlosapaduarte (approved):

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -102,7 +102,7 @@ This `button` element has a [child][] `span` element. Because the `span` element
 ```html
 <button>
 	Save
-	<span aria-label="save options" aria-expanded="false" tabindex="0">▼</span>
+	<span role="button" aria-label="save options" aria-expanded="false" tabindex="0">▼</span>
```
Jym77, line 105: I'm not sure about this.

* The `button` element has `button` role, with presentational children. Hence the `span` should be presentational whatever its explicit role. So adding a `role` attribute probably doesn't change anything.
* OTOH, I agree that `aria-expanded` is not a global ARIA attribute, so it shouldn't be used on a roleless element. But adding a `role` attribute shouldn't change that since it should be ignored.

* It seems browser mostly ignore presentational children. In this case, both Chrome and Firefox expose the `span` has a "generic" without `role` attribute, or has a `button` with it. Which goes against the ARIA specs as far as I understand them (but it is a "SHOULD NOT expose descendants"…) This however does mean that this example might not cause any accessibility issue (if the tabbable content is nonetheless exposed). I've encountered similar situations with an `a` element within a `tab` role which was exposed and listed in NVDA list of links. This may be a browser heuristic to ignore presentational children in case of tabbable content, or altogether, or for some roles…

So, it feel there may be some problem with the rule and we may need to investigate UA/AT behaviour further; but this is independent from this specific change, so I'm accepting it.

Jym77 (approved):

colabottles (approved):

Issue #1552, opened 2021-03-05
"ARIA required owned elements" (bc4a75): Passed example 3 is incorrect


kasperisager:
https://act-rules.github.io/testcases/bc4a75/c26f9f87576ea7fbeb2db67a559e78501d1d232a.html

```html
<table role="grid">
	<tr role="row">
		<span role="cell">Item 1</span>
	</tr>
</table>
```

When parsed as HTML, the above turns into the following:

```html
<span role="cell">Item 1</span>
<table role="grid">
  <tbody>
    <tr role="row"></tr>
  </tbody>
</table>
```

That code fails the rule due to the row missing a cell.

kasperisager:
I also couldn't help but notice that both SortSite, QualWeb, and axe-core pass that case. When I run axe-core directly against the test case, however, it does correctly fail, contrary to what's reported in the implementation data. QualWeb passes it for some reason.

WilcoFiers:
I'm not so sure we should make them have required children TBH. Is there anything inaccessible about an empty row?

kasperisager:
Without it, I don't think the rule should exist at all: https://github.com/act-rules/act-rules.github.io/issues/1426#issuecomment-680838986

Jym77:
We should probably only write examples that are correct DOM/HTML. So not an example like that is parsed into a DOM different from what the HTML looks like.

Jym77:
Closed by #1850.

PR #1850, opened 2022-05-19
ARIA required owned elm; fix parser issue in pass 3

WilcoFiers:
Browsers only allow `th` and `td` elements as children of `tr`. The `span` we used in passed example 3 gets booted out by the browser, and is instead rendered above the table inside the DOM tree. If you inspect the DOM of this code example, it'll look like this:

- `<body>`
    - `<span role="cell">`
    - `<table role="grid">`
        - `<tbody>`
            - `<tr role="row">`

Wrapping the span in a td with role=none should fix the issue, and I think, preserve the intent of the example.

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-owned-element-bc4a75.md
+++ b/_rules/aria-required-owned-element-bc4a75.md
@@ -92,12 +92,14 @@ This element with the `tablist` role only owns elements with the `tab` role. The
 
 #### Passed Example 3
 
-This element with the `grid` role only owns elements with the `row` role, and the element with the `row` role only owns elements with the `cell` role. The `row` role is one of the [required owned elements][] for `grid`, and `cell` is one of the [required owned elements][] for `row`.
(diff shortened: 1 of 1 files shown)


Jym77:
This should close #1552. Can you link it from the description?

daniel-montalvo (dismissed):

HelenBurge (approved):

tombrunet:
I'm having a hard time resolving Pass 3 as a good example when it conflicts with the ARIA in HTML guidance - https://www.w3.org/TR/html-aria/#docconformance . The ARIA in HTML group noted that role may not be used on table elements within a table, grid, or treegrid. They had approached us about flagging role on these elements as violations. I think the redundant "row" role is less of an issue than the td with role="none".

colabottles (approved):

scottaohara:
This allowance would require some bugs to be filed to make it work consistently across browsers/AT pairings. Results in codepen from testing this morning.
https://codepen.io/scottohara/pen/poapLqW

tombrunet (dismissed):
See Scott's codepen notes. There are several browser/AT combinations that don't support this.

tombrunet (approved):

HelenBurge (approved):

Jym77 (approved):

colabottles (approved):

carlosapaduarte (approved):

PR #1854, opened 2022-06-01
Removing 1.3.1 mapping from role attribute rule

tbostic32:
- Removing mapping to 1.3.1
- Updated assumptions
- Added link to DPUB in examples

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


HelenBurge (approved):

WilcoFiers (approved):

carlosapaduarte (approved):

Jym77 (commented):
I'm not a fan of removing  the mapping.

I do agree that technically a mispelled `role` attribute, does not always cause a failure of 1.3.1 and therefore as per our own rule format specification we shouldn't have the mapping.

However, a mispelled `role` attribute is still extremely suspicious (at the best) and extremely likely to be an actual mistake that should be fixed.

As a tool vendor, I see a lot of customers that are focused on compliance and start by disabling all rules that are not WCAG A or AA. By removing this mapping, we effectively remove this check for many people, even though mispelled `role` attribute is probably causing problems. Moreover, ignoring this check can make it harder to investigate **other** errors. (e.g. we won't flag an image for missing an accessible name because it has `role="image"` and doesn't trigger [Image has non-empty accessible name](https://act-rules.github.io/rules/23a2a8)…)

---
Of course, the rule still maps to ARIA. But sadly, ARIA is not required for legal compliance that only focuses on WCAG (at least in US and EU), so many of our customers incorrectly believe that ARIA rules are only Best Practice 😢 

Of course, that's a wrong view of things. incorrect ARIA does not necessarily causes WCAG failure but often makes finding and investigating WCAG problems that much harder. (the same way that checking grammar of a text riddled with misspellings is hard)

---
So, on the technical level I do not have an objection to this change.
But on the "moral" level, I feel that we should fin a way to still points out that this rule (and several others) are extremely important and not mere "best practices". I am not sure how to do that.

WilcoFiers:
Closing, we're going a different direction.

PR #1855, opened 2022-06-02
Actioning feedback from Survey on iFrames

### LINKED ISSUE #1843: Iframe negative tabindex; applicability does not work [akn7bn]
In a recent update, the applicability of this rule was changed to:
> This rule applies to any non-focusable `iframe` element that has [focusable][] content.

In the definition of "focusable" there is this exception:

> Exception: Elements that lose focus during a period of up to 1 second after gaining focus, without the user interacting with the page the element is on, are not considered focusable.

Since most browsers redirect focus from iframes, i.e. when the iframe is focused, focus is immediately moved to the first element inside that frame, iframes essentially are never focusable. I.e. in browsers that redirect focus, I think iframes are never focusable. I think we need to return to the previous wording of " a negative number as a `tabindex` attribute value".

**secondly**, there is a mismatch between the applicability and the examples. In particular passed example 1 seems like it should now be inapplicable, since the iframe content does not have focusable elements. I don't really understand why "has focusable content" was added to the applicability. That doesn't seem right.

**Thirdly**: It seems to me that passed example 4 should be inapplicable. In all other rules, content that is both visually hidden and hidden to AT is always inapplicable. Having that pass in this rule seems inconsistent.

HelenBurge:
Updated the content from the issue raised

Closes issue(s):

- closes #1843

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -109,11 +93,26 @@ This `iframe` element does not have a `tabindex` [attribute value][] that is a n
 <iframe tabindex="0" srcdoc="<a href='/'>Home</a>"></iframe>
 ```
 
+#### Inapplicable Example 2
+
+This `iframe` element contains no content that is part of [sequential focus navigation][].
```
Jym77, line 98: The Applicability does not mention "content that is part of sequential focus navigation".
So this example is currently Applicable as it is an `iframe` with negative tabindex, and it passes the Expectation as there is no focusable content in it.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -109,11 +93,26 @@ This `iframe` element does not have a `tabindex` [attribute value][] that is a n
...
+
+```html
+<iframe tabindex="-1" srcdoc="<h1>Hello world</h1>"></iframe>
+```
+
+#### Inapplicable Example 3
+
+This `iframe` element contains no [visible][] content because the iframe is hidden.
```
Jym77, line 106: The applicability does not mention the `iframe` to be visible or not.

Jym77 (dismissed):

giacomo-petri:
Might be an edge case, but if developers would like to move the focus onto the iframe via Javascript thanks to the tabindex="-1" attribute and then, for some reason, would like to trap the user inside using a temporary keyboard trap (e.g. similarly to a modal dialog), based on this rule it should be a failure but the content inside is currently behaving as expected via keyboard.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
HelenBurge: I have switched the expectation and applicability around to help make these valid.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
HelenBurge: I have switched the expectation and applicability around to help make these valid.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,13 @@ acknowledgments:
...
+This rule applies to any `iframe` element that [contains](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
+
+An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
 
 ## Expectation
 
-For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
-
-An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
+For each test target, the [nested browsing context][] does not have a negative number as a `tabindex` [attribute value][].
```
Jym77, line 37: Agree with @giacomo-petri The negative tabindex is set on the `iframe` element, aka the browsing context container, not on the nested browsing context (which is an abstract notion not really "existing" in the HTML code and not having attributes).

```suggestion
None of the test target has a negative number as a `tabindex` [attribute value][].
```

Jym77 (dismissed):

Jym77 (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,36 +59,20 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
...
 
-This `iframe` element contains no content that is part of [sequential focus navigation][].
-
-```html
-<iframe tabindex="-1" srcdoc="<h1>Hello world</h1>"></iframe>
-```
-
-#### Passed Example 2
-
 This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.
```
giacomo-petri, line 62: Should we mark also this example as inapplicable now, due to the new definition "This rule applies to any `iframe` element that [contains](#akn7bn:contain) elements that are both [visible][] and part of the [sequential focus navigation][]."?

We should add instead an example with an iframe with no tabindex attribute (or tabindex="0", or both) and both visible and SFN  content.

giacomo-petri (changes requested):
Should we mark also new Passed Example 1 (This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.) as inapplicable now, due to the new applicability definition "This rule applies to any `iframe` element that [contains](#akn7bn:contain) elements that are both [visible][] and part of the [sequential focus navigation][]."?

Edit: 
I've edited my previous comment removing last paragraph:
"We should add instead an example with an iframe with no tabindex attribute (or tabindex="0", or both) and both visible and SFN  content." 
as it doesn't match neither rule title nor description.

Should we instead add "with negative tabindex value" to the applicability section?
This rule applies to any `iframe` element **_with negative tabindex value_** that [contains](#akn7bn:contain) elements that are both [visible][] and part of the [sequential focus navigation][].

HelenBurge:
> Should we mark also new Passed Example 1 (This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.) as inapplicable now, due to the new applicability definition "This rule applies to any `iframe` element that contains elements that are both [visible][] and part of the [sequential focus navigation][]."?
> 
> Edit: I've edited my previous comment removing last paragraph: "We should add instead an example with an iframe with no tabindex attribute (or tabindex="0", or both) and both visible and SFN content." as it doesn't match neither rule title nor description.
> 
> Should we instead add "with negative tabindex value" to the applicability section? This rule applies to any `iframe` element **_with negative tabindex value_** that contains elements that are both [visible][] and part of the [sequential focus navigation][].

Hi Giacomo, after reviewing your feedback, this is applicable, but not relevant to the work in this pull request. Could you possibly create a new pull request to action your feedback?

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,13 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any non-focusable `iframe` element that has [focusable][] content.
+This rule applies to any `iframe` element that [contains](#akn7bn:contain) elements that are both [visible][] and part of the [sequential focus navigation][].
```
Jym77: Expliciting the logic.
```suggestion
This rule applies to any `iframe` element that [contains](#akn7bn:contain) at least one element for which all the following are true:
- the element is [visible][]; and
- the element is part of the [flattened tabindex-ordered focus navigation scope][] of the `iframe`.
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,13 @@ acknowledgments:
...
 
-## Expectation
+An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][]. The elements (inside the iframe) are part of the [flattened tabindex-ordered focus navigation scope].
```
Jym77:
```suggestion
An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][]. 

The [flattened tabindex-ordered focus navigation scope][] of an `iframe` is essentially the [sequential focus navigation][] order inside its content.
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,54 +59,54 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
 
 #### Passed Example 1
 
-This `iframe` element contains no content that is part of [sequential focus navigation][].
+This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.
 
 ```html
-<iframe tabindex="-1" srcdoc="<h1>Hello world</h1>"></iframe>
+<iframe tabindex="-1" srcdoc="<a href='/' tabindex='-1'>Home</a>"></iframe>
```
Jym77: I agree that with the change, this is now Inapplicable and should be moved accordingly.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,54 +59,54 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
...
 ```
 
-#### Passed Example 2
+### Failed
 
-This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.
+#### Failed Example 1
+
+This `iframe` element has a link that is part of [sequential focus navigation][].
```
Jym77: Maybe we can boost the description with the technical mumbo-jumbo…
```suggestion
This `iframe` element contains a [visible][] link that is part of its [flattened tabindex-ordered focus navigation scope][]. Note that the link is not part of the document's [sequential focus navigation][] because the containing `iframe` is excluded from the document's [tabindex-ordered focus navigation scope](https://html.spec.whatwg.org/multipage/interaction.html#tabindex-ordered-focus-navigation-scope) due to its negative `tabindex`.
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,54 +59,54 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
...
+### Inapplicable
 
-This `iframe` element contains no [visible][] content because of the small size of the iframe.
+#### Inapplicable Example 1
+
+This `iframe` element does not have a `tabindex` [attribute value][] that is a negative number
 
 ```html
-<iframe tabindex="-1" width="1" height="1" srcdoc="<a href='/'>Home</a>"></iframe>
+<iframe tabindex="0" srcdoc="<a href='/'>Home</a>"></iframe>
```
Jym77: This one has no become a Passed example since the iframe has visible and tabbable content.

Jym77 (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,54 +63,54 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
...
+<iframe tabindex="-1" srcdoc="<a href='/'>Home</a>"></iframe>
 ```
 
-#### Passed Example 3
+### Inapplicable
 
-This `iframe` element contains no [visible][] content because of the small size of the iframe.
+#### Inapplicable Example 1
+
+This `iframe` element contains no content that is part of [sequential focus navigation][].
```
Jym77:
```suggestion
This `iframe` element contains no content that is part of its [flattened tabindex-ordered focus navigation scope][].
```
(in Failed Example 1, the link is part of the FTOFNS (!) but not of the sequential focus order, due to the negative tabindex of the iframe; the difference is that here there is nothing in th FTOFNS of the iframe, so we don't care about the negative tabindex).

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,54 +63,54 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
...
+<iframe tabindex="-1" width="1" height="1" srcdoc="<a href='/'>Home</a>"></iframe>
 ```
 
-### Inapplicable
-
-#### Inapplicable Example 1
+#### Inapplicable Example 4
 
-This `iframe` element does not have a `tabindex` [attribute value][] that is a negative number
+This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.
```
Jym77:
```suggestion
This `iframe` element contains a link that is not part of its [flattened tabindex-ordered focus navigation scope][] because of its `tabindex`.
```

Jym77 (changes requested):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,17 @@ acknowledgments:
...
-## Expectation
+An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][]. 
+
+The [flattened tabindex-ordered focus navigation scope][] of an `iframe` is essentially the [sequential focus navigation][] order inside its content.
 
-For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
+## Expectation
 
-An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
+None of the test target has a negative number as a `tabindex` [attribute value][].
```
Jym77: I'm actually considering whether we should rather request the iframe to be part of the tab-order of the top-level document. That would be a bit more generic for ways other than negative tabindex  to remove stuff from tab-order. But it also seems that currently negative tabindex (on the iframe) is the only way to do so (for tabbable inside an iframe). So that would complicate the rule with technical mumbo-jumbo without really improving it 🤔

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -59,54 +63,54 @@ By setting the `tabindex` [attribute value][] of an `iframe` element to `-1` or
...
+<iframe tabindex="0" srcdoc="<a href='/'>Home</a>"></iframe>
 ```
 
-#### Passed Example 2
+### Failed
 
-This `iframe` element contains a link that is not part of [sequential focus navigation][] because of its `tabindex`.
+#### Failed Example 1
+
+This `iframe` element contains a [visible][] link that is part of its [flattened tabindex-ordered focus navigation scope][]. Note that the link is not part of the document's [sequential focus navigation][] because the containing `iframe` is excluded from the document's [tabindex-ordered focus navigation scope](https://html.spec.whatwg.org/multipage/interaction.html#tabindex-ordered-focus-navigation-scope) due to its negative `tabindex`.
```
Jym77: I'm also considering whether that inlined note should be made into a full Background note (slightly changed), since it explain the full intend of the rule: stuff that is in the FTOFNS of an iframe should be in the tab-order of the web page.

Jym77 (commented):
Couple of more generic comments. Not sure about them. Leaving them to see what others thing about them.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
HelenBurge: I think this might be something we can discuss in the next meeting?

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
HelenBurge: I think this might be something we can discuss in the next meeting?

Jym77 (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
daniel-montalvo: +1 to adding this as a background note

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
daniel-montalvo: +1 to keeping as-is, if we are going to complicate things unnecessarily.

daniel-montalvo (dismissed):
Thanks @HelenBurge  and @Jym77  for your work on this.

I lean towards agreeing with both comments from Jean-Yves.

Let's see what others think of this.

Jym77:
I've:
* Updated the title and description of the rule to match the new content.
* Beefed up the background note with technical mumbo-jumbo, cleaned up the description of Failed Example.
* Added a Passed example where there is no `tabindex` on the `iframe` (we do not want to imply that all `iframe` need a `tabindex`).

---
I've ultimately decided to leave the Expectation about "negative `tabindex`".

The local tab-order ([tabindex-ordered focus navigation scope](https://html.spec.whatwg.org/multipage/interaction.html#tabindex-ordered-focus-navigation-scope)) contain both focusable areas (second point) + all [focus navigation scope owners](https://html.spec.whatwg.org/multipage/interaction.html#focus-navigation-scope-owner) (first point). Later when [flattening it](https://html.spec.whatwg.org/multipage/interaction.html#flattened-tabindex-ordered-focus-navigation-scope), items that are **not** focusable areas (i.e. the scope owners) are replaced by their (tabbable content) (2nd point in the algorithm).

So, if I get that correctly, an `iframe` is a scope owner and therefore included in the scope of its container, but it is not necessarily a focusable area (this may be browser dependent) and therefore not in the flattened scope. So, we cannot have an expectation that the `iframe` is in the flattened scope.

This does also answer @giacomo-petri question from CG call yesterday: `iframe` are not necessarily [focusable areas](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area) (this is actually UA dependent due to the the "or the element is determined by the user agent to be focusable" bit 😞), and therefore are (possibly) dropped out of the flattened scope (and the final tab-order) of the full page.

---
Concrete case:
```html
<html>
<iframe id="A" srcdoc="<button>A</button>" />
<iframe id="B" tabindex="0" srcdoc="<button>B</button>" />
<iframe id="C" tabindex="-1" srcdoc="<button>C</button>" />
<button>D</button>
</html>
```

1. Each of the 3 `iframe` has one element in its tabindex-ordered focus navigation scope: the `button` inside it.
2. The main document has 3 elements in its TOFNS: the first and second `iframe`, and `button` D; the 3rd `iframe` is excluded from it due to its negative `tabindex` (first point of [TOFNS](https://html.spec.whatwg.org/multipage/interaction.html#tabindex-ordered-focus-navigation-scope)).
3. When [flattening the document's TOFNS](https://html.spec.whatwg.org/multipage/interaction.html#flattened-tabindex-ordered-focus-navigation-scope), to build the actual sequential navigation order:
    *  the first `iframe` is not a focusable area (UA dependent), so it is replaced by its own TOFNS (point 2.2)
    * the second `iframe` is a focusable area (due to its own `tabindex`), so its TOFNS is inserted after it (point 2.3)
    * the third `iframe` is not even considered because it in not in the document's TOFNS.
    * the `button` D stays here, it's part of the clone built in step 1 and is just skipped over at step 2.1.

So, the TOFNS are:
* `iframe#A`: {`button A`}
* `iframe#B`: {`button B`}
* `iframe#C`: {`button C`}
* document: {`iframe#A`, `iframe#B`, ~`iframe#C` removed due to negative `tabindex`,~ `button D`}

And the flattening (assuming `iframe#A` is not a focusable area) yield { ~`iframe#A` removed as not focusable,~ `button A`, `iframe#B`, `button B`, `button D`}

Anyway, `iframe#A` should quite certainly pass the rule, but is not necessarily part of the final tab-order, so I kept the "no negative tabindex" expectation 😁

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: akn7bn
-name: Iframe with negative tabindex has no interactive elements
+name: Iframe with interactive elements are in tab-order
```
giacomo-petri:
```suggestion
name: Interactive elements within iframe are in tab-order
```

@Jym77, per HTML specs and your previous feedback:

> iframe are not necessarily [focusable areas](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area) (this is actually UA dependent due to the the "or the element is determined by the user agent to be focusable" bit 😞), and therefore are (possibly) dropped out of the flattened scope (and the final tab-order) of the full page.

from title seems that we leave browsers with the responsibility to decide if an iframe is passing or failing this rule.
In fact, if you create an iframe with a simple input as content (e.g. `<input type="text" name="test" aria-label="test">`),opening it in FireFox the iframe is tabbable (and consistent with this rule title), if instead you open it in Chrome the iframe is not tabbable (and seems not consistent with this rule title).

The final goal is to allow users interacting with the iframe content, so I would suggest to use "interactive elements" as subject of this rule.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -1,9 +1,9 @@
...
 rule_type: atomic
 description: |
-  This rule checks that `iframe` elements with a negative `tabindex` attribute value contain no interactive elements.
+  This rule checks that `iframe` elements which contain an interactive (tabbable) element are themselves tabbable.
```
giacomo-petri:
```suggestion
  This rule checks that interactive elements contained in `iframe` elements are tabbable.
```

same as above

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
giacomo-petri: On the other hand, this leads to a bit of discrepancy between rule title and "test target"/applicability 🤔

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,14 +28,19 @@ acknowledgments:
...
 ## Applicability
 
-This rule applies to any non-focusable `iframe` element that has [focusable][] content.
+This rule applies to any `iframe` element that [contains](#akn7bn:contain) at least one element for which all the following are true:
 
-## Expectation
-
-For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
+- the element is [visible][]; and
+- the element is part of the [flattened tabindex-ordered focus navigation scope][] of the `iframe`.
```
carlosapaduarte: Do we need to keep the following sentence about nested browsing context?

carlosapaduarte (commented):
Changes seem okay. Questioning the need to retain the sentence about nested browsing contexts.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: I think we do.
There are some technical subtelties in what owns what and what is inside what with browsing context containers, browsing contexts, and documents. The clarification makes it explicit what is meant here.
I also feel that putting it as a "next line clarification" does not really block readability and is only providing details for the technically savvy readers.

Jym77:
@giacomo-petri Agreed with your comments. I try to manage both sides (frontloading test target and being correct) by negating stuff (matching the expectation). Let me know what you think of it.

giacomo-petri (approved):
Perfect!

carlosapaduarte (approved):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: akn7bn
-name: Iframe with negative tabindex has no interactive elements
+name: Iframe with interactive elements are not excluded from tab-order
```
kengdoj:
```suggestion
name: Iframe with interactive elements is not excluded from tab-order
```

kengdoj:
```suggestion
name: Iframe with interactive elements does not have a negative tabindex
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
WilcoFiers: I think this got unnecessarily complicated. 

```suggestion
- the element is part of the [sequential focus navigation][] of the `iframe`.
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,14 +28,19 @@ acknowledgments:
...
 
-## Expectation
-
-For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
+- the element is [visible][]; and
+- the element is part of the [flattened tabindex-ordered focus navigation scope][] of the `iframe`.
 
 An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
 
+The [flattened tabindex-ordered focus navigation scope][] of an `iframe` is essentially the [sequential focus navigation][] order inside its content.
```
WilcoFiers: You don't need this. All documents have sequential focus navigation, an iframe has a document, ergo. Iframes can have sequential focus navigation.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -46,7 +51,7 @@ There are no major accessibility support issues known for this rule.
...
+Inside a page, each nested document (mostly, content of an `iframe` element) has its own [tabindex-ordered focus navigation scope][] which is the local tab-order inside this document. Then, all of these are grouped together in a [flattened tabindex-ordered focus navigation scope][] which is effectively the tab-order of the full page. By setting the `tabindex` attribute of a [focus navigation scope owner][] (mostly, an `iframe` element) to a negative value, it is excluded from its containing document's [tabindex-ordered focus navigation scope][]; and therefore its own [tabindex-ordered focus navigation scope][] is excluded from the global [flattened tabindex-ordered focus navigation scope][]. That is, a `button` may be in tab-order of a document, but if that document is included in another page via an `iframe` with negative `tabindex`, then the `button` is not in the tab-order of that page; and therefore not keyboard accessible.
```
WilcoFiers: This text is far more complicated than it needs to be to explain the problem; Tabindex=-1 on iframes means things in the iframe cannot be focused with keyboard navigation.

WilcoFiers (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
HelenBurge: To be discussed in the ACT-R meeting

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: Indeed 🙈 
I somehow got lost in technicalities and forgot that the sequential focus navigation is indeed exactly the FTOFNS 🙃 
(well, not exactly, because the former is an order and the later is a list 🤓, but close enough)

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: I've simplified this note.
I kept a bit of the technicalities and links because this corresponds to discussions we had during [CG call](https://www.w3.org/2022/07/28-act-r-minutes.html#item05) and [in this PR](https://github.com/act-rules/act-rules.github.io/pull/1855#issuecomment-1199064988), so I feel we need to keep trace of it.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -46,7 +49,7 @@ There are no major accessibility support issues known for this rule.
...
+Inside a page, each nested document (mostly, content of an `iframe` element) has its own [sequential focus navigation order][] which is the local tab-order inside this document. Then, all of these are grouped together in a [flattened tabindex-ordered focus navigation scope][] which is effectively the tab-order of the full page. Setting the `tabindex` attribute of an `iframe` element to a negative value effectively excludes its content from the global tab-order of the page (when building the [tabindex-ordered focus navigation scope][] of the page). That is, a `button` may be in tab-order of a document, but if that document is included in another page via an `iframe` with negative `tabindex`, then the `button` is not in the tab-order of that page; and therefore not keyboard accessible.
```
WilcoFiers: I think this is still a bit tricky. I tried to rephrase this, to hopefully improve the readability.

```suggestion
Setting the `tabindex` attribute of an `iframe` element to a negative value effectively excludes the content of that `iframe` from the tab-order of the page. A `button` may be in the tab-order of inside an `iframe`, but if the `iframe` is taken from the tab-order, the `button` is effectively keyboard inaccessible. 

Each document, including document inside an `iframe` has its own [sequential focus navigation order][]. These focus orders are combined to get the page's global tab-order (called the [flattened tabindex-ordered focus navigation scope][]). For an `iframe` with a negative tabindex, its sequential focus navigation order is not included in the page's global tab-order.
```

WilcoFiers (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,14 +28,17 @@ acknowledgments:
...
-
-For each test target, the [nested browsing context][] does not [contain](#akn7bn:contain) elements that are [visible][] and part of the [sequential focus navigation][].
+- the element is [visible][]; and
+- the element is part of the [sequential focus navigation order][] of the `iframe`'s [document][].
 
 An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
 
+## Expectation
+
+The test target does not have a negative number as a `tabindex` [attribute value][].
```
WilcoFiers, line 40: LGTM, although I want to note that `inert` might also now be able to trigger this situation. I opened a separate issue for it: #1965

WilcoFiers (approved):

PR #1858, opened 2022-06-09
image-no-text: Propose updated applicability

WilcoFiers:
In revisiting this rule, I think the `rendered image resources` definition makes the rule difficult to read. I tried to take some of the wording from that definition and put it into the applicability itself. Would like thoughts on this before I continue refining this.

---

## How to Review And Approve

### DIFF
--- a/_rules/image-no-text-0va7u6.md
+++ b/_rules/image-no-text-0va7u6.md
@@ -32,7 +32,13 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [rendered image resources][] in a [web page][].
+This rule applies to the [visible][] [image resource][] loaded on a [web page][]. Image resources can be loaded on a page through the following elements:
+
+- `img` elements with a non-empty [source set][]; or
+- `input` elements with a `type` attribute value of `image` and its `src` attribute value is not empty; or
+- `object` elements with a data attribute value referencing a resource with an image MIME type; or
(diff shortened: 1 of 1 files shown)


_rules/image-no-text-0va7u6.md
```diff
@@ -32,7 +32,13 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [rendered image resources][] in a [web page][].
+This rule applies to the [visible][] [image resource][] loaded on a [web page][]. Image resources can be loaded on a page through the following elements:
```
Jym77, line 35: 1. I'm not sure what `[image resource]` is linking to. Is it the same as "image source" (in our current definition of "rendered image resources")?
2. I'm not a fan of the "can be" formulation because I think it is a bit ambiguous whether this is an exhaustive list or not (can it be loaded in a different way). If it is not exhaustive, then the text is ambiguous. If it is exhaustive, we should say so ("image resources are loaded when one of the following is true" or something similar).
3. The current definition of "rendered image resources" also has a notion of visibility of (pixels of) the image which makes sense (if the image is fully covered by some other image, who cares…)

Jym77 (commented):
I guess the current definition is a bit cumbersome with many links to more stuff…

Maybe instead of expanding it in the rule, it would work to expand it in the definition (i.e. move the bullet list in the definition)?

Issue #1810, opened 2022-03-10
Is ARIA required owned elements failed example 7 really a fail? [bc4a75]


dd8:
Although having treeitems inside a menu doesn't make sense, I don't think this violates the required owned elements conditions in ARIA 1.1 and 1.2 so example 7 shouldn't be a fail for this rule.

https://act-rules.github.io/rules/bc4a75#failed-example-7

```
<!DOCTYPE html> 
<div role="menu">
 <div role="group">
  <span role="menuitem">Item 1</span>
  <div role="group">
   <span role="treeitem">Item 1</span>
   <span role="treeitem">Item 2</span>
  </div>
 </div>
</div>
```

The ARIA 1.1 and 1.2 specs requires "at least one element"  to satisfy the condition. It does not exclude other roles:
https://www.w3.org/TR/wai-aria-1.1/#mustContain

For example, you want to have elements with role=img for document type icons inside a Mac Finder or Windows Explorer style file system tree. And [Example 12 in the ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#example-12) has role=presentation interleaved with tree and treeitem roles.

For the code above I think the "at least one required owned element" condition is satisfied for all elements:

role=group - required owned elements: none
role=menu - required owned elements: group -> menuitemradio, menuitem, menuitemcheckbox, menuitemradio
role=menuitem - required owned elements: none
role=treeitem - required owned elements: none

It's not a failure of the required context role rule either, since the ARIA spec allows this test case because treeitem is contained by a group:

> Authors MUST ensure [elements](https://www.w3.org/TR/wai-aria-1.1/#dfn-element) with [role](https://www.w3.org/TR/wai-aria-1.1/#dfn-role) treeitem are contained in, or [owned](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element) by, an element with the role group **or** tree.

This is different to the requirement for role=menuitem which must be owned by role=menu or role=menubar:

https://www.w3.org/TR/wai-aria-1.1/#menuitem
> Authors MUST ensure that menu items are [owned](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element) by an element with role menu or menubar in order to identify that they are related [widgets](https://www.w3.org/TR/wai-aria-1.1/#dfn-widget)

carlosapaduarte:
If I recall correctly the reasoning behind that example is the following:
- the "menu" role has a "group" role to meet the required owned elements condition
[...]

PR #1862, opened 2022-06-16
"Role attribute has valid value" [674b10]: how to enforce WCAG mappings

Jym77:
[Role attribute has valid value](https://act-rules.github.io/rules/674b10) has a mapping to 1.3.1 (4.1.2 would be another candidate) which is not always respected. But in man of the cases, a mispelled `role` is still an accessibility error.

After TF discussion, we tried to remove the mapping in #1854 but this opened concerned from tool vendors that it will effectively remove the rule from the checks that their users care about since they are usually focused on compliance. See also [the CG call discussion about removing mappings](https://www.w3.org/2022/06/09-act-r-minutes.html#t05).

As per the CG call discussion, this tries to add a non-automatable exception to the Expectation to ensure the mapping is correct. I've tried both with 1.3.1 and 4.1.2. The later sounds much better imho, but the mapping is also weaker…

These propositions should likely be taken to the TF for further discussions at some point.

Closes issue(s):

- N/A

Need for Call for Review:
N/A, this is mostly for opening discussion.

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


Jym77:
Side question is whether we want to add examples for these exceptions, knowing that it would effectively prevents automated tools to claim a complete implementation 🤔

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -37,9 +37,14 @@ This rule applies to any `role` attribute for which all the following are true:
...
 - the attribute is specified on an [HTML or SVG element][] which is not [programmatically hidden][].
 
-## Expectation
+## Expectation (mapping to 4.1.2)
+
+Each test target either has at least one token which is a valid value corresponding to a non-abstract role from [WAI-ARIA Specifications][]; or the test target is not a [user interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components)
+
+## Expectation (mapping to 1.3.1)
+
+Each test target either has at least one token which is a valid value corresponding to a non-abstract role from [WAI-ARIA Specifications][]; or the test target has not a [structure](https://www.w3.org/TR/WCAG21/#dfn-structure) conveyed through presentation.
```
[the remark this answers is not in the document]
carlosapaduarte: I'm struggling to understand how any test target could meet the second part of the expectation version 1.3.1. If the test target must not be programatically hidden then it always have structure conveyed through presentation (because it will always be organised in relation to other parts of the web page).

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -37,9 +37,14 @@ This rule applies to any `role` attribute for which all the following are true:
 - the attribute has a value that is neither empty ("") nor only [ASCII whitespace][]; and
 - the attribute is specified on an [HTML or SVG element][] which is not [programmatically hidden][].
 
-## Expectation
+## Expectation (mapping to 4.1.2)
+
+Each test target either has at least one token which is a valid value corresponding to a non-abstract role from [WAI-ARIA Specifications][]; or the test target is not a [user interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components)
```
WilcoFiers, line 42: I think both the UIC and the structure definitions are ambiguous. UIC because it is not clear who is meant by "users", and "structure" for not specifying what "organize" means.

Jym77: We can replace UIC by our definition of [Instrument](https://act-rules.github.io/glossary/#instrument-to-achieve-an-objective) but that would require to define the objective, so I'm not sure it would actually work 😕

Jym77:
Another possibility that we've discussed would be to change the ACT rules format to have both mandatory and optional mappings. For this rule, it would be something like:
> Implementation of "Role attribute has valid value" **must** fail ARIA conformance and **may** fail 1.3.1 and/or 4.1.2

So, implementation that fail one or the other SC on some or even all test cases would be considered consistent, but implementations that fail 2.1.1 would not be considered consistent.
We could revisit some earlier changes in mapping to get them back as optional mappings.
This could be use elsewhere, for example to fix #1875.

This could also be used to streamline the button/image button rules by removing the "button not image button" exception from "button has accessible name" and having instead a mandatory mapping to 4.1.2 and an optional to 1.1.1.

dd8:
> Another possibility that we've discussed would be to change the ACT rules format to have both mandatory and optional mappings. For this rule, it would be something like:
> 
> > Implementation of "Role attribute has valid value" **must** fail ARIA conformance and **may** fail 1.3.1 and/or 4.1.2

That seems like a good solution. It might also solve the mapping problem in https://github.com/act-rules/act-rules.github.io/issues/1810#issuecomment-1180888267

Might also be worth considering conditional mappings. For a lot of ARIA issues 4.1.2: Name, Role, Value seems like the best option, but then has problems because it only applies to [user interface components] (https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html#dfn-user-interface-component)

So you could say something like 

> Must fail 1.3.1 if the widget is visible
> Must fail 4.1.2 if the widget is a user interface component

That could also fix https://github.com/act-rules/act-rules.github.io/issues/1875. The conditional mapping for an area without alt would be: 

> Must fail 1.1.1 for area elements

Jym77:
The ACT rule format is moving toward having secondary requirements, which would be a better way to handle this kind of cases. Closing this.

PR #1864, opened 2022-06-16
"Links with identical accessible names and context serve equivalent purpose" [fd3a94]: precise that the rule only looks for "same" context

### LINKED ISSUE #1784: "Links with identical accessible names and context serve equivalent purpose" [fd3a94]: "same" or "identical" contexts?
While reviewing #1765, I got into this:
[Links with identical accessible names and context serve equivalent purpose](https://act-rules.github.io/rules/fd3a94) title uses "**identical** (…) context" but the Expectation uses "**same** context".

To me, this rings a bell of difference between "structural" equality (identical) vs "reference" equality (same). I.e.:
```html
<p>Lorem <a href="foo">ipsum</a> </p>
<p>Lorem <a href="bar">ipsum</a> </p>
```
I would say that these links have different (= "not same") contexts because these are two different `p` element, but they have "identical" context because they look exactly the same.

I'm not full sure which one the rule should consider, but we should probably use the same word in title and expectation. Also, it seems that the test cases are all built with the **same** context, so that might be the one to keep.

---
"same" is obviously stronger than "identical". I do feel that "identical" is the comparison we actually want (and that the previous example should be failing). However, this also opens a can of worms of defining "identical" (because an empty `div`, a `span` wrapping a word to no effect, a different text colour, …) should all arguably be ignored by "identical" in this rule (essentially, only the text content of the context should matter 🤔 )
```html
<p>Lorem <span></span> <a href="foo">ipsum</a> <span style="color: red">dolor</span> </p>
<p>Lorem <span><a href="bar">ipsum</a></span> <span style="color: blue">dolor</span> </p>
```
I'd argue that these two should still be considered "identical contexts" even though the structure starts being quite different…

Jym77:
There was a bit of confusion whether the rule was looking for "same" contexts (= same elements) or "identical" ones (= similar content? maybe also similar styling?) Since "identical" content is hard to define, we restrict the rule on "same" context and make it explicit in several places.

Closes issue(s):

- closes #1784 

Need for Call for Review:
This will require a 1 week Call for Review (changing title and description to match Applicability)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
+++ b/_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
@@ -1,9 +1,9 @@
 ---
 id: fd3a94
-name: Links with identical accessible names and context serve equivalent purpose
+name: Links with identical accessible names and same context serve equivalent purpose
 rule_type: atomic
 description: |
-  This rule checks that links with identical accessible names and context resolve to the same or equivalent resources.
+  This rule checks that links with identical accessible names in the same context resolve to the same or equivalent resources.
 accessibility_requirements:
   wcag20:2.4.4: # Link Purpose (In Context) (A)
     forConformance: true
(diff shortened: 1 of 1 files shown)


carlosapaduarte (approved):

HelenBurge (approved):

tbostic32 (approved):

Jym77:
Call for review ends on June 30th.

colabottles (approved):

Issue #1861, opened 2022-06-14
"Image button has non-empty accessible name" [59796f] - clarification about input type image with empty alt but non-empty title


giacomo-petri:
- **Starting URL**: https://www.w3.org/WAI/standards-guidelines/act/rules/59796f/
- **Highlighted element**: input type image, empty alt attribute and non-empty title attribute 
`<input type="image" src="test.png" alt="" title="non-empty value">`
- **Description**: should we add an example for input type image with empty alt attribute and non-empty title attribute that clearly identify it as a failure?
    - per failed example 2
    _The image button has an empty alt attribute, and no other attributes that can give it an [accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/59796f/#accessible-name), hence its name is the default “Submit Query”._
`<input type="image" src="/test-assets/shared/search-icon.svg" alt="" />`
the input is failing due to the absence of an accessible name.
    - per passed example 3
_The image button has an [accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/59796f/#accessible-name) through the title attribute._
_note: The title attribute may not always be [accessibility supported](https://www.w3.org/WAI/standards-guidelines/act/rules/59796f/#accessibility-support)._
`<input type="image" src="/test-assets/shared/search-icon.svg" title="Search" />`
input has an accessible name thanks to the title attribute and title attribute can give it an accessible name.
- **Concerns**:
    - from failed example 2
    _The image button has an empty alt attribute, **and no other attributes** that can give it an [accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/59796f/#accessible-name),_
    actually, per passed example 3, the title attribute seems a valid attribute to give the input type image an accessible name. I understand that in this context, _can give it_ means that the resulting accessible name is non-empty, but might not be fully clear.
    - following the [accessible name computation](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) instructions, the empty alt attribute should force the accessible name to be empty. That said, on Firefox, the title attribute is used as fallback even if the alt attribute is empty, as per screenshots attached:
    <img width="1792" alt="screenshot highlighting input type image code with empty alt attribute and non-empty title attribute in Firefox inspector" src="https://user-images.githubusercontent.com/106394951/173630662-484ab0e5-228a-4f94-aaa2-57132cc326db.png">
[...]

PR #1865, opened 2022-06-16
Update image-button-non-empty-accessible-name-59796f.md

giacomo-petri:
Added:
* failed example 3
* updated "Accessibility Support" section

Closes issue: #1861

Need for Call for Review:

Note: Sorry if I made mistakes while creating the pull request, this is my first pull request and first time I contribute to a GitHub project.

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,6 +56,8 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+The image button with non-empty `title` attribute but empty `alt` attribute results in a button with empty accessible name on many browsers. On Firefox, instead, the `title` attribute is used as fallback, resulting in a button with `title` attribute value as accessible name.
```
carlosapaduarte, line 59: I think we should try to write this without naming any specific browsers (like we do in the previous sentence).

carlosapaduarte (changes requested):
No mistakes when creating the PR! 👍 

I just left a request to do the updated accessibility support section without naming any specific browsers.

_rules/image-button-non-empty-accessible-name-59796f.md
giacomo-petri: Thanks Carlos,
updated using a generic _known popular browser_  as above.

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,6 +56,8 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+The image button with non-empty `title` attribute but empty `alt` attribute results in a button with empty accessible name on many browsers. There is a known popular browser instead, that uses the `title` attribute as fallback even with empty `alt`, resulting in a button with `title` attribute value as accessible name.
```
carlosapaduarte, line 59:
```suggestion
The image button with non-empty `title` attribute but empty `alt` attribute results in a button with empty accessible name on many browsers. There is a known popular browser that uses the `title` attribute as fallback even with empty `alt`, resulting in a button with `title` attribute value as accessible name. 
```

carlosapaduarte (changes requested):
Editorial suggestion

carlosapaduarte (approved):

WilcoFiers (approved):

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,6 +56,8 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+The image button with non-empty `title` attribute but empty `alt` attribute results in a button with empty accessible name on many browsers. There is a known popular browser that uses the `title` attribute as fallback even with empty `alt`, resulting in a button with `title` attribute value as accessible name.
```
Jym77, line 59:
```suggestion
An image button with non-empty `title` attribute but empty `alt` attribute results in a button with an empty accessible name on many browsers. There is a known popular browser that uses the `title` attribute as fallback even with empty `alt`, resulting in a button with `title` attribute value as accessible name. 
```

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -128,6 +130,14 @@ The image button has an empty `alt` attribute, and no other attributes that can
 
 #### Failed Example 3
 
+The image button has a non-empty `title` attribute, but an empty `alt` attribute. This gives the button an empty [accessible name][].
```
Jym77, line 133:
```suggestion
This image button has a non-empty `title` attribute, but an empty `alt` attribute. This gives the button an empty [accessible name][].
```

Jym77 (dismissed):
Typos.

Jym77 (approved):

tombrunet (approved):

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,6 +56,8 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+An image button with non-empty `title` attribute but empty `alt` attribute results in a button with an empty accessible name on many browsers. There is a known popular browser that uses the `title` attribute as fallback even with empty `alt`, resulting in a button with `title` attribute value as accessible name.
```
scottaohara: an image and an image button are different things.  they have two different naming algorithms for a reason.  why are they being merged into one test?

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -128,6 +130,14 @@ The image button has an empty `alt` attribute, and no other attributes that can
 
 #### Failed Example 3
 
+This image button has a non-empty `title` attribute, but an empty `alt` attribute. This gives the button an empty [accessible name][].
+
+```html
+<input type="image" src="/test-assets/shared/search-icon.svg" title="Search" alt="" />
```
scottaohara: firefox says search button
chrome has no name (bug)
safari says submit button

firefox is correct here
chrome is wrong by providing no name
safari skips title but at least provides the fallback text.

_rules/image-button-non-empty-accessible-name-59796f.md
giacomo-petri: I agree with you; we were almost talking about the need of a clearer definition for "usable text string" in accessible name computation rules.

That said, this rule is specific for image button elements (we didn't merge image and image button elements int one single test).

As per [Inapplicable Example 4](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/59796f/ebd0080bacb8debc7ad069072240657df38c3e2c.html)
> The img element is not a user interface component, and so is not tested for [Success Criterion 4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value).

that confirms that images are not part of this rule.

_rules/image-button-non-empty-accessible-name-59796f.md
giacomo-petri: In my opinion this might be subjective (and that's probably also why browsers render this example in different ways).

From [4.3.1 input type="image" Accessible Name Computation](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation)

> 4.3.1 input type="image" Accessible Name Computation
    1. If the control has an [aria-label](https://www.w3.org/TR/wai-aria-1.2/#aria-label) or an [aria-labelledby](https://www.w3.org/TR/wai-aria-1.2/#aria-labelledby) attribute the [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name) is to be calculated using the algorithm defined in [Accessible Name and Description: Computation and API Mappings](https://www.w3.org/TR/accname-1.2/).
    2. Otherwise use alt attribute.
    3. Otherwise use title attribute.
    4. Otherwise if the previous steps do not yield a usable text string, the [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name) is a localized string of the word "Submit Query".
If none of the above yield a usable text string there is no [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name).

In our example the alt attribute has been clearly defined as empty (alt=""). Depending on interpretations, we might stop at point 2, considering the alt="" as a choice to make the image decorative, or we can move forward reviewing the title.

Even more important is that for images, the HTML Accessibility API Mapping contemplate (rightly) the specific scenario ([4.10 img Element](https://www.w3.org/TR/html-aam/#img-element-accessible-name-computation))

> 4.10.1 img Element Accessible Name Computation
    1. If the img element has an [aria-label](https://www.w3.org/TR/wai-aria-1.2/#aria-label) or an [aria-labelledby](https://www.w3.org/TR/wai-aria-1.2/#aria-labelledby) attribute the [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name) is to be calculated using the algorithm defined in [Accessible Name and Description: Computation and API Mappings](https://www.w3.org/TR/accname-1.2/).
    2. Otherwise use alt attribute, even if its value is the empty string.
    NOTE
    An img with an alt attribute whose value is the empty string is mapped to the [presentation](https://www.w3.org/TR/core-aam-1.2/#role-map-presentation) role. It has no accessible name.
    3. Otherwise, if there is no alt attribute use the title attribute.
    4. Otherwise there is no [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name).

Combining these two items, might result unclear which are expectations.

Interesting also the example provided by @Jym77 when images are not loaded.

giacomo-petri: In addition, if these are the expectations, it is necessary that all browsers move in this direction; right now, this example might be an important barrier for some user depending on the technology the user is using.

_rules/image-button-non-empty-accessible-name-59796f.md
Jym77: In light of @scottaohara clarification at https://github.com/w3c/html-aam/issues/414#issuecomment-1178384158, I think we can rephrase this in the line of
> The naming algorithm (+link to relevant HTML AAM part) uses the first non-empty name, but some AU/AT stop at the first existing one, even if empty.

This would make it clear that it is not just some browsers quirks deciding what to do, but an actual bug in some implementation (and would explain why our examples are built following specs and may behave differently on some browsers).

---
Maybe we can also add a background note in the line of:
> Contrarily to images, empty `alt` does not make image button decorative. Image buttons are button and therefore exposed as interactive elements.

which is one on these things that is obvious once stated but maybe not so obvious before seeing it once…

_rules/image-button-non-empty-accessible-name-59796f.md
Jym77: In the light of the recent clarifications, this should be a Passed Example indeed (since the empty `alt` does not provide a "usable string").
Description can say that the image button has name "Search", but some UA/AT (incorrectly) either use the empty `alt` or the fallback "Submit query".

Jym77 (dismissed):

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,8 +56,13 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some AU/AT stop at the first existing one, even if empty.
+
 ## Background
 
+Empty `alt` does not provide a "usable string" for `input` element with a `type` [attribute value][] of `image`.
+Therefore, contrarily to images, empty `alt` does not make image button decorative. Image buttons have a button role and therefore are exposed as interactive elements.
```
Jym77:
```suggestion
Contrarily to images, an empty `alt` attribute does not make image button decorative; image buttons have a button role and are therefore exposed as interactive elements. Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation default to other means of providing a name.
```

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,8 +56,13 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some AU/AT stop at the first existing one, even if empty.
```
Jym77:
```suggestion
The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some user agents and assistive technologies combinations stop at the first existing one, even if empty.
```

Jym77 (changes requested):
Polishing a bit the notes.

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,12 +56,11 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
...
 
-The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some AU/AT stop at the first existing one, even if empty.
+The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some user agents and assistive technologies combinations stop at the first existing one, even if empty.
 
 ## Background
 
-Empty `alt` does not provide a "usable string" for `input` element with a `type` [attribute value][] of `image`.
-Therefore, contrarily to images, empty `alt` does not make image button decorative. Image buttons have a button role and therefore are exposed as interactive elements.
+Contrarily to images, an empty `alt` attribute does not make image button decorative; image buttons have a button role and are therefore exposed as interactive elements. Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation default to other means of providing a name.
```
Jym77:
```suggestion
Contrarily to images, an empty `alt` attribute does not make image button decorative; image buttons have a button role and are therefore exposed as interactive elements. Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation defaults to other means of providing a name.
```

Jym77 (changes requested):
Fixing a typo i my previous suggestion 🙈

Jym77 (approved):

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -56,8 +56,12 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 There is a known combination of a popular browser and assistive technology that does not by default support `title` as an [accessible name][].
 
+The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some user agents and assistive technologies combinations stop at the first existing one, even if empty.
+
 ## Background
 
+Contrarily to images, an empty `alt` attribute does not make image button decorative; image buttons have a button role and are therefore exposed as interactive elements. Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation defaults to other means of providing a name.
```
WilcoFiers: 1. I don't think you should use `Contrarily to images`, you're talking specifically about the `img` element. There are different types of images in HTML.
2. I'm not sure what that second sentence means. Can you make it clearer?

_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -108,6 +112,14 @@ The image button has an [accessible name][] through the `aria-labelledby` attrib
...
 ```
 
+#### Passed Example 5
+
+This image button has an empty `alt` attribute, but a non-empty `title` attribute. This gives the button an non-empty [accessible name][] through the `title` attribute.
+
+```html
+<input type="image" src="/test-assets/shared/search-icon.svg" title="Search" alt="" />
+```
+
```
WilcoFiers: Since this does indeed cause significant accessibility support issues, I don't think we can include it as a passing example. It's perfectly understandable / correct for testers to fail such image buttons.

```suggestion
```

WilcoFiers (changes requested):

_rules/image-button-non-empty-accessible-name-59796f.md
giacomo-petri: 1. I agree with the 1st item. I'll update "images" with `img` elements.

----

2. About the second sentence
> Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation defaults to other means of providing a name.

it refers to the previous adjacent Accessibility support paragraph

> The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some user agents and assistive technologies combinations stop at the first existing one, even if empty.

Per [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation), if the alt attribute value consists in a non-usable string the title attribute should be considered, and again, if the title attribute value consists in a non-usable string, the accessible name is a localized string of the word "Submit Query".

For the sake of clarity I'll update 
> Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation defaults to other means of providing a name.

in 

> Consequently, an empty `alt` attribute does not provide a "usable string" for image buttons and the computation defaults to other means of providing a name, as defined in [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation).

Does it make sense?

_rules/image-button-non-empty-accessible-name-59796f.md
giacomo-petri: My idea was to clearly define a way to test this scenario, showing a pass/fail example to better understand how to tackle it. It was a real case found in a website but the definition was ambiguous, without clearly explaining if it was a pass or a fail.

If specs clearly define it as a pass, in my opinion we should reiterate our feedback, asking techs to support it as per specs, adding maybe a clear note in Accessibility Support section.

giacomo-petri:
As per our conversation, I've removed Passed example 5 and I've opened an issue about how to handle those cases: #1954

WilcoFiers (approved):

PR #1868, opened 2022-06-20
Update ARIA required properties rule to WAI-ARIA 1.2

### LINKED ISSUE #1282: Element with `role` attribute has required states and properties (4e8ab6): Make combobox examples correct
Many examples in [Element with `role` attribute has required states and properties](https://act-rules.github.io/rules/4e8ab6) are using elements with role of `combobox` but alone in the void.

These examples should probably be updated to show good practice of building `combobox`:
* at the very least adding the required `textbox` and likely also the `listbox` popup when relevant;
* avoid `aria-control` referencing to a non-existing ID.

### LINKED ISSUE #1373: Element with role has required states and properties - Failed example 1 [4e8ab6]
More ARIA versioning issues. 

```
<div role="combobox" aria-expanded="true"></div>
```

**ARIA 1.0 combobox pattern** - `aria-expanded` always required , `aria-owns` not listed as required - should be used when aria-autocomplete=list (but should rather than must)
https://www.w3.org/TR/wai-aria-1.0/roles#combobox
https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-6

**ARIA 1.1 combobox pattern** - `aria-expanded` always required, `aria-controls` only set to set to popup element
when expanded
https://www.w3.org/TR/wai-aria-1.1/#combobox
https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-6

**ARIA 1.2 combobox pattern** - `aria-expanded` and `aria-controls` always required  with `aria-controls` always set to popup element
https://www.w3.org/TR/wai-aria-1.2/#combobox

The problem is that the ARIA 1.1 spec supports both the 1.0 and 1.1 combobox:

> The ARIA 1.0 specification describes a combobox pattern where a text input element has the combobox role and owns a listbox element. User agents, assistive technologies, and conformance checkers SHOULD continue to support the ARIA 1.0 pattern so that existing implementations of the ARIA 1.0 pattern remain functional.
https://www.w3.org/TR/wai-aria-1.1/#combobox

**tl;dr;** until ARIA 1.2 moves to CR the only required attribute for `role=combbox` is `aria-expanded` because ARIA 1.1 supports both the 1.1 and 1.0 combobox patterns, so failed example 1 is a pass

WilcoFiers:
- Add several new related to properties that in ARIA 1.2 no longer have a default value
- Add an example for the separator, which has a required prop only if it is focusable
- Improve the combobox examples, borrowing code from the WAI-ARIA spec itself
- Add accessible names to elements that require one

Closes #1282, closes #1373

---

This is one part of migrating ACT Rules over to WAI-ARIA 1.2. We'll have all WAI-ARIA 1.2 PRs go to the wai-aria 1.2 branch, which we'll merge into the develop branch once everything is ready. This lets us chunk up the work while ensuring the upgrade to WAI-ARIA 1.2 happens all at once.

The final WAI-ARIA 1.2 PR will be a 2 week call for review. Everything going into that branch will require the usual 3 approvals, but will not be sent out as call for review.

---

## How to Review And Approve


_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -97,13 +97,28 @@ This `input` element is not part of the [sequential focus navigation][] because
 
 #### Passed Example 4
 
-This `a` element is part of the [sequential focus navigation][], but moves focus to the `input` element whenever it receives focus.
+This `a` element is not [focusable][] because it moves focus to the `input` element whenever it receives focus. These elements
```
Jym77, line 100: This example (and the next one) does not look like it is related to ARIA 1.2 (nor to the rule that the PR mentions in its title).
Should it be moved to another PR?
(or maybe it's because something has been merged but the `wai-aria-1.2` branch hasn't been updated to the latest `develop` so git get confused)

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -169,6 +184,27 @@ This `summary` element is part of the [sequential focus navigation][].
 </details>
 ```
 
+#### Failed Example 6
```
Jym77, line 187: This example looks unrelated to this PR.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +64,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
 ```
 
 ### Failed
 
 #### Failed Example 1
 
-This `combobox` is missing the required `aria-controls` property.
-
-**Note**: In [WAI-ARIA 1.2][], `combobox` will also require `aria-expanded`.
+This `heading` does not have the required `aria-level` property. Prior to [WAI-ARIA 1.2][] the `heading` role had an implicit default `aria-level` value of `2`. As of WAI-ARIA 1.2 this property must be explicitly set.
```
Jym77, line 132: I'm not sure about putting "historical" details in the descriptions 🤔 Maybe just saying that the property has no default (no need to say how it was in the previous versions, people can look at them on their own).
```suggestion
This `heading` does not have the required `aria-level` property, which has no default value.
```

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +64,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
-<div role="combobox" aria-expanded="true"></div>
+<div role="heading">
+	My First Heading
+</div>
 ```
 
 #### Failed Example 2
 
-This `combobox` has an empty value for the required `aria-controls` property.
+This `switch` does not have the required `aria-checked` property. Prior to [WAI-ARIA 1.2][] the `switch` role had an implicit default `aria-checked` value of `false`. As of WAI-ARIA 1.2 this property must be explicitly set.
```
Jym77, line 142:
```suggestion
This `switch` does not have the required `aria-checked` property, which has no default value.

File: _rules/role-required-states-and-properties-4e8ab6.md [hunk 5]
```diff
@@ -60,47 +64,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
+
+```html
+<div role="switch">
+	Toggle me
+</div>
+```
+
+#### Failed Example 3
+
+This `checkbox` does not have the required property `aria-checked`. Prior to [WAI-ARIA 1.2][] the `checkbox` had an implicit default `aria-checked` value of `false`. As of WAI-ARIA 1.2 this property must be explicitly set.
```
Jym77, line 152:
```suggestion
This `checkbox` does not have the required property `aria-checked`, which has no default value.
```

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +64,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
-This `combobox` has an empty value for the required `aria-controls` property.
+This `switch` does not have the required `aria-checked` property. Prior to [WAI-ARIA 1.2][] the `switch` role had an implicit default `aria-checked` value of `false`. As of WAI-ARIA 1.2 this property must be explicitly set.
+
+```html
+<div role="switch">
+	Toggle me
+</div>
+```
+
+#### Failed Example 3
```
Jym77, line 150: It also sounds that these 3 examples are here to highlight the differences between ARIA 1.1 and 1.2 (with default disappearing), which I don't think is the goal of ACT rules. We can probably remove one or two of them.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +64,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
+
+```html
+<p>My first HTML</p>
+<div role="separator" tabindex="0"></div>
+<p>My last HTML</p>
+```
+
+#### Failed Example 5
+
+This `combobox` does not have the required `aria-expanded` property. Prior to [WAI-ARIA 1.2][] the `combobox` had an implicit default `aria-expanded` value of `false`. As of WAI-ARIA 1.2 this property must be explicitly set.
```
Jym77, line 171:
```suggestion
This `combobox` does not have the required `aria-expanded` property, which has no default value.
```

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +64,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
+<input type="text" id="tag_combo" role="combobox" aria-controls="popup_listbox" />
+<ul role="listbox" id="popup_listbox">
+	<li role="option">Zebra</li>
+	<li role="option" id="selected_option">Zoom</li>
+</ul>
+```
+
+#### Failed Example 6
+
+This `combobox` uses `aria-owns` where it should use `aria-controls` property. This property is required as of [WAI-ARIA 1.1][].
```
Jym77, line 184: I don't really understand this description.
Does it mean that `aria-owns` used to be required on `combobox` but isn't anymore? It seems that [combobox in ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#combobox) was already requiring `aria-controls` and `aria-expanded` (unchanged in ARIA 1.2) and wasn't required `aria-owns`.
So, I don't understand this test case (why is it needed and what does it test?)

Jym77 (changes requested):

PR #1872, opened 2022-06-21
Fix git+ssh on wai-deploy

WilcoFiers:
I broke the release script. This fixes it

---

## How to Review And Approve


.github/workflows/wai-deploy.yml
```diff
@@ -19,12 +19,13 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
+      - run: git config --global url."https://github.com/".insteadOf ssh://git@github.com/
```
WilcoFiers, line 22: I tried to combine this with the credential setup on line 27. Couldn't figure it out so I'll just make the call twice.

Jym77 (approved):

Issue #1853, opened 2022-05-31
Should img without role be failed for Element marked as decorative? [46ca7f]


WilcoFiers:
I'm concerned about implementing [failed example 2](https://wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/46ca7f/#failed-example-2) in our implementation:

```html
<img src="/test-assets/shared/w3c-logo.png" alt="" aria-labelledby="label" />
<span hidden id="label">W3C logo</span>
```

I can easily imagine component code that ends up looking like this. Take this react component for example:

```jsx
const Image = ({ alt = '', ...props }) => (<img alt={alt} {...props} />)

export default () => (<figure>
  <Image aria-labelledby="caption" src="my-image.png" />
  <figcaption id="caption">My wonderful caption</figcaption>
</figure>
```

This is a perfectly reasonable way to write a component, which then ends up being failed by the Element marked as decorative rule. I think this rule previously only reported on the explicit use of role=none / presentation. I'd prefer to go back to that, and turn failed example 2 into a pass.

Jym77:
Plot twist…
In ARIA 1.2, `aria-label` becomes prohibited on role of `presentation`.

So, the first thing is that `<img alt="" aria-label="foo" />` is not valid anymore.

The second thing is that I understand that prohibited attributes should not trigger the presentational role conflict resolution, and this example is therefore having a role of presentation and not failing the rule…

[presentational role conflict resolution](https://www.w3.org/TR/wai-aria-1.2/#conflict_resolution_presentation_none):
> If an element has **global WAI-ARIA states or properties**, user agents MUST ignore the presentation role and expose the element with its implicit role. 

But [global properties](https://www.w3.org/TR/wai-aria-1.2/#global_states):
> The following global states and properties are supported by all roles and by all base markup elements **unless otherwise prohibited**.

(emphasises mine)
I understand that prohibited properties should not be considered as global, and not trigger the conflict. We may want to run some tests to be sure.

---
I suggest this is taken by the ARIA 1.2 committee since it relates to it.

WilcoFiers:
Wouldn't have guessed this from the text, but you seem to be correct. At least as far as Chrome / Edge are concerned, aria-label and aria-labelledby no longer trigger conflict resolution for role=none / role=presentation. I.e. If you put aria-label on an h4 with role=none, Chrome no does not include that heading in the accessibility tree. Firefox and Safari still do though.

Jym77:
[...]

PR #1877, opened 2022-06-23
Adds prohibited attributes expectation to rule 5c01ea

### LINKED ISSUE #1535: Consider prohibited ARIA attributes in ARIA props permitted [5c01ea]
WAI-ARIA 1.2 comes with [Prohibited States and Properties](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes). When ARIA 1.2 is published as candidate rec, I think we should add this as an expectation to the [ARIA states and properties permitted](https://act-rules.github.io/rules/5c01ea) rule.

schne324:
This patch adds a second expectation regarding usage of prohibited attributes for a given role. Based on [ARIA 1.2's statement](https://www.w3.org/TR/wai-aria-1.2/#state_property_processing):

> Global states and properties are supported on any element in the host language. However, authors MUST only use non-global states and properties on elements with a role supporting the state or property; either defined as an explicit WAI-ARIA role, or as defined by the host language implicit WAI-ARIA semantic matching an appropriate WAI-ARIA role.

Closes issue(s):

- closes #1535

### DIFF
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -183,6 +183,14 @@ The `aria-orientation` property may not be used on `audio` element, nor it can b
 <audio src="/test-assets/moon-audio/moon-speech.mp3" controls aria-orientation="horizontal"></audio>
 ```
 
+#### Failed Example 3
+
+The `aria-label` property may not be used on a `generic` element as it [prohibited from being named](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited).
+
+```html
+<div aria-label="Bananas"></div>
+```
+
 ### Inapplicable
 
 #### Inapplicable Example 1
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -42,6 +42,8 @@ For each test target, one of the following is true:
 - **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
+For each test target there are no prohibited attributes for the target's role.
+
 ## Assumptions
 
 _There are currently no assumptions_
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -34,7 +34,7 @@ acknowledgments:
 
 This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is [included in the accessibility tree][].
 
-## Expectation
+## Expectation 1
 
 For each test target, one of the following is true:
 
@@ -42,7 +42,9 @@ For each test target, one of the following is true:
 - **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
-For each test target there are no prohibited attributes for the target's role.
+## Expectation 2
+
+Each test target is not prohibited on the role of the element on which it is specified.
 
 ## Assumptions
 
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -44,7 +44,7 @@ For each test target, one of the following is true:
 
 ## Expectation 2
 
-Each test target is not prohibited on the role of the element on which it is specified.
+Each test target is not [prohibited](https://www.w3.org/TR/wai-aria-1.2/#state_property_processing) on the role of the element on which it is specified.
 
 ## Assumptions
 
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -187,7 +187,7 @@ The `aria-orientation` property may not be used on `audio` element, nor it can b
 
 #### Failed Example 3
 
-The `aria-label` property may not be used on a `generic` element as it [prohibited from being named](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited).
+The `aria-label` property may not be used on an element with a `generic` role as it is [prohibited from being named](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited).
 
 ```html
 <div aria-label="Bananas"></div>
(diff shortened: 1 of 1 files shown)


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -42,6 +42,8 @@ For each test target, one of the following is true:
 - **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
+For each test target there are no prohibited attributes for the target's role.
```
Jym77, line 45: 1. In this rule, the "test targets" are the ARIA states and properties, not the element. So they do not have a role. This should be phrased as "the test target is not prohibited on the role of the element on which it is specified" or something like that (see the way the second bullet is phrased).
2. This sounds like a completely different expectation, should they be numbered as "Expectation 1", "Expectation 2" instead?
3. What is a "prohibited attribute"? This needs to link to some specification. "role" should also link to our definition of "semantic role".

Jym77, line 45: …and I actually do not understand the need for this new expectation 🤔 

If the state or property is either global, inherited, supported, required, or explicitly allowed by ARIA in HTML, how can it be prohibited?
This needs at least a new test case to show something that fails because of this new expectation and wouldn't fail without it (see [Test every condition](https://act-rules.github.io/pages/design/test-cases/#test-every-condition) in our test cases design document).

Jym77 (changes requested):

_rules/aria-state-or-property-permitted-5c01ea.md
WilcoFiers: It does need new tests. As for why this is needed; ARIA 1.2 prohibits aria-label(ledby) on a bunch of things, even though it is a global attribute. And aria-roledescription on `generic`.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -34,14 +34,18 @@ acknowledgments:
...
 
 For each test target, one of the following is true:
 
 - **global**: the test target is a [global state or property][global]; or
 - **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
+## Expectation 2
+
+Each test target is not [prohibited](https://www.w3.org/TR/wai-aria-1.2/#state_property_processing) on the role of the element on which it is specified.
```
schne324:
```suggestion
Each test target is not [prohibited](https://www.w3.org/TR/wai-aria-1.2/#state_property_processing) on the [semantic role][] of the element on which it is specified.
```

_rules/aria-state-or-property-permitted-5c01ea.md
carlosapaduarte: I'm not sure I agree that it does not need new tests. Why not have a failed example with an `aria-label` on `generic` container? It passes expectation one (because it is a global property) and fails expectation 2 (because it is prohibited).

carlosapaduarte (dismissed):
See https://github.com/act-rules/act-rules.github.io/pull/1877#discussion_r932079032

_rules/aria-state-or-property-permitted-5c01ea.md
Jym77: @carlosapaduarte "It does need new tests" and that is also what @WilcoFiers wrote 😃 
(and I also had to read it twice to see there is no negation in the first sentence, I have no clue why this sentence is apparently so hard to parse 🤔)

carlosapaduarte: Damn... I read it more than twice and it always registered in my brain as "It does not need new tests" 🫣

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -34,14 +34,18 @@ acknowledgments:
...
 For each test target, one of the following is true:
 
 - **global**: the test target is a [global state or property][global]; or
 - **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
+## Expectation 2
+
+Each test target is not [prohibited](https://www.w3.org/TR/wai-aria-1.2/#state_property_processing) on the [semantic role][] of the element on which it is specified.
+
```
Jym77: I'm not fully sure we need to have it as a separate expectation. The [ARIA 1.2 definition of global](https://www.w3.org/TR/wai-aria-1.2/#global_states) states:
> The following global states and properties are supported by all roles and by all base markup elements **unless otherwise prohibited**.
(emphasis mine)
and later lists:
> * [aria-label](https://www.w3.org/TR/wai-aria-1.2/#aria-label) (Except where prohibited)
> * [aria-labelledby](https://www.w3.org/TR/wai-aria-1.2/#aria-labelledby) (Except where prohibited)

So, I do understand that prohibited attributes are not "global" for the roles on which they are prohibited (see also my comment at https://github.com/act-rules/act-rules.github.io/issues/1853#issuecomment-1195326291).

---
At the same time, I do agree that it is not completely obvious to derive that fact. This would at least require a note to explain that prohibited attributes are not global…

OTOH, I feel the current formulation also requires a note since it has a redundant expectation.

I think the less confusing thing to do would be to keep Expectation 2 as is, but add a Background note in the line of "prohibited attributes are not global, but we kept Expectation 2 for the sake of clarity"

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -181,6 +185,14 @@ The `aria-orientation` property may not be used on `audio` element, nor it can b
 <audio src="/test-assets/moon-audio/moon-speech.mp3" controls aria-orientation="horizontal"></audio>
 ```
 
+#### Failed Example 3
+
+The `aria-label` property may not be used on an element with a `generic` role as it is [prohibited from being named](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited).
```
Jym77: Since the expectation is in term of "prohibited" attributes rather than roles "prohibited from being named", I'd rather have the example description using the same language too.

Jym77 (dismissed):
I think we should add an explanation on the link between global and prohibited.

_rules/aria-state-or-property-permitted-5c01ea.md
WilcoFiers: While yes, prohibited attributes might not be global... depending on how you read that, I think making this explicit is important for the clarity of the rule.

Jym77 (approved):

lbrnpm (approved):

carlosapaduarte (approved):

carlosapaduarte:
Merging

PR #1884, opened 2022-07-05
Update ARIA required properties rule to WAI-ARIA 1.2

### LINKED ISSUE #1282: Element with `role` attribute has required states and properties (4e8ab6): Make combobox examples correct

### LINKED ISSUE #1373: Element with role has required states and properties - Failed example 1 [4e8ab6]

WilcoFiers:
- Add several new related to properties that in ARIA 1.2 no longer have a default value
- Add an example for the separator, which has a required prop only if it is focusable
- Improve the combobox examples, borrowing code from the WAI-ARIA spec itself
- Add accessible names to elements that require one

Closes #1282, closes #1373

---

This is one part of migrating ACT Rules over to WAI-ARIA 1.2. We'll have all WAI-ARIA 1.2 PRs go to the wai-aria 1.2 branch, which we'll merge into the develop branch once everything is ready. This lets us chunk up the work while ensuring the upgrade to WAI-ARIA 1.2 happens all at once.

The final WAI-ARIA 1.2 PR will be a 2 week call for review. Everything going into that branch will require the usual 3 approvals, but will not be sent out as call for review.

---

## How to Review And Approve

### DIFF
--- a/_rules/role-required-states-and-properties-4e8ab6.md
+++ b/_rules/role-required-states-and-properties-4e8ab6.md
@@ -222,6 +222,6 @@ This `combobox` is not [included in the accessibility tree][] due to its styling
 [included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of Included in The Accessibility Tree'
 [wai-aria required states and properties]: https://www.w3.org/TR/wai-aria-1.2/#requiredState
 [wai-aria implicit value for role]: https://www.w3.org/TR/wai-aria-1.2/#implictValueForRole
-[wai-aria 1.1]: https://www.w3.org/TR/wai-aria-1.1/
 [wai-aria 1.2]: https://www.w3.org/TR/wai-aria-1.2/
 [html or svg element]: #namespaced-element
+[focusable]: #focusable
--- a/_rules/role-required-states-and-properties-4e8ab6.md
+++ b/_rules/role-required-states-and-properties-4e8ab6.md
@@ -89,7 +89,7 @@ This `scrollbar` has the required properties `aria-controls` and `aria-valuenow`
 <main id="content"></main>
 ```
 
-### Passed Example 4
+#### Passed Example 4
 
 These `option` nodes do not need the required `aria-selected` property because it has a default value of `false`.
 
@@ -101,7 +101,7 @@ These `option` nodes do not need the required `aria-selected` property because i
 </ul>
 ```
 
-### Passed Example 5
+#### Passed Example 5
 
 This `separator` is not a `widget` because it is not [focusable][]. The `separator` role only requires the `aria-valuenow` property when the element is focusable.
 
--- a/_rules/role-required-states-and-properties-4e8ab6.md
+++ b/_rules/role-required-states-and-properties-4e8ab6.md
@@ -180,7 +180,7 @@ This `combobox` does not have the required `aria-expanded` property. Prior to [W
 
 #### Failed Example 6
 
-This `combobox` uses `aria-owns` where it should use `aria-controls` property. This property is required as of [WAI-ARIA 1.1][].
+This `combobox` uses `aria-owns` instead of using the required `aria-controls` property.
 
 ```html
 <label for="tag_combo">Tag</label>
--- a/_rules/role-required-states-and-properties-4e8ab6.md
+++ b/_rules/role-required-states-and-properties-4e8ab6.md
@@ -15,6 +15,11 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  aria12:requiredState: # 5.2.2 Required States and Properties
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: satisfied
 input_aspects:
   - DOM Tree
 acknowledgments:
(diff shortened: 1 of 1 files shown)


_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +63,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
 ```
 
 ### Failed
 
 #### Failed Example 1
 
-This `combobox` is missing the required `aria-controls` property.
-
-**Note**: In [WAI-ARIA 1.2][], `combobox` will also require `aria-expanded`.
+This `heading` does not have the required `aria-level` property. Prior to [WAI-ARIA 1.2][] the `heading` role had an implicit default `aria-level` value of `2`. As of WAI-ARIA 1.2 this property must be explicitly set.
```
Jym77, line 131: I'm not a huge fan of the "historical" notes of "things were like that at some point but not anymore" 😕 
I feel they only bring value for people who know the details of ARIA 1.1 and wonder about this, but these are likely also the people who can look into the spec directly and figure out the difference by themselves.

Won't block anything because of these, however.

Jym77 (dismissed):

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -60,47 +63,132 @@ This rule relies on browsers and assistive technologies to support leaving out [
...
+<input type="text" id="tag_combo" role="combobox" aria-controls="popup_listbox" />
+<ul role="listbox" id="popup_listbox">
+	<li role="option">Zebra</li>
+	<li role="option" id="selected_option">Zoom</li>
+</ul>
+```
+
+#### Failed Example 6
+
+This `combobox` uses `aria-owns` where it should use `aria-controls` property. This property is required as of [WAI-ARIA 1.1][].
```
WilcoFiers:
```suggestion
This `combobox` uses `aria-owns` instead of using the required `aria-controls` property.
```

_rules/role-required-states-and-properties-4e8ab6.md
WilcoFiers: I get your point, but I think many people do know there's a default level for the heading role. If we don't say this I'm concerned people could think we made a mistake. In my experience, people are more likely to question the quality of the example than they are the quality of their ARIA knowledge.

This doesn't need to be here forever, but I think to start off, it's helpful.

_rules/role-required-states-and-properties-4e8ab6.md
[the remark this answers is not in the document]
WilcoFiers: Probably right on this one. This has been required for years, people who knew about aria-owns being required probably know aria-controls is now too. I'll take that out.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -115,7 +203,7 @@ This `div` does not have a [semantic role](#semantic-role).
 
 #### Inapplicable Example 2
 
-This `checkbox` has an [implicit semantic role](#implicit-role) that is identical to the [explicit semantic role](#explicit-role).
+This `checkbox` has an [implicit semantic role](#implicit-role) that is identical to the [explicit semantic role](#explicit-role). This allows native HTML `checked` state to apply.
```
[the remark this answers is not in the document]
WilcoFiers: Good idea, but seems out of scope. Would you mind opening a separate PR, or an issue on this?

Jym77 (approved):

HelenBurge (approved):

wcreedle (approved):

carlosapaduarte (approved):

PR #1886, opened 2022-07-08
Image-accessible-name: Ensure SVG accessibility by adding role

WilcoFiers:
This was brought to my attention by @schne324. Thanks Harris. Specifically, this comes from NVDA in Firefox, where the entire SVG element is ignored. It's significant enough of a problem that I think we need to make note of it in the rule.

Here's the NVDA issue I opened for it: https://github.com/nvaccess/nvda/issues/13863

---

## How to Review And Approve


_rules/image-accessible-name-descriptive-qt1vmo.md
```diff
@@ -295,6 +295,7 @@ This is a `div` element with a background image. Background images must be teste
 ```
 
 [accessible name]: #accessible-name 'Definition of accessible name'
+[explicit semantic role]: #explicit-semantic-role
```
WilcoFiers:
```suggestion
[explicit semantic role]: #explicit-role
```

colabottles (approved):

kengdoj (approved):

_rules/image-accessible-name-descriptive-qt1vmo.md
```diff
@@ -58,7 +58,7 @@ This rule assumes that the language of the [accessible name][] of each test targ
 
 ## Accessibility Support
 
-_There are no major accessibility support issues known for this rule._
+Some popular browser / screen reader combinations do not pronounce the accessible names of `svgs` elements. This can be resolved by adding an [explicit semantic role][] of `img` to the `svg` element.
```
WilcoFiers:
```suggestion
Some popular browser / screen reader combinations do not pronounce the accessible names of `svg` elements. This can be resolved by adding an [explicit semantic role][] of `img` to the `svg` element.
```

Jym77 (commented):
I'm uncomfortable about this.
We're patching test cases for a specific UA/AT combination that is broken.
* Is our goal that all tests cases behave correctly with all UA/AT despite their bugs?
* Will we remove these explicit roles once the bug is fixed?
* If is was, say, `img` elements that where not correctly announced in that combination, would we add a `role="img"` on all `img` elements in our test cases?

I'm all in favour of the AS note, which is specifically here to call out UA/AT combination that are buggy. But I don't like adding the explicit role

HelenBurge (dismissed):
Should passing examples have role="img" added? As it is a browser issue and not always required. 

`aria-label` is not an appropriate accessible name for an SVG, you can have it on the preceding URL element, but not the SVG. Instead to label an SVG you use the `title` element and `aria-labelledby` to the title element to follow the SVG specifications.

carlosapaduarte (dismissed):
I agree with @Jym77 
I think we should the Accessibility Support note. I don't think we should change the examples because one specific UA/AT combination is buggy.

WilcoFiers:
**With my chair hat**:
Our passed examples need to satisfy the success criteria it maps to, in addition to passing the rule. By leaving this in we're implicitly saying that it is incorrect for implementors to fail this under 1.1.1. That doesn't seem correct, given the support issue.

In the past we've said we weren't going to account for accessibility support problems in the applicability and expectation, but I don't recall we've said anything about the examples. Given where we are on implementations, I suspect we may have to actually adopt a practice where we make sure examples don't include problems mentioned in accessibility support, so that implementors have the flexibility to report those. WCAG 2 allows this, so I think we have to as well.

**With my implementor hat**:
I just want to be clear, axe DevTools Pro reports this as a separate issue. So there is no inconsistency for us. For us the issue comes from the failed example, which we report as an accessibility support problem, since from the perspective of the accessibility support baseline we use, the element has no accessible name.

tombrunet (approved):

ChrisLoiselle:
If this increases the ability for a screen reader and vendor to work properly to provide the end user with the important information, then I agree with this request. It is still up to AT and user agent to code and validate to standards and resolve their own bugs related to code and validating to WCAG and how this relates to accessibility supported , etc.

HelenBurge (approved):

dd8:
Not an issue for us either way. We flag AT support issues, but don't map these to WCAG SCs, and say something along the lines of :

>"This isn't a compliance failure but `xyz` doesn't work in the following screen readers..."

PR #1903, opened 2022-08-04
Handle `content-visibility` and `hidden="until-found"`

### LINKED ISSUE #1860: Programmatically hidden doesn't handle CSS content-visibility: and hidden=until-found
There are new ways to show/hide content:

## CSS content-visibility: property

- `content-visibility: hidden` removes an element contents from the screen and a11y tree (seems easy to resolve)
- `content-visibility: auto` adds elements with `display: none` back into a11y tree (might be harder to address)
 
See https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility#accessibility 

Supported from Chrome/Edge 85 onwards. Not supported in Firefox/Safari

The current definition of Programmatically Hidden doesn't take `content-visibility` into account

## HTML hidden=until-found attribute

A new value has been added to the hidden attribute in HTML - it's no longer a boolean attribute. This makes hidden elements searchable via the browser Find command, and automatically shows then when found. This lets you search text in accordions and similar widgets.

https://github.com/WICG/display-locking/blob/main/explainers/hidden-content-explainer.md
https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute
https://developer.chrome.com/articles/hidden-until-found/

This was added to the Living Standard on 23 March 2022, and is not yet documented on MDN. Supported from Chrome 102.

How should elements with hidden=until-found be treated in rules?

Jym77:
Parking as draft for now.
Both `content-visibility` and `hidden="until-found"` have poor browser support. This means at least that any test case using them would hit an Accessibility Support problem…

---

* Add `content-visibility: hidden` to the reasons for being Programmatically Hidden.
* Add a not that `hidden: until-found` is recommended to be handled via `content-visibility: hidden`.
* Add inapplicable examples to rules that directly use "Programmatically Hidden".

There are many more rules that indirectly use "Programmatically hidden", because it is mentioned in "Included in the Accessibility tree". And indeed, programmatically hidden elements meet the conditions for being [excluded from the Accessibility Tree](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion).

I'm however not sure we need to add examples with these new properties or value on half the ruleset 🤔 
OTOH, it may be worth doing it like here with one single Inapplicable Example grouping many cases of exclusion. This does not clutter rules with a lot of extra examples, but still allow to exercice the definition (and it becomes the task of the tool vendors to understand why they fail a given test cases)

Closes issue(s):

- closes #1860 

Need for Call for Review:
This will require a 2 weeks Call for Review (updating a definition with impact in many rules)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

f


pages/glossary/programmatically-hidden.md
```diff
@@ -11,11 +11,12 @@ input_aspects:
...
 
 **Note**: Contrary to the other conditions, the `visibility` CSS property may be reverted by descendants.
 
-**Note**: The [HTML standard suggests](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements) setting the CSS `display` property to `none` for elements with the `hidden` attribute. While not required by HTML, all modern browsers follow this suggestion. Because of this the `hidden` attribute is not used in this definition. In browsers that use this suggestion, overriding the CSS `display` property can reveal elements with the `hidden` attribute.
+**Note**: The [HTML standard suggests](https://html.spec.whatwg.org/multipage/rendering.html#hiddenCSS) setting the CSS `display` property to `none` for elements with the `hidden` attribute; and the CSS `content-visibility` property to `hidden` for elements with the `hidden` attribute set to `until-found`. While not required by HTML, all modern browsers follow this suggestion. Because of this the `hidden` attribute is not used in this definition. In browsers that use this suggestion, overriding the CSS `display` or `content-visibility` properties can reveal elements with the `hidden` attribute.
```
carlosapaduarte:
```suggestion
**Note**: The [HTML standard suggests](https://html.spec.whatwg.org/multipage/rendering.html#hiddenCSS) setting the CSS `display` property to `none` for elements with the `hidden` attribute set to the `hidden` state; and the CSS `content-visibility` property to `hidden` for elements with the `hidden` attribute set to `until-found`. While not required by HTML, all modern browsers follow this suggestion. Because of this the `hidden` attribute is not used in this definition. In browsers that use this suggestion, overriding the CSS `display` or `content-visibility` properties can reveal elements with the `hidden` attribute.
```

carlosapaduarte (dismissed):

colabottles (approved):

carlosapaduarte (approved):

HelenBurge (approved):

giacomo-petri (approved):

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -170,10 +170,15 @@ This `role` attribute is only [ASCII whitespace][].
...
-This `role` attribute is specified on an element which is [programmatically hidden][].
+These `role` attributes are specified on elements which are [programmatically hidden][].
 
 ```html
 <div aria-hidden="true" role="banner">Some Content</div>
+<div hidden role="banner">Some Content</div>
+<div hidden="until-found" role="banner">Some Content</div>
+<div style="display: none" role="banner">Some Content</div>
+<div style="visibility: hidden" role="banner">Some Content</div>
+<div style="content-visibility: hidden" role="banner">Some Content</div>
```
WilcoFiers: I'm not sure these are all necessary. We're not writing unit tests for the `programmatically hidden` definition here. For what is necessary we need separate examples. We should only have one test per test case.

WilcoFiers (dismissed):

pages/glossary/programmatically-hidden.md
```diff
@@ -11,11 +11,12 @@ input_aspects:
 An HTML element is _programmatically hidden_ if either it has a [computed][] CSS property `visibility` whose value is not `visible`; or at least one of the following is true for any of its [inclusive ancestors][] in the [flat tree][]:
 
 - has a [computed][] CSS property `display` of `none`; or
+- has a [computed][] CSS property `content-visibility` of `hidden`; or
```
WilcoFiers: This is wrong, an element with `content-visibility:hidden` is not hidden from the accessibility tree. Its content is, but the element itself isn't. To make this more complicated. Even with `content-visibility:hidden`, child nodes can (in Chrome) still be used in the accessible name. For example, not only is the following still a button in the accessibility tree, its accessible name is still "Hello world", even though the text isn't visible.

```html
<button style="content-visibility:hidden">
  Hello world
</button>
```

dd8: There are various caveats on implementing content-visibility: to avoid screen reader detection:
https://drafts.csswg.org/css-contain/#cv-a11y

And there is an outstanding issue on this:
https://github.com/w3c/csswg-drafts/issues/5857

It may be worth holding off on this until there's consensus on how AT should implement this.

**Edit:** The spec seems to suggest that what gets exposed to AT depends on timings of CSS layout. They want to avoid AT detection due to slow layout with AT vs. fast layout without AT. There are definitely big timing differences between Firefox (multi-thread, breadth-first CSS layout written in Rust) and Chrome/WebKit (single thread, depth-first layout written in C++)

Jym77:
@dd8 Tried to sum up the discussion about `auto` in an assumption + note in the definition.

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -170,12 +170,28 @@ This `role` attribute is only [ASCII whitespace][].
...
+<div hidden role="banner">Some Content</div>
+```
+
+#### Inapplicable Example 7
+
+This `role` attributes is specified on an element which is [programmatically hidden][].
+
+```html
+<div hidden="until-found" role="banner">Some Content</div>
+```
```
WilcoFiers, line 193: I think I'd prefer these test cases in different rules. This rule is has been somewhat contested. How about adding the `hidden` example to `Form field has non-empty accessible name`, and have the image rule above use `hidden="until-found"`?

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -246,6 +246,16 @@ This `img` element is hidden with `visibility: hidden`.
...
 ```
 
+#### Inapplicable Example 6
+
+This `img` element is [programmatically hidden][] with `content-visibility: hidden`.
+
+```html
+<div style="content-visibility: hidden">
+  <img src="/test-assets/shared/w3c-logo.png" />
+</div>
```
WilcoFiers, line 253: Until this is better supported I don't think we should add this. This example fails in Firefox and Safari. I think its better to use hidden=until-found here. If you wanted to pass in Chromium you'd still need to support content-visibility, but the example wouldn't be a failure in browsers that don't support it yet. What's also neat specifically about using it in this rule is that there's nothing to be found here, so this couldn't become a visible image through some user action.

```suggestion
<div hidden="until-found">
  <img src="/test-assets/shared/w3c-logo.png" />
</div>
```

pages/glossary/programmatically-hidden.md
```diff
@@ -8,15 +8,29 @@ input_aspects:
   - DOM tree
 ---
 
-An HTML element is _programmatically hidden_ if either it has a [computed][] CSS property `visibility` whose value is not `visible`; or at least one of the following is true for any of its [inclusive ancestors][] in the [flat tree][]:
+An HTML element is _programmatically hidden_ if any of the following is true:
```
WilcoFiers, line 11: Not relevant to this PR, but I thought I'd share. I recently discovered that you can't hide area elements with display:none. Probably not something we need to worry about, I just thought it was interesting.

pages/glossary/programmatically-hidden.md
```diff
@@ -8,15 +8,29 @@ input_aspects:
...
 
-- has a [computed][] CSS property `display` of `none`; or
-- has an `aria-hidden` attribute set to `true`
+- the element has a [computed][] CSS property `visibility` whose value is not `visible`; or
+- at least one [ancestor][] of the element has a [computed][] CSS property `content-visibility` of `hidden`; or
+- at least one [inclusive ancestors][] of the element has a [computed][] CSS property `display` of `none`; or
+- at least one [inclusive ancestors][] of the element has an `aria-hidden` attribute set to `true`
```
WilcoFiers, line 16: Not new to the PR, we could open a separate issue, but this is incorrect I think. If you use aria-owns to grab a node that has an ancestor with aria-hidden, that node itself isn't hidden.

WilcoFiers (changes requested):

pages/glossary/programmatically-hidden.md
Jym77: Yes, that's also the conclusion that was reach [in this ARIA PR](https://github.com/w3c/aria/pull/1839#issuecomment-1341754741). 

I'm recording that in #1991 .

pages/glossary/programmatically-hidden.md
Jym77: `area`  elements are not really rendered (and do have a default `display: none`: [HTML rendering guidelines](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements)).
It is the associated image which is rendered, but neither the `map` nor the `area` is (they are mostly abstractions for interactions).

_rules/image-non-empty-accessible-name-23a2a8.md
Jym77: `hidden="until-fonud"` also seems to have poor support: [MDN browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden#browser_compatibility).
I guess it's best to just park this until there is better support…

Jym77:
Looks like both `content-visibility` and `hidden="until-found"` have poor browser support for now.
Parking these to revisit later.

PR #1904, opened 2022-08-04
Text contrast enhanced: Fix passed example 3 not fully passing

WilcoFiers:
Reported by Kathy. Colors around the "k" in that text dropped below the 7:1 contrast.

---

## How to Review And Approve


HelenBurge (approved):
I would prefer to remove px and use rem... but that is not something for this PR :)

Jym77 (approved):

kengdoj (approved):

_rules/text-contrast-enhanced-09o5cg.md
```diff
@@ -102,11 +102,16 @@ This dark gray text has a contrast ratio between 12.6:1 and 7:1 on the white to
 This light gray text has a contrast ratio between 18:1 and 7:1 on the background image.
```
giacomo-petri:
```suggestion
This white text has a contrast ratio between 18:1 and 7:1 on the background image.
```

New color set is white, no longer light grey 😄

PR #1905, opened 2022-08-04
"Focusable element has no keyboard trap via standard navigation" [a1b64e]: fix Failed example and add clarification

### LINKED ISSUE #1885: No Keyboard Trap [a1b64e] - note and Fail Ex 1
Question about the Expectation for the [rule](https://act-rules.github.io/rules/a1b64e) for no keyboard trap:

> For each target element, focus can cycle to the browser UI by using standard keyboard navigation.
> Note: Cycling back to the browser UI can be done both by moving forward through the tab order and by moving backwards.

Does the expectation require moving forward _and_ backward or _either_ forward or backward to pass?

The SC requires "focus can be moved away" (and does not specify in both directions):

> Success Criterion [2.1.2 No Keyboard Trap](https://www.w3.org/TR/WCAG21/#no-keyboard-trap) (Level A): If keyboard focus can be moved to a component of the page using a keyboard interface, then focus can be moved away from that component using only a keyboard interface, and, if it requires more than unmodified arrow or tab keys or other standard exit methods, the user is advised of the method for moving focus away.

There seems to be difference in browsers and OS for [Failed Example 1](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/80af7b/780388a837915960bca7651bd80743fb2cafdbcb.html):

- I tested on Chrome/Firefox/Edge + Windows: I am able to tab forward and move focus away from the button to the browser.  I cannot move focus away from the button when going backwards.

- A colleague tested on Chrome + Mac: focus could not move away from the button in either direction. 

- Whether the link gets focus is also inconsistent in different browsers.

Jym77:
* Fix Failed Example 1 to ensure there actually is a keyboard trap
* Add a clarification about forward and backward navigation

Closes issue(s):

- closes #1885 

Need for Call for Review:
This will require a 1 week Call for Review (updating one example, adding a background note)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


colabottles (approved):

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
```diff
@@ -42,6 +42,8 @@ There are no major accessibility support issues known for this rule.
 
 ## Background
 
+[Standard keyboard navigation](#standard-keyboard-navigation) can move focus both forward and backward. Therefore, cycling back to the browser UI can be done both by moving forward through the tab order and by moving backwards. It is not fully clear whether [Success Criterion 2.1.2 No keyboard trap][sc212] requires navigation in both directions. Therefore, this rule does not constrain navigation direction, to stay on the safe side.
```
carlosapaduarte:
```suggestion
[Standard keyboard navigation](#standard-keyboard-navigation) can move focus both forward and backward. Therefore, cycling back to the browser UI can be done both by moving forward through the tab order and by moving backwards. It is not fully clear whether [Success Criterion 2.1.2 No keyboard trap][sc212] requires navigation in both directions. Therefore, to stay on the safe side, this rule does not constrain navigation direction.
```

carlosapaduarte (approved):
Proposing an editorial change, but I'm okay if you don't change it, so I'm approving already.

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
```diff
@@ -29,7 +29,7 @@ This rule applies to any [HTML or SVG element][] that is [focusable][].
 
 For each target element, focus can cycle to the browser UI by using [standard keyboard navigation](#standard-keyboard-navigation).
```
giacomo-petri, line 30:
```suggestion
For each target element, focus can be moved away from that element using only [standard keyboard navigation](#standard-keyboard-navigation).
```

I would suggest a better adherence with SC ([2.1.2: No Keyboard Trap](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html)). 
- "focus can cycle to the browser UI" is related to the entire flattened tabindex ordered focus navigation scope, not to the single target element;
- "focus can cycle to the browser UI" navigating on Safari with the TAB, focus moves from the last tabbable element to the body element, skipping the "Browser UI". On the contrary, navigating backward (SHIFT+TAB) focus can be moved to the address-bar. Anyway, the new proposal should avoid any confusion.

Jym77: I'm a bit reluctant to change the expectation in this PR 🤔 Especially since that changes its scope quite a bit.
Can you open an issue for that? And we'll take it in a separate PR.

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
```diff
@@ -80,13 +82,14 @@ This element is made [focusable][] by the `tabindex` attribute, even if it is no
...
 
-This [focusable][] element creates a keyboard trap bringing focus to the `button`.
+This [focusable][] element creates a keyboard trap bringing focus to the `button`. Note that if one of the links is removed, the focus may jump to the browser UI before the timeout expires, at which point the `this.focus()` trap cannot trigger anymore.
 
 ```html
 <a href="#">Link 1</a>
 <button onblur="setTimeout(() => this.focus(), 10)">
 	Button1
 </button>
+<a href="#">Link 1</a>
```
giacomo-petri: Irrelevant for understanding, but...

According to [focusable](https://www.w3.org/WAI/standards-guidelines/act/rules/a1b64e/proposed/#focusable) definition

> Exception: Elements that lose focus during a period of up to 1 second after gaining focus, without the user interacting with the page the element is on, are not considered focusable.

when focus is on the button element, first link (navigating backward) and second link (navigating forward) are not considered focusable. So in this scenario, when focus is on the "button" element, we've just one "focusable" element within the page that is the button itself. 

Should we replace "[focusable][]" with "part of [sequential focus navigation](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation)" in Applicability?

giacomo-petri: @Jym77, per your answer above, I'm opening a new issue.

Jym77: I don't think that's needed 🤔 
I agree it is a bit weird that the focusability of the links depends on the state of the page (they are "focusable" only if the button isn't currently focused…) That's more a problem with the definition, I'd say.

But in any cases, the button still has a keyboard trap sa defined by this rule (focus can't move away).

HelenBurge (approved):

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
WilcoFiers:
```suggestion
<a href="#">Link 2</a>
```

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
```diff
@@ -42,6 +42,8 @@ There are no major accessibility support issues known for this rule.
 
 ## Background
 
+[Standard keyboard navigation](#standard-keyboard-navigation) can move focus both forward and backward. Therefore, cycling back to the browser UI can be done both by moving forward through the tab order and by moving backwards. It is not fully clear whether [Success Criterion 2.1.2 No keyboard trap][sc212] requires navigation in both directions. Therefore, to stay on the safe side, this rule does not constrain navigation direction.
```
WilcoFiers: This is an assumption. I think we put it there. Also, might be good to open an issue with AGWG to just ask this question.

My recollection of this though was that we'd said that forward navigation was required, but backward navigation was not. Am I misremembering that?

WilcoFiers (dismissed):

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
kengdoj: I agree with Wilco's suggested change.

kengdoj (approved):
Line 92's suggested change is a good one.

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
Jym77: > My recollection of this though was that we'd said that forward navigation was required, but backward navigation was not. Am I misremembering that?

As per [CG discussion of July 14th](https://www.w3.org/2022/07/14-act-r-minutes.html#item05) and the following [summary of action](https://github.com/act-rules/act-rules.github.io/issues/1885#issuecomment-1184535392), we decided that either forward or backward was enough (to stay on the safe side).

I agree this should be an Asumption 👍

Jym77: > This is an assumption. I think we put it there. Also, might be good to open an issue with AGWG to just ask this question.

https://github.com/w3c/wcag/issues/2678

Putting this on hold, waiting to see if we have an answer.

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
```diff
@@ -29,12 +29,13 @@ This rule applies to any [HTML or SVG element][] that is [focusable][].
 
 For each target element, focus can cycle to the browser UI by using [standard keyboard navigation](#standard-keyboard-navigation).
 
-**Note:** Cycling back to the browser UI can be done both by moving forward through the tab order and by moving backwards. It is not possible to fulfill this expectation by using browser specific shortcuts to return to the browser UI.
+**Note:** It is not possible to fulfill this expectation by using browser specific shortcuts to return to the browser UI.
```
WilcoFiers, line 32: I'm not so sure that things that can only be exited with shift+tab should pass this SC. I think we need to get clarity on that. This note shouldn't have been a note. Knowing whether its backward, forward, or both needs to be in the expectation, but then by taking this out of the note without changing the expectation we're making the rule less clear.

WilcoFiers (dismissed):

_rules/focusable-no-keyboard-trap-standard-nav-a1b64e.md
Jym77: > This note shouldn't have been a note. 

Which is why we upgrade it to an Assumption 😁 

The "one way is enough" view has been discussed on [CG call on July 14th](https://www.w3.org/2022/07/14-act-r-minutes.html#item05), check also the [action point summary](https://github.com/act-rules/act-rules.github.io/issues/1885#issuecomment-1184535392).

---
Thus said, I do agree that our definition of "standard keyboard navigation" lists 6 keys (or combination) and the expectation is not fully clear whether one of these is enough (technically, it is because the def is "one or more of [these 6 keys]", so cycling back to UI using only one of the listed method does fulfil the definition).

giacomo-petri (approved):

WilcoFiers (approved):

Jym77:
Call for Review neds on January 16th.

bruce-usab (changes requested):
I would appreciate having this rule discussed for 10 min on AGWG call or adding more of caveat regarding support for Shift-Tab not being absolutely necessary. 

I am okay with _this rule_  rule assuming that only requires navigation in one direction (either forward or backward), not both, and not a specific one.

I am of the opinion that  [Failed Example](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/80af7b/780388a837915960bca7651bd80743fb2cafdbcb.html) _is_ nominally a SC 2.1.2 failure.

I am not okay with assuming SC 2.1.2 generally requires Tabbing through UI but Shift-Tab not working is permissible.  I am not comfortable that the rule-as-written implies otherwise.

Jym77:
@bruce-usab I've moved this assumption to a background note (given that the result is already covered by the "further testing needed" mapping of the parent composite rule). And I've tuned it down on what WCAG means, making it clearer that this is an assumption/simplification of the rule without implying anything about WCAG meaning. 
I think this should cover your hesitation, let me know if this sounds good for you.

Jym77:
Call for Review has ended. Merging.

PR #1912, opened 2022-08-17
c4a8a4 -- Updates to web page definition and editorials

daniel-montalvo:
Updates definitions, updates applicability, and removes technique H25 from assumptions as it is already in the bibliography section.

Closes issue(s):

Need for Call for Review:
This will require a 2 weeks Call for Review: substantial changes affecting description, applicability, and background.

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -37,18 +37,15 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to the first [HTML][] `title` element that
-
-- is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the `html` element of a [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), and
-- contains [children](https://dom.spec.whatwg.org/#concept-tree-child) that are [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
+This rule applies to the first `title` element on an [html web page][] that is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the [document element](https://dom.spec.whatwg.org/#document-element) that contains [children](https://dom.spec.whatwg.org/#concept-tree-child) that are [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```
WilcoFiers: I think we can simplify this a good bit:

```suggestion
This rule applies to the [document title][] of each [html web page][].
```

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -175,4 +172,4 @@ This `title` element is a child of an `svg` element.
 </svg>
 ```
 
-[html]: #namespaced-element
+[html web page]: #web-page-html 'Definition of Web Page (HTML)'
```
WilcoFiers, line 176:
```suggestion
[document title]: https://html.spec.whatwg.org/multipage/dom.html#document.title 'HTML definition of Document Title'
[html web page]: #web-page-html 'Definition of Web Page (HTML)'
```

WilcoFiers (dismissed):

HelenBurge (approved):

WilcoFiers (approved):

tbostic32 (approved):

tombrunet (approved):

colabottles (approved):

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -175,4 +172,5 @@ This `title` element is a child of an `svg` element.
 </svg>
 ```
 
-[html]: #namespaced-element
+[document title]: https://html.spec.whatwg.org/multipage/dom.html#document.title 'HTML definition of Document Title'
```
Jym77: This should rather point to
```suggestion
[document title]: https://html.spec.whatwg.org/multipage/dom.html#the-title-element-2 'HTML definition of Document Title'
```
(i.e. "the first `title` element …")

And we might want to accordingly change or remove the AS "assumption" that browsers do use it correctly if we take a definition from the specs.

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -37,18 +37,15 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to the first [HTML][] `title` element that
-
-- is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the `html` element of a [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), and
-- contains [children](https://dom.spec.whatwg.org/#concept-tree-child) that are [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```
Jym77, line 43: Removing the "non empty" condition does change the Applicability a bit.
I don't remember why we had this in the first place (and the split with [HTML page has non-empty title](https://act-rules.github.io/rules/2779a5) which is automatable).

Jym77 (dismissed):

_rules/html-page-title-descriptive-c4a8a4.md
daniel-montalvo: > Removing the "non empty" condition does change the Applicability a bit. I don't remember why we had this in the first place (and the split with [HTML 
 page has non-empty title](https://act-rules.github.io/rules/2779a5) which is automatable).

Thanks @Jym77 

I don't remember either :-)

But do you think we should not get rid of that condition? Does not the browser automatically discard the first title elements if it's empty?

Jym77: I'm not sure about discarding or not the condition since I do not remember why it was here in the first place (so I'm afraid there was a good reason and it would be a mistake to remove it).

I would at least want the removal to be a conscious decision in its own PR rather than a side effect hidden in an editorial PR (this is not purely editorial, and we want the changelog for the rule to list the Applicability change).

WilcoFiers: Interesting! Good catch @Jym77.  Doing some quick testing, empty `title` elements do not get ignored, so this update is correct. So we just accidentally fixed a bug!

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -37,18 +37,15 @@ htmlHintIgnore:
...
+This rule applies to the [document title][] of each [html web page][].
```
daniel-montalvo: Hey @Jym77  and @WilcoFiers 

What about something like this?

```suggestion
This rule applies to the first [document title][] of each [html web page][] that contains [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```

_rules/html-page-title-descriptive-c4a8a4.md
Jym77: 👍 

Given the change in Applicability, I think we need to add a Failed Example with an empty `title`.

Jym77 (dismissed):

Jym77 (dismissed):
I'm getting confused 🙈 
At [this commit](https://github.com/act-rules/act-rules.github.io/pull/1912/commits/e950cbea2679cbe7320f427624607a71df772623) we did put back the "non-empty" condition.

But that seems to contradict @WilcoFiers comment (https://github.com/act-rules/act-rules.github.io/pull/1912/files#r963601609) that empty titles are picked up, and are indeed a failure (which is already caught by the automatable [HTML page has non-empty title](https://act-rules.github.io/rules/2779a5)).

And in the current state, empty titles are inapplicables (again) (and we can't have them as Inapplicable examples because they fail the requirement).

…anyway…
1. Is an empty first `title` a problem => yes, and it's already caught by [HTML page has non-empty title](https://act-rules.github.io/rules/2779a5)
2. Should this rule applies to them (and fail) or not => I don't have a strong opinion 🤷 

But we definitely need to make sure that the Applicability and Examples match.
(sorry, I feel I've been losing context and focus at some point and made conflicting reviews…)

daniel-montalvo:
> 1. Is an empty first `title` a problem => yes, and it's already caught by [HTML page has non-empty title](https://act-rules.github.io/rules/2779a5)
> 2. Should this rule applies to them (and fail) or not => I don't have a strong opinion 
> 🤷

If empty titles were picked up by browsers, there could be a chance for a subsequent title to be descriptive. But since browsers still don't pick up empty titles, when the first title element is empty, it's also not descriptive.

I'd say they should be failed also in this rule.

> But we definitely need to make sure that the Applicability and Examples match. (sorry, I feel I've been losing context and focus at some point and made conflicting reviews…)

Fair enough. I think we should revert e950cbea2679cbe7320f427624607a71df772623

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -37,18 +37,15 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to the first [HTML][] `title` element that
-
-- is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the `html` element of a [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), and
-- contains [children](https://dom.spec.whatwg.org/#concept-tree-child) that are [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
+This rule applies to the first [document title][] of each [html web page][] that contains [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```
WilcoFiers: Probably fine to remove the word "first", since there can only be one document title.

```suggestion
This rule applies to the [document title][] of each [html web page][] that contains [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```

_rules/html-page-title-descriptive-c4a8a4.md
WilcoFiers: @Jym77 I'm not sure that's necessary, but I don't see the harm in it either. Daniel added the example, can you have another look?

WilcoFiers (dismissed):

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -175,4 +172,22 @@ This `title` element is a child of an `svg` element.
 </svg>
 ```
 
-[html]: #namespaced-element
+#### Failed Example 3
```
daniel-montalvo: This example is now removed again since this would fail the SC

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -37,18 +37,15 @@ htmlHintIgnore:
 
 ## Applicability
 
-This rule applies to the first [HTML][] `title` element that
-
-- is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the `html` element of a [web page](https://www.w3.org/TR/WCAG21/#dfn-web-page-s), and
-- contains [children](https://dom.spec.whatwg.org/#concept-tree-child) that are [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
+This rule applies to the [document title][] of each [html web page][] that contains [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```
Jym77: I feel the relative doesn't work 🤔 It sounds like "that" refers to "html web page" rather than "document title".
```suggestion
This rule applies to the [document title][] of each [html web page][] if this [document title][] contains [text nodes](https://dom.spec.whatwg.org/#text) that are not only [whitespace](#whitespace).
```

_rules/html-page-title-descriptive-c4a8a4.md
```diff
@@ -37,18 +37,15 @@ htmlHintIgnore:
...
 
 ## Expectation
 
 The target element describes the topic or purpose of the overall content of the [document](https://dom.spec.whatwg.org/#concept-document).
 
 ## Assumptions
 
-- A prerequisite for this rule is that Technique H25: Providing a title using the title element is used for living up to 2.4.2 Page Titled
+There are currently no assumptions.
```
Jym77, line 48: I think the Assumption still exists, but maybe in a slightly different form. See #1924 (an email written as HTML is an "html web page", the HTML document might have an empty `title` element but the email's Subject is actually used as the "title" (in WCAG sense), so it wouldn't fail the SC).
OTOH, we might let that as is and tackle that in another PR when solving #1924.

Jym77 (dismissed):

_rules/html-page-title-descriptive-c4a8a4.md
daniel-montalvo: > I think the Assumption still exists, but maybe in a slightly different form. See #1924 (an email written as HTML is an "html web page", the HTML document might have an empty `title` element but the email's Subject is actually used as the "title" (in WCAG sense), so it wouldn't fail the SC). OTOH, we might let that as is and tackle that in another PR when solving #1924.

+1 to keeping the assumption until we resolve that issue.

Jym77 (approved):

WilcoFiers (approved):

PR #1914, opened 2022-09-05
Error field is not identified using color differences only

giacomo-petri:
Proposal for a new rule to check errors are not identified using color differences only.

Closes issue(s):
- new rule

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
+---
+id:
```
Jym77: You need to provide an id, and to use it in the filename, plus the filename should be `.md`
As per [Rule design](https://act-rules.github.io/pages/design/rule-design/#rule-identifier), you can use [random.org](https://www.random.org/strings/?num=1&len=6&digits=on&loweralpha=on&format=html&rnd=new) to generate random ids.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+name: Error field is not identified using color differences only
+rule_type: atomic
+description: |
+  This rule checks that error fields can be identified not only by color differences but through another visual means.
+accessibility_requirements: # Remove whatever is not applicable
```
Jym77:
```suggestion
accessibility_requirements:
```

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+  -  DOM Tree
+  -  CSS Styling
+acknowledgments:
+  authors:
+    -  Giacomo Petri
+---
+
+## Applicability
+
+This rule applies to any error field that is [visible][].
```
Jym77: What is an "error field"? This needs an unambiguous and objective definition.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+## Expectation (1)
```
Jym77:
```suggestion
## Expectation
```

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+Each test target, together with its visual context, identify the presence of errors not only through color differences.
```
Jym77: What does "not only through color differences" mean?
Notably, does color refers to the full RGB color or only to the hue? Is a difference in contrast good enough (from which treshold).

Look at #1010 to see some other discussions around 1.4.1 and what constitute a good enough difference, OTOH there might be other thing this rule needs to consider.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+## Assumptions
+
+If the presence of errors is conveyed through the use of colors that differ not only in their hue, but that also have a difference in lightness and the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater, then this counts as an additional visual mean.
```
Jym77: This is not really an assumption, but part of the expectation.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+*Note*: It is possible for an error field to be distinguishable from adjacent fields but still not have a meaningful error message. In that case, it would pass Success Criterion 1.4.1: Use of Color but still fail [Success Criterion 3.3.1: Error Identification](https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html) and, if applicable, [Scuccess Criterion 3.3.3: Error Suggestion](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html).
```
Jym77: That belongs more the "Background" section, I think. This is not an assumption that makes the rule behave incorrectly when it's not met.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+## Accessibility Support
+
+_No accessibility support issues known._
```
Jym77:
```suggestion
There are no major accessibility support issues known for this rule.
```
seems to be the way it is in other rules.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+- [Understanding Success Criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)
+- [F81 - Failure of Success Criterion 1.4.1 due to identifying required or error fields using color differences only](https://www.w3.org/WAI/WCAG21/Techniques/failures/F81.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The presence of errors is conveyed through text error message.
```
Jym77: We usually try to front-load the target in the test case descriptions.
```suggestion
This error message (the `p` element) conveyed the presence of an error through its text.
```
(similar remark apply to other descriptions)

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+```html
+<label for="first_name">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:1px solid #dd0000" aria-invalid="true" aria-describedby="first_name_error_msg">
+<p id="first_name_error_msg" class="error_msg" style="color:#dd0000">This field is empty. Enter your first name.</p>
```
Jym77: I also feel this example (actually all examples) would require a non-errored field so that we have something to compare to.
Maybe this form is super bad and says "This field is empty" on every field, just in red for errors and green for valid (that would fail both 1.4.1 and 3.3.1, of course).

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+The presence of errors is conveyed also through other CSS properties:
+- font-weight: bold AND;
+- border-width: 3px.
+
+```html
+<label for="first_name" style="color:#dd0000">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:3px solid #dd0000" aria-invalid="true">
+<label for="middle_name" style="color:#333333">Middle Name (optional)</label>
+<input type="text" id="middle_name" autocomplete="additional-name" style="border:1px solid #b0b0b0" value="Michael">
```
Jym77: I'm not sure the middle name adds much to the example. There already is one errored and one valid field.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+<label for="first_name">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:1px solid #dd0000" aria-invalid="true" aria-describedby="first_name_error_msg">
+<p id="first_name_error_msg" class="error_msg" style="color:#dd0000">This field is empty. Enter your first name.</p>
+```
+
+#### Passed Example 2
+
+The presence of errors is conveyed also through other CSS properties:
+- font-weight: bold AND;
+- border-width: 3px.
```
Jym77: If only one of these is enough, the example should be shrinked down (and maybe split).
This looks like both are required to pass the rule.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+```html
+<label for="first_name" style="color:#dd0000">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:3px solid #dd0000" aria-invalid="true">
+<label for="middle_name" style="color:#333333">Middle Name (optional)</label>
+<input type="text" id="middle_name" autocomplete="additional-name" style="border:1px solid #b0b0b0" value="Michael">
+<label for="last_name" style="color:#333333">Last Name</label>
+<input type="text" id="last_name" autocomplete="family-name" style="border:1px solid #b0b0b0" value="Doe">
```
Jym77: In that example, there is no error message, while in the first one there is.
This is why I feel the Applicability needs to be made more precise. Are we targeting:
*  errored form fields (i.e., `input`, … elements as listed in [Form field label is descriptive](https://act-rules.github.io/rules/cc0f0a) for example)? Then Passed Example 1 needs to change its explication a bit (the error is indicated by a message programatically linked to the errored field);
* error messages? Then Passed Example 2 is not applicable (there is no error message)?
* [programmatic label](https://act-rules.github.io/rules/cc0f0a#programmatic-label) of errored form field? Then neither Passed Example 1 (it's a description, not a programmatic label) nor Passed Example 2 (it's the field, not the label that indicates the error);
* something else?

---
Also, related to that, what is an errored field in the first place?
I like what you do with `aria-invalid` in the example. And I think we can actually use that and define "errored field" as "with `aria-invalid` attribute value of `true` [or better different from `false` since it can also be `spelling` or `grammar`]". Plus an assumption that this is used correctly.
Then we do have an objective an unambiguous definition of which fields should be looked at. Additionally, we can likely rely on UA to set it correctly when some other mechanism says the field is invalid 🤞 

Having a clear definition of errored field, we can work on defining the visual error indicator. From examples, it seems like it should be either the field, or its programmatic label, or visibles elements used in its accessible name or description, or the element referred to by `aria-errormessage`. As far as I understand, at least one of these must have a "visual, non-hue indication".

I'm not sure we want to provide a full list of valid indications 🤔 This might be too restrictive.

---
See also what is done in [Error message describes invalid form field value](https://act-rules.github.io/rules/36b590) to work with error messages, although I'm not sure this is the best way to handle it 😕

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+### Passed
+
+#### Passed Example 1
+
+The presence of errors is conveyed through text error message.
+
+```html
+<label for="first_name">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:1px solid #dd0000" aria-invalid="true" aria-describedby="first_name_error_msg">
```
Jym77:
```suggestion
<input type="text" id="first_name" autocomplete="given-name" style="border:1px solid #dd0000" aria-invalid="true" aria-errormessage="first_name_error_msg">
```
It also feels like this field should be [`required`](https://html.spec.whatwg.org/multipage/input.html#attr-input-required)

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+#### Passed Example 2
+
+The presence of errors is conveyed also through other CSS properties:
+- font-weight: bold AND;
+- border-width: 3px.
+
+```html
+<label for="first_name" style="color:#dd0000">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:3px solid #dd0000" aria-invalid="true">
```
Jym77: Making it `required` would make it more obvious the reason why this is an error, especially with an explicitly optional field in the form. Makin the last name also `required` could make sense.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+## Background
+
+- [Understanding Success Criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)
+- [F81 - Failure of Success Criterion 1.4.1 due to identifying required or error fields using color differences only](https://www.w3.org/WAI/WCAG21/Techniques/failures/F81.html)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
```
Jym77: I feel we need examples that use something else than text inputs and other errors than absence of a required field. That can be both other built-in errors (check the list of [constraint validation errors](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#validity-states) for inspiration); or custom errors (e.g. "password must contain at least 8 characters, …" that is usually checked by some script).

Notably, that would show that the rule also considers custom errors.
Also, I do believe that if built-in constraint validations are failing, the UA should automagically set `aria-invalid` (needs to be tested).

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+#### Failed Example 1
+
+The presence of errors is conveyed through color (hue) differences only.
+
+```html
+<label for="first_name" style="color:#dd0000">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" style="border:1px solid #dd0000" aria-invalid="true">
+<label for="last_name" style="color:#333333">Last Name</label>
+<input type="text" id="last_name" autocomplete="family-name" style="border:1px solid #b0b0b0" value="Doe">
```
Jym77: Could be an example where the color difference is only on the field (the labels don't even have a color difference), and one where it is only on the label.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
```
Jym77: There should be an example where there are only non-errored form fields.

_rules/error-field-is-not-identified-using-color-only
```diff
@@ -0,0 +1,107 @@
...
+
+There are no error fields.
+
+```html
+<p>This is a paragraph.</p>
+```
```
Jym77: You need to add the reference list of the terms used in the rule.
For example, you use `[visible][]`, so there should be here a definition:
```
[visible]: #visible 'Definition of Visible'
```
(see how it's done in other rules files

Jym77 (changes requested):
Good work. There are still quite a bunch of precision to add to the Applicability and Expectation to meet the requirements of ACT rules 😃

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: Passed 2-3 and failed 1 examples already contain both input with errors and input without errors.
Fixed example 1.

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: Removed all instances of middle name

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: At the beginning, the example included just the bold CSS property; on the other hand I thought the error indicator was more clear and visually more prominent highlighting both. I've added both considering we should try to aim as possible best practices.

Would be ok to specify that "(just one of the CSS properties is sufficient)"?

Happy to split it if it's more clear.

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: updated keeping both label and border colors and added 2 failed examples for only label color and only input border color

giacomo-petri: Added also inapplicable example with hidden form.

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: @Jym77, I've added the required attribute to all required fields. With the required attribute the browser should manage automatically the error validation for values that are no matching requirements (e.g. empty input with required attribute). Should we add in the Assumptions section that we assume form has novalidate attribute or some javascript that prevent browsers to generate native error messages?

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: same as above

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: The idea was to follow the WCAG 1.4.1 requirement

> If content is conveyed through the use of colors that differ not only in their hue, but that also have a significant difference in lightness, then this counts as an additional visual distinction, as long as the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater.

So color differences was intended as differences only in their hue, as also specified in Assumptions section. Per other feedbacks I've moved assumption notes in expectation.

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: Updated with different examples and added more details in what's failing.
See also my previous feedback about "required" and browser form validation (something to add in assumptions).

_rules/error-field-is-not-identified-using-color-only
giacomo-petri: That's a good point I didn't think about.

>  I think we can actually use that and define "errored field" as "with aria-invalid attribute value of true [or better different from false since it can also be spelling or grammar]". Plus an assumption that this is used correctly.

In my opinion, even with assumptions about proper use of aria-invalid (which is not mandatory for compliancy), the risk to have many inapplicable implementations is high, unless we clearly state in assumptions that we assume every input with errors must have the aria-invalid="true" (or different from false) attribute.


What about replacing 
"This rule applies to any error field that is [visible][]."
with 
"This rule applies to any [visible][] input error with its context."

starting from the WCAG definition of input error

> input error
>
> information provided by the user that is not accepted
> 
> NOTE
> This includes:
> 1. Information that is required by the [Web page](https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html#dfn-web-page) but omitted by the user
> 2. Information that is provided by the user but that falls outside the required data format or values

Then we need to explain what context means, but maybe we can re-use the definition of [Form Field Error Indicator](https://act-rules.github.io/rules/36b590#form-field-error-indicator); anyway, in addition to the input element already covered by definition above, it should include:
* its label (and again we can re-use something from WCAG)
    > label
    > 
    > text or other component with a text alternative that is presented to a user to identify a component within Web content
    > 
    > NOTE
    > A label is presented to all users whereas the name may be hidden and only exposed by assistive technology. In many (but not all) cases the name and the label are the same.
    > 
    > NOTE
    > The term label is not limited to the label element in HTML.
* elements visually associated to the input (e.g. error messages).

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Applicability
+
+This rule applies to any [visible][] [input error][] together with its visual context.
+
+## Expectation
+
+Each test target, together with its visual context, identify the presence of errors not only through color differences.
+
+If the presence of errors is conveyed through the use of colors that differ not only in their hue, but that also have a difference in lightness and the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater, then this counts as an additional visual mean.
```
Jym77:
```suggestion
A difference in hue resulting in a contrast ratio of 3:1 or greater is not considered as difference in color only.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+
+It is possible for an error field to be distinguishable from adjacent fields but still not have a meaningful error message. In that case, it would pass Success Criterion 1.4.1: Use of Color but still fail [Success Criterion 3.3.1: Error Identification](https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html) and, if applicable, [Scuccess Criterion 3.3.3: Error Suggestion](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html).
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The first name `input` value is empty. Its error message (the `p` element) conveyed the presence of an error through its text.
```
Jym77:
```suggestion
The "first name" `input` value is empty. Its error message (the `p` element) conveys the presence of an error through its text.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+<label for="first_name">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" required style="border:1px solid #dd0000" aria-invalid="true" aria-errormessage="first_name_error_msg">
+<p id="first_name_error_msg" class="error_msg" style="color:#dd0000">This field is empty. Enter your first name.</p>
+<label for="last_name" style="color:#333333">Last Name</label>
+<input type="text" id="last_name" autocomplete="family-name" required style="border:1px solid #b0b0b0" value="Doe">
+```
+
+#### Passed Example 2
+
+The phone number `input` value doesn't match the `type` value. Its visual error indicator (the first couple of `label` and `input` elements) conveyed the presence of an error through its CSS properties:
```
Jym77:
```suggestion
The "phone number" `input` value doesn't match the `type` value. Its visual error indicator (the first couple of `label` and `input` elements) conveys the presence of an error through its CSS properties:
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+- font-weight: bold AND;
+- border-width: 3px.
```
Jym77:
```suggestion
- `font-weight: bold`; and
- `border-width: 3px`.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+<input type="tel" id="phone_number" autocomplete="tel" required style="border:3px solid #dd0000" aria-invalid="true" value="John Doe">
+<label for="email" style="color:#333333">Email</label>
+<input type="email" id="email" autocomplete="email" required style="border:1px solid #b0b0b0" value="john.doe@example.com">
+```
+
+#### Passed Example 3
+
+The email `input` value is missing the "@" symbol. Its visual error indicator (the first couple of `label` and `input` elements) conveyed the presence of an error through its CSS properties:
+- lightness (difference in relative luminance between the colors leads to a contrast ratio greater than 3:1) AND;
+- font-weight: bold.
```
Jym77:
```suggestion
The email `input` value is missing the "@" symbol. Its visual error indicator (the first couple of `label` and `input` elements) conveyed the presence of an error through its CSS properties:
- lightness (difference in relative luminance between the colors leads to a contrast ratio greater than 3:1); and
- `font-weight: bold`.
```
I think it would make sense to have the different contrast being the **only** difference. Otherwise, tools or methodologies that ignore it but accept `font-weight` would still be considered consistent.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+<input type="email" id="email" autocomplete="email" required style="border:1px solid #dd0000" aria-invalid="true" value="john.doeexample.com">
+<label for="address" style="color:#000">Address</label>
+<input type="text" id="address" autocomplete="address-line1" required style="border:1px solid #000" value="5th Example Street">
+```
+
+### Failed
+
+#### Failed Example 1
+
+The first name `input` value is empty. Its visual error indicator (the red border color of the first `input` element) conveyed the presence of an error through color (hue) differences only.
```
Jym77:
```suggestion
The "first name" `input` value is empty. Its visual error indicator (the red border color of the first `input` element) conveys the presence of an error through color (hue) differences only.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+```html
+<label for="first_name" style="color:#333333">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" required style="border:1px solid #dd0000" aria-invalid="true">
+<label for="last_name" style="color:#333333">Last Name</label>
+<input type="text" id="last_name" autocomplete="family-name" required style="border:1px solid #b0b0b0" value="Doe">
+```
+
+#### Failed Example 2
+
+The phone number `input` value doesn't match the `type` value. Its visual error indicator (the red color of the first `label` element) conveyed the presence of an error through color (hue) differences only.
```
Jym77:
```suggestion
The "phone number" `input` value doesn't match the `type` value. Its visual error indicator (the red color of the first `label` element) conveys the presence of an error through color (hue) differences only.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+```html
+<label for="phone_number" style="color:#dd0000">Phone number</label>
+<input type="tel" id="phone_number" autocomplete="tel" required style="border:1px solid #b0b0b0" aria-invalid="true" value="John Doe">
+<label for="email" style="color:#333333">Email</label>
+<input type="email" id="email" autocomplete="email" required style="border:1px solid #b0b0b0" value="john.doe@example.com">
+```
+
+#### Failed Example 3
+
+The email `input` value is missing the "@" symbol. Its visual error indicator (for the first couple of `label` and `input` elements, respectively the red text color and the red border color) conveyed the presence of an error through color (hue) differences only.
```
Jym77:
```suggestion
The email `input` value is missing the "@" symbol. Its visual error indicator (for the first couple of `label` and `input` elements, respectively the red text color and the red border color) conveys the presence of an error through color (hue) differences only.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+```html
+<label for="first_name" style="color:#000;">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" required style="border:1px solid #000" value="John">
+<label for="last_name" style="color:#000">Last Name</label>
+<input type="text" id="last_name" autocomplete="family-name" required style="border:1px solid #000" value="Doe">
+```
+
+#### Inapplicable Example 3
+
+Error field is not visible.
```
Jym77:
```suggestion
The error field is not visible.
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+```html
+<div style="display:none">
+<label for="phone_number" style="color:#dd0000">Phone number</label>
+<input type="tel" id="phone_number" autocomplete="tel" required style="border:1px solid #b0b0b0" aria-invalid="true" value="John Doe">
+<label for="email" style="color:#333333">Email</label>
+<input type="email" id="email" autocomplete="email" required style="border:1px solid #b0b0b0" value="john.doe@example.com">
+</div>
+```
+
+[input error]: https://www.w3.org/WAI/WCAG22/Understanding/error-identification.html#dfn-input-error 'Definition of input error from WCAG 2.1 success criterion 3.3.1 Error Identification'
```
Jym77:
```suggestion
[input error]: https://www.w3.org/TR/WCAG21/#dfn-input-error 'Definition of input error from WCAG 2.1 success criterion 3.3.1 Error Identification'
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,162 @@
...
+    -  Giacomo Petri
+---
+
+## Description
+
+This rule checks that errors are not identified only through color differences.
+
+## Applicability
+
+This rule applies to any [visible][] [input error][] together with its visual context.
```
Jym77: I'm not sure that "visual context" is objective enough for Applicability 🤔 
(see the discussions we have on the "link context" definition)

Jym77 (changes requested):
We need to define "visual context" (or find a way to do without it...)

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
giacomo-petri: I was trying to maintain best practices to the passed example. Do you think could be acceptable to remove `font-weight: bold` from the passed example description, but keep it in the passed example code as per new edits?

Jym77: We also need to exert the rule to its limit (and the tools). Since the rule considers that a single difference is enough, we should have test cases with that, even if they are not best practice. This is part of the [One thing at a time](https://act-rules.github.io/pages/design/test-cases/#one-thing-at-a-time) guideline in our design guide.
(actually, the case could even has a contrast ratio of exactly 3:1 to ensure that tools do accept that limit)
We can add to the background that while contrast alone is enough, it is a better practice to also use something else.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
giacomo-petri: @Jym77,

I've removed the generic "visual context" with a more detailed list of elements that should be included in the evaluation.
I'm not sure about the "closest error message that refers to it" is objective enough.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
giacomo-petri: font-weight removed and added a paragraph in background section.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Description
+
+This rule checks that errors are not identified only through color differences.
+
+## Applicability
+
+This rule applies to any [visible][] [input error][] together with:
+- its [programmatic label]; and
+- the closest error message that refers to it.
```
Jym77: Indeed, "closest" is ambiguous. It could notably be distance in the DOM tree, or physical distance on screen between the painted elements.

Jym77: I'm also not sure that "error message" is unambiguous.
For example, putting a red cross next to errored field and a green check mark next to good ones, is a red cross an "error message" (it's not really a message)? (or the red border of Passed Example 3)

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Expectation
+
+Each test target, together with its visual context, identify the presence of errors not only through color differences.
```
Jym77: We need to change "visual context" here also.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+    -  Giacomo Petri
+---
+
+## Description
+
+This rule checks that errors are not identified only through color differences.
+
+## Applicability
+
+This rule applies to any [visible][] [input error][] together with:
```
Jym77: Not fully convinced that "input error" is unambiguous enough.
> information provided by the user that is not accepted

Does that mean that the field in which it was inputed is put in a validation error state? Or just that some code checked it and rejected it, without changing the validation state? Actually, this refers to the information, not even to the form field, is it what you want the rule to target?

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+## Accessibility Support
+
+There are no major accessibility support issues known for this rule.
+
+## Background
+
+- [Understanding Success Criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)
+- [F81 - Failure of Success Criterion 1.4.1 due to identifying required or error fields using color differences only](https://www.w3.org/WAI/WCAG21/Techniques/failures/F81.html)
+
+It is possible for an error field to be distinguishable from adjacent fields but still not have a meaningful error message. In that case, it would pass Success Criterion 1.4.1: Use of Color but still fail [Success Criterion 3.3.1: Error Identification](https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html) and, if applicable, [Scuccess Criterion 3.3.3: Error Suggestion](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html).
```
Jym77:
```suggestion
It is possible for an error field to be distinguishable from adjacent fields but still not have a meaningful error message. In that case, it would pass Success Criterion 1.4.1: Use of Color but still fail [Success Criterion 3.3.1: Error Identification](https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html) and, if applicable, [Success Criterion 3.3.3: Error Suggestion](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html).
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+<label for="first_name">First Name</label>
+<input type="text" id="first_name" autocomplete="given-name" required style="border:1px solid #dd0000" aria-invalid="true" aria-errormessage="first_name_error_msg">
+<p id="first_name_error_msg" class="error_msg" style="color:#dd0000">This field is empty. Enter your first name.</p>
+<label for="last_name" style="color:#333333">Last Name</label>
+<input type="text" id="last_name" autocomplete="family-name" required style="border:1px solid #b0b0b0" value="Doe">
+```
+
+#### Passed Example 2
+
+The "phone number" `input` value doesn't match the `type` value. Its visual error indicator (the first couple of `label` and `input` elements) conveys the presence of an error through its CSS properties:
```
Jym77: This description uses "visual error indicator", which doesn't match the "error message" used in the Applicability. What does "visual error indicator" refers to?
(same on other descriptions)

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+<input type="email" id="email" autocomplete="email" required style="border:1px solid #dd0000" aria-invalid="true" value="john.doeexample.com">
+<label for="address" style="color:#000">Address</label>
+<input type="text" id="address" autocomplete="address-line1" required style="border:1px solid #000" value="5th Example Street">
+```
+
+### Failed
+
+#### Failed Example 1
+
+The "first name" `input` value is empty. Its visual error indicator (the red border color of the first `input` element) conveys the presence of an error through color (hue) differences only.
```
Jym77: Maybe add the contrast ratio with the other border color, to drive the nail in.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,166 @@
...
+<input type="email" id="email" autocomplete="email" required style="border:1px solid #dd0000" aria-invalid="true" value="john.doeexample.com">
+<label for="address" style="color:#333333">Address</label>
+<input type="text" id="address" autocomplete="address-line1" required style="border:1px solid #b0b0b0" value="5th Example Street">
+```
+
+### Inapplicable
+
+#### Inapplicable Example 1
+
+There are no error fields.
```
Jym77: These descriptions should talk about "input error", not "error field", in order to match the Applicability.

Jym77 (dismissed):

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,175 @@
...
+    -  Giacomo Petri
+---
+
+## Description
+
+This rule checks that errors are not identified only through color differences.
+
+## Applicability
+
+This rule applies to each [HTML element][] that is [visible][] and has one of the following [semantic roles][]:
```
Jym77, line 27:
```suggestion
This rule applies to any [HTML element][] that is [visible][] and has one of the following [semantic roles][]:
```

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,175 @@
...
+- switch or
+- textbox.
+
+## Expectation
+
+Each test target either has no [visible][] [form field error indicators][], or at least one of the [visible][] [form field error indicators][] identifies the presence of errors not only through color differences.
+
+## Assumptions
+
+When content is communicated using colors that vary not just in hue but also have a substantial difference in brightness, this constitutes an extra visual differentiation provided that the difference in the colors' relative luminance creates a contrast ratio of 3:1 or more.
```
Jym77, line 46: I think we somehow need to put in the Expectation the fact that "contrast of 3:1 or more" is **not** a color difference.
We probably need to keep an assumption that this is indeed enough, but we also want to say in the "normative" part that the rule accepts this.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
```diff
@@ -0,0 +1,175 @@
...
+- radio,
+- searchbox,
+- slider,
+- spinbutton,
+- switch or
+- textbox.
+
+## Expectation
+
+Each test target either has no [visible][] [form field error indicators][], or at least one of the [visible][] [form field error indicators][] identifies the presence of errors not only through color differences.
```
Jym77, line 42: Seeing a potential pitfall here (also when looking at examples):
This somewhat implicitly consider that we have something to compare the error indicator to, in order to decide whether it is color differences or more (i.e. "differences" with **what**?)

This could either be another field, not in error, in the same form, or the same field when not errored. 
The second solution somewhat requires a statefull rule and the tester to act (and put the field in error). This hopefully will be doable soonish (once we've solved #1953 / #2045), but that may still take time.
The first point could be doable by targeting pairs of field, passing if they are both errored or none errored, and comparing if only one is errored 🤔 

Put otherwise: if there is only one form field in the page, and it is in error, how do we check the rule?

Jym77 (changes requested):
I think we need to find a way to tell in the Expectation what we are comparing the errored field with.

_rules/error-field-is-not-identified-using-color-only-i3z4v8.md
giacomo-petri: > This could either be another field, not in error, in the same form

Comparing a pair of inputs, where one fails and the other does not, is not an optimal approach. Additionally, inputs may have different styles, such as checkboxes or text fields. Therefore, it is advisable to only compare inputs of the same type and with the same CSS.

> Put otherwise: if there is only one form field in the page, and it is in error, how do we check the rule?

Limiting the rule to a specific pair of inputs (assuming the previous concern is addressed) would, in my opinion, be too restrictive.

I think we should go with the second option. @Jym77, should we put this rule on hold till https://github.com/act-rules/act-rules.github.io/issues/1953 / https://github.com/act-rules/act-rules.github.io/issues/2045 are solved?

Jym77: I think it's a list worth bringing to attention that this is another rule proposal that would benefit from being statefull, so that we are sure the solution can handle it.

WilcoFiers:
Very out of date. Closing on behalf of @giacomo-petri

PR #1916, opened 2022-09-07
Adds a defintion for essential text change

tbostic32:
In [Text content that changes automatically can be paused, stopped or hidden](https://www.w3.org/WAI/standards-guidelines/act/rules/efbfc7/proposed/) it was determined that we need to have `essential` in the expectation. However, `essential` is an ambiguous term, so we need to create a new definition that handles the various cases where text changing is an essential function of a page. 

Need for Call for Review:

## How to Review And Approve


pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
...
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An _essential text change_ occurs when the text content of an HTML element is required to convey accurate information to the user. An _essential text change_ may appear in any of the following forms:
+
+1. Providing context - Text is changed to provide users with additional information required to understand the current state of the page. For example, while searching for results, text may change to convey the page is *loading* to provide context to the user.
+2. Live updates of important information - The content automatically updates to show the most recent information available from a data source and should not be ignored by the user.
```
tbostic32: This feels very awkward, namely specifying that the information must be important in order for this to qualify. I felt the need to do this since I wouldn't qualify the examples given in Understand 2.2.2 like weather or stock ticker information as important enough to be exempt from having controls. *However*, there could be information such as safety alerts/updates that I would consider to be exempt.

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
...
+key: essential-text-change
+unambiguous: true
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An _essential text change_ occurs when the text content of an HTML element is required to convey accurate information to the user. An _essential text change_ may appear in any of the following forms:
```
tbostic32: I find this to be awkward, but I also couldn't figure out a good way to say it without doing something like:
> An essential text change occurs when the changing of text on a page is essential.

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Essential Text Change
+key: essential-text-change
+unambiguous: true
+objective: true
```
WilcoFiers:
```suggestion
objective: false
```

pages/glossary/essential-text-change.md
WilcoFiers: Yeah, this one's difficult. Who decides what should and what should not be ignored? I think we should maybe just try to give a list of the types of things that shouldn't be ignored? Maybe, financial data, information on the status of a user session, that kind of thing.

WilcoFiers (dismissed):

tbostic32:
@carlosapaduarte From our review we found that the applicability is somewhat hard to understand, so I would like to propose the change below:

1. Remove the **no-child-changed** and **not-alone** applicability conditions
2. Modify the **change** condition to only consider the text content of child text nodes (only direct children, not all descendants)

At least my understanding was that the **no-child-changed** condition was meant to prevent an parent element from being applicable when a the text content of a child was changed. If we only consider the text content of child text nodes then the an ancestor node will never be applicable when one of its descendants text changes.

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
...
+objective: false
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An _essential text change_ occurs when the modification of the text content of an HTML element is necessary to convey accurate and timely information to the user. An _essential text change_ may appear in any of the following forms:
+
+1. Providing context - Text is changed to provide users with additional information required to understand the current state of the page. For example, while searching for results, text may change to convey the page is *loading* to provide context to the user.
```
Jym77:
```suggestion
1. Providing context - Text is changed to provide users with additional information required to understand the current state of the page. For example, while searching for results, text may change to convey that the page is *loading* to provide context to the user.
```

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
...
+2. Live updates of important information - The content automatically updates to show the most recent information available from a data source and should not be ignored by the user. For example, updates to a users financial information or browsing sessions status would constitute an `essential-text-change`.
```
Jym77: Maybe adding something like live news updates in case of danger (e.g. safety/evacuation instructions in case of natural disaster or terror attack).

Jym77: Should we add examples that are **not essential**?

Jym77:
```suggestion
2. Live updates of important information - The content automatically updates to show the most recent information available from a data source and should not be ignored by the user. For example, updates to a users financial information or browsing sessions status would constitute an Essential Text Change. 
```

_rules/auto-update-text-efbfc7.md
```diff
@@ -34,7 +34,7 @@ This rule applies to any [HTML element][] that has a [visible][] [text node][] a
 
 ## Expectation
 
-For each test target there is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][], to achieve at least one of the following objectives:
+For each test target that is not an [essential text change][] there is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][], to achieve at least one of the following objectives:
```
Jym77: We need to also say what happen for the essential ones.
```suggestion
For each test target, either it is an [essential text change][], or there is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][], to achieve at least one of the following objectives:
```

Jym77 (dismissed):

pages/glossary/essential-text-change.md
tbostic32: Agreed (and very relevant given the hurricanes hitting the US right now), I will add something to that effect.

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
...
+key: essential-text-change
+unambiguous: true
+objective: false
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An _essential text change_ occurs when the modification of the text content of an HTML element is necessary to convey accurate and timely information to the user. An _essential text change_ may appear in any of the following forms:
```
WilcoFiers: Would something like a clock or a count-down be considered essential. For example on a new years calendar, an auctioning site counting down when the auction closes, or a ticket site counting down to when tickets go on sale?

tbostic32: I think if it is a core functionality of the website then it would be considered essential. E.g., in a web page that is only a large clock counting down to new year then the text change of the clock would be essential.

HelenBurge (approved):

Jym77 (approved):

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,15 @@
...
+
+1. Providing context - Text is changed to provide users with additional information required to understand the current state of the page. For example, while searching for results, text may change to convey that the page is *loading* to provide context to the user.
+2. Live updates of important information - The content automatically updates to show the most recent information available from a data source and should not be ignored by the user. For example, updates about the users browsing sessions status or important safety information like evacuation instructions in the case of a natural disaster would constitute an Essential Text Change. Oppositely, information such as stock prices or advertisements are not important information and would not be considered an Essential Text Change.
```
WilcoFiers: This still feels fairly open-ended. Who decides that something is important, and that the user shouldn't ignore it? I think I'd like to see a list here of scenarios what we think are essential.

WilcoFiers (dismissed):

pages/glossary/essential-text-change.md
tbostic32: I would agree, though the fact that it isn't objective helps with that some. I could add a few more examples so our total list would be the examples below. Would that be sufficient?

Important information:
1. Browsing session status (from definition)
2. Safety information (from definition)
3. Changing text which represents a core functionality of the web page, such as the time readings at https://www.time.gov/ or a countdown timer for an auction site (I'm not sure if this one works, as pausing it you should still be able to get something from the site)
4. A change in text when using collaborative systems, such as a chat messenger, forum, or other collaboration software.
5. A text change triggered by user input, such as selecting an option in a text-based adventure game.

Non-Important Information:
1. Stock ticker information (from definition)
2. Advertisements (from definition)
3. Live sports scores
4. "Rolling" updates such as job postings, or news article titles and links
5. Countdown timer to an event that does lose opportunity for the user, such as a countdown to the start of a sports game.

WilcoFiers: I like that, yeah that sounds good to me.

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,28 @@
...
+    *Important*
+    - Information on the user's browsing session status
+    - Information pertaining to the safety of the user, such as a natural disaster alert
+    - A change in text when using collaborative systems, such as a chat messenger or forum
+    - A change in text triggered by user input, such as selecting an option in a text-based adventure game
+    - A change in text that represents a core functionality of the web page, such as a countdown timer for an auction site
+    
+    *Not Important*
+    - Text changes that do not require immediate user action, such as stock ticker information, sport scores, or advertisements
+    - "Rolling" updates containing informmation such as job postings or news article titles and links
```
tbostic32:
```suggestion
    - "Rolling" updates containing information such as job postings or news article titles and links
```

Jym77 (approved):

HelenBurge (approved):

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,28 @@
...
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An _essential text change_ occurs when the modification of the text content of an HTML element is necessary to convey accurate and timely information to the user. An _essential text change_ may appear in any of the following forms:
+
+1. Providing context - Text is changed to provide users with additional information required to understand the current state of the page. For example, while searching for results, text may change to convey that the page is *loading* to provide context to the user.
+2. Live updates of important information - The content automatically updates to show the most recent information available from a data source and should not be ignored by the user. Importance of information is often subjective and requires some amount of human understanding, so we have provided examples below to help illustrate when text should be considered important.
```
WilcoFiers: This is still open ended. What I was really hoping for here is a semi-definitive list of things that are essential. Anything that isn't on the list isn't essential. I tried to do something like that with the definition of [essential text presentation](https://github.com/act-rules/act-rules.github.io/pull/1969/files#diff-4a01a104eef4098b17b94e7fdf93e7720a0da132098bd10a77e8624e9a0081c3R2).

WilcoFiers (changes requested):

pages/glossary/essential-text-change.md
```diff
@@ -0,0 +1,28 @@
...
+key: essential-text-change
+unambiguous: true
+objective: false
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An _essential text change_ occurs when the modification of the text content of an HTML element is necessary to convey accurate and timely information to the user. An _essential text change_ may appear in any of the following forms:
```
carlosapaduarte:
```suggestion
An _essential text change_ occurs when the modification of the text content of an HTML element is necessary to convey accurate and timely information to the user. An _essential text change_ may appear for any of the following purposes:
```

pages/glossary/essential-text-change.md
carlosapaduarte: Agree that It is open-ended, but this is not an objective definition. Does it really have to be close-ended?

carlosapaduarte (changes requested):

_rules/auto-update-text-efbfc7.md
```diff
@@ -36,7 +36,7 @@ This rule applies to any [HTML element][] that has a [visible][] [text node][] a
 
 ## Expectation
 
-For each test target there is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][], to achieve at least one of the following objectives:
+For each test target, the text change is a [non essential text change][] and there is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][], to achieve at least one of the following objectives:
```
WilcoFiers: If I'm reading the logic right, aren't you now saying that essential text changes don't meet the expectation? This is going to be super awkward, so we need to find a better way to phrase this, but what this should be I think is:

```suggestion
For each test target, the text change is not a[non essential text change][] or there is at least one set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][], to achieve at least one of the following objectives:
```

WilcoFiers: We should have some examples for essential text changes passing the rule.

pages/glossary/non-essential-text-change.md
```diff
@@ -0,0 +1,21 @@
...
+A _non-essential text change_ occurs when the modification of the text content of an HTML element is not required for the purpose of providing context or live updated information to the user. A _non-essential text change_ usually occurs for one of the following purposes:
+
+1. **Action Not Required** - Text changes on the page that does not require the immediate action of the user.
+2. **No Context Included** - Text changes on the page that do not provide any additional information on the context of the page or provide additional help to the user.
+
+Examples of a _non-essential text change_ would include:
+
+    - Text changes that do not require immediate user action, such as stock ticker information, sport scores, or advertisements.
+    - "Rolling" updates containing information such as job postings or news article titles and links.
+    - A countdown timer to an event that does not cause a loss of opportunity for the user, such as a countdown to the start of a sporting event. An event like an auction, where timing is a key component, may not be considered a _non-essential text change_.
```
WilcoFiers: I still think you're trying to define what "essential" mean here, instead of taking advantage of what we can do with "non-essential", which is to just list out things we know definitely won't be essential. I took another stab at it, here's what I'm thinking might work:

```suggestion
A text change is considered non-essential if it occurs in any of the following scenarios:

- **Carousels**: The text change is part of a fixed set of slides that is updated as part of a carousel; or
- **Rolling Ads**: The text change occurs when an ad replaced with another ad on a timer; or
- **Article feed**: The text change is part of a feed in which new articles are added as they become available such as news articles or posts on a social media site. This does not include chat feeds; or
- **Live data**: The text change is an update of a live event or part of live information such as the scores of a sports match, an exchange rate, and weather measurements; or
- **Countdown timer**: A timer that counts down to an event that occurs at a predetermined time.

**Note**: Other scenarios in which a text change is not essential may exist, but are not covered by this definition. A text change is only "essential" when pausing, stopping or hiding it would fundamentally alter the meaning of information, or the functionality available on the page.

**Note**: While pausing a live feed or live data does change the functionality, it is not a fundamental change. An equivalent alternative for live information would be information along with a particular date and time, to clearly distinguish it from live data. Similarly, a countdown timer is equivalent to providing the exact time at which the event occurs.
```

WilcoFiers (changes requested):

_rules/auto-update-text-efbfc7.md
tbostic32: My goal was actually to exclude essential text changes from the expectation, but since we can't use the definition inside of the applicability I don't really have a way to do that. I was really trying to avoid the `not a non-essential text change` language since the double negative is somewhat confusing, but I'm not sure we can avoid it.

tbostic32: Coming back to this after way too long, I think this rule will actually be subject to the changes we are suggesting in the subjective applicability discussion. I think this would fall into the third category of "[Thing] is/has [subjective attribute]". In this case, filling in the blanks, it would be 'Text change is not non-essential'.

With this, we could move this into the applicability and not have to worry about having that test in the expectation which should really be there, and we wont need to include an example in the test cases.

WilcoFiers:
Waiting for subjective applicability / state discussions to resolve before picking this back up.

PR #1917, opened 2022-09-07
Update scrollable-element-keyboard-accessible-0ssw9k.md

kengdoj:
Changes to address survey comments:

1. Update inapp example 6 description (Tom's comments:)
2. Rename scrollable elements to non-embedded scrollable elements
3. Cross-link to the negative tabindex on iframe rule in the background
4. Clarify inapp example 1 description (Kathy's comment)

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -32,10 +32,10 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML element][] that has [visible][] [children][] in the [flat tree][] for which at least one of the following is true:
+This rule applies to any non-embedded [scrollable element][] that has [visible][] [children][] in the [flat tree][] for which at least one of the following is true:
```
WilcoFiers: I'm not sure there's actually anything wrong with the applicability. I think we just need to better explain the difference between a scrollable element and an element that shows a scrollable browsing context. What you're proposing here doesn't really work. Our definition of scrollable element means iframes are not considered scrollable elements. There are no "embedded scrollable elements", or at least, not in that sense.

The only addition I would make is this one, as it's not clear whether this "one of the following" applies to the element or to the children.

```suggestion
This rule applies to any [HTML element][] that has [visible][] [children][] in the [flat tree][], and for which at least one of the following is true:
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -160,7 +162,7 @@ This [horizontally scrollable][scrollable] `section` element is not included in
 
 #### Inapplicable Example 1
 
-This `section` element has a [computed][] [overflow][] of `visible`.
+This `section` element is not [scrollable][] because it has a [scroll distance][scrollable] of 0 in both directions.
```
WilcoFiers: Revert

```suggestion
This `section` element has a [computed][] [overflow][] of `visible`.
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -246,7 +248,7 @@ This `section` element has a [horizontal scroll distance][scrollable] that is le
 
 #### Inapplicable Example 6
 
-This `iframe` element is not a [scrollable element][scrollable].
+This `iframe` element is an embedded [scrollable element][scrollable].
```
WilcoFiers:
```suggestion
This `iframe` element is not a scrollable element, but instead contains a nested browsing context that is scrollable. These must be tested separately.
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -55,6 +55,8 @@ Some browsers restrict scrolling to the [content box](https://drafts.csswg.org/c
 
 To ensure there is some element from which arrow keys can be used to control the scroll position, focus must be on or in a scrollable region. If scripts are used to prevent the keyboard events from reaching the scrollable region, this could still cause a keyboard accessibility issue. This must be tested separately.
 
+This rule is only applicable to non-embedded scrollable elements. Embedded scrollable elements, such as iframe or object elements, are covered in other rules such as [Iframe with negative tabindex has no interactive elements](iframe-not-focusable-has-no-interactive-content-akn7bn.md).
```
WilcoFiers:
```suggestion
This rule only applies to elements scroll content in the same document. Elements such as iframes that embed other documents may also show be scrollable, but for them it is the embedded document that scrolls, not the content in the same document. Such scenarios are tested separately wit rules such as [Iframe with negative tabindex has no interactive elements](https://act-rules.github.io/rules/akn7bn).
```

WilcoFiers (changes requested):

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
JenniferChadwick: I am good with the changes.  Spelling correction:  "...separately wit**h** rules such as [**i**Frame with negative tabindex has no interactive elements]...

JenniferChadwick (approved):

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
Jym77: We can also update the list by saying "the element has…" rather than "it has…" which would make it more obvious, and also would better comply with [our own guidelines](https://act-rules.github.io/pages/design/rule-design/#listed-conditions).

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -32,10 +32,10 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML element][] that has [visible][] [children][] in the [flat tree][] for which at least one of the following is true:
+This rule applies to any non-embedded [scrollable element][] that has [visible][] [children][] in the [flat tree][] for which at least one of the following is true:
 
 - It has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right] of the element; or
```
Jym77:
```suggestion
- the element has a [horizontal scroll distance][scrollable] greater than the [computed][] [left][padding-left] or [right padding][padding-right] of the element; or
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -32,10 +32,10 @@ acknowledgments:
...
-- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom] of the element
+- It has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom] of the element.
```
Jym77:
```suggestion
- the element has a [vertical scroll distance][scrollable] greater than the [computed][] [top][padding-top] or [bottom padding][padding-bottom] of the element.
```

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
Jym77:
```suggestion
This rule only applies to elements who scroll content in the same document. Elements such as iframes that embed other documents may also be scrollable, but for them it is the embedded document that scrolls, not the content in the same document. Such scenarios are tested separately with rules such as [Iframe with negative tabindex has no interactive elements](https://act-rules.github.io/rules/akn7bn).
```
(correcting some typos on @WilcoFiers' suggestion)

Jym77 (dismissed):
Seconding Wilco's opinion.

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
kengdoj: A survey comment was to use terms from the Applicability to describe why this example is inapplicable. Is there a way to say "computed overflow of visible" using terms from Applicability?

WilcoFiers: Well... the descriptions of 1 and 2 are the same as you're suggesting them. I'm not exactly sure what you're looking for. The thing example 1 does is that the container isn't tall enough to contain its content, but because it has overflow: visible, the content doesn't become scrollable. It just renders outside the container. Inapplicable example 2 on the other hand, isn't scrollable because the content isn't long enough. Because of overflow:auto on example 2, if it had more content, it would be scrollable.

The current descriptions are saying that... maybe this would help?

```suggestion
This `section` element has a [computed][] [overflow][] of `visible`. Because of this its [scroll distance][scrollable] will be 0, regardless of the element's content.
```

And then for example 2 you could use:

```md
This `section` element has a [scroll distance][scrollable] of 0 in both directions because the content is smaller than the `section` element.
```

WilcoFiers (changes requested):

WilcoFiers (approved):

Jym77 (approved):

tbostic32 (approved):

schne324 (approved):

PR #1919, opened 2022-09-08
Add target size enhanced rule

WilcoFiers:

---

## How to Review And Approve


dd8:
The lack of a CSS hit testing spec is listed as an open issue:
https://github.com/w3c/csswg-drafts/issues/2325

Hit testing doesn't seem fully interoperable. - though things may have improved since this was written:
https://lists.w3.org/Archives/Public/www-style/2010Aug/0407.html

There's a W3 Wiki article here detailing the hit testing in IE:
https://www.w3.org/wiki/Hit_Testing

Here's the code for handling touch events in Chromium:
https://github.com/chromium/chromium/blob/be0d3937d6ccc1a4e0d27315d2e827151850d3fa/android_webview/renderer/aw_render_frame_ext.cc#L233
(so Chrome actually does hit testing on rects, rather than the circle used in WCAG 2.2 2.5.8)

**Edit:** elements with opacity:0 are clickable. You can try this out with the "back to top" link at bottom right of https://www.w3.org/WAI/ (it sets opacity:0 when scroll thumb is at top of document - but you can still click on it in Chrome/FF/Safari)

**Edit:** SVG strokes/fills/markers are only clickable if they have a color specified, so the borders of the following `rect` are clickable, but the middle isn't due to fill=none:
```
    <svg height="400px" width="400px" viewBox="-5 -5 30 30">
    <a href="https://example.com"><rect x="0" y="0" width="20" height="10" stroke="black" stroke-width="3" fill="none"></rect></a>
    </svg>
```
In SVG 2.0 this is controlled by the pointer-events property, but it's unclear if any browsers have implemented this:
https://www.w3.org/TR/SVG2/interact.html#PointerEventsProperty

Jym77:
@dd8 Been trying another direction with the rule. See also https://w3ccommunity.slack.com/archives/C01167KLA73/p1700745811381929 for details.
I'd appreciate if you could get a look at the current state and test cases. You have both the technical expertise and tool insight to figure out whether this is going anywhere.

dd8:
My email doesn't seem to have access to the Slack, but I've had a quick look at the changes and have a few comments. If somebody can give me access to the Slack channel I'll take a look at that as well.

I think excluding `area` elements is a good idea since they're often essential, because they're tied to the image associated with the image map (e.g.bio links on a group photo, places on a satellite photo or map). 

Given the lack of spec for clickable areas I would probably restrict the applicability to the simplest cases initially (then introduce more complex clickable areas in future versions of the rule). By simple I mean:

- HTML elements using the CSS box model (no SVG, canvas etc)
- exclude elements that have been transformed (there have been browser interop issues with CSS transforms and clickable areas for shapes like this https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)

The current applicability of 'widget' role *and* focusable looks wrong to me - it's a common a11y problem to have clickable widgets you can't tab to because they're not focusable.

The trickiest problem is knowing what's clickable. It's easy for built in widgets but not for DIVs with a 'click' event. Even if you can find which elements have a 'click' event registered there are some frameworks which put a generic 'click' event on *every element* which dispatches to handlers the author registers with the framework (e.g. the framework gives every `div` a `click` handler but the author only registers framework handlers for `div.button`, so only those trigger actions)

You may also get something like this where the clickable area is smaller than the widget role area:

```
<div role="button' style='padding:4em;'><div onclick='fireButton()'>Click me</div></div>
```

Jym77:
> My email doesn't seem to have access to the Slack, but I've had a quick look at the changes and have a few comments. If somebody can give me access to the Slack channel I'll take a look at that as well.

@dd8 It is the W3C Slack. See [the wiki page](https://www.w3.org/wiki/Slack) for joining.

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,731 @@
...
+Exception: not for `area` element (due to weird shapes + often essential size).
+Exception: not if a descendant is focusable (hard to define the clickable area).
+Exception: the target is a [shape-shifted element][] (weird shape).
+
+## Expectation
+
+For each test target, at least one of the following is true:
+
+- the target element has a [clickable area][] containing a rectangle aligned with the screen border whose width and height are at least 44 CSS pixels; or
+- though scrolling, the element can be brought into viewport with a [clickable area][] width and height of at least 44 CSS pixels; or
```
carlosapaduarte: These are not  applicable given the current definition of 'Can be target by a pointer event' because it requires that 'the element's [border box][] intersects the viewport'

pages/glossary/can-be-targeted-by-pointer-event.md
```diff
@@ -0,0 +1,37 @@
...
+  - CSS styling
+  - DOM tree
+---
+
+An element _can be targeted by pointer events_ when all the following conditions are true:
+
+- the element is a [semantic `widget`][semantic role]; and
+- the element is [focusable][]; and
+- the element's [border box][] intersects the viewport; and
+- the element's [border box][] is not entirely covered by the [border boxes][] of elements with greater `z-index` and `pointer-events: auto` CSS properties.
```
carlosapaduarte: Want to consider also open dialogs?
https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-dialog-open

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,731 @@
...
+Exception: the target is a [User Agent controlled component][] (WCAG exception).
+Exception: not for `area` element (due to weird shapes + often essential size).
+Exception: not if a descendant is focusable (hard to define the clickable area).
+Exception: the target is a [shape-shifted element][] (weird shape).
+
+## Expectation
+
+For each test target, at least one of the following is true:
+
+- the target element has a [clickable area][] containing a rectangle aligned with the screen border whose width and height are at least 44 CSS pixels; or
```
WilcoFiers: "Aligned with the screen border" feels off to me. Screens aren't necessarily rectangles. Some smart watches have circular screens for example. But this also seems like it would require some new input aspect of having to know the something about the edge of the screen?

_rules/target-size-enhanced-gi8qkf.md
WilcoFiers: This is a really neat idea! ❤️

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,731 @@
...
+- though scrolling, the element can be brought into viewport with a [clickable area][] width and height of at least 44 CSS pixels; or
+- the element has an empty [clickable area][], and its [clickable area][] cannot be made non-empty through scrolling; or
+  > Note: this is mostly for totally covered or off-screen elements. This should probably rather be handled in the Applicabilty and make these Inapplicable, but I'm not sure it can be done with the current definition…
```
WilcoFiers: What is an "empty" clickable area?

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,731 @@
...
+- the target is part of [inline text][]; or
+  > comment: I feel this is going to be difficult to define objectively, so it is better in Expectation for now. #1010 has some work in that direction that we can probably reuse: https://github.com/act-rules/act-rules.github.io/pull/1010/files#diff-32079a0602a5a909b242b4e0961e7c5ddd6b6f5c9906b216d5bf21cf2ba13a77R28-R29 > https://github.com/act-rules/act-rules.github.io/blob/4b64bba6cb77a8d4dc0649c83c55372f513d979f/pages/glossary/rendered-on-a-line.md
+- The size is [essential target size][]
+  > comment: this is always a bit tricky. I guess we can do as in #1916 and list cases that are considered essential (list can grow).
```
WilcoFiers: I would be all in favor of an assumption for this one. Do we know of real-world examples?

pages/glossary/can-be-targeted-by-pointer-event.md
Jym77: I'm not sure exactly what you mean 🤔 

Is it about elements "behind" the modal being blocked by it when it's opened (even if they are visible)?
I think that this is covered because in that case, their "clickable area" would be empty (the blocked elements are never topmost target, until the modal is closed and stops blocking them).

_rules/target-size-enhanced-gi8qkf.md
Jym77: "clickable area" are defined as sets of pixels. So, it is the empty set.

_rules/target-size-enhanced-gi8qkf.md
Jym77: I've added the definition.
This is mostly for pins on a map, and dots on a graph.

Jym77:
I've updated the rule with comments for review, and cleaned it a bit. It now starts to look like an actual rule 🎉

* Ultimately, I've ditched the "weird shape" exceptions. These are only really needed for the 2.5.8 rule (due to distance), and there was no real need to restrict this rule because of that. I made sure that the examples do not rely on the weird shape (i.e., no 50×50 button clipped to 40×40), so that automated tools who rely on border boxes or `getBoundingClientRect` can still claim a consistent implementation (and added a note to explain that).
* Also simplified a bit some of the definitions…

WilcoFiers:
Superseded by #2279

PR #1920, opened 2022-09-08
Add secondary requirements to rules

WilcoFiers:
This pull request makes it so that reporting AAA success criteria as failed on rules for AA becomes optional for implementors. I.e. implementors may fail them, but aren't required to.

Anticipating work on https://github.com/w3c/wcag-act/pull/531/, this pull request lists a number of success criteria as "secondary". These are criteria that fail when the rule fails, but don't necessarily pass when the rule passed. 

This PR is jumping the gun a little on this, but I think this is important to get done because it is causing problems for at least three of the implementors currently. Their implementations are reported as partially consistent for not failing a AAA success criterion, even though the implementor does not support AAA to begin with.

Closes issue(s): https://github.com/w3c/wcag-act-rules/issues/134

---

## How to Review And Approve

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ yarn.lock
 
 # Ignore JetBrains IDE file
 .idea
+
+# Tmp build directory
+wcag-act-rules-tmp/
--- a/_rules/image-no-text-0va7u6.md
+++ b/_rules/image-no-text-0va7u6.md
@@ -10,8 +10,9 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
-  wcag20:1.4.9: # Images of Text (No Exception) (AA)
+  wcag20:1.4.9: # Images of Text (No Exception) (AAA)
     forConformance: true
+    secondary: true
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
--- a/_rules/link-in-context-descriptive-5effbb.md
+++ b/_rules/link-in-context-descriptive-5effbb.md
@@ -5,13 +5,14 @@ rule_type: atomic
 description: |
   This rule checks that the accessible name of a link together with its context describes its purpose.
 accessibility_requirements:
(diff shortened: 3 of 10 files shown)


_rules/image-no-text-0va7u6.md
```diff
@@ -10,8 +10,9 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
-  wcag20:1.4.9: # Images of Text (No Exception) (AA)
+  wcag20:1.4.9: # Images of Text (No Exception) (AAA)
     forConformance: true
+    secondary: true
```
WilcoFiers, line 15: This name of "secondary" isn't final yet, and will probably change. This'll be easy to change though so I'm not going to worry about it for now.

.gitignore
```diff
@@ -69,3 +69,6 @@ yarn.lock
 
 # Ignore JetBrains IDE file
 .idea
+
+# Tmp build directory
+wcag-act-rules-tmp/
```
WilcoFiers, line 74: I accidentally pushed the tmp dir to the repo. Turns out that wasn't in the ignore. It is now.

Jym77 (commented):
That looks good. I have some caveat on the format change that I mentioned there.

---
Also, we had discussions around, e.g., the 1.3.1/4.1.2 mapping for "role attribute has valid value" (and some other rules whose mappings have change this year). Should we include them here?

It feels like they are more "Satisfy Mapping" in the lingo of https://github.com/w3c/wcag-act/pull/531/files#r954924761 (if the role attribute value exists, things are likely OK ("assuming no other problems") bit if the role is mispelled, we still need further testing to figure out if it does cause a problem). 

Not sure if we should handle both kind of secondary mappings in one go or two…

HelenBurge (approved):

tbostic32 (approved):

Jym77 (approved):

tombrunet (approved):

kengdoj (approved):

PR #1923, opened 2022-09-15
Text spacing rewrite

### LINKED ISSUE #1687: Text spacing not !important rules should check text is actually affected [78fd32, 24afc2, 9e45ec]

### LINKED ISSUE #1688: Text spacing not !important does not account for actual font-size [78fd32, 24afc2, 9e45ec]

Jym77:
With #1804, the text spacing rules becomes way too complex.
This simplifies them by looking at the text directly rather than at the `style` attribute. This automatically solves all the problems about inheritance, changed values, … by making sure that we only look at values that are actually used.

I feel this version is much simpler even than the current version.

- Failed testcases stay the exact same.
- Some Passed cases are now Inapplicable.
- Added 3 Passed/Inapplicable cases corresponding to the ones listed in #1687 and #1688.
- Added an assumption about the text using the corresponding property. This matches an exception in WCAG.
- Copied the assumption from "line height" rule, that `font-size` isn't changed through `::first-line` or other, to other rules.

Closes issue(s):

- closes #1687
- closes #1688 
- supersedes #1804 

Need for Call for Review:
This will require a 2 weeks Call for Review (new versions of the rules, including heavy changes and simplification)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/important-letter-spacing-wide-enough-fsy8c7.md
```diff
@@ -0,0 +1,301 @@
+---
+id: fsy8c7
```
WilcoFiers: While this is a significant update, I don't think the difference is significant enough that we should create them as new rules. I think I'd rather see them as updates to the existing rules.

_rules/important-letter-spacing-wide-enough-fsy8c7.md
```diff
@@ -0,0 +1,301 @@
...
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jean-Yves Moyen
```
WilcoFiers: Regardless of if this is a new rule or if this changes the existing rule, I think Jey should be listed as a previous author.

_rules/important-letter-spacing-wide-enough-fsy8c7.md
```diff
@@ -0,0 +1,301 @@
...
+---
+
+## Applicability
+
+This rule applies to any [HTML element][] for which all the following are true:
+
+- the element is has at least one [visible][] text node children; and
```
WilcoFiers:
```suggestion
- the element is has at least one [visible][] text node; and
```

_rules/important-letter-spacing-wide-enough-fsy8c7.md
```diff
@@ -0,0 +1,301 @@
...
+    inapplicable: further testing needed
+input_aspects:
+  - DOM Tree
+  - CSS Styling
+acknowledgments:
+  authors:
+    - Jean-Yves Moyen
+---
+
+## Applicability
```
WilcoFiers: I'm personally not a fan of taking these rules in this direction. As far as I've heard, this hasn't come from any false positives, so the simplest solution could be to just add an assumption. If we did want to fix this, I think the way to do it would be to adjust the expectation to be something like this:

> the [computed]() value of its [word-spacing]() property is at least 0.16 times the [computed]() value of the [font-size]() of the smallest [visible][] text node descendant; or

It's more intuitive (to me) for the rule to continue selecting the element with the style attribute. If you were to literally implement this rule, a single style attribute could generate a dozen issues because some paragraph with no text size changes could have a couple links in them.

WilcoFiers (dismissed):

Jym77:
Target text nodes rather than elements (+ note that CSS property is actually on parent).
Replace old rules rather than create new ones.

Jym77:
@WilcoFiers I've updated the `letter-spacing` rule. Can you look at it and see if it's OK? If so, I'll keep going with the other two.

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,40 +16,48 @@ input_aspects:
...
 
 ## Applicability
 
-This rule applies to any [HTML element][] that is [visible][] and for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+This rule applies to any [visible][] [text node][] child of an [HTML element][], when all the following are true for its `letter-spacing` property:
+
+- the [specified][] value is [declared][] in a `style` attribute; and
+- the [computed][] value is [important][].
+
+Where the `letter-spacing` property of a text node means the one of its parent.
```
WilcoFiers: I think you can move this to the background, rather than repeat this in the applicability and expectation. I think this is sufficiently unambiguous on its own.

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -59,31 +67,23 @@ CSS specifications define each declaration as being either [important][] (if it
 
 ## Test Cases
 
+Test cases description abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```
WilcoFiers: I'm not sure we can actually have text here. The act-tools script might not transfer this to the WAI site. Could you either have a look to make sure it does, or move this to the background?

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -219,48 +203,108 @@ There is no HTML element.
 
 #### Inapplicable Example 2
 
-This `p` element is not [visible][] because of `display: none`.
+There is no text node.
+
+```html
+<div style="letter-spacing: 0.1em !important; border-top: 1px solid black;">
+	<!-- empty div, border to make it "visible" -->
```
WilcoFiers: I don't think this border's really needed.

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -97,71 +97,55 @@ This `p` element has a [computed][] [letter-spacing][] of `3px`, which is **wide
...
-#### Passed Example 7
+#### Passed Example 5
 
-The [computed][] value of the `letter-spacing` property of this `p` element is **not [important][]**. The [computed][] value of the `letter-spacing` property of this `span` element is the [inherited][] value, that is the [computed][] value of its parent and therefore also **not [important][]**.
+This text node has a [computed][] `letter-spacing` of `2px`, 0.2 times its [computed][] `font-size` of `10px`
 
 ```html
-<p style="letter-spacing: 0.1em">
-	<span style="letter-spacing: inherit !important;">
+<div style="font-size: 16px; letter-spacing: 2px !important">
```
WilcoFiers, line 123: Don't think font-size is needed on this one.

WilcoFiers (dismissed):

_rules/important-letter-spacing-wide-enough-24afc2.md
Jym77: 🤔 I prefer keeping it rather than relying on default `font-size`, especially given that the example is about changing the `font-size` between the definition of `letter-spacing` and the text.
Otherwise, we might need a somewhat clumsy description to point out the rewrite (e.g. saying that `2px` is not big enough for the default `16px` but that is OK because the actual one is `10px`).

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,74 +16,74 @@ input_aspects:
...
+  previous_authors:
     - Jey Nandakumar
   funding:
     - WAI-Tools
 ---
 
 ## Applicability
 
-This rule applies to any [HTML element][] that is [visible][] and for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+This rule applies to any [visible][] [text node][] child of an [HTML element][], when all the following are true for the `letter-spacing` property of the element:
```
WilcoFiers: I still feel that targeting the element with the attribute is better here than to target individual text nodes. I think targeting the test is more complicated than targeting the element.

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,74 +16,74 @@ input_aspects:
...
+
+- the [specified][] value is [declared][] in a `style` attribute; and
+- the [computed][] value is [important][].
 
 ## Expectation
 
-For each test target, at least one of the following is true:
+For each test target, the [computed][] value of its `letter-spacing` property is at least 0.12 times the [computed][] value of its `font-size` property.
```
WilcoFiers, line 34: So... bad news. I think I found a new problem with this. Or maybe, I found that the old problem hasn't been solved. We were previously assuming that you could get the font-size from the element with the style attribute, we're now assuming you can get font-size from the parent of the text node. I don't think that's true though. Someone could use pseudo-elements like `::first-line` to adjust font-size, and this rule would not spot that.

WilcoFiers (dismissed):

_rules/important-letter-spacing-wide-enough-24afc2.md
Jym77: There is an assumption precisely to handle these cases.
In the current version, the assumption is here for the `line-height` rule, now copied in all three.

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,74 +16,72 @@ input_aspects:
...
 ---
 
 ## Applicability
 
-This rule applies to any [HTML element][] that is [visible][] and for which the `style` attribute [declares][declared] the [letter-spacing][] CSS property.
+This rule applies to any [HTML element][] with one or more [visible][] [text node][] child, when all the following are true for the `letter-spacing` property of the element:
 
-## Expectation
+- the [specified][] value is [declared][] in a `style` attribute; and
+- the [computed][] value is [important][].
```
carlosapaduarte, line 30: Is the computed value `important`?
Or is it the declaration of the `letter-spacing` property that is `important`?
The [`important` declaration](https://www.w3.org/TR/css-cascade-4/#important) influences how the computed value is obtained, but I'm not sure we can say the computed value is `important`....

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,74 +16,72 @@ input_aspects:
...
 
 - This rule assumes that WCAG's meaning for the "Letter spacing style property" is the value of the CSS `letter-spacing` property rather than the actual space between letters. The value of the CSS property is _added_ to whichever spacing already exist (for example, due to kerning). Thus, the actual space between letters can be larger than the value of the `letter-spacing` property. If [Success Criterion 1.4.12 Text Spacing][sc1412] is concerned by the actual space between letters, then this rule may fail (with the `letter-spacing` property being too small) while the Success Criterion is still satisfied (with the actual space being enough).
 
 - This rule assumes that when inter-letters space is changed because of justification, the `letter-spacing` property is not changed. Therefore, whether a text is justified or not doesn't change the result of this rule. Note that justifying text is a failure of [Success Criterion 1.4.8 Visual Presentation][sc148].
 
+- The target text node expresses something in a human language written in a script that that uses the `letter-spacing` property.
```
carlosapaduarte: I'm not sure I understand what you mean with this assumption... the text expresses something written in a script?

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,74 +16,72 @@ input_aspects:
...
 
-CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for this property, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
+CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for some properties, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```
carlosapaduarte:
```suggestion
CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for some properties, and that `!important` is wider known than this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```

_rules/important-letter-spacing-wide-enough-24afc2.md
```diff
@@ -16,74 +16,72 @@ input_aspects:
...
 
 ### Bibliography
 
 - [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
 - [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
 - [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
 
+### About test cases
+
+Test cases description abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```
carlosapaduarte:
```suggestion
Test cases descriptions abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```

_rules/important-line-height-wide-enough-78fd32.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Background
+
+Styles [declared][] in a `style` attribute have higher [cascade specificity][] than any selector; therefore, they "win" the [cascade sort] over any other style from [author origin][], i.e. it cannot be overridden by any of these. On the other hand, if such a style is [declared][] in a style sheet, it can still "lose" the [cascade sort][] to declarations with higher [specificity][] or simply coming from a later style sheet (such as ones injected by assistive technologies). This rule ensures that the element is not in the first case and that the style can be overridden by users, unless it is already at least the minimum required threshold. [Important][] styles that are declared with the [user][user origin] or [user agent][user agent origin] can win the [cascade sort][] over styles with the [author origin][].
+
+CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for some properties, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```
carlosapaduarte:
```suggestion
CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for some properties, and that `!important` is wider known than this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```

_rules/important-line-height-wide-enough-78fd32.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Bibliography
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+### About test cases
+
+Test cases description abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```
carlosapaduarte:
```suggestion
Test cases descriptions abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```

_rules/important-word-spacing-wide-enough-9e45ec.md
```diff
@@ -0,0 +1,307 @@
...
+
+## Background
+
+Styles [declared][] in a `style` attribute have higher [cascade specificity][] than any selector; therefore, they "win" the [cascade sort] over any other style from [author origin][], i.e. it cannot be overridden by any of these. On the other hand, if such a style is [declared][] in a style sheet, it can still "lose" the [cascade sort][] to declarations with higher [specificity][] or simply coming from a later style sheet (such as ones injected by assistive technologies). This rule ensures that the element is not in the first case and that the style can be overridden by users, unless it is already at least the minimum required threshold. [Important][] styles that are declared with the [user][user origin] or [user agent][user agent origin] can win the [cascade sort][] over styles with the [author origin][].
+
+CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for some properties, and that `!important` is wider known that this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```
carlosapaduarte:
```suggestion
CSS specifications define each declaration as being either [important][] (if it has the `!important` annotation) or [normal][]. Given that `normal` is also a keyword for some properties, and that `!important` is wider known than this distinction, this rule rather uses "[important][]"/"not [important][]" to avoid confusion.
```

_rules/important-word-spacing-wide-enough-9e45ec.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Bibliography
+
+- [Understanding Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)
+- [CSS Text Module Level 3 - Spacing](https://www.w3.org/TR/css-text-3/#spacing)
+- [CSS Visual formatting model details](https://drafts.csswg.org/css2/visudet.html)
+
+### About test cases
+
+Test cases description abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```
carlosapaduarte:
```suggestion
Test cases descriptions abusively refer to the CSS properties of text nodes, meaning the one of their parent.
```

carlosapaduarte (dismissed):

_rules/important-letter-spacing-wide-enough-24afc2.md
Jym77: Good point. We do need to somehow look at the computed value due to Inapplicable Example 8, where the `inherit !important` declaration is important but we don't care because the computed value is the inherited one, coming from a non-important declaration and thus can be overwritten.
I do agree that only declarations seems to be important 🙈 

Given that this wording is already present in [the current rule](https://www.w3.org/WAI/standards-guidelines/act/rules/24afc2/) (first bullet in Expectation + Passed Example 7), I suggest we treat that in another Issue/PR: https://github.com/act-rules/act-rules.github.io/issues/2002

_rules/important-letter-spacing-wide-enough-24afc2.md
Jym77: "(text expresses something) in a (human language written in a script that use letter spacing)".

For (as far as I know) written Arabic (and other languages), the letters are connected together and `letter-spacing` does not apply. ([MDN warning](https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing#internationalization_concerns))
This matches WCAG's exception:
> Exception: Human languages and scripts that do not make use of one or more of these text style properties in written text can conform using only the properties that exist for that combination of language and script.

So, essentially trying to say "if the target is text in Arabic, then `letter-spacing: 0 !important` is OK, but the rule cannot check that". (note that this is also a problem with the existing rules, so we can as well take this assumption out and deal with the problem in a separate PR…)

carlosapaduarte: Got it. It was the "script" that threw me off... I was in another context!

carlosapaduarte:
```suggestion
- The target text node expresses something in a human language written in a script that uses the `letter-spacing` property.
```

_rules/important-line-height-wide-enough-78fd32.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Assumptions
+
+- There is no mechanism available on the page to adjust `line-height`. If there is such a mechanism, it is possible to fail this rule while [Success Criterion 1.4.12 Text Spacing][sc1412] is still satisfied.
+
+- The font size is constant for all text in the element. If font-size changes (e.g., through use of the `::first-line` pseudo-element) then the required line height would also change throughout the element. This is untested by the current rule.
+
+- No other style attributes are used to increase or decrease the distance between lines of text. For example, style attributes such as `position`, `padding`, and `margin` could be used to increase the distance between lines of text to meet [Success Criterion 1.4.12 Text Spacing][sc1412]. Oppositely, those style attributes could also be used to reduce the distance between lines of text. Thus, it is possible to pass this rule, but still fail [Success Criterion 1.4.12 Text Spacing][sc1412] due to other styling choices.
+
+- The text in the element express something in a human language written in a script that that uses the `line-height` property.
```
carlosapaduarte:
```suggestion
- The text in the element express something in a human language written in a script that uses the `line-height` property.
```

_rules/important-word-spacing-wide-enough-9e45ec.md
```diff
@@ -0,0 +1,307 @@
...
+
+- This rule assumes that WCAG's meaning for the "Word spacing style property" is the value of the CSS `word-spacing` property rather than the actual space between words. The value of the CSS property is _added_ to whichever spacing already exist (for example, the size of the space character). Thus, the actual space between words is larger than the value of the `word-spacing` property. If [Success Criterion 1.4.12 Text Spacing][sc1412] is concerned by the actual space between words, then this rule may fail (with the `word-spacing` property being too small) while the Success Criterion is still satisfied (with the actual space being enough).
+
+- This rule assumes that when inter-words space is changed because of justification, the `word-spacing` property is not changed (the change occurs on the width of the space character between the words). Therefore, whether a text is justified or not doesn't change the result of this rule. Note that justifying text is a failure of [Success Criterion 1.4.8 Visual Presentation][sc148].
+
+- At least one text node child of the element express something in a human language written in a script that that uses the `word-spacing` property.
```
carlosapaduarte:
```suggestion
- At least one text node child of the element express something in a human language written in a script that uses the `word-spacing` property.
```

carlosapaduarte (dismissed):
Still a spelling issue

carlosapaduarte (approved):

colabottles (approved):

_rules/important-line-height-wide-enough-78fd32.md
```diff
@@ -0,0 +1,371 @@
...
+    - Jean-Yves Moyen
+  previous_authors:
+    - Jey Nandakumar
+  funding:
+    - WAI-Tools
+---
+
+## Applicability
+
+This rule applies to any [HTML element][] with one or more [visible][] [text node][] child, when the text inside the elements includes a [soft wrap break][] and all the following are true for the `line-height` property of the element:
```
tbostic32:
```suggestion
This rule applies to any [HTML element][] with one or more [visible][] [text node][] children, when the text inside the elements includes a [soft wrap break][] and all the following are true for the `line-height` property of the element:
```
Could also do `with one or more [visible][] child [text nodes][]`, but I think a plural needs to be in there somewhere. Goes for the important word spacing rule as well.

tbostic32 (approved):
Little editorial to the applicabilities, but otherwise I think these look good.

_rules/important-letter-spacing-wide-enough-24afc2.md
dd8: I've added a comment on https://github.com/act-rules/act-rules.github.io/issues/2002 on a way to deal with this

giacomo-petri (approved):

Jym77:
Call for reviews end on October 12.

Jym77:

PR #1926, opened 2022-09-26
Update heading is descriptive ACT-TF feedback.

colabottles:
- Added aria-level to Failed example 3.
- Passed example 3 was removed.
- The applicability and expectations were simplified.
- Empty headings are excluded from the applicability.
- Passed example 7 is now a failed example.

Closes issue(s):

N/A

Need for Call for Review:
This will require a two-week Call for Review. 

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/heading-descriptive-b49b2e.md
```diff
@@ -219,6 +190,17 @@ Heading marked up with `h1` element that does not describe the topic or purpose
...
 #### Failed Example 5
 
+Heading marked up with `h1` element that describes the topic or purpose of the following [palpable content][]. The heading is [visible][], but is not [included in the accessibility tree][].
+
+```html
+<html lang="en">
+	<h1 aria-hidden="true">Opening Hours</h1>
+	<p>We are open Monday through Friday from 10 to 16</p>
+</html>
+```
```
Jym77: This is now Inapplicable since the Applicability only considers heading included in the accessibility tree.

_rules/heading-descriptive-b49b2e.md
```diff
@@ -27,21 +27,11 @@ acknowledgments:
...
-- the target element is not [visible][]; or
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [visible][], and after the target element in tree order in the [flat tree][].
-
-## Expectation 2
-
-For each target element at least one of the following is true:
-
-- the target element is not [included in the accessibility tree][]; or
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [included in the accessibility tree][], and after the target element in tree order in the [flat tree][].
+Each target element describes the topic or purpose of the first non-[decorative][] [palpable content][], [included in the accessibility tree][], and after the target element in tree order in the [flat tree][].
```
Jym77: I think I did prefer the previous formulation
> Each target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [included in the accessibility tree][], and after the target element in tree order in the [flat tree][].

because it groups the conditions together and I feel the sentence flows better.
Not a blocker for me, however. Feel free to ignore.

_rules/heading-descriptive-b49b2e.md
```diff
@@ -91,17 +84,6 @@ Heading marked up with `role="heading"` that describes the topic or purpose of t
...
 #### Passed Example 3
 
-Heading marked up with `role="heading"` that describes the topic or purpose of the following [palpable content][], with a default aria-level assigned.
-
-```html
-<html lang="en">
-	<span role="heading">Opening Hours</span>
-	<p>We are open Monday through Friday from 10 to 16</p>
-</html>
-```
```
Jym77, line 101: I'm not sure why this example is removed. I feel it is still a valid one. Maybe adding an explicit `aria-level` to avoid the default one (as done in Failed Example 3).
Having the `role="heading"` element **also** visible makes it a bit different from Passed Example 5 where it is off-screen.

Jym77 (changes requested):

_rules/heading-descriptive-b49b2e.md
colabottles: @Jym77 This was discussed in the ACT-TF meeting and decided that the example should be removed. There was talk about perhaps making this example inapplicable as I look through the notes here.

It seemed to me that the example was not coherent with the second expectation since the heading was still in the accessibility tree even though it has no accessible name. My thought was an empty accessible name would be sufficient, but I'm open to discussion and curious to hear what others would say.

_rules/heading-descriptive-b49b2e.md
colabottles: @Jym77 @WilcoFiers and I had worked on this and had concluded that it made sense to combine the two expectations.

_rules/heading-descriptive-b49b2e.md
colabottles: I can move this to Inapplicable when I get a moment. Thanks @Jym77

_rules/heading-descriptive-b49b2e.md
WilcoFiers: Yeah, we tried to simplify, but I don't think it works. What if we take it the other direction and expand on the text:

```suggestion
Each target element describes the topic or purpose of the first node after the test target for which all the following are true:
- The node is not [decorative][]; and
- The node is [palpable content][] (such as text nodes or elements like `h1`, `p`, `img`. etc.); and
- The node is [included in the accessibility tree].

The order of elements is determined by the [flat tree][].
```

_rules/heading-descriptive-b49b2e.md
```diff
@@ -27,21 +27,11 @@ acknowledgments:
...
-
-- the target element is not [visible][]; or
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [visible][], and after the target element in tree order in the [flat tree][].
-
-## Expectation 2
-
-For each target element at least one of the following is true:
-
-- the target element is not [included in the accessibility tree][]; or
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [included in the accessibility tree][], and after the target element in tree order in the [flat tree][].
```
WilcoFiers, line 44: @Jym77 Your thought... palpable content as I read it could be something like an empty `p` element.... sort of arguable, an empty `p` element probably isn't decorative either, so wouldn't something like the following fail this rule, because the heading doesn't describe the first p element?

```html
<h1>ACT Rule</h1>
<p></p>
<p>ACT Rules are great</p>
```

_rules/heading-descriptive-b49b2e.md
WilcoFiers: The only difference between this one and passed example 2 is the omission of `aria-level`. The level isn't relevant to this rule, and since omitting `aria-level` is no longer allowed in ARIA 1.2, it didn't seem we should keep the example.

_rules/heading-descriptive-b49b2e.md
WilcoFiers: I would suggest if we're going to have an inapplicable example of a heading with aria-hidden, make it an empty heading, so that there isn't a 1.3.1 violation.

WilcoFiers (changes requested):

_rules/heading-descriptive-b49b2e.md
colabottles: @WilcoFiers Agreed. I will make this inapplicable with the empty heading.

colabottles: The commit message should have read **"Passed example 5"** not **"Passed example 3"** @WilcoFiers

_rules/heading-descriptive-b49b2e.md
colabottles: I'm good with this change if @Jym77 is, @WilcoFiers

_rules/heading-descriptive-b49b2e.md
```diff
@@ -55,8 +45,11 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 ## Background
 
+Visible headings that are not in the accessibility tree are a failure of [Success Criterion 1.3.1 Info and Relationships][sc131]. These are not tested by this rule but they can still fail [Success Criterion 2.4.6 Headings and Labels][sc246].
```
kengdoj, line 48: I think this is saying that visible headings that are not in the accessibility tree are not tested in this rule, but they are. Perhaps:

```suggestion
Visible headings that are not in the accessibility tree are a failure of [Success Criterion 1.3.1 Info and Relationships][sc131]. These can also fail [Success Criterion 2.4.6 Headings and Labels][sc246]. 
```

kengdoj: Maybe just switch the order of wording? 

```suggestion
Headings that are in the accessibility tree and not visible are a failure of [Success Criterion 1.3.1 Info and Relationships][sc131]. These are not tested by this rule but they can still fail [Success Criterion 2.4.6 Headings and Labels][sc246]. 
```

colabottles: Committed this suggestion.

_rules/heading-descriptive-b49b2e.md
colabottles: Committed this suggestion.

WilcoFiers (approved):

kengdoj (dismissed):
With visible removed from the Applicability, would inapplicable examples 3 and 4 be applicable?

Inapp Exs 2 and 5 still have visibility in their descriptions. Suggest removing to just say not included in the accessibility tree.

_rules/heading-descriptive-b49b2e.md
Jym77: Good question.
I think [Palpable content](https://html.spec.whatwg.org/multipage/dom.html#palpable-content) tries to circumvent that with these bits:
> As a general rule, elements whose content model allows any [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content-2) or [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) should have at least one node in its [contents](https://html.spec.whatwg.org/multipage/dom.html#concept-html-contents) that is palpable content and that does not have the [hidden](https://html.spec.whatwg.org/multipage/interaction.html#attr-hidden) attribute specified.

> Note: [Palpable content](https://html.spec.whatwg.org/multipage/dom.html#palpable-content-2) makes an element non-empty by providing either some descendant non-empty [text](https://html.spec.whatwg.org/multipage/dom.html#text-content), or else something users can hear ([audio](https://html.spec.whatwg.org/multipage/media.html#the-audio-element) elements) or view ([video](https://html.spec.whatwg.org/multipage/media.html#the-video-element), [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element), or [canvas](https://html.spec.whatwg.org/multipage/canvas.html#the-canvas-element) elements) or otherwise interact with (for example, interactive form controls).

> This requirement is not a hard requirement, however, as there are many cases where an element can be empty legitimately, for example when it is used as a placeholder which will later be filled in by a script, or when the element is part of a template and would on most pages be filled in but on some pages is not relevant.

I read the fist bit as "elements listed here that allow flow content should be non-empty", which sort of cover your point (the embedded content elements such as `img` being likely assumed to also be non-empty).

But then the last paragraph states that this is not always true.

---
Maybe we can use our [perceivable content][https://act-rules.github.io/glossary/#perceivable-content) instead, which sort of tries to handle that?

Or go at length of "describe at least one node between this heading and the next of same or lower level", but that feels tricky…

Or we can try and use "visible palpable content", but adding a border to your `p` would also break that…

_rules/heading-descriptive-b49b2e.md
```diff
@@ -27,21 +27,16 @@ acknowledgments:
...
-- the target element is not [visible][]; or
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [visible][], and after the target element in tree order in the [flat tree][].
-
-## Expectation 2
-
-For each target element at least one of the following is true:
-
-- the target element is not [included in the accessibility tree][]; or
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [included in the accessibility tree][], and after the target element in tree order in the [flat tree][].
+The order of elements is determined by the [flat tree][].
```
Jym77: I'd rather keep the list short and focused and move the precision below it:
```
Each target element describes the topic or purpose of the first node after the test target for which all the following are true:
- The node is not [decorative][]; and
- The node is [palpable content][]; and
- The node is [included in the accessibility tree].

The order of elements is determined by the [flat tree][].
Palpable content is text nodes or elements like `h1`, `p`, `img`, etc.
```
the reason being that we already have a link to the actual definition in the list item, and the precision then kind of breaks the flow of reading.
Not a strong feeling on that; I'm a technical person and have no problem reading further HTML specs…

_rules/heading-descriptive-b49b2e.md
```diff
@@ -55,8 +50,11 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 ## Background
 
+Headings that are in the accessibility tree and not visible are a failure of [Success Criterion 1.3.1 Info and Relationships][sc131]. These are not tested by this rule but they can still fail [Success Criterion 2.4.6 Headings and Labels][sc246].
```
Jym77: I don't think the first sentence is true. 1.3.1 state:
> Information, [structure](https://www.w3.org/TR/WCAG21/#dfn-structure), and [relationships](https://www.w3.org/TR/WCAG21/#dfn-relationships) conveyed through [presentation](https://www.w3.org/TR/WCAG21/#dfn-presentation) can be [programmatically determined](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) or are available in text.

Headings that are not visible do not convey anything through presentation and therefore cannot fail 1.3.1 🤔

Jym77 (dismissed):

_rules/heading-descriptive-b49b2e.md
giacomo-petri: Many times headings are visually implicit thanks to the design and layout of the page (and might be visually unpleasant), but might be very helpful for AT users. SR-only headings are not necessary failures of WCAG.

giacomo-petri (changes requested):
> This rule applies to any [semantic][semantic role] `heading` element that is [included in the accessibility tree][].

Per applicability, Failed Example 4 

> `<h1 aria-hidden="true">Weather</h1>`

should be inapplicable, as the heading is not exposed to the acc tree and is ignored by AT.

giacomo-petri (dismissed):
> This rule applies to any [semantic][semantic role] `heading` element that is [included in the accessibility tree][].

Per applicability, element visibility is no longer a requirement to test this rule; consequently inapplicable examples 2 3 4 5 should be reviewed.

_rules/heading-descriptive-b49b2e.md
WilcoFiers: @colabottles and I talked about this, we prefer the current one. As you don't seem to feel strongly on it, we're leaving this as is.

_rules/heading-descriptive-b49b2e.md
colabottles: @WilcoFiers worked on this today and the changes have been made.

_rules/heading-descriptive-b49b2e.md
colabottles: @WilcoFiers and I worked on this today and changes have been made.

colabottles:
Changes left to be made are to the examples to make sure they are correct.

Jym77 (approved):

_rules/heading-descriptive-b49b2e.md
```diff
@@ -195,11 +168,11 @@ Heading marked up with `role="heading"` that does not describe the topic or purp
 
 #### Failed Example 3
 
-Heading marked up with `role="heading"` that does not describe the topic or purpose of the following [palpable content][]. The heading is positioned off screen and is [included in the accessibility tree][].
+Heading marked up with `role="heading"` that does not describe the topic or purpose of the following [perceivable content][]. The heading is positioned off screen and is [included in the accessibility tree][].
```
giacomo-petri:
```suggestion
Heading marked up with `role="heading"` that does not describe the topic or purpose of the following [perceivable content][]. The heading is positioned off screen but is [included in the accessibility tree][].
```

Editorial change, just to be consistent with passed example 5 and focus the user attention on the idea that even if the element is off screen it is still available in the acc tree.

giacomo-petri (dismissed):

_rules/heading-descriptive-b49b2e.md
colabottles: Changes have been made to reflect @giacomo-petri comment for consistency.

colabottles:
@WilcoFiers I think this may be ready to be merged.

giacomo-petri:
Yesterday I was trying to apply this new rule to a website and I've found a scenario that fails but should not: a set of accordions.

The same example is present in ARIA-PRACTICES: https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion.html

The second accordion h3-button, as is on page load, is failing because the next perceivable content is another h3-button which is not described by the current h3-button.

The third accordion h3-button, as is on page load, is failing because the next perceivable content is a h2 which initiate a completely new context.

carlosapaduarte (changes requested):
@colabottles, what @giacomo-petri [found out](https://github.com/act-rules/act-rules.github.io/pull/1926#issuecomment-1399204681) is correct. The expectation needs to be adjusted for those instances where the first perceivable content is a heading at the same or lower level.

_rules/heading-descriptive-b49b2e.md
```diff
@@ -27,21 +27,13 @@ acknowledgments:
...
 ## Applicability
 
-This rule applies to any [semantic][semantic role] `heading` element that is either [visible][] or [included in the accessibility tree][].
+This rule applies to any [semantic][semantic role] `heading` element that is [included in the accessibility tree][] and has a non-empty (`””`) [accessible name][].
 
-## Expectation 1
+## Expectation
 
-For each target element at least one of the following is true:
+Each target element describes the topic or purpose of the first [perceivable content][] after the test target that is not [decorative][].
```
tbostic32, line 34:
```suggestion
Each target element describes the topic or purpose of the first [perceivable content][] after the test target that is not
- Only [decorative][] content
- A [semantic][semantic role] `heading` element that is [included in the accessibility tree][] and has a non-empty (`””`) [accessible name][]
```

tbostic32: @giacomo-petri @carlosapaduarte This is more or less the fix you were requesting. However, what would happen if there was a sentence after the third accordion, but before the next heading? Then the heading would go back to failing and I'm not sure that would seem correct either.

carlosapaduarte: Agreed @tbostic32 The problem remains. Shall we add an assumption that the content the heading is describing is not invisible and hidden from the AT? Otherwise, I believe we will end with with a fairly complex rule...

_rules/heading-descriptive-b49b2e.md
```diff
@@ -27,36 +27,33 @@ acknowledgments:
...
-- the target element describes the topic or purpose of the first [palpable content][] which is non-[decorative][], [included in the accessibility tree][], and after the target element in tree order in the [flat tree][].
+The order of elements is determined by the [flat tree][].
 
 **Note:** Headings do not need to be lengthy. A word, or even a single character, may suffice.
 
 ## Assumptions
 
 This rule assumes that the [flat tree][] order is close to the reading order, as elements are rendered on the page. Due to positioning, it is possible to render a document in an order that greatly differs from the tree order, in which case the content which is visually associated with a heading might not be the content following it in tree order and this rule might fail while [Success Criterion 2.4.6 Headings and Label][sc246] is still satisfied.
 
+This rule also assumes that the content the heading is intended to describe is visible and not hidden from AT. Otherwise, cases such as expandable content using a heading might fail this rule while [Success Criterion 2.4.6 Headings and Label][sc246] is still satisfied.
```
tbostic32: @colabottles @carlosapaduarte @giacomo-petri Here is the new assumption discussing the issue we have been wrestling with. From a conversation with Wilco, we would like to get this through, if there are additional edge cases to handle we would like to do so in a different PR.

tbostic32:
```suggestion
This rule also assumes that the content the heading is intended to describe is [visible][] and not hidden from AT. Otherwise, cases such as expandable content using a heading might fail this rule while [Success Criterion 2.4.6 Headings and Label][sc246] is still satisfied. 
```

colabottles: @tbostic32 I've made the change you suggested. Thank you.

colabottles (commented):
I have made the latest suggested changes to the applicability section as requested.

carlosapaduarte (approved):

tbostic32 (approved):

Jym77:
Call for review ended… merging.

PR #1931, opened 2022-10-04
Update object rule  decorative assumption

WilcoFiers:
Based on AGWG feedback from AWK, I went back to look at decorative objects. In testing I discovered that VoiceOver on Safari will happily announce object elements with no text alternative as "Image". So at least in Safari, any decorative object without role=none will be announced, and so fails the SC. In which case I don't think we need that assumption.

---

## How to Review And Approve


_rules/object-has-acessible-name-8fc3b6.md
```diff
@@ -38,17 +38,19 @@ Each target element has an [accessible name][] that is not empty (`""`).
 
 ## Assumptions
 
-The `object` element is not rendered for presentational purposes. If the `object` is decorative and not [marked as decorative][] then the rule might fail but the success criterion might still be satisfied.
+_There are currently no assumptions_
 
 ## Accessibility Support
 
+Some screen readers announce `object` elements even if they do not have an accessible name, while other skip the element. If an `object` is used to render decorative content, to ensure it is ignored by all major screen readers a presentational role is necessary.
```
WilcoFiers:
```suggestion
Some screen readers announce `object` elements even if they do not have an accessible name, while other skip the element. If an `object` is used to render decorative content, to ensure it is [marked as decorative][] and can be ignored by all major screen readers a presentational role is necessary.
```

colabottles (approved):

Jym77 (approved):

carlosapaduarte (approved):

PR #1933, opened 2022-10-04
aria-hidden not focusable, add hidden svg in link example

WilcoFiers:
Request from AGWG to add an example showing aria-hidden used on an SVG that is inside a link.

---

## How to Review And Approve


colabottles (approved):

_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -131,6 +139,19 @@ This `button` element is [focusable][], but not part of the [sequential focus na
 </div>
 ```
 
+#### Passed Example 6
+
+This `svg` element with `aria-hidden` is a descendant of a focusable link, rather than an ancestor.
```
Jym77:
```suggestion
This `svg` element with `aria-hidden` does not have a focusable descendant, focusable ancestors are not a problem for this rule.
```
I usually prefer description that follow the wording of expectation.

Jym77 (dismissed):

Jym77 (approved):

carlosapaduarte (approved):

PR #1950, opened 2022-10-07
Update ACT Rules to use WAI-ARIA 1.2

### LINKED ISSUE #1867: Upgrade to WAI-ARIA 1.2
- [x] Update WAI-ARIA specifications definition to use WAI-ARIA 1.2 @WilcoFiers 
- [ ] Accessibility Support for Explicit Semantic Role mentions ARIA in HTML is a working draft, it is no longer one @schne324 
- [ ] Update acc requirements mapping to go to ARIA 1.2 @WilcoFiers 
- [ ] https://act-rules.github.io/rules/6a7281#applicability needs updating, has ARIA 1.1 things @i01000001 
- [x] [Element with role attribute has required states and properties](https://act-rules.github.io/rules/4e8ab6) Update expectation, remove note on default values, and remove exception for default values @WilcoFiers 
- [ ] Update references to HTML in ARIA, indicating it is no longer a draft @schne324

WilcoFiers:
This pull request makes various changes to update all ACT rules to use WAI-ARIA version 1.2.

Closes #1867 

---

## How to Review And Approve


_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -268,7 +268,7 @@ Element has ARIA role, but no ARIA states or properties
 
 `aria-checked` state with empty value
 
-**Note**: The HTML validator flags an `aria-checked` attribute with an empty value as an issue. However WAI-ARIA 1.1 indicates `aria-checked` has a default value of `undefined`.
+**Note**: The HTML validator flags an `aria-checked` attribute with an empty value as an issue. However, since WAI-ARIA 1.1, `aria-checked` has a default value of `undefined`.
```
HelenBurge, line 271: Should this reference be updated to 1.2?

HelenBurge (changes requested):
As per my comment

_rules/aria-state-or-property-valid-value-6a7281.md
carlosapaduarte: I don't think so. This is true since 1.1 and it's still true in 1.2, therefore "since 1.1" is correct, right?

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -214,29 +196,30 @@ This element with the `list` role owns an element with the `tab` role through th
...
+			<span role="treeitem">Item 2</span>
+		</div>
 	</div>
 </div>
 ```
 
 #### Failed Example 7
 
-This element with the `menu` role only owns an element with a `group` role. The `group` in turn owns an element with the `menuitem` role, and an element with the `group` role, in which each element has the `treeitem` role. ARIA `group` roles are allowed to own other elements with a `group` role, but those nested `group` nodes must still meet the requirements.
+This element with the `list` role owns an element with the `listitem` role and an element with the `group` role, in which each element has the `listitem` role. The `group` role is no longer [required owned element][] for the `list` role.
```
tombrunet:
```suggestion
This element with the `list` role owns an element with the `listitem` role and an element with the `group` role, in which each element has the `listitem` role. The `group` role is no longer a [required owned element][] for the `list` role.
```

_rules/form-field-label-descriptive-cc0f0a.md
```diff
@@ -78,9 +65,21 @@ Each test target, together with its [visual context][], describes the purpose of
 
 ## Background
 
+The list of applicable [semantic roles][semantic role] is derived by taking all the [ARIA 1.2][aria12] roles that:
+
+- inherit from the `input`, `menuitem` or `select` role, and
+- are form field controls (this notably excludes `menu`, `option` or `tree`).
+-
```
tombrunet: Is this a dangling -?

tombrunet (approved):

_rules/aria-state-or-property-valid-value-6a7281.md
HelenBurge: I guess :)

HelenBurge (approved):

_rules/form-field-label-descriptive-cc0f0a.md
WilcoFiers:
```suggestion
```

WilcoFiers: Thanks, updated.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -160,7 +142,7 @@ This element with the `menu` role only owns an element with a `group` role. The
 
 #### Failed Example 1
 
-This element with the `list` role owns an element without any of its [required owned elements][] (`listitem` or `group` containing `listitem`).
+This element with the `list` role owns an element without any of its `listitem` [required owned elements][].
```
Jym77:
```suggestion
This element with the `list` role owns an element which is not a `listitem` [required owned elements][].
```

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -34,14 +34,18 @@ acknowledgments:
...
 
 For each test target, one of the following is true:
 
 - **global**: the test target is a [global state or property][global]; or
 - **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
+## Expectation 2
+
+Each test target is not [prohibited][] on the [semantic role][] of the element on which it is specified.
```
Jym77: I find the negation inside the quantification a bit weird and feel it would flow better with
```suggestion
No test target is [prohibited][] on the [semantic role][] of the element on which it is specified.
```
but we may want to stick with the "each test target" formulation (even though there are a ew rules where it's tweaked a bit). So no strong opinion here.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -115,7 +203,7 @@ This `div` does not have a [semantic role](#semantic-role).
 
 #### Inapplicable Example 2
 
-This `checkbox` has an [implicit semantic role](#implicit-role) that is identical to the [explicit semantic role](#explicit-role).
+This `checkbox` has an [implicit semantic role](#implicit-role) that is identical to the [explicit semantic role](#explicit-role). This allows native HTML `checked` state to apply.
```
Jym77: Technically, that should probably be the "`checked` content attribute" (state/property are an ARIA notion). Or maybe even the [checkedness](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-checked) (which is the same as the `checked` IDL attribute).

Technicalities aside, I'd still prefer
```suggestion
This `checkbox` has an [implicit semantic role](#implicit-role) that is identical to the [explicit semantic role](#explicit-role). This allows native HTML `checked` attribute to apply.
```

Jym77 (dismissed):

Jym77 (approved):

PR #1958, opened 2022-10-17
meta refresh (no exceptions) examples should not fail SC 2.2.1

WilcoFiers:
/me Chair hat off, product owner hat on

I'm having difficulties mapping to the Meta element refresh (no exceptions) ACT rule. Axe-core has two rules on meta elements. One that tests for failures of SC 2.2.1, and one that tests for failures of SC 2.2.4 that are not failures of SC 2.2.1. I.e. two axe-core rules combined cover all of this single ACT rule.

The problem is that this "no exceptions" ACT rule has examples that also fail SC 2.2.1. When axe-core reports those as failures of both SC 2.2.1 and 2.2.4, that implementation ends up as partially consistent, because SC 2.2.1 isn't listed in the accessibility requirements.

I think there are two solutions for it:
1. Ensure the failed examples don't fail SC 2.2.1; or
2. Add SC 2.2.1 as a secondary requirement

I noticed we don't have this problem on color contrast (enhanced) because all its failed examples pass level AA. As a quick solve, I think this will work.

We'll need a broader conversation about which option we'll want for scenarios like this, as part of discussing secondary requirements. My guess is we'll end up with option 2, but I don't think we're ready to commit to that just yet, and I since this is a priority issue, I don't think we should wait.

---

## How to Review And Approve

### DIFF
--- a/_rules/meta-refresh-no-delay-no-exception-bisz58.md
+++ b/_rules/meta-refresh-no-delay-no-exception-bisz58.md
@@ -129,16 +129,6 @@ The first `meta` element is not valid (because of the colon instead of a semi-co
 </head>
 ```
 
-#### Failed Example 4
-
-This `meta` element redirects the user after 20 hours.
-
-```html
-<head>
-	<meta http-equiv="refresh" content="72001; http://example.com" />
-</head>
-```
-
 ### Inapplicable
 
 #### Inapplicable Example 1
--- a/_rules/meta-refresh-no-delay-no-exception-bisz58.md
+++ b/_rules/meta-refresh-no-delay-no-exception-bisz58.md
@@ -92,7 +92,7 @@ The first valid `meta` element redirects immediately.
 ```html
 <head>
 	<meta http-equiv="refresh" content="0; https://w3.org" />
-	<meta http-equiv="refresh" content="5; https://w3.org" />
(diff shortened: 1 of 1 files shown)


carlosapaduarte:
@WilcoFiers Wouldn't a solution similar to what was done in #1920 work for this rule also?

WilcoFiers:
@carlosapaduarte It's not the same. On PR 1920 I set some requirements as secondary. Here the problem is that the SC that is getting failed isn't listed as an accessibility requirement. It's not allowed to be by the rules format, since SC 2.2.1 isn't always failed when this rule is failed.

This is the only way I think this can be solved under ACT Rules Format 1.0. Once we have an editors draft of 1.1, we'll have other options, but since this problem is hindering an implementation, and there is a way around the issue, I don't think we should wait.

HelenBurge (approved):

Jym77 (dismissed):
Can you add a background note that refresh below 20h are also bad but not in the examples due to mappings?
Maybe also adding the other rule in the bibliography.

WilcoFiers:
@Jym77 That feels inconsistent with what we did for Text has enhanced contrast. I think I'd rather keep them consistent TBH.

Jym77 (approved):

colabottles (approved):

tombrunet (approved):

PR #1963, opened 2022-10-25
Add GitHub script to set rules to approved

WilcoFiers:
This script sets a rule as approved on the WAI website. It  can be triggered from the "Actions" tab, but only by one of the ACT TF leads. When activating the action, it'll ask for a rule ID and branch.

This cannot be used to update a rule that's already approved. I'll need to create a separate action for that.

### DIFF
--- a/.github/scripts/approve-rule.mjs
+++ b/.github/scripts/approve-rule.mjs
@@ -1,0 +1,82 @@
+#!/usr/bin/env zx
+import 'zx/globals';
+import assert from 'assert';
+import moment from 'moment';
+
+import {
+  config,
+  cloneWcagActRules,
+  createOrCheckoutBranch,
+  commitAndPush
+} from './commons.mjs';
+
+const w3cDataFormat = 'D MMMM YYYY';
+const isoDateFormat = 'YYYY-MM-DD';
+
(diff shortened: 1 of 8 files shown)


.github/scripts/approve-rule.mjs
```diff
@@ -0,0 +1,82 @@
...
+
+if (!argv['skip-clone']) {
+  await cloneWcagActRules(config);
+}
+
+await createOrCheckoutBranch(config, argv.branch);
+await generateApprovedRulePages(config, argv.ruleId);
+await updateRuleVersionsYaml(config, argv.ruleId);
+await approveTestCaseJson(config, argv.ruleId);
+await commitAndPush(config, `Set ${argv.ruleId} to approved`);
```
WilcoFiers, line 28: This script basically automated the workflow documented here:

https://github.com/w3c/wcag-act-rules#adding-new-approved-rules

tbostic32 (approved):
Only comment would be adding some documentation on running it manually (basically the last line from your github workflow) and then maybe even a tidbit about triggering it from the git workflow.  Otherwise the code looks solid and right in line with expectations from the act-tools workflow you linked to.

carlosapaduarte (approved):

PR #1971, opened 2022-11-16
Table headers - No ARIA tables  and TF feedback

### LINKED ISSUE #1517: "Table header cell has assigned data cells" (d0f69e): test tables and grids separately?

### LINKED ISSUE #1545: Should all table header cells have assigned data cells?

ajanec01:
A few of changes to [Table header cell has assigned cells](https://act-rules.github.io/rules/d0f69e). 

The only reliably way of determining how table headers are assigned is in the HTML specification. It is unclear how ARIA table headers are assigned to cells so they are now omitted from the rule.

The update includes changes made in response to [the 1st Task Force feedback](https://www.w3.org/2002/09/wbs/93339/ACTTableHeaders/results) and [the 2nd Task Force feedback](https://docs.google.com/spreadsheets/d/1OSkPFocXk4K3zYLnwS78WLsWO4PvE5yRcsauyefuIUI/edit#gid=1850701681).

Closes #1545 and closes #1517 

Call for review- 2 weeks

---

## How to Review And Approve

### DIFF
--- a/_rules/table-header-cell-has-assigned-cells-d0f69e.md
+++ b/_rules/table-header-cell-has-assigned-cells-d0f69e.md
@@ -26,12 +26,12 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any non-empty `th` element that is [included in the accessibility tree][], has a [semantic role][] of [rowheader][] or [columnheader][], and for which all of the following are true:
+This rule applies to any [non-empty][] `th` element that is [included in the accessibility tree][], has a [semantic role][] of [rowheader][] or [columnheader][], and for which all of the following are true:
 
 - **in a table** the `th` element is a [descendant][] of a `table` element for which all of the following are true:
   - **visible** the `table` element is [visible][]; and
   - **minimum rows** the `table` element's [`rows` collection](https://html.spec.whatwg.org/multipage/tables.html#dom-table-rows) has 2 or more `tr` elements; and
-  - **non-empty data cell** the `table` element has at least one non-empty `td` element that is a [child][] of a `tr` element with the [`rowIndex`][rowindex] value of 1 or more if the `tr` element with the value of 0 has at least two `th` [children][child] or the value of 0 or more if the `tr` element with the value of 0 has one or less `th` [children][child].
+  - **non-empty data cell** the `table` element has at least one [non-empty][] `td` element that is a [child][] of a `tr` element with the [`rowIndex`][rowindex] value of 1 or more if the `tr` element with the value of 0 has at least two `th` [children][child] or the value of 0 or more if the `tr` element with the value of 0 has one or less `th` [children][child].
 - **no explicit role** none of the elements that make up the [table model][] have an [explicit semantic role][].
 
 ## Expectation
@@ -102,7 +102,7 @@ Each of the 2 `th` elements is assigned to the `td` element because this `td` el
 
 #### Passed Example 3
 
-Each of the 4 non-empty `th` elements is assigned to a `td` element, within the same `table` element.
+Each of the 4 [non-empty][] `th` elements is assigned to a `td` element, within the same `table` element.
 
 ```html
 <table>
@@ -371,7 +371,7 @@ This `table` element is not [visible][].
 
 #### Inapplicable Example 11
 
-This `table` element does not have a non-empty `td` element.
+This `table` element does not have a [non-empty][] `td` element.
 
 ```html
 <table>
@@ -388,7 +388,7 @@ This `table` element does not have a non-empty `td` element.
 
 #### Inapplicable Example 12
 
-There are no non-empty `th` elements in the `table` element.
+There are no [non-empty][] `th` elements in the `table` element.
 
 ```html
 <table>
@@ -404,6 +404,7 @@ There are no non-empty `th` elements in the `table` element.
 ```
 
 [attribute value]: #attribute-value 'Definition of attribute value'
+[non-empty]: #non-empty 'Definition of non-empty'
 [child]: https://dom.spec.whatwg.org/#concept-tree-child 'Definition of child'
 [explicit semantic role]: #explicit-role 'Definition of explicit semantic role'
 [rowindex]: https://html.spec.whatwg.org/multipage/tables.html#dom-tr-rowindex 'Definition of rowIndex'
--- a/_rules/table-header-cell-has-assigned-cells-d0f69e.md
+++ b/_rules/table-header-cell-has-assigned-cells-d0f69e.md
@@ -1,9 +1,9 @@
 ---
 id: d0f69e
-name: Table header cell has assigned cells
+name: Table header cell is assigned to a cell
 rule_type: atomic
 description: |
-  This rule checks that each table header has assigned cells in a table element.
+  This rule checks that each table header is assigned to at least one cell.
 accessibility_requirements:
   wcag20:1.3.1: # Info and Relationships (A)
     forConformance: true
@@ -24,16 +24,17 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML element][] with a [semantic][semantic role] [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any `th` element that is [included in the accessibility tree][], has a [semantic role][] of [rowheader][] or [columnheader][], and for which all of the following are true:
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][] is [included in the accessibility tree][].
+- **in a row** the element is a child of a `tr` element; and
+- **table** the element is an [inclusive descendant][] of a `table` element; and
+- **minimum rows** the `table` element has at least two `tr` elements as its [inclusive descendant][] elements; and
+- **visible** the `table` element is [visible][]; and
+- **no explicit role** the `th`, `tr`, `table`, and another elements in between them do not have an [explicit semantic role][].
 
 ## Expectation
 
-Each target element is [assigned][] to at least one element with an [inheriting semantic][] [cell][].
+Each target element is [assigned][] to at least one `td` or `th` element.
 
 ## Assumptions
 
@@ -341,18 +342,15 @@ This `th` element is part of a table which is not [included in the accessibility
 </table>
 ```
 
+[explicit semantic role]: #explicit-role 'Definition of explicit semantic role'
 [semantic role]: #semantic-role 'Definition of semantic role'
 [visible]: #visible 'Definition of visible'
-[inheriting semantic]: #inheriting-semantic 'Definition of Inheriting Semantic Role'
 [included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of included in the accessibility tree'
+[inclusive descendant]: https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant 'Definition of inclusive descendant'
 [assigned]: https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics 'Forming relationships between data cells and header cells'
-[cell]: https://www.w3.org/TR/wai-aria-1.1/#cell 'ARIA cell role'
 [flat tree]: https://drafts.csswg.org/css-scoping/#flat-tree 'Definition of flat tree'
-[table]: https://www.w3.org/TR/wai-aria-1.1/#table 'ARIA table role'
-[grid]: https://www.w3.org/TR/wai-aria-1.1/#grid 'ARIA grid role'
 [columnheader]: https://www.w3.org/TR/wai-aria-1.1/#columnheader 'ARIA columnheader role'
 [rowheader]: https://www.w3.org/TR/wai-aria-1.1/#rowheader 'ARIA rowheader role'
 [explicit role]: #explicit-role 'Definition of Explicit Role'
 [presentational roles conflict resolution]: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none 'Presentational Roles Conflict Resolution'
 [sc1.3.1]: https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html 'Understanding Success Criterion 1.3.1: Info and Relationships'
-[html element]: #namespaced-element
--- a/_rules/table-header-cell-has-assigned-cells-d0f69e.md
+++ b/_rules/table-header-cell-has-assigned-cells-d0f69e.md
@@ -18,6 +18,7 @@ acknowledgments:
   authors:
     - Aron Janecki
     - Audrey Maniez
+    - Jean-Yves Moyen
     - Jey Nandakumar
   funding:
     - WAI-Tools
--- a/_rules/table-header-cell-has-assigned-cells-d0f69e.md
+++ b/_rules/table-header-cell-has-assigned-cells-d0f69e.md
@@ -48,12 +48,13 @@ Some popular combinations of browsers and assistive technologies disregard the `
 
 ## Background
 
-The roles inheriting from `cell` are `columnheader`, `gridcell`, and `rowheader`.
+The [HTML specification](https://html.spec.whatwg.org/) contains the [internal algorithm for scanning and assigning header cells](https://html.spec.whatwg.org/multipage/tables.html#internal-algorithm-for-scanning-and-assigning-header-cells) which indicates that the `th` elements with a `scope` [attribute value][] of `auto`, and both a data cell in their row and one in their column are neither row header nor column header. However, some browsers give those `th` elements a table header role and this process is not standardized. One browser may evaluate such `th` elements as [rowheader][] whereas other browsers may evaluate the same `th` elements as [columnheader][]. This is likely not going to convey the relationship conveyed through presentation.
 
-### Bibliography
+This rule does not apply to tables and grids built using [WAI-ARIA 1.1][] [explicit semantic roles][explicit semantic role] as it is not clear how the process of assigning table headers works for such elements.
 
 - [Understanding Success Criterion 1.3.1: Information and relationships][sc1.3.1]
 - [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H43)
+- [H63: Using the scope attribute to associate header cells and data cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H63)
 - [Forming relationships between data cells and header cells][assigned]
 
 ## Test Cases
@@ -343,6 +344,7 @@ This `th` element is part of a table which is not [included in the accessibility
 </table>
 ```
 
+[attribute value]: #attribute-value 'Definition of attribute value'
 [explicit semantic role]: #explicit-role 'Definition of explicit semantic role'
 [semantic role]: #semantic-role 'Definition of semantic role'
 [visible]: #visible 'Definition of visible'
@@ -355,3 +357,4 @@ This `th` element is part of a table which is not [included in the accessibility
 [explicit role]: #explicit-role 'Definition of Explicit Role'
 [presentational roles conflict resolution]: https://www.w3.org/TR/wai-aria-1.1/#conflict_resolution_presentation_none 'Presentational Roles Conflict Resolution'
 [sc1.3.1]: https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html 'Understanding Success Criterion 1.3.1: Info and Relationships'
+[wai-aria 1.1]: https://www.w3.org/TR/wai-aria-1.1/ 'WAI-ARIA 1.1'
--- a/_rules/table-header-cell-has-assigned-cells-d0f69e.md
+++ b/_rules/table-header-cell-has-assigned-cells-d0f69e.md
@@ -80,31 +80,6 @@ This `th` element is [assigned][] to at least one `td` element.
 
 #### Passed Example 2
 
-Each of the 2 `span` elements with role of `columnheader` has assigned `span` elements with a role of `cell`.
-
-```html
-<div role="table">
-	<div role="rowgroup">
-		<div role="row">
-			<span role="columnheader">Month</span>
-			<span role="columnheader">Top Temperature</span>
-		</div>
-	</div>
-	<div role="rowgroup">
-		<div role="row">
-			<span role="cell">July</span>
-			<span role="cell">40 C</span>
-		</div>
-		<div role="row">
-			<span role="cell">August</span>
-			<span role="cell">45 C</span>
-		</div>
-	</div>
-</div>
-```
-
-#### Passed Example 3
-
 Each of the 2 `th` elements has an assigned `td` element because this `td` element spans 2 columns.
 
 ```html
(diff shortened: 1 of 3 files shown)


_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -24,34 +25,37 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML element][] with a [semantic][semantic role] [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any non-empty `th` element that is [included in the accessibility tree][], has a [semantic role][] of [rowheader][] or [columnheader][], and for which all of the following are true:
```
Jym77: "non-empty" will need some definition.
E.g. is `<th> <span></span> </th>` empty even though it has children?

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -24,34 +25,37 @@ acknowledgments:
...
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][] is [included in the accessibility tree][].
+- **in a table** the `th` element is a [descendant][] of a `table` element for which all of the following are true:
+  - **visible** the `table` element is [visible][]; and
+  - **minimum rows** the `table` element's [`rows` collection](https://html.spec.whatwg.org/multipage/tables.html#dom-table-rows) has 2 or more `tr` elements; and
+  - **non-empty data cell** the `table` element has at least one non-empty `td` element that is a [child][] of a `tr` element with the [`rowIndex`][rowindex] value of 1 or more if the `tr` element with the value of 0 has at least two `th` [children][child] or the value of 0 or more if the `tr` element with the value of 0 has one or less `th` [children][child].
+- **no explicit role** none of the elements that make up the [table model][] have an [explicit semantic role][].
```
Jym77, line 35: What if the explicit role is the same as the implicit one? (`<th role="rowheader">`)
Plus, this seems to be a bullet in the outer list (used by the `th`) but rather refers to the `table` (i.e. the inner list).

Jym77: It seems that the `th` element is no more required to be visible or included in the accessibility tree 🤔

Jym77: I'm afraid the nested lists will prove difficult to read.
What about defining a "proper table" or something as a `table` element matching all the conditions here, and just require the `th` to be "part of a proper table"? The def might be re-usable by other rules about HTML tables.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -24,34 +25,37 @@ acknowledgments:
...
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][] is [included in the accessibility tree][].
+- **in a table** the `th` element is a [descendant][] of a `table` element for which all of the following are true:
+  - **visible** the `table` element is [visible][]; and
+  - **minimum rows** the `table` element's [`rows` collection](https://html.spec.whatwg.org/multipage/tables.html#dom-table-rows) has 2 or more `tr` elements; and
+  - **non-empty data cell** the `table` element has at least one non-empty `td` element that is a [child][] of a `tr` element with the [`rowIndex`][rowindex] value of 1 or more if the `tr` element with the value of 0 has at least two `th` [children][child] or the value of 0 or more if the `tr` element with the value of 0 has one or less `th` [children][child].
```
Jym77: "non-empty" will also need to be clarified.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -24,34 +25,37 @@ acknowledgments:
...
-This rule applies to any [HTML element][] with a [semantic][semantic role] [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any non-empty `th` element that is [included in the accessibility tree][], has a [semantic role][] of [rowheader][] or [columnheader][], and for which all of the following are true:
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][] is [included in the accessibility tree][].
+- **in a table** the `th` element is a [descendant][] of a `table` element for which all of the following are true:
+  - **visible** the `table` element is [visible][]; and
+  - **minimum rows** the `table` element's [`rows` collection](https://html.spec.whatwg.org/multipage/tables.html#dom-table-rows) has 2 or more `tr` elements; and
```
Jym77, line 33: 🤔 Not fully sure about that one.
```html
<table>
<tr>
  <th>Opening hours</th> <td>8-17</td>
</tr>
</table>
```
sounds like a reasonable case.
And
```html
<table>
<tr>
  <th>Opening hours</th> <th>Mon-Fri</th> <td>8-17</td>
</tr>
</table>
```
sounds like a bad one (where the first `th` is not assigned to anything due to the second one (or maybe it is still assigned to the second `th`, can't remember…))

(OTOH, restricting Applicability won't create false positives, so it won't really hurt…)

_rules/table-header-cell-has-assigned-cells-d0f69e.md
Jym77: Also not sure about that due to multi-headers.
```html
Opening hours:
<table>
<tr>
  <th rowspan="2">Morning</th> <th>Mon-Fri</th> <td>8-12</td>
</tr>
<tr>
  <!-- no first column --> <th>Sat-Sun</th> <td>10-12</td>
</tr>
<table>
```
would have at least 2 `th` children in the first row without it being a "headers row".

(here also, no real harm in restricting the Applicability)

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -24,34 +25,37 @@ acknowledgments:
...
 
 ## Expectation
 
-Each target element is [assigned][] to at least one element with an [inheriting semantic][] [cell][].
+Each target element is [assigned][] to at least one `td` or `th` element.
 
 ## Assumptions
 
-This rule assumes that table header cells have a relationship conveyed through presentation with other cells within the same table. This excludes edge cases such as a table definition where there is only one header cell, or a table definition where there are multiple headers and no other cells. In such scenarios the rule fails, but [success criterion 1.3.1 Info and Relationships][sc1.3.1] could still be satisfied.
+This rule assumes that table header cells have a relationship conveyed through the presentation with other cells within the same table. It excludes edge cases such as a table definition with only one header cell or a table definition with multiple headers and no other cells that would correspond to them. The rule fails in such scenarios, but [success criterion 1.3.1 Info and Relationships][sc1.3.1] could still be satisfied.
```
Jym77: I think both of these corner cases are currently removed by the **non-empty data cell** condition.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -24,34 +25,37 @@ acknowledgments:
...
 
 ## Accessibility Support
 
-- Table markup and header cell association is not well supported by some popular assistive technologies. Passing this rule can still cause issues for users of those assistive technologies.
-- Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have one of the applicable [semantic roles][semantic role] and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
+Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have one of the applicable [semantic roles][semantic role] and fail this rule with some technology but users of other technologies would not experience any accessibility issue.
```
Jym77: Our definition of "semantic role" is just the `role` attribute. Which means that the **no explicit role** effectively prevents the conflict to happen 🤔

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -193,27 +177,45 @@ Each of the 5 `th` elements in this table has assigned cells, whether data or he
...
+				</tr>
+			</thead>
+			<tbody>
+				<tr>
+					<td>15%</td>
+				</tr>
+			</tbody>
+		</table>
+	</body>
+</html>
```
Jym77, line 212:
```suggestion
		<style>
			table {
				border-collapse: collapse;
				border: hidden;
			}
			tr td,
			tr th {
				border: 1px solid grey;
			}
		</style>
		<table>
			<thead>
				<tr>
					<th>Rate</th>
					<th>Value</th>
				</tr>
			</thead>
			<tbody>
				<tr>
					<td>15%</td>
				</tr>
			</tbody>
		</table>
```
(not sure how much the style is needed, though)

Jym77 (changes requested):

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: Seconded.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: I've chosen the safer version as we don't know for sure what browsers do when aria roles are added as it's not in the spec. UA have their own way of assigning headers when aria is used and at the moment we can only guess. I understand that the same implicit role would work fine in practice but I'm slightly concerned we may expose the rule to some weird, uncommon example that will be a false positive.

Do you think the rule will benefit from adding an exception for the same explicit role?

Hmm... this role applies to all the elements that make up the test target's table element, so `table`,`thead`, `tbody`,`tr`, etc.  I thought that the [table model](https://html.spec.whatwg.org/multipage/tables.html#table-model) definition was explicit enough. Open to suggestions on this.

ajanec01: Sorry, not sure what you mean by this. Are you suggesting that the "included in the accessibility tree" part in the first paragraph of the applicability section is not necessary?

ajanec01: Sure, sounds good to me!

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: Hey @Jym77 ,
Your second example is actually a common thing where the data is yet to be added. As long as visually it is clear that there are no other rows in this table I don't think that would be a failure of 1.3.1. I think that we are going to catch more issues with this rule in the actual two dimensional tables.

Happy to work on it though if you think this is the direction we should follow.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: Correct. Will amend.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: What if the _actual_ table was nested inside a layout table with `role="presentation"`? For example, if the table comes inside of an email (which often use layout tables). This would apply, or would it not?

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: I thought it should be added in response to @kengdoj comment

> From https://www.w3.org/TR/wai-aria-1.1/#columnheader: The columnheader establishes a relationship between it and all cells in the corresponding column. When there are no cells under a column header (as in Fail Ex 1), I don't think 1.3.1 requires that it can't be a column header without a data cell.

My point here is, visually looks like there is a cell for the second column header but programmatically there isn't one so it is a failure. Would you agree?

kengdoj: That pasted comment is from Q5 in the [survey](https://docs.google.com/spreadsheets/d/1OSkPFocXk4K3zYLnwS78WLsWO4PvE5yRcsauyefuIUI/edit#gid=1850701681) so I was a bit confused. 

My survey comment for this example was in Q4 and is below. It didn't look like there's a cell to me, though I can understand that someone may assume one:

> Fail Ex 1's table has no data cell under the second column and programmatically matches the presentation. Not sure this is a failure of 1.3.1.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
carlosapaduarte: Agree with Jean-Yves. The first case is reasonable, and I can't come up with a good reason for us to ignore it.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
carlosapaduarte: Then why not request the table to be present in the accessibility tree for the rule to be applicable?

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -16,42 +16,47 @@ input_aspects:
...
 
 ## Background
 
-The roles inheriting from `cell` are `columnheader`, `gridcell`, and `rowheader`.
+The [HTML specification](https://html.spec.whatwg.org/) contains the [internal algorithm for scanning and assigning header cells](https://html.spec.whatwg.org/multipage/tables.html#internal-algorithm-for-scanning-and-assigning-header-cells) which indicates that the `th` elements with a `scope` [attribute value][] of `auto`, and both a data cell in their row and one in their column are neither row header nor column header. However, some browsers give those `th` elements a table header role and this process is not standardized. One browser may evaluate such `th` elements as [rowheader][] whereas other browsers may evaluate the same `th` elements as [columnheader][]. This is likely not going to convey the relationship conveyed through presentation.
 
-### Bibliography
+This rule does not apply to tables and grids built using [WAI-ARIA 1.1][] [explicit semantic roles][explicit semantic role] as it is not clear how the process of assigning table headers works for such elements.
```
carlosapaduarte: Should update this to ARIA 1.2

carlosapaduarte (changes requested):

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: The question is how much value we get out of that example? Not too much if you think that there is only one row and row headers are useful when you navigate between the rows. This change corresponds to Wilco's [survey](https://docs.google.com/spreadsheets/d/1OSkPFocXk4K3zYLnwS78WLsWO4PvE5yRcsauyefuIUI/edit#gid=1850701681) comments. I'm trying to avoid the below scenario:
 ``` 
      <table>
        <thead>
          <tr>
          <td>N/a
          <th>2008
          <th>2007
          <th>2006
        <tbody>
          <tr>
          <td>
          <td>
          <td>
          <td>
      </table>
```

Because tables are two dimensional the most we will get out of this rule if there will be at least two rows and there will be a non-empty data cell in a row that follows a column header row or in any of the rows providing that there are row headers in them.

Would you agree?

File: _rules/table-header-cell-has-assigned-cells-d0f69e.md (hunk 6 above)
Comment (ajanec01) [reply _rules/table-header-cell-has-assigned-cells-d0f69e.md]: The target needs to be in the accessibility tree which I think also means that the table itself will be in the acc tree.

File: _rules/table-header-cell-has-assigned-cells-d0f69e.md (hunk 8 above)
Comment (ajanec01) [reply _rules/table-header-cell-has-assigned-cells-d0f69e.md]: It's still in Candidate Recommendation. So we are ok to do that?

File: _rules/table-header-cell-has-assigned-cells-d0f69e.md [hunk 9]
```diff
@@ -16,42 +16,47 @@ input_aspects:
...
 ## Applicability
 
-This rule applies to any [HTML element][] with a [semantic][semantic role] [rowheader][] or [columnheader][] for which all of the following is true:
+This rule applies to any [non-empty][] `th` element that is [included in the accessibility tree][], has a [semantic role][] of [rowheader][] or [columnheader][], and for which all of the following are true:
 
-- the element is [visible][]; and
-- the element is [included in the accessibility tree][]; and
-- the element has at least one ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][]; and
-- the element's closest ancestor in the [flat tree][] that is a [semantic][semantic role] [table][] or [grid][] is [included in the accessibility tree][].
+- **in a table** the `th` element is a [descendant][] of a `table` element for which all of the following are true:
```
carlosapaduarte, line 31:
```suggestion
- **in a table**: the `th` element is a [descendant][] of a `table` element for which all of the following are true:
```

Need to add the colon after the name of all conditions

_rules/table-header-cell-has-assigned-cells-d0f69e.md
carlosapaduarte: I don't understand how your example applies to this condition (your example has 2 rows, therefore this conditions applies) :-)

But I can live with this condition. As Jean-Yves mentioned, it will only limit what is applicable, not create false positives. Although I still can't see why we should not consider the one row example Jean-Yves described.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -16,42 +16,47 @@ input_aspects:
...
+  - **visible** the `table` element is [visible][]; and
+  - **minimum rows** the `table` element's [`rows` collection](https://html.spec.whatwg.org/multipage/tables.html#dom-table-rows) has 2 or more `tr` elements; and
+  - **non-empty data cell** the `table` element has at least one [non-empty][] `td` element that is a [child][] of a `tr` element with the [`rowIndex`][rowindex] value of 1 or more if the `tr` element with the value of 0 has at least two `th` [children][child] or the value of 0 or more if the `tr` element with the value of 0 has one or less `th` [children][child].
```
carlosapaduarte, line 34:
```suggestion
  - **non-empty data cell** the `table` element has at least one [non-empty][] `td` element that is a [child][] of a `tr` element with the [`rowIndex`][rowindex] value of 1 or more if the `tr` element with the value of 0 has at least two `th` [children][child]; or the value of 0 or more if the `tr` element with the value of 0 has one or less `th` [children][child].
```

Interestingly, this condition would make Jean-Yves' one row example, applicable.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
```diff
@@ -16,42 +16,47 @@ input_aspects:
...
 
 ## Expectation
 
-Each target element is [assigned][] to at least one element with an [inheriting semantic][] [cell][].
+Each target element is [assigned][] to at least one `td` or `th` element.
 
 ## Assumptions
 
-This rule assumes that table header cells have a relationship conveyed through presentation with other cells within the same table. This excludes edge cases such as a table definition where there is only one header cell, or a table definition where there are multiple headers and no other cells. In such scenarios the rule fails, but [success criterion 1.3.1 Info and Relationships][sc1.3.1] could still be satisfied.
+This rule assumes that table header cells have a relationship conveyed through the presentation with other cells within the same table.
```
carlosapaduarte:
```suggestion
This rule assumes that table header cells have a relationship with other cells within the same table, conveyed through presentation .
```

_rules/table-header-cell-has-assigned-cells-d0f69e.md
carlosapaduarte: Then, your example (of a layout table with `role` `presentation`) is not applicable, and Jean-Yves' comment holds, therefore, no conflict resolution kicks in.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
carlosapaduarte: We have moved a lot of rules already :-) Check #1950

carlosapaduarte (changes requested):
Still some issues to keep working on

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: Resolved in b93861a47a99b89a4b21ef64a64b4ca45ac97bf6

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: The problem that the above example introduces is that there is no use of the `<th>` elements, technically they do not have an assigned `<td>` element and they should not be column headers but the implementation varies. My condition requires at least one non-empty `<td>` element in the second row, meaning that it is clear visually that there is a relationship between a data cell and a column header cell.

_rules/table-header-cell-has-assigned-cells-d0f69e.md
ajanec01: I see, right, will remove that part.

HelenBurge:
Hi @ajanec01 - I was wondering if there was any updates for this as waiting on it for a TF assignment I need to action.

FYI @kengdoj

ajanec01:
Hi @HelenBurge 
Apologies but I won't be able to work on this for some time.

PR #1974, opened 2022-11-18
Form field name passed example should actually pass

WilcoFiers:
Labels with aria-hidden do **not** provide an accessible name. So passed example 8 was actually a failure. Using aria-labelledby does work because it specifically does allow hidden labels.

Need for Call for Review: **NONE** Please note, normally we do send this for review but I think this is a simple mistake, and we need it fixed before this goes to AGWG next week.

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/form-field-non-empty-accessible-name-e086e5.md
+++ b/_rules/form-field-non-empty-accessible-name-e086e5.md
@@ -133,15 +133,15 @@ This [semantic][semantic role] `checkbox` element has the text content as its [a
 
 #### Passed Example 8
 
-These `menuitemcheckbox` elements have an [accessible name][] because its aria-labelledby attribute references a `label` element.
+These `menuitemcheckbox` elements have an [accessible name][] because its aria-labelledby attribute references a `span` element.
 
 ```html
 <p id="dip">Add one or more dip:</p>
 <div role="menu" aria-labelledby="dip">
-	<input type="checkbox" role="menuitemcheckbox" aria-labelledby="ketchup" /><label id="ketchup" aria-hidden="true">Ketchup</label
+	<input type="checkbox" role="menuitemcheckbox" aria-labelledby="ketchup" /><span id="ketchup" aria-hidden="true">Ketchup</span
 	><br />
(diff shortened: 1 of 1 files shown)


giacomo-petri:
Thank you @WilcoFiers .

in your edit, I suggest to replace in passing example description the `label` with the WCAG 2.1 label definition. Right now the accessible name is provided by the relationship given by the aria-labelledby and not by the `<label>` itself.

ajanec01 (approved):

WilcoFiers:
@giacomo-petri Can you put in a suggestion? I'm not sure what element you'd like to see it changed to. Personally I think label is most appropriate here, even using aria-labelledby, but I don't mind using something else. Not a big deal.

giacomo-petri:
> @giacomo-petri Can you put in a suggestion? I'm not sure what element you'd like to see it changed to. Personally I think label is most appropriate here, even using aria-labelledby, but I don't mind using something else. Not a big deal.

I don't know why but I cannot put suggestions for the specific line of code.
I would suggest to change

> These `menuitemcheckbox` elements have an [accessible name][] given by their associated `label`.

with 

> These `menuitemcheckbox` elements have an [accessible name][] because of its aria-labelledby attribute.

only to emphasize that the accessible name is not given by the "natural" relationship between input and labels (such as using for and id, or wrapping the input inside its label), but is given by the aria-labelledby attribute that is pointing to the label element.

tombrunet (approved):

WilcoFiers:
Aaahhh. I see. Agreed, yes. Will fix.

Jym77 (dismissed):
* I second updating the description of the test to not include the `label` element.
* I also think we should maybe use another element than `label` to make thinks clearer. A `label` referenced to using an `aria-labelledby` (rather than `for`) sounds super wonky and I foresee it creating questions and misunderstanding (at best). This may be tackled later, though.
* Technically, hidden `label` should be used, but UA don't do it 🙈 I've gathered some ARIA discussion in [the related Alfa Issue](https://github.com/Siteimprove/alfa/issues/1266)… We might want an Accessibility Support note here (again, may be delayed to later).

giacomo-petri (approved):

Jym77 (approved):

PR #1975, opened 2022-11-21
ARIA state or property is permitted- adding FE4 with aria-checked on a checkbox- 5c01ea

ajanec01:
I feel that the added failed example happens quite often out there and it is not commonly understood why the native states override explicit states so I thought that it would be worthing adding a common example. The example is already covered by the "language feature" point of the expectation section so I only added some background information.

Need for Call for Review:
This will not require a Call for Review.

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -32,14 +32,18 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is [included in the accessibility tree][].
+This rule applies to any [WAI-ARIA state or property][] that is:
+
+- **namespaced element**: specified on an [HTML or SVG element][namespaced element]; and
+- **perceivable**: the element is [included in the accessibility tree][]; and,
+- **not conflicting**: [WAI-ARIA state or property][] is not specified on an element that has the same [implicit WAI-ARIA semantic](https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics).
 
 ## Expectation 1
 
 For each test target, one of the following is true:
 
 - **global**: the test target is a [global state or property][global]; or
-- **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
+- **semantic role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
 ## Expectation 2
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -217,13 +217,21 @@ This `div` element is not [included in the accessibility tree][], hence its [WAI
 
 #### Inapplicable Example 3
 
-The `aria-checked` state is the same as the implicit state of the `input type="checkbox"` element.
+The `input type="checkbox"` element has an implicitly determined state that corresponds to the `aria-checked` state.
 
 ```html
 <label>
 	All
 	<input type="checkbox" aria-checked="mixed" />
 </label>
+```
+
+#### Inapplicable Example 4
+
+The `h3` element has an implicitly determined property that corresponds to the `aria-level` property.
+
+```html
+<h1 aria-level="5">Introduction</h1>
 ```
 
 [attribute value]: #attribute-value 'Definition of attribute value'
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -58,8 +58,6 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 In HTML, there are language features that do not have corresponding implicit WAI-ARIA semantics. As per [ARIA in HTML](https://www.w3.org/TR/html-aria/), those elements can have [global states or properties][global]. Some of those elements can also have [inherited][], [supported][], or [required][] [states][state] or [properties][property] that correspond to a [WAI-ARIA role](https://www.w3.org/TR/wai-aria-1.2/#introroles). For example, the `audio` element has no corresponding ARIA semantics but it can have [inherited][], [supported][], or [required][] [states][state] or [properties][property] of the [`application` role](https://www.w3.org/TR/wai-aria-1.2/#application).
 
-As per the [conflicts with host language semantics](https://www.w3.org/TR/wai-aria-1.2/#host_general_conflict) section of [WAI-ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/), user agents must ignore [WAI-ARIA state or property][] that have the same [implicit WAI-ARIA semantic](https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics) in order to avoid providing conflicting states and properties to assistive technologies.
-
 Assessing the value of the attribute is out of scope for this rule.
 
 ### Related rules
@@ -193,17 +191,6 @@ The `aria-label` property is [prohibited][] for an element with a `generic` role
 
 ```html
 <div aria-label="Bananas"></div>
-```
-
-#### Failed Example 4
-
-The `aria-checked` state is [prohibited][] for an `input type="checkbox"` element to avoid providing conflicting states.
-
-```html
-<label>
-	<input type="checkbox" aria-checked="mixed" />
-	All
-</label>
 ```
 
 ### Inapplicable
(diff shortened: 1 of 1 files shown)


colabottles (approved):

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -58,6 +58,8 @@ Implementation of [Presentational Roles Conflict Resolution][] varies from one b
 
 In HTML, there are language features that do not have corresponding implicit WAI-ARIA semantics. As per [ARIA in HTML](https://www.w3.org/TR/html-aria/), those elements can have [global states or properties][global]. Some of those elements can also have [inherited][], [supported][], or [required][] [states][state] or [properties][property] that correspond to a [WAI-ARIA role](https://www.w3.org/TR/wai-aria-1.2/#introroles). For example, the `audio` element has no corresponding ARIA semantics but it can have [inherited][], [supported][], or [required][] [states][state] or [properties][property] of the [`application` role](https://www.w3.org/TR/wai-aria-1.2/#application).
 
+As per the [conflicts with host language semantics](https://www.w3.org/TR/wai-aria-1.2/#host_general_conflict) section of [WAI-ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/), user agents must ignore [WAI-ARIA state or property][] that have the same [implicit WAI-ARIA semantic](https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics) in order to avoid providing conflicting states and properties to assistive technologies.
+
```
Jym77: I'm not sure this (and new example) is actually an ARIA problem 🤔 

---
The WAI-ARIA section you reference mostly mention user agents behaviour, and host languages specs. It seems the only part that refers authors behaviour is:
> When a feature in the host language with identical role semantics and values is available, and the author has no compelling reason to avoid using the host language feature, authors SHOULD use the host language features rather than repurpose other elements with WAI-ARIA.

but then the "SHOULD" bit make it not mandatory.
So I do feel that ARIA does not consider `aria-checked` as prohibited when `checked` is here.

Similarly, HTML AAM for `checked` states:
> If the element includes both the checked attribute and the aria-checked attribute with a valid value, User Agents MUST expose only the checked attribute value.

this corresponds to the WAI-ARIA bit "host languages MUST explicitly declare where the use of WAI-ARIA attributes on each host language element conflicts with native attributes for that element". But again, this (and the paragraph this comes from) is about host language and UA requirements, not author requirements.

So, I feel it does have some value to flag it (as in "it won't act as you expect"), but this is more of a Best Practice than really a requirement. Notably, `<input type="checkbox" aria-checked="mixed" />` won't cause any accessibility problem per se with conforming UAs that are just going to discard the `aria-checked` (it may cause some if the author relies on `aria-checked` somewhere else, but that checkbox alone is OK.

---
Would it make sense to keep that out of this rule and have another rule "conflicting property is not used"?

ajanec01: Hey @Jym77 
I used the quote from ARIA as it seems to explain the issue well. I should've probably used ARIA in HTML to make it clearer. I believe that it is a requirement. ARIA in HTML says:

> [WAI-ARIA](https://www.w3.org/TR/html-aria/#bib-wai-aria-1.2) identifies roles which have [prohibited states and properties](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes). These roles do not allow certain WAI-ARIA attributes to be specified by authors. HTML elements which expose these implicit WAI-ARIA roles also prohibit authors from specifying these WAI-ARIA attributes.

And then it has the following next to the `input type="checkbox"` in the _Rules of ARIA attribute usage by HTML element_ table:

> Authors [MUST NOT use the aria-checked attribute on input type=checkbox elements](https://www.w3.org/TR/html-aria/#att-checked).

... and the following next to "Any element where the [checked](https://html.spec.whatwg.org/multipage/input.html#attr-input-checked) attribute is allowed" in the _Rules of ARIA attribute usage by HTML feature_ table:

> Authors MUST NOT use the [aria-checked](https://www.w3.org/TR/wai-aria-1.2/#aria-checked) attribute on any element where the [checkedness](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-checked), or the indeterminate checked value of the element can be in opposition to the current value of the aria-checked attribute.

It causes an issue where the appropriate state may not be exposed to the user which in my view may be a failure of 4.1.2. As the rule's expectation number one references ARIA in HTML I think that it is valid.

Open to any suggestions. Do you still feel strongly that this should be a separate rule or should I just make it clearer in the background section and reference ARIA in HTML?

Jym77: Good points 👍 
I agree that ARIA in HTML makes this an author requirement 👍 
And indeed, the rule seems to already include ARIA in HTML stuff 👍 
(this does raise the question of the requirement mapping only going to WAI-ARIA and some ARIA technique, but that's a bit unrelated).

I think the background note should also refer to ARIA in HTML (the existing one does so already).

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -193,6 +195,17 @@ The `aria-label` property is [prohibited][] for an element with a `generic` role
 <div aria-label="Bananas"></div>
 ```
 
+#### Failed Example 4
+
+The `aria-checked` state is [prohibited][] for an `input type="checkbox"` element to avoid providing conflicting states.
```
WilcoFiers: aria-checked is not prohibited for checkbox. This is a different type of problem, that should probably go into a separate rule.

WilcoFiers (changes requested):

_rules/aria-state-or-property-permitted-5c01ea.md
ajanec01: Hey Wilco
That's true, the definition of prohibited states and properties does not apply in this case.

Initially I thought there was not enough to make it a separate rule. I thought mainly of `aria-checked` and `aria-disabled` but things such as `aria-level` will also come into play in here and that could be a direct failure of WCAG.

In this case shouldn't we make a change to the **language feature** point of Expectation 1 to not include conflicting states and properties and write a separate rule?

ajanec01: Hey @WilcoFiers 
I'm planning to write a separate rule about this. In the meantime I thought it was worth adding the exception to the applicability.

_rules/aria-state-or-property-permitted-5c01ea.md
ajanec01: Hey @Jym77 
Please review again as per my and Wilco's conversation. Thanks!

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -32,14 +32,18 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is [included in the accessibility tree][].
+This rule applies to any [WAI-ARIA state or property][] that is:
+
+- **namespaced element**: specified on an [HTML or SVG element][namespaced element]; and
+- **perceivable**: the element is [included in the accessibility tree][]; and,
+- **not conflicting**: [WAI-ARIA state or property][] is not specified on an element that has the same [implicit WAI-ARIA semantic](https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics).
```
Jym77, line 39: Following our [Listed conditions](https://act-rules.github.io/pages/design/rule-design/#listed-conditions) guidelines
```suggestion
This rule applies to any [WAI-ARIA state or property][] for which all the following is true:

- **namespaced element**: the property is specified on an [HTML or SVG element][namespaced element]; and
- **perceivable**: the property is specified on an element which is [included in the accessibility tree][]; and,
- **not conflicting**: the property is not specified on an element that has the same [implicit WAI-ARIA semantic](https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics).
```
(also, you can use `<dfn>` element for the condition names and refer to them in the example descriptions)

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -211,6 +215,25 @@ This `div` element is not [included in the accessibility tree][], hence its [WAI
 <div role="button" aria-sort="" style="display:none;"></div>
 ```
 
+#### Inapplicable Example 3
+
+The `input type="checkbox"` element has an implicitly determined state that corresponds to the `aria-checked` state.
```
Jym77, line 220:
```suggestion
This `input type="checkbox"` element has an implicitly determined state that corresponds to the `aria-checked` state.
```

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -211,6 +215,25 @@ This `div` element is not [included in the accessibility tree][], hence its [WAI
...
+```html
+<label>
+	All
+	<input type="checkbox" aria-checked="mixed" />
+</label>
+```
+
+#### Inapplicable Example 4
+
+The `h3` element has an implicitly determined property that corresponds to the `aria-level` property.
```
Jym77, line 231:
```suggestion
This `h1` element has an implicitly determined property that corresponds to the `aria-level` property.
```

Jym77 (changes requested):
I'm not sure of the pertinence of these changes now 🤔 
Treating them as Inapplicable (rather than Passed) makes the rule a bit more complex (notably implementation-wise) without adding much value to it (since Passed and Inapplicable are mostly the same thing) (tool vendor hat on, I'm not sure I'm going to bother making that change).

I do see that these cases are problematic, but then if we write another rule to catch them we should be good and we maybe don't need to make this rule more complex.

We're also getting in a bad situation with the new examples, since we do agree that they are failing the requirement, and Inapplicable examples should not fail the requirement as per our [Consistency With Accessibility Requirements](https://act-rules.github.io/pages/design/test-cases/#consistency-with-accessibility-requirements) guidelines.

PR #1987, opened 2022-12-08
Link rules to WAI site rather than CG site

WilcoFiers:
Fix hard-coded links so they go to the WAI website rather than the CG website.

Need for Call for Review: none, changes are editorial.

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-hidden-no-focusable-content-6cfa84.md
+++ b/_rules/aria-hidden-no-focusable-content-6cfa84.md
@@ -54,7 +54,7 @@ The 1 second time span introduced in the exception of the definition of [focusab
 
 ### Related rules
 
-- [Element with presentational children has no focusable content](https://act-rules.github.io/rules/307n5z)
+- [Element with presentational children has no focusable content](https://www.w3.org/WAI/standards-guidelines/act/rules/307n5z/proposed/)
 
 ### Bibliography
 
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -62,7 +62,7 @@ Assessing the value of the attribute is out of scope for this rule.
 
 ### Related rules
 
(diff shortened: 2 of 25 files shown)


_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -54,7 +54,7 @@ The 1 second time span introduced in the exception of the definition of [focusab
 
 ### Related rules
 
-- [Element with presentational children has no focusable content](https://act-rules.github.io/rules/307n5z)
+- [Element with presentational children has no focusable content](https://www.w3.org/WAI/standards-guidelines/act/rules/307n5z/proposed/)
```
Jym77, line 57: I'm not a fan of the fact that we need to add `proposed` at the end of the URL ourselves 🤔 
This means that if the rule get accepted, we need to update rules that link to it to got to the accepted version (or keep linking to the proposed version, but this is not what this PR does for rules that link to accepted rules).

And in this case, the rule is accepted and `https://www.w3.org/WAI/standards-guidelines/act/rules/307n5z` exists…

---
Would it be possible for the non-accepted rules to have, e.g. `https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea` redirect to `https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea/proposed`, so that `https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea` acts as an URI for the rule which stays stable whatever the status of the rule is? 
Then we could use these URIs to link to from other rules.

colabottles (approved):

_rules/aria-hidden-no-focusable-content-6cfa84.md
WilcoFiers: I've had that on my wishlist for a while now. But I don't want to hold up fixing bad links for it. I created an issue here: https://github.com/act-rules/act-tools/issues/14

carlosapaduarte (approved):

Jym77 (approved):

PR #1835, opened 2022-04-19
b4f0c3: Adjust meta viewport applicability based on properties defined

tombrunet:
Adjust the applicability of the rule to be more narrow and only apply to meta elements whose content attributes defines one of the affected values. Move Pass 1 and Pass 4 to the N/A section since these examples do not define these attributes. Discussed briefly in taskforce call April 14.

Closes issue(s): 

Need for Call for Review: 1 week

---

### **When creating PR:**

- [x] Make sure you're requesting to **pull a branch** (right side) to the `develop` branch (left side).
- [x] Make sure you do not remove the "How to Review and Approve" section in your pull request description

### **After creating PR:**

- [x] Add yourself (and co-authors) as "Assignees" for PR.
- [x] Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- [ ] [Link the PR to any issue it solves](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue). This will be done automatically by [referencing the issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) at the top of this comment in the indicated place.
- [ ] Optionally request feedback from anyone in particular by assigning them as "Reviewers".

### **When merging a PR:**

- [ ] Close any issue that the PR resolves. This will happen automatically upon merging if the PR was correctly linked to the issue, e.g. by referencing the issue at the top of this comment.

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Call for Review period. In case of disagreement, the longer period wins.

PR #1994, opened 2022-12-15
"Meta viewport allows for zoom" (b4f0c3): Explicit meaning of 'has'

### LINKED ISSUE #1982: "Meta viewport allows for zoom" (b4f0c3) should explain what "having" a property means
[Meta viewport allows for zoom](https://www.w3.org/WAI/standards-guidelines/act/rules/b4f0c3/) states:
> Applicability: (…) the content attribute value **has** the maximum-scale property.
> Expectation: (…) the attribute value does not **have** a maximum-scale property with a value less than 2.

(emphasises mine)

The [CSS spec for `maximum-scale`](https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale) mentions that accepted "specified" values are numbers or some strings (`yes`, `no`, …) and mentions how to interpret (translate) these. It also says that "Negative number values are dropped".

---
Now, where the confusion is, is that I am not sure if in our rule "having" a property refers to the value written in the `content` attribute (what would be the "specified" value if this was a CSS property), or the one obtained after the translation in the spec (what would be the "computed" or "used" value if this was a CSS property).

Notably, the Expectation is using a raw number, thus suggesting that "have" in the Expectation refers to the "computed" (post-translation) value of the property.
But Passed Example 3 uses `maximum-scale=-1` which, post-translation, does not "have" the `maximum-scale` property since it has been dropped, thus suggesting that "has" in the Applicability refers to the "specified" value (pre-translation) of the property.

---
Alfa choses to uses the post-translation value for deciding if the attribute "has" a property and therefore considers Passed Example 3 as Inapplicable.
This is not a huge deal since Passed and Inapplicable are very similar in meaning, but should still be clarified in the rule.

Jym77:
[Meta viewport allows for zoom](https://www.w3.org/WAI/standards-guidelines/act/rules/b4f0c3/) was using sentence like "the `content` attribute value **has** the property…" with different meaning of "has" in the Applicability and Expectation. As exemplified in Passed Example 3, the Applicability is concerned about the presence of the substring in the `content` value (as written); while the Expectation is concerned about the presence of the property in the key-value pairs list after parsing and dropping the value as written.

I've chose to keep the different meanings and make them explicit in order to keep PE3 as a Passed example. Another possibility would have been to always used the parsed and translated key-value pairs list and make PE3 an Inapplicable example, which I feel wasn't the intention of the latest rewrite of the rule in #1835.

I've also added an AS note since it is not clear all UAs behave the same way.

I'm not sure the links to the suggested algorithm are fully needed, but I felt it was hard to justify why certain values are dropped without some reference.

Closes issue(s):

- closes #1982

Need for Call for Review:
This will require a 1 week Call for Review (Clarification to Applicability and Expectation without changing the meaning).

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,20 +24,26 @@ acknowledgments:
...
+- the `content` attribute specified value includes the text `user-scalable`; or
+- the `content` attribute specified value includes the text `maximum-scale`.
+
+**Note:** The Applicability refers to string inclusion of the indicated strings in the value of the `content` attribute, as written.
 
 ## Expectation 1
 
 For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `no`.
 
+The [attribute value][] of the `content` attribute is a list of key-value pairs, obtained by parsing the specified value, then applying the [translations into `@viewport` descriptors][descriptor translation], notably [the translation for `user-scalable`](https://www.w3.org/TR/css-device-adapt-1/#user-scalable).
```
carlosapaduarte: I'm finding this has the potential to confuse people. By referencing the transplantation to @viewport blocks, it becomes harder to understand what really is checked. A specified attribute of `no` originates an attribute value of `no` which gets translated into `fixed` for the `user-zoom` descriptor. 

The descriptor will never have a value of `no`, and is not what the expectation looks at. Therefore, I'm not sure this sentence adds value to understanding the expectation. 

Same applies to Expectation 2.

carlosapaduarte (dismissed):

_rules/meta-viewport-b4f0c3.md
Jym77: Good points.
* Expectation 2, and the way it is written ("does not have a value less than 2") is pretty much the reason why we need to refer to these translations, since we want, e.g. `maximum-scale=-1` to pass (Passed Example 3). I am not sure we can do without referring to the translation one way or another.
* Expectation 1 is more interesting since it seems to be yet another meaning of "have" 🙈
* Notably, there are no example to show whether we wanted to accept `user-scalable=0`. It is "specified" as `0`, but "computed" as `fixed` and therefore passes the rule (it does not have a value of `no` in any decent sense of the term because "translated into the same value as `no` is translated" is definitely not a decent sense for "have"…) but has the exact same effect as `user-scalable=no` which fails it 💥 

I think we messed up in the latest update and Expectation 1 should say "does not have a `user-scalable` property with a value of **`fixed`**." Plus keep the note about "have" meaning "after translation" (which would match Expectation 2)

carlosapaduarte: But from what I understand (please, correct me if I'm wrong), it's not the `user-scalable` property that gets the value fixed. It's the `user-zoom` descriptor that gets that value. So, I guess we have two choices:

1. We mention the `user-zoom` descriptor in the expectation
2. We update it to say that `user-scalable` property must not have a value of `no`, or in the range of -1 to 1, or an unknown value.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,20 +24,26 @@ acknowledgments:
...
+
 ## Expectation 2
 
 For each test target, the [attribute value][] does not have a `maximum-scale` property with a value less than 2.
 
+The [attribute value][] of the `content` attribute is a list of key-value pairs, obtained by parsing the specified value, then applying the [translations into `@viewport` descriptors][descriptor translation], notably [the translation for `maximum-scale`](https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale) (including dropping the key-value pair completely for invalid values).
```
carlosapaduarte: Same issue here. What we want is the `max-zoom` description to not have a value less than 2. So, we can either say that in the expectation, or say that the `maximum-scale` property, must not have a numeric value under 2, or a literal value that is different from `device-width` and `device-height`

carlosapaduarte (dismissed):
Still not clear about the expectation targeting the correct thing

_rules/meta-viewport-b4f0c3.md
Jym77: Ah, yes, I see your point.
I think we should avoid rephrasing the translation, and that was a bit the goal of #1822 (no need to duplicate that and if CSS decides to change it, the rule should probably follow suit).

What about adding something like "…and using the value obtained for `user-zoom`" at the end of the sentence?

carlosapaduarte: That could work

_rules/meta-viewport-b4f0c3.md
carlosapaduarte: Still, if we have an example like 

```
<meta name="viewport" content="maximum-scale=device-width">
```

how do I compare `device-width` with `2`?

carlosapaduarte (dismissed):

_rules/meta-viewport-b4f0c3.md
Jym77: After translating it and using the value for `max-zoom`, `device-width` results in a `max-zoom` of 10, which is greater than 2, so this is OK 🤔

carlosapaduarte: I think only now it clicked in my head. I was using the definition of `attribute value` to compute the value. But the second paragraph adds to the definition. Should we, instead, update the definition of `attribute value` to compute the value in this instance that refers to the value of a CSS property instead of the value of the HTML attribute?

Jym77: I'm not sure because this is currently the only rule doing that, so I'd prefer to keep locality.
OTOH, we can definitely  improve the situation with some examples. I'll give it a go.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,20 +24,26 @@ acknowledgments:
...
+- the `content` attribute specified value includes the text `maximum-scale`.
+
+**Note:** The Applicability refers to string inclusion of the indicated strings in the value of the `content` attribute, as written.
 
 ## Expectation 1
 
-For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `no`.
+For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `fixed`.
+
+The [attribute value][] of the `content` attribute is a list of key-value pairs, obtained by parsing the specified value, then applying the [translations into `@viewport` descriptors][descriptor translation], notably [the translation for `user-scalable`](https://www.w3.org/TR/css-device-adapt-1/#user-scalable), and using the value obtained for `user-zoom`.
```
carlosapaduarte: Trying to make the algorithm clearer. Also adding the "For the purpose of this rule" at the beginning so that it is clear that the attribute value is not computed using our standard definition. 

```suggestion
For the purpose of this rule, the [attribute value][] of the `content` attribute is a list of key-value pairs, obtained by:

1. parsing the specified value for each key-value pair;
2. applying the [translations into `@viewport` descriptors][descriptor translation];
3. using the value of the corresponding `@viewport` descriptor, notably using the value obtained for the `user-zoom` descriptor from [the translation for `user-scalable`](https://www.w3.org/TR/css-device-adapt-1/#user-scalable).
```

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,20 +24,26 @@ acknowledgments:
...
 
 ## Expectation 2
 
 For each test target, the [attribute value][] does not have a `maximum-scale` property with a value less than 2.
 
+The [attribute value][] of the `content` attribute is a list of key-value pairs, obtained by parsing the specified value, then applying the [translations into `@viewport` descriptors][descriptor translation], notably [the translation for `maximum-scale`](https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale) (including dropping the key-value pair completely for invalid values), and using the value obtained for `max-zoom`.
```
carlosapaduarte:
```suggestion
For the purpose of this rule, the [attribute value][] of the `content` attribute is a list of key-value pairs, obtained by:

 1. parsing the specified value for each key-value pair;
 2. applying the [translations into `@viewport` descriptors][descriptor translation];
 3. using the value of the corresponding `@viewport` descriptor  (including dropping the key-value pair completely for invalid values), notably using the value obtained for the `max-zoom` descriptor from [the translation for `maximum-scale`](https://www.w3.org/TR/css-device-adapt-1/#min-scale-max-scale).
```

carlosapaduarte (dismissed):
I suggest a revision of the way the value computing is specified, but I'm approving already because I believe you current description is sufficient.

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,20 +24,34 @@ acknowledgments:
...
-- the `content` [attribute value][] has the `maximum-scale` property.
+- the `content` attribute specified value includes the text `user-scalable`; or
+- the `content` attribute specified value includes the text `maximum-scale`.
+
+**Note:** The Applicability refers to string inclusion of the indicated strings in the value of the `content` attribute, as written.
 
 ## Expectation 1
 
-For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `no`.
+For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `fixed`.
```
WilcoFiers: Should this have been `user-zoom`?

I'm a little hesitant for us to start using the CSS property names / values here in place of the actual key/value pairs on the content attribute. It creates adds a layer of abstraction to the rule 
which makes it more difficult to understand. Since rules on how to process these props are pretty easy, I would suggest we put them straight into the rule in the same way that the autocomplete rule has a pseudo-copy of the HTML spec's description of what makes a valid autocomplete attribute value.

```suggestion
For each test target, when the [attribute value][] is parsed into key/value pairs, one of the following is true:
- there is no `user-scalable` key; or
- the `user-scalable` value is `yes`; or
- the `user-scalable` value is a number less than -1, or greater than 1.
```

We can then put something in the background about how the expectation follows the steps described in CSS. You also wouldn't have to add those translate statements to the example descriptions.

WilcoFiers (dismissed):

_rules/meta-viewport-b4f0c3.md
Jym77: Good idea. I gave it a shot 👍

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,22 +24,37 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to each `content` attribute on a `meta` element with a `name` [attribute value][] of `viewport` for which at least one of the following is true:
+This rule applies to each `content` attribute for which all of the following are true:
 
-- the `content` [attribute value][] has the `user-scalable` property; or
-- the `content` [attribute value][] has the `maximum-scale` property.
+- the attribute is defined on a `meta` element with a `name` [attribute value][] of `viewport`; and
+- the [attribute value][] has either the `maximum-scale` or `user-scalable` key.
```
carlosapaduarte: Doesn't this make the following attribute inapplicable?

```html
<meta name="viewport" content="maximum-scale=2, user-scalable=5" />
```

It doesn't have either of the attributes. It has both. 
And I believe this should be applicable.

carlosapaduarte (dismissed):

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,22 +24,37 @@ acknowledgments:
...
+For the purpose of this rule, the [attribute value][] of the `content` attribute is a list of key/value pairs.
 
 ## Expectation 1
 
-For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `no`.
+For each test target's [attribute value][], at least one of the following is true:
+
+- the `user-scalable` key is not defined; or
+- the `user-scalable` key is paired with a value of `yes`, `device-width` or `device-height`; or
+- the `user-scalable` key is paired with a numeric value which is not in the range ]-1;1[.
```
WilcoFiers: I would prefer we simplify this a bit. I don't think this is any less ambiguous than what you had:

```suggestion
- `user-scalable` is not defined; or
- `user-scalable` is `yes`, `device-width` or `device-height`; or
- `user-scalable` is a number which is not between -1 and 1.
```

WilcoFiers (dismissed):
I like this direction much better. I do think you can be a bit less verbose.

colabottles (approved):

_rules/meta-viewport-b4f0c3.md
```diff
@@ -24,22 +26,37 @@ acknowledgments:
...
+- the [attribute value][] has at least one of the `maximum-scale` or `user-scalable` keys.
+
+For the purpose of this rule, the [attribute value][] of the `content` attribute is a list of key/value pairs.
 
 ## Expectation 1
 
-For each test target, the [attribute value][] does not have a `user-scalable` property with a value of `no`.
+For each test target's [attribute value][], at least one of the following is true:
+
+- `user-scalable` is not defined; or
```
WilcoFiers, line 40: I'm okay with this as you write it, but I think it might be better if eventually we split this up. Have one rule for user-scalable, and a second for maximum-scale. This split expectation is more complicated than it needs to be I think because of this.

WilcoFiers (approved):

_rules/meta-viewport-b4f0c3.md
Jym77: Good point.

Related to that, and given what we semi-often do, I wonder if it would be a good idea to allow expectations like:
> For each test target [which verify some condition], …

So that test targets that do not verify the condition would automatically pass that expectation.

In that case, that would allow to write:
> For each test target that defines `user-scalable`, …

And in general, anytime we have Applicability with an alternative, that would allow somewhat simpler expectation. And result in less rules overall, I feel these splittings are often diluting rules that should stay close together.

Alternatively, this could also be done with Composite rules. Having one Atomic rule for each property, and a Composite requesting both to not fail. It is a bit redundant, but avoids splitting rules that are really working together.

carlosapaduarte (approved):

_rules/meta-viewport-b4f0c3.md
```diff
@@ -83,6 +102,24 @@ This viewport `meta` element does not prevent user scaling because it has `user-
...
+		<title>Simple page showing random text</title>
+		<meta name="viewport" content="user-scalable=5" />
+	</head>
+	<body>
+		<p>
+			Lorem ipsum
+		</p>
+	</body>
+</html>
+```
```
tombrunet, line 119: Is there a reason that this new example was added in the middle rather than the end? I'm not sure that we're strict about this, but if there's something out there already that refers to Pass Example 2, that reference would be bad since Pass Example 2 is now Pass Example 3. I'm sure it happens anyway when we remove examples, but I don't see a reason to do that in this case.

tombrunet (dismissed):
Somewhat minor comment, but hoping to clean this up before approving.

_rules/meta-viewport-b4f0c3.md
Jym77: Yes, I wanted to keep the `user-scalable` tests together, and the `maximum-scale` together. I think it makes the rule more readable than if we mix them (less need to switch context).

Our test cases numbers are absolutely not stable, and we should not link to them directly. This is acceptable when referring to them in issues, … but we shouldn't, e.g., have a rule linking to an example with a `passed-example-2` id. If we need to link inside rule, we should manually had an anchor.

_rules/meta-viewport-b4f0c3.md
Jym77: Tracking the idea in https://github.com/w3c/wcag-act/issues/548

Jym77:
Call for Review ends on October 12th.

tombrunet (approved):

Jym77:

PR #2005, opened 2023-01-16
ARIA attr defined; Editorial improvements

WilcoFiers:
Mostly editorial work. Tweaked one example which had 2 ARIA violations on it, to only have one.

Need for Call for Review: 1 week, updated an example

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-attr-defined-5f99a7.md
+++ b/_rules/aria-attr-defined-5f99a7.md
@@ -60,7 +60,7 @@ This `div` element with a role of `dialog` has an `aria-modal` attribute which i
 
 #### Passed Example 3
 
-This `div` element with a role of `textbox` has multiple `aria-*` attributes which are defined in [WAI-ARIA Specifications][].
+This `div` element with a role of `textbox` has multiple `aria-` attributes which are defined in [WAI-ARIA Specifications][].
 
 ```html
 <div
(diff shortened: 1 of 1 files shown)


_rules/aria-attr-defined-5f99a7.md
```diff
@@ -93,17 +93,15 @@ This `input` element with a role of `spinbutton` has multiple `aria-*` attribute
...
-<ul>
-	<li role="menuitemcheckbox" aria-not-checked="true">List Item</li>
-</ul>
+<div role="checkbox" aria-not-checked="true">All conditions are met</div>
 ```
 
 #### Failed Example 2
 
-This `div` element with a role of `searchbox` has an `aria-*` attribute (`aria-labelled`) which is not defined in [WAI-ARIA Specifications][].
+This `div` element with a role of `searchbox` has an `aria-` attribute (`aria-labelled`) which is not defined in [WAI-ARIA Specifications][].
```
carlosapaduarte:
```suggestion
This `div` element with a role of `searchbox` has an `aria-labelled` attribute which is not defined in [WAI-ARIA Specifications][].
```

To be coherent with the previous example.

carlosapaduarte (dismissed):

_rules/aria-attr-defined-5f99a7.md
WilcoFiers:
```suggestion
This `div` element with a role of `searchbox` has an `aria-labelled` attribute (instead of `aria-labelledby`) which is not defined in [WAI-ARIA Specifications][].
```

daniel-montalvo: I would mention the non-existing attribute explicitly, as in #2009 but I don't feel strongly.

carlosapaduarte (approved):

HelenBurge (approved):

ajanec01 (approved):

PR #1986, opened 2022-12-07
Add more secondary requirements

WilcoFiers:
- Add secondary requirements to ARIA rules
- Add level A criteria as secondary for level AAA rules
- List 1.1.1 as secondary for link name, because of `area` elements.
- Update act-tools so that secondary requirements are no longer included in the "Rule mapping" section at the page top
- Update the test to allow omitting the mapping on secondary requirements

Need for Call for Review: 1 week

---

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Call for Review period. In case of disagreement, the longer period wins.

PR #2006, opened 2023-01-16
role-valid: Remove the assumption, clarify in background

WilcoFiers:
Resolving ACT TF review feedback. This assumption should have been removed when we removed SC 1.3.1 from the requirements mapping in #1986

Need for Call for Review: none, this is pretty much just editorial IMO.

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -177,7 +177,7 @@ This `input` element does not have an [explicit role][] of `textbox`, but the `a
 
 #### Failed Example 1
 
-The `aria-sort` [property][] may not be used on a [semantic][semantic role]] `button`.
+The `aria-sort` [property][] may not be used on a [semantic][semantic role] `button`.
 
 ```html
 <button aria-sort="">Sort by year</button>
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -43,7 +43,7 @@ This rule applies to any [WAI-ARIA state or property][] that is specified on an 
 For each test target, one of the following is true:
 
 - **global**: the test target is a [global state or property][global]; or
-- **semantic Role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
+- **semantic role**: the test target is an [inherited][], [supported][], or [required][] [state][] or [property][] of the [semantic role][] of the element on which the test target is specified; or
 - **language feature**: the test target is specified on an [HTML element][namespaced element] and is allowed on that element. Which ARIA states or properties may be used on which element is described in [ARIA in HTML](https://w3c.github.io/html-aria/).
 
 ## Expectation 2
@@ -85,7 +85,7 @@ Assessing the value of the attribute is out of scope for this rule.
 
 #### Passed Example 1
 
-The `aria-pressed` [state][] is [supported][] for a [semantic][semantic role] `button`, which is the [implicit role][] for `button` elements.
+The `aria-pressed` [state][] is [supported][] with `button`, which is the [implicit role][] for `button` elements.
 
 ```html
 <button aria-pressed="false">My button</button>
@@ -93,7 +93,7 @@ The `aria-pressed` [state][] is [supported][] for a [semantic][semantic role] `b
 
 #### Passed Example 2
 
-The `aria-pressed` [state][] is [supported][] for the [semantic][semantic role] `button`, which is the [explicit role][] of this `div` element.
+The `aria-pressed` [state][] is [supported][] with `button`, which is the [explicit role][] of this `div` element.
 
 ```html
 <div role="button" aria-pressed="false">My button</div>
@@ -109,7 +109,7 @@ The `aria-busy` [state][] is a [global][] [state][] that is [supported][] by all
 
 #### Passed Example 4
 
-The `aria-label` [property][] is a [global][] [property][] and thus [inherited][] for all [semantic role][].
+The `aria-label` [property][] is a [global][] [property][]. It is allowed on any [semantic role][].
 
 ```html
 <div role="button" aria-label="OK">✓</div>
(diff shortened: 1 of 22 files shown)


_rules/role-attribute-valid-value-674b10.md
```diff
@@ -42,20 +42,20 @@ Each test target has at least one token which is a valid value corresponding to
...
 
 The `role` attribute is a set of [space separated tokens][]. Having a [whitespace](#whitespace) separated list of more than one token in the value of the role attribute is used for what is known as _fallback roles_. If the first token is not accessibility supported (or valid), the next one will be used for determining the [semantic role][] of the element, and so forth. Having the rule target attributes containing at least one non-[ASCII whitespace][] character ensures that there is at least one token in the set.
 
+Not every rule should be used on every element. Which ARIA roles may be used on which HTML elements is defined in [ARIA in HTML](https://www.w3.org/TR/html-aria/). Testing this is not part of this rule.
```
carlosapaduarte:
```suggestion
Not every role should be used on every element. Which ARIA roles may be used on which HTML elements is defined in [ARIA in HTML](https://www.w3.org/TR/html-aria/). Testing this is not part of this rule.
```

carlosapaduarte (dismissed):
Typo

carlosapaduarte (dismissed):

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -42,19 +42,19 @@ Each test target has at least one token which is a valid value corresponding to
 
 ## Assumptions
 
-- This rule assumes that the `role` attribute is used to provide an ARIA [semantic role][] to the elements. If it is used for other purposes, this rule shouldn't be used.
-- This rule assumes that elements with a `role` attribute have their intended structure and relationship conveyed through some sort of presentation. If it is not the case, it is possible to fail this rule while still satisfying [Success Criterion 1.3.1 Info and Relationship][sc131].
-- This rule assumes that the intended role of the element is not its [implicit role][]. If no token is valid, User Agents will default to the [implicit role][] for the element; if that role is the intended one, it is possible to fail this rule but still satisfy [Success Criterion 1.3.1 Info and Relationships][sc131].
+There are no assumptions.
```
ajanec01: I'm not sure what the discussions were so not marking it as changes requested but I'd rather leave the first paragraph of the Assumptions section as is. As pointed out by @Jym77 some time ago `role` attribute may have some other use cases. The [Role Attribute spec](https://www.w3.org/TR/role-attribute/) says the following 

> Use cases include accessibility, device adaptation, server-side processing, and complex data description
...

> The role attribute is necessary to support Accessible Rich Internet Applications (WAI-ARIA) ... Although this is the reason the Role Attribute is published by the Protocols and Formats Working Group, the attribute has more general use cases as well.

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -42,19 +42,19 @@ Each test target has at least one token which is a valid value corresponding to
...
 
 ## Accessibility Support
 
 Older browsers do not support more than one token in the value for a role attribute. If multiple values are used in the role attribute, the attribute is ignored in these browsers.
 
 ## Background
 
-Using an invalid role is often the result of a typo or other developer error. Such roles are ignored by browsers and assistive technologies, and the element's default semantics is used. This often means that a role that should exist is missing. This can cause issues under [success criterion 1.3.1 Info and Relationships][sc131] or [4.1.2 Name, Rule Value][sc412].
+Using an invalid role is often the result of a typo or other developer error. Unknown roles are ignored by browsers and assistive technologies, and the element's [implicit role][] is used. This often means that a role that should exist is missing. This can cause issues under [success criterion 1.3.1 Info and Relationships][sc131] or [4.1.2 Name, Role, Value][sc412]. If the element's implicit semantics are sufficient to communicate its intent, an invalid role may not cause an accessibility issue.
```
ajanec01, line 53: To me this is not entirely the same as the first paragraph of the assumptions section, if that's what the intent was. This refers more to typos and not using the `role` attribute for different purposes; hence why I'm in favour of leaving the first paragraph of the assumption section.

_rules/role-attribute-valid-value-674b10.md
WilcoFiers: The reason we're moving this is because these rules no longer explicitly fail WCAG:
https://github.com/act-rules/act-rules.github.io/pull/2006/files#diff-b2c6c65374b4ddd9a243e403522e4c9237f44c8ef0c7bc5e5d601b1b12b023f6R18-R21


In ACT RF 1.1 we've introduced the concept of secondary requirements. These are requirements that can, but don't always fail the success criterion. So instead assuming that these things are never true, we're in the background explaining that if these things are true, the rule fails SC 1.3.1, but regardless if it is or isn't true, this is invalid ARIA. Hope that makes sense, if not happy to jump on a call to talk through this. This is important for the direction of ACT, so I want to get a broad perspective on this.

ajanec01:
```suggestion
This rule assumes that the `role` attribute is used to provide an ARIA [semantic role][] to the elements. If it is used for other purposes, this rule shouldn't be used.
```

Hey @WilcoFiers , as per our call, it's worth leaving the first paragraph in as the rule is mapping to the technique and there might be, however small, possibility of using the `role` attribute intentionally not for accessibility.

ajanec01 (dismissed):

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -177,7 +177,7 @@ This `input` element does not have an [explicit role][] of `textbox`, but the `a
 
 #### Failed Example 1
 
-The `aria-sort` [property][] may not be used on a [semantic][semantic role]]  `button`.
+The `aria-sort` [property][] may not be used on a [semantic][semantic role]] `button`.
```
Jym77:
```suggestion
The `aria-sort` [property][] may not be used on a [semantic][semantic role] `button`.
```
(I'm also not sure these typo are related with what the PR says it does and whether it should be part of another PR instead)

Jym77 (approved):
It looks like the typos correction in other files are a bit unrelated, which might pollute a bit their changelog in the end 🤔

ajanec01 (approved):

carlosapaduarte (approved):

Issue #1999, opened 2023-01-05
"Links with identical accessible names and context serve equivalent purpose" [fd3a94]: Failed Example 1 should pass.


Jym77:
(splitting from #1879 since it is taking a different direction)
[Failed Example 1](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/fd3a94/228c0a3d78557fb48a855d6733d50848a86f0d62.html) is:
```html
<html lang="en">
	<p>
		We are on social media:
		<a href="https://act-rules.github.io/">ACT rules</a>
		<a href="https://www.w3.org/community/act-r/">ACT rules</a>
	</p>
</html>
```
However, in that case, the links purposes are [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general), which is a valid exception of [2.4.4](https://www.w3.org/TR/WCAG21/#link-purpose-in-context)

We've discussed this a bit while discussing #1879, but that issue took the direction of refining the definition of "context", which won't fix this problem.

giacomo-petri:
I'll work on creating a non-semantic structure with image and "Read more" text that allow sighted users to visually associate the elements, but keeping the same semantic context for AT users.

giacomo-petri:
Something like:
```
<div style="display:flex; justify-content:space-around;">
    <img src="..." alt="image context 1">
    <img src="..." alt="image context 2">
</div>
<div style="display:flex; justify-content:space-around;">
    <a href="...">Read more</a>
    <a href="...">Read more</a>
</div>
```

Jym77:
But I don't think that updating the Failed Example solves the problem.
As long as the code (currently) in Failed Example 1 is failing the rule, the rule is problematic. If we update the example to be an actual failure, but the rules keep failing the old code, then it is still a False Positive.

carlosapaduarte:
But the rule has an assumption stating: 

[...]

Issue #1622, opened 2021-05-26
Equivalent resource definition seems ambiguous


WilcoFiers:
This one effects rather a lot of rules:

- [Block Of Repeated Content Is Collapsible](https://act-rules.github.io/rules/3e12e1)
- [Bypass Blocks Of Repeated Content](https://act-rules.github.io/rules/cf77f2)
- [Document Has Heading For Non-Repeated Content](https://act-rules.github.io/rules/047fe0)
- [Document Has An Instrument To Move Focus To Non-Repeated Content](https://act-rules.github.io/rules/ye5d6e)
- [Document Has A Landmark With Non-Repeated Content](https://act-rules.github.io/rules/ye5d6e)
- [Iframe Elements With Identical Accessible Names Have Equivalent Purpose](https://act-rules.github.io/rules/b40fd1)
- [Links With Identical Accessible Names Have Equivalent Purpose](https://act-rules.github.io/rules/4b1c6c)
- [Links With Identical Accessible Names And Context Serve Equivalent Purpose](https://act-rules.github.io/rules/b20e66)

I think the following phrase is ambiguous:

> If all resources cover the user's expectations equally well, the resources are considered to be equivalent.

Which users is this about? How can I know their expecations? How can it ever be "covered equally well" if the test is different. Even minor differences, even differences that have nothing to do with the text could arguably be sufficient to say one "covers" the expectation slightly better.

PR #2007, opened 2023-01-17
Update Links with identical accessible names and context serve equivalent purpose rule 

giacomo-petri:
Our current failed scenarios are ambiguous to users in general, which conflicts with 1st assumption.
With these changes I've:
* Moved "not ambiguous to users in general" from Assumptions to Applicability and updated Expectations adding a new one.
* Updated failed scenarios, providing a clear visual context for sighted users, that doesn't reflect in an equal semantic context for AT users, resulting in links with same acc name and context, but not ambiguous to users in general.
* Added 2 png files to make the examples more "realistic" (they have been created from scratch, so we have full rights).
* Updated the definition of "programmatically determined link context, adding note about punctuation (which in my opinion is relevant).
* Created 2 inapplicable example for:
    * scenario where one of the links in set is not in the acc tree.
    * scenario which is ambiguous to users in general.

Closes issue(s): #1999

Need for Call for Review:
I'm note sure about the timing here as multiple failed cases have been updated. I've set 2 weeks but feel free to update if needed.

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


giacomo-petri:
Just reviewed Carlos' feedback here https://github.com/act-rules/act-rules.github.io/issues/1999#issuecomment-1385162708.

I'll update it asap.

giacomo-petri:
@carlosapaduarte, updated per your request in https://github.com/act-rules/act-rules.github.io/issues/1999#issuecomment-1385162708. 

Ready for review

giacomo-petri:
TODO: create very simple static demo pages for failed examples that reflect the link acc name (Contact Us).

Note: we still need to update the context definition to handle scenarios where the meaning of the sentence make the purpose of the link clear without ambiguity (with a different PR?).

e.g. 
`<div><a href="#">contact us</a> via chat, or <a href="#">contact us</a> via email.</div>`

Punctuation might be covered by the note I've added in "ambiguous to users in general" definition, but link might be enough meaningful and distinguishable within its context also without punctuation.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,20 +42,23 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
 **Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
 
-## Expectation
+## Expectation 1
 
 When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource).
 
 **Note**: Resolving the links includes potential redirects, if the redirects happen instantly.
 
+## Expectation 2
```
carlosapaduarte: I don't think this expectation is needed, because links that are ambiguous to users in general will not be applicable. Therefore, all applicable test targets will always meet expectation 2.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -217,92 +221,120 @@ These two SVG `a` and HTML `a` elements have the same [accessible name][], same
...
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They are visually distinguishable thanks to the relationships conveyed through CSS, but go to different resources.
 
 ```html
 <html lang="en">
 	<p>
-		We are on social media:
-		<a href="https://act-rules.github.io/">ACT rules</a>
-		<a href="https://www.w3.org/community/act-r/">ACT rules</a>
+		<h2>Contact us:</h2>
+		<a href="https://act-rules.github.io/" style="display:inline-block; background-image: url(images/chat.png); background-position: 0; background-repeat: no-repeat; background-size: 40px; padding: 20px 0 20px 50px;">Contact Us</a>
```
carlosapaduarte: The URL of the image resources is not pointing to the correct test-asset.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -217,92 +221,120 @@ These two SVG `a` and HTML `a` elements have the same [accessible name][], same
...
+		<a href="https://www.w3.org/community/act-r/" style="display:inline-block; background-image: url(images/phone.png); background-position: 0; background-repeat: no-repeat; background-size: 40px; padding: 20px 0 20px 50px; margin-left:40px;">Contact Us</a>
```
carlosapaduarte: The URL of the image resources is not pointing to the correct test-asset.

pages/glossary/programmatically-determined-link-context.md
```diff
@@ -16,6 +16,8 @@ The _programmatically determined context_ of a link (or _programmatically determ
 - being a header cell [assigned][] to the closest [ancestor][] of the link in the [flat tree][] that has a [semantic role][] of `cell` or `gridcell`; or
 - being referenced by an `aria-describedby` attribute of the link.
 
+**Note**: Since screen readers interpret punctuation, they can also provide the context from the current sentence, when the focus is on a link in that sentence.
```
carlosapaduarte: I'm not sure why you added this note to the definition. Why is mentioning the sentence relevant here?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -217,92 +221,120 @@ These two SVG `a` and HTML `a` elements have the same [accessible name][], same
...
-		Learn more (<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/index.html"
-			>Contact us</a
-		>) and get in touch (<a
-			href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/redirect1.html"
-			>Contact us</a
-		>)
+		<span style="text-align:center;">Contact us</span>
+		<span style="display:flex; justify-content:space-around;">
+			<img src="/test-assets/shared/chat.png" alt="Chat" style="width:50%;">
+			<img src="/test-assets/shared/phone.jpg" alt="Phone" style="width:50%;">
```
carlosapaduarte:
```suggestion
			<img src="/test-assets/shared/phone.png" alt="Phone" style="width:50%;">
```

carlosapaduarte (changes requested):
Nice work, but some issues still left to address 

Feel free to add yourself to the list of authors of this rule

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Thanks @carlosapaduarte, I really don't know why I put it there.

Do you think might be worth to put it in background notes, just to clarify why we put this point here?
I'll do that, but if you feel it is not necessary at all, please let me know.

pages/glossary/programmatically-determined-link-context.md
giacomo-petri: Per my previous content

> Note: we still need to update the context definition to handle scenarios where the meaning of the sentence make the purpose of the link clear without ambiguity (with a different PR?).
> 
> e.g. `<div><a href="#">contact us</a> via chat, or <a href="#">contact us</a> via email.</div>`
> 
> Punctuation might be covered by the note I've added in "ambiguous to users in general" definition, but link might be enough meaningful and distinguishable within its context also without punctuation.

I was trying to partially address the example provided, but I understand that as is, it is only more confusing. I remove it but I think we should address it in a separate PR.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,20 +42,20 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
 - the elements are [included in the accessibility tree][included in the accessibility tree]; and
 - the elements have [matching][] [accessible names][accessible name] that are not empty (`""`); and
-- have the same [programmatically determined link context][].
+- the elements have the same [programmatically determined link context][]; and
+- the purpose of the elements is not [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general). 
 
 **Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
 
-## Expectation
+## Expectation 1
```
Jym77:
```suggestion
## Expectation
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -68,10 +68,13 @@ This rule is designed specifically for [2.4.4 Link Purpose (In Context)][sc244],
 
 This rule specifically targets links within the exact same context. Links with identical name that are in identical (but not the same) contexts also fail [2.4.4 Link Purpose (In Context)][sc244]. However, defining "identical context" unambiguously was not really possible and was left out of this rule.
 
+Elements that are [inheriting semantic][] `link` nodes within each test target are distinguishable each other by some user. This means that this rule applies to links with identical accessible names and context that are not [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general), which is the exception mentioned in Success Criterion 2.4.4 Link Purpose (In Context). If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links, which makes it more of a general user experience concern than an accessibility issue.
```
Jym77: I don't really understand the first sentence in this paragraph 🤔

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -217,92 +220,120 @@ These two SVG `a` and HTML `a` elements have the same [accessible name][], same
...
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They are visually distinguishable thanks to the relationships conveyed through CSS, but go to different resources.
 
 ```html
 <html lang="en">
 	<p>
-		We are on social media:
-		<a href="https://act-rules.github.io/">ACT rules</a>
-		<a href="https://www.w3.org/community/act-r/">ACT rules</a>
+		<h2>Contact us:</h2>
+		<a href="https://act-rules.github.io/" style="display:inline-block; background-image: url(/test-assets/shared/chat.png); background-position: 0; background-repeat: no-repeat; background-size: 40px; padding: 20px 0 20px 50px;">Contact Us</a>
```
Jym77:
```suggestion
		<a href="https://act-rules.github.io/" style="display:inline-block; background: url(/test-assets/shared/chat.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px;">Contact Us</a>
```
Should have the same effect.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -217,92 +220,120 @@ These two SVG `a` and HTML `a` elements have the same [accessible name][], same
...
+		<a href="https://www.w3.org/community/act-r/" style="display:inline-block; background-image: url(/test-assets/shared/phone.png); background-position: 0; background-repeat: no-repeat; background-size: 40px; padding: 20px 0 20px 50px; margin-left:40px;">Contact Us</a>
```
Jym77:
```suggestion
		<a href="https://www.w3.org/community/act-r/" style="display:inline-block; background: url(/test-assets/shared/phone.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px; margin-left: 40px;">Contact Us</a>
```

Jym77 (changes requested):
I think that would work 💯

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: The target of this rule is a set of two or more elements that are [inheriting semantic][] `link` nodes.

These 2 or more element that are part of the target, that have the same acc name, same context, etc. are somehow distinguishable by some user, resulting in something that is not ambiguous to users in general.

I agree, it's a bit complex as written. 

Would be better to remove the first part, so:
```suggestion
This rule applies to links with identical accessible names and context that are not [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general), which is the exception mentioned in Success Criterion 2.4.4 Link Purpose (In Context). If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links, which makes it more of a general user experience concern than an accessibility issue.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,7 +42,8 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
 - the elements are in the same [web page (HTML)][]; and
 - the elements are [included in the accessibility tree][included in the accessibility tree]; and
 - the elements have [matching][] [accessible names][accessible name] that are not empty (`""`); and
-- have the same [programmatically determined link context][].
+- the elements have the same [programmatically determined link context][]; and
+- the purpose of the elements is not [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general).
```
carlosapaduarte: I'm having second thoughts about this. The definition is 

> the purpose cannot be determined from the link and all information of the Web page presented to the user simultaneously with the link

I don't think this is objective. I agree it is unambiguous, so we might rewrite the rule in a way that moves this condition from the applicability to the expectation (and update the examples accordingly). 

What do you think?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
carlosapaduarte: I prefer the revision you propose. However, I do not understand why this is in the Accessibility Support section instead of the Background section. This is not related to assistive technology limitations.

carlosapaduarte (changes requested):

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Right, moving to background notes.

Edit: uhm, this is already in the background notes :D
I'm only replacing it with the new revision.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: @carlosapaduarte ,

I agree, even more considering that the definition of "ambiguous to users in general" is far away from our definition of "programmatically determined link context".

On it

giacomo-petri: @carlosapaduarte,

I've tried with multiple solutions, but for all of them, removing "not ambiguous to users in general" from the applicability, makes the expectation even more complicated to understand.

In addition, the rule title is "Links with identical accessible names and same context serve equivalent purpose", but removing the "not ambiguous to users in general" from the applicability makes the title no longer valid; in fact links with identical accessible names and same context (but ambiguous to users in general) might serve different purposes, resulting in a pass for the rule that contrasts with the title.

What do you think?

carlosapaduarte: My suggestion would be to update the current expectation to something like 

> For the links in each set of target elements, one of the following is true:
> - The links are not [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general) and, when followed, resolve to the [same resource][] or to [equivalent resources](#equivalent-resource); or
> - The links are [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general)

Jym77: The we don't even need the exception in the first bullet:
> either
> * the links resolve to the same resource; or
> * the links are ambiguous

giacomo-petri: - Links not ambiguous to users in general has been removed from Applicability
- Added expectation as per suggestions
- Slightly changed the background note about ambiguous links to reflect new applicability
- Moved Inapplicable example n.8 to passed example 9 (changing a bit the description).

carlosapaduarte (approved):

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +232,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+			<img src="/test-assets/shared/phone.png" alt="Phone" style="width:50%;">
+		</span>
+		<span style="display:flex; justify-content:space-around;">
+			<span
 			role="link"
 			tabindex="0"
-			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html'"
-			>Contact us</span
-		>) and get in touch (<span
+			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html'">Contact Us</a>
```
carlosapaduarte:
```suggestion
			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html'">Contact Us</span>
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +232,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+			<span
 			role="link"
 			tabindex="0"
-			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html'"
-			>Contact us</span
-		>)
-	</p>
+			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html'">Contact Us</a>
```
carlosapaduarte:
```suggestion
			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html'">Contact Us</span>
```

carlosapaduarte (changes requested):
Tests are failing because of incorrect HTML

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,20 +42,22 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
 
 **Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
 
 ## Expectation
 
-When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource).
+For the links in each set of target elements, one of the following is true:
+
+- when followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource); or
+- the purpose of the links is [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general).
```
Jym77: "ambiguous to user in general" refers to a single link, not several.
So we should rather say something like
> For each target set, the destination of links that are not ambiguous to users in general are pairwise different

(i.e. remove the ambiguous links from the set and then check that the rest is different)

But then I'm not sure "ambiguous to users in general" get un-ambiguous enough:
> the purpose cannot be determined from the link and all information of the Web page presented to the user simultaneously with the link (i.e., readers without disabilities would not know what a link would do until they activated it)

(see also #1622 on a related ambiguity)

So, maybe wrap it up the other way around:
> For each pair of links in each target set, either the links go to the same place, or there is no visual information on the page to determine they go to different places.

---
(I do believe we need the expectation to looks at every pair of links, rather than the set together, because there may be a mix of ambiguous links and links going to different places which makes the alternative hard to handle)

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +232,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
-			>Contact us</a
-		>)
+		<span style="text-align:center;">Contact us</span>
+		<span style="display:flex; justify-content:space-around;">
+			<img src="/test-assets/shared/chat.png" alt="Chat" style="width:50%;">
+			<img src="/test-assets/shared/phone.png" alt="Phone" style="width:50%;">
+		</span>
+		<span style="display:flex; justify-content:space-around;">
+			<a href="https://act-rules.github.io/">Contact Us</a>
+			<a href="https://www.w3.org/community/act-r/">Contact Us</a>
```
Jym77: Actually, also wondering what happens with browsers that show the link URL when hovering them (e.g., Chrome shows it in the bottom left corner of the window). I do personally use that info sometimes, essentially to disambiguate links before clicking 🤔
But then of course, if the URL is enough to disambiguate links, then they are never really ambiguous to users in general unless the URL itself is non-informative (e.g. `contact1.html` vs `contact2.html`).
Should we have an Accessibility Support note to mention that?

Jym77 (changes requested):

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: That's an interesting point to consider while proceeding with your suggested proposal for expectation: 

> For each pair of links in each target set, either the links go to the same place, or there is no visual information on the page to determine they go to different places.

as "visual information on the page" might be not enough clear and might include also the browser visual experience (e.g. as you said Chrome shows the URL on hover or focus).

We might say something like:

> For each pair of links in each target set, one of the following is true:
> - either the links resolve to the [same resource][] or to [equivalent resources](#equivalent-resource); or 
> - links resolve to different resources but there is no visual information within the content of the page to predict they go to specific different contents.

With "within the content" we restrict the scope to the visual presentation of the HTML content, preventing people to include visual browsers functionalities that give suggestions/hint.

----

I'm not really concerned about the hover functionality, as it behaves consistently with assistive technologies as well; for example, using VO, if you press ctrl+opt+shift+U you can listen the href attribute. Similarly, if you use VoiceControl, you can say "Move mouse to $number" to display the Chrome "hint" with url info.

On the other hand, how many users are aware about that?

And, even if the URL is different and might help differentiating the destinations, there might be url structures that are more or less complex to be understood, especially for non-expert people; if you think about something like:

* https://demo.com/contact-us
* https://demo.com/about-us

or 

* https://demo.com/en-us/category/client-service/contact-us
* https://demo.com/en-us/category/client-service/about-us

or 

* https://demo.com/en-us/category/client-service/contact-us/?j=432847382974328937428942896&f=438243892748329743829&name=demopage
* https://demo.com/en-us/category/client-service/about-us/?j=432847382974328937428942896&f=438243892748329743829&name=demopage

they are very different in terms of understanding.
I would say that, even if the differences might be perceived and understood by someone using the url, the acc name and visual context within the page are still ambiguous to users in general.

Jym77: Good point. Maybe that should be an assumption (that reading the URL as per hovering in a browser or invoking some shortcut in a screen reader isn't enough to disambiguate as far as WCAG is concerned).
Maybe also strengthening the examples by using URLs that are somewhat confusing. On that direction, I think we can build on top of your last example with query parameters that are a SHA to a customers support ticket or products something like that `<a href="../catalog/product.html?product=432847382974328937428942896">View product</a> <a href="../catalog/product.html?product=438243892748329743829">View product</a>`
That would totally be ambiguous even with the URLs… But could be visually disambiguated by aligned pictures of the products (the same way you did for the chat/phone icons).

giacomo-petri: - updated expectation to every pair of links, as per @Jym77 suggestion;
- added assumption to clarify that browser/AT support to display/announce the URL is not enough to disambiguate links.

TODO:
update some example by using confusing URL

giacomo-petri: Added examples with query parameters to emphasize ambiguity.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,21 +42,24 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
 
 **Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
 
 ## Expectation
 
-When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource).
+For each pair of links in each target set, one of the following is true:
+
+- either the links resolve to the [same resource][] or to [equivalent resources](#equivalent-resource); or
+- links resolve to different resources but there is no visual information within the content of the page to predict this behaviour.
```
carlosapaduarte: Failing the spelling check 

```suggestion
- links resolve to different resources but there is no visual information within the content of the page to predict this behavior.
```

carlosapaduarte (dismissed):
Spelling check still failing

carlosapaduarte:
@WilcoFiers There are multiple instances of the test checking the existence of the resources failing in this rule. Giacomo added test cases with query parameters in the URL (following [Jean-Yves suggestion](https://github.com/act-rules/act-rules.github.io/pull/2007#discussion_r1087843874)), and now the test complains the resource does not exist. The resource is there, but without the query parameter, of course. Can you update the test to not look at query parameters in the URL of the resource?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,21 +43,24 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
 
 **Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
 
 ## Expectation
 
-When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource).
+For each pair of links in each target set, one of the following is true:
+
+- either the links resolve to the [same resource][] or to [equivalent resources](#equivalent-resource); or
+- links resolve to different resources but there is no visual information within the content of the page to predict this behavior.
```
Jym77:
```suggestion
- both links resolve to the [same resource][]; or 
- both links resolve to [equivalent resources][]; or
- there is no visual information within the content of the page to let users know that both links resolve to [non-equivalent resources][equivalent resources].
```
Streamlining a bit the list:
* avoid the nested "or";
* start each item with "**both** links" to emphasise the fact that this is about pairs of links;
* remove the pre-condition in the last item since it is an "otherwise" clause for the rest;
* trying to replace "predict" which sounds a bit crystal-bally 🔮 

(there might be need to add `[equivalent resource]: #equivalent-resource 'Definition of Equivalent Resource'` to the references list at the end)

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+
+```html
+<html lang="en">
+	<p>
+		<h2>Contact us:</h2>
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=1" style="display:inline-block; background: url(/test-assets/shared/chat.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px;">Contact Us</a>
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=2" style="display:inline-block; background: url(/test-assets/shared/phone.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px; margin-left: 40px;">Contact Us</a>
+	</p>
+</html>
+```
```
Jym77: I feel this also fails 1.3.1 by not conveying programmatically the info conveyed by the background images. Likely also fails1.1.1 since these are informative images without text alternative…

That's not really a problem, but then I guess we should have 1.1.1 and 1.3.1 as secondary requirements with a Background note explaining why.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+
 #### Failed Example 2
 
-These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They link to web pages that are similar, but have different information in their content.
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], but go to different resources.
```
Jym77:
```suggestion
These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], but go to different resources. Their purpose is disambiguated for sighted users by the alignement of the links with the images above.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=60e950cff70bf1ec60a702086748ab4dec361514">Contact Us</a>
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=c1d4e0f067462f4b28716f028d9213a25eb82f28">Contact Us</a>
+		</span>
 	</div>
 </html>
 ```
 
 #### Failed Example 3
 
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They link to web pages that are similar, but have different information in their content.
```
Jym77:
```suggestion
These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They link to web pages that are similar, but have different information in their content. Their purpose is disambiguated for sighted users by the alignement of the links with the images above.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+	<div>
+		<span style="text-align:center;">Contact us</span>
+		<span style="display:flex; justify-content:space-around;">
+			<img src="/test-assets/shared/chat.png" alt="Chat" style="width:50%;">
+			<img src="/test-assets/shared/phone.png" alt="Phone" style="width:50%;">
+		</span>
+		<span style="display:flex; justify-content:space-around;">
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html">Contact Us</a>
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html">Contact Us</a>
+		</span>
```
Jym77: I'm not sure the chat/phone disambiguation is good enough given the links.
That is, as a sighted user, how would I know that "Contact us (phone)" is the link to click to access the Admission office rather than general reception?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+	</div>
+</html>
+```
+
+#### Failed Example 4
+
 These two HTML `span` elements have an [explicit role][] of link, same [accessible name][] and [context][programmatically determined link context], but link to resources that offer different content.
```
Jym77:
```suggestion
These two HTML `span` elements have an [explicit role][] of link, same [accessible name][] and [context][programmatically determined link context], but link to resources that offer different content. Their purpose is disambiguated for sighted users by the alignement of the links with the images above.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
-		>) and get in touch (<span
+			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/about/contact.html'">Contact Us</span>
+			<span
 			role="link"
 			tabindex="0"
-			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html'"
-			>Contact us</span
-		>)
-	</p>
+			onclick="location='/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/admissions/contact.html'">Contact Us</span>
```
Jym77: Same remark about the images.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +234,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+			<text>Contact us</text>
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=1" aria-label="Contact Us">
+				<path d="m212.0806 68.0717c-10.3917 10.3852-22.4311 20.3239-27.1905 15.5646-6.8075-6.8075-11.0088-12.7418-26.0285-.6696-15.0132 12.0657-3.4792 20.1139 3.1182 26.7047 7.6149 7.6149 36.0001.407 64.0571-27.6434 28.0504-28.057 35.2386-56.4422 27.6172-64.0571-6.5974-6.604-14.6062-18.1314-26.6719-3.1182-12.0723 15.0132-6.1444 19.2145.6761 26.0285 4.7397 4.7593-5.1925 16.7988-15.5777 27.1905z"/>
 			</a>
-
-			<a href="https://www.w3.org/community/act-r/">
-				<text x="50" y="90" text-anchor="middle">
-					ACT rules
-				</text>
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=2" aria-label="Contact Us">
```
Jym77, line 353: It seems the "contact us" page doesn't act on `?page=1/2` but only on the more complex values.

Jym77 (changes requested):
It feels we're getting closer 🎉

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Uhm... it should thanks to 
`var page = urlParams.get('page');`

Are you sure it doesn't work? I'll check again.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: I've added 2 blocks in _/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html_ file to manage this scenario. I've created similar breadcrumbs, similar services navigation, similar headings but different content, to reflect the failure description.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Applied same solution for this one.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
Jym77: 😓 Haven't followed that bit of the logic. Then it is probably OK…

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Added the following background note:

> This rule contains Failed Examples that use background informative images without an alternative text to convey information. This leads also [success criterion 1.1.1 Non-text Content][sc111] to fail. Similarly, the visual position of the elements allows sighted users to disambiguate the target links thanks to their vertical alignment with the images above. This causes issues under [success criterion 1.3.1 Info and Relationships][sc131] as relationships conveyed visually are not programmatically determined.

giacomo-petri: Should I add also 1.1.1 and 1.3.1 to the _Bibliography_ section?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -68,10 +77,15 @@ This rule is designed specifically for [2.4.4 Link Purpose (In Context)][sc244],
...
 
+Links that are [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general) are covered by the exception mentioned in Success Criterion 2.4.4 Link Purpose (In Context). If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links, which makes it more of a general user experience concern than an accessibility issue.
+
+This rule contains Failed Examples that use background informative images without an alternative text to convey information. This leads also [success criterion 1.1.1 Non-text Content][sc111] to fail. Similarly, the visual position of the elements allows sighted users to disambiguate the target links thanks to their vertical alignment with the images above. This causes issues under [success criterion 1.3.1 Info and Relationships][sc131] as relationships conveyed visually are not programmatically determined.
```
Jym77: I'd put that in a more generic way.
```suggestion
Pages with links that are not [ambiguous to users in general][], but are ambiguous to some users are likely to fail [success criterion 1.3.1 Info and Relationships][sc131] is the disambiguation information is conveyed through presentation, but not visually. Moreover, when this information is non-text content, such a page is likely to fail [success criterion 1.1.1 Non-text Content][sc111]. Therefore, both these success criteria are secondary requirement for this rule.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
Jym77: Indeed, looking at the other rules it seems we are putting the secondary requirements in the Bibliography.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
Jym77: Yes, it should indeed work . Sorry about that.

Jym77 (changes requested):

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: I've added them into the Bibliography section adding "(secondary requirement)" to clarify why they are there. Maybe we can find a more meaningful way, but I just put it there for now.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Thank you, suggestion committed. I've replaced "but not visually" with "but not semantically", assuming you meant this.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,21 +47,25 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
+- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of a link can only accurately describe one resource (notably, homonyms alone are not used as link names). Thus, if two or more links have the same [accessible name][] but resolve to different resources, at least one of them does not accurately describe its purpose.
+- This rule assumes that assistive technologies are exposing all links on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its links, then it is possible for two links to have identical name and context but resolve to different resources without failing [Success Criterion 2.4.4 Link Purpose (In Context)][sc244] (if said links are in separate [documents][document] or [shadow trees][shadow tree]).
+- This rule assumes that reading the URL as per hovering/focusing in a browser or invoking AT shortcuts isn't enough to disambiguate links.
```
Jym77:
```suggestion
- This rule assumes that reading the URL as per hovering/focusing in a browser or invoking an assistive technology shortcuts isn't enough to disambiguate links. 
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +243,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+			<img src="/test-assets/shared/chat.png" alt="Chat" style="width:50%;">
+			<img src="/test-assets/shared/phone.png" alt="Phone" style="width:50%;">
+		</span>
+		<span style="display:flex; justify-content:space-around;">
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=3">Contact Us</a>
+			<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=4">Contact Us</a>
+		</span>
+	</div>
+</html>
+```
```
Jym77: I'm not sure I remember why we kept both example 2 and 3 😓 
They seem to be the same, except for the query parameters in the URLs 🤔

Jym77 (approved):
Looks good 👍 
Couple of minor thnigs, approving already.

Jym77:
@giacomo-petri It seems the tests are failing because you're not in the contributors list.
I suggest you do so in a separate PR (since it affects all of yours…), see, e.g. #2048 
(plus a typo on "disambiguated" 🤔 )

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: - example 2 refers to completely different resources (?page=1 and ?page=2);
- example 3 refers to similar resources with the same page structure but different values for specific fields (page=3 and ?page=4 e.g. phone number and opening hours).

This distinction was already present at the time of the edits, so I kept it also for the new adjustments. If it's not necessary we can remove it.

Jym77: Ah, yes, thanks for the refresh.
Then, I guess I'm challenging the complex query parameter which seems unnecessarily and totally took me off when looking at them.
I do remember we did discuss these as an example of "the URL is not always enough to disambiguate". But in the end this is not something that is present in the rule (mentioned in the assumption, but not really discussed in the rule). I guess example 2 would work as well with `page1` and `page2` as query parameters 🤔 (it would also simplify the code by removing the test). But OTOH, we may want to keep a "realistic" hash query parameter.

giacomo-petri: @Jym77, 
I agree with above.
Maybe a more clear understanding (page1 and page2 as query params) is better than covering a "realistic" case but both the solutions work for me.

Jym77: Let's see if others have opinions about it 😃

colabottles (approved):

carlosapaduarte (approved):

giacomo-petri:
This PR is in call for review for 2 weeks. Review ends Sept 29th.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,21 +47,25 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
 **Note:** The test target for this rule is the full set of link elements that share the same [matching][] [accessible name][] and [programmatically determined link context][].
 
 ## Expectation
 
-When followed, the links in each set of target elements resolve to the [same resource][] or to [equivalent resources](#equivalent-resource).
+For each pair of links in each target set, one of the following is true:
+
+- both links resolve to the [same resource][]; or 
+- both links resolve to [equivalent resources][equivalent resource]; or
+- there is no visual information within the content of the page to let users know that both links resolve to [non-equivalent resources][equivalent resource].
```
WilcoFiers: I'm not keen on this "content of the page" phrase. At the very least this should link to our web page (HTML) definition. Better would be if we had a definition of "content", but I'm okay opening a separate issue for that and merging this PR. We have that problem in a few other rules too.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,21 +47,25 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
-- This rule assumes that assistive technologies are exposing all links on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its links, then it is possible for two links to have identical name and same context but resolve to different resources without failing [Success Criterion 2.4.4 Link Purpose (In Context)][sc244] (if said links are in separate [documents][document] or [shadow trees][shadow tree])
+- This rule assumes that, within the context of the test subject, the description provided by the [accessible name][] of a link can only accurately describe one resource (notably, homonyms alone are not used as link names). Thus, if two or more links have the same [accessible name][] but resolve to different resources, at least one of them does not accurately describe its purpose.
```
WilcoFiers, line 58: This is pretty difficult to read, would you mind trying to simplify the language of these?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -42,21 +47,25 @@ This rule applies to any set of two or more [HTML or SVG elements][] for which a
...
+- This rule assumes that assistive technologies are exposing all links on the page in the same way no matter which [document tree](https://dom.spec.whatwg.org/#document-trees) they are in. If an assistive technology requires the user to "enter" an `iframe` or a [shadow tree][] before exposing its links, then it is possible for two links to have identical name and context but resolve to different resources without failing [Success Criterion 2.4.4 Link Purpose (In Context)][sc244] (if said links are in separate [documents][document] or [shadow trees][shadow tree]).
+- This rule assumes that reading the URL as per hovering/focusing in a browser or invoking an assistive technology shortcuts isn't enough to disambiguate links.
```
WilcoFiers: How about this?

```suggestion
- This rule assumes that reading the URL, for example from the status bar when the link is focused is not part of the context, and so does not disambiguate links.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +243,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+#### Failed Example 1
+
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They are visually distinguishable thanks to the relationships conveyed through CSS, but go to different resources.
+
+```html
+<html lang="en">
+	<p>
+		<h2>Contact us:</h2>
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=1" style="display:inline-block; background: url(/test-assets/shared/chat.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px;">Contact Us</a>
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=2" style="display:inline-block; background: url(/test-assets/shared/phone.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px; margin-left: 40px;">Contact Us</a>
```
WilcoFiers, line 270: These resolve to the same page with a different URL. That's not a failure of this rule. Same for other examples this PR adds.

test-assets/shared/chat.png
WilcoFiers, line 1: Where did these images come from? Are we free to use these? Please add credit for these images to the acknowledgement section.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +243,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
 </html>
 ```
 
+### Failed
+
+#### Failed Example 1
```
WilcoFiers, line 236: I'm reluctant to add this example TBH. This is first and foremost a 1.1.1 issue. Informative images are missing a text alternative. I think that this then results in a 2.4.4 failure is kind of incidental. In the same way that putting aria-hidden on the text of a link makes it a 1.3.1 issue, but that then also fits under 2.4.4.

This feels a bit more like a "gotcha" test case for tools than something that actually adds something important edge case. Even if we do keep the example I would prefer not to see it as the first one, as its far from the most obvious failure you can see get under this criterion.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -16,13 +16,18 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:1.1.1: # Non-text Content (A)
+    secondary: true
```
WilcoFiers: These need text now, not a boolean. Please check other rules for examples.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -229,80 +243,124 @@ These two HTML `a` elements have the same [accessible name][] and [context][prog
...
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=1" style="display:inline-block; background: url(/test-assets/shared/chat.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px;">Contact Us</a>
+		<a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=2" style="display:inline-block; background: url(/test-assets/shared/phone.png) 0 / 40px no-repeat; padding: 20px 0 20px 50px; margin-left: 40px;">Contact Us</a>
+	</p>
+</html>
+```
+
 #### Failed Example 2
 
-These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context]. They link to web pages that are similar, but have different information in their content.
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], but go to different resources. Their purpose is disambiguated for sighted users by the alignment of the links with the images above.
```
WilcoFiers, line 277: Kinda similar to the previous case, I feel this is primarily a reading order failure.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -398,8 +456,23 @@ These two HTML `a` elements have the same [accessible name][] but different [pro
...
+#### Inapplicable Example 7
+
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], but the second one is not [included in the accessibility tree][included in the accessibility tree].
+
+```html
+<html lang="en">
+	<p>
+		We are on social media:
+		<a href="https://act-rules.github.io/">ACT rules</a>
+		<a aria-hidden="true" href="https://www.w3.org/community/act-r/">ACT rules</a>
```
WilcoFiers, line 481: This fails the success criterion I think. I don't think we can put an example like this in the inapplicability section. What was the reason you wanted something like this in the rule?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -16,13 +16,18 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:1.1.1: # Non-text Content (A)
+    secondary: true
+  wcag20:1.3.1: # Info and Relationships (A)
```
WilcoFiers: I don't think I understand why you added 1.3.1 in here. Do any of the failed examples fail 1.3.1 in your opinion? I didn't think so.

WilcoFiers (dismissed):
I think the update to the applicability and expectation are good. I'm not keen on the new failed examples though. I think they mostly show failures of different success criteria. That failure does propagate to 2.4.4, but I think it would be better if our examples focus specifically on what makes 2.4.4 failures unique. That also takes away your need for secondary requirements here I suspect.

test-assets/shared/chat.png
giacomo-petri: Images have been created by UsableNet designers and I have the approval to publish them. Please, let me know if something else is needed.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: The purpose was to clarify that this situation does not fit within the scope of the rule, as elements excluded from the accessibility tree are not part of the applicability. I'm open to removing it if it causes any confusion.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: They are not the same resource.

This is the content of page 1:
screenshot showing chat with us content

and this is the content of page 2:
screenshot showing call us content

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: I don't totally agree.

If you fix the reading sequence it does not guarantee that these links will pass 2.4.4

```
<div>
    <img src="/test-assets/shared/chat.png" alt="Chat" style="width:50%;">
    <a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=60e950cff70bf1ec60a702086748ab4dec361514">Contact Us</a>
    <img src="/test-assets/shared/phone.png" alt="Phone" style="width:50%;">
    <a href="/test-assets/links-with-identical-names-serve-equivalent-purpose-b20e66/contact-us.html?page=c1d4e0f067462f4b28716f028d9213a25eb82f28">Contact Us</a>
</div>
```

The reading sequence is now correct, but it's still failing 2.4.4

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: I'm fine moving the example below, but I don't agree that the 2.4.4 is "incidental" here.

If we have an empty link due to child image with empty alt attribute such as:

`<a href="test"></a>`

the empty link is still failing 2.4.4. Of course it's "incidental" to the empty alt attribute and the image is failing 1.1.1, but both the issues are valid and should be triggered by tools.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: The main idea behind is that if two links have identical acc name and context they should, in most cases, appear ambiguous to users in general. 
But due to their "context" (not the programmatic one), that for example might include a wrong structure, background images that convey supplementary information, etc., they are not.
This "context" provides information and relationships through presentation. Consequently, I have included SC 1.3.1 here.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: I'm happy to make those changes, but the existing content is currently structured this way. 
In comparison to the current version (available at https://www.w3.org/WAI/standards-guidelines/act/rules/fd3a94/proposed/#assumptions), I've made two edits: 
- I removed the initial bullet point, which is no longer relevant given the modifications made
- and I introduced the last point to explicitly state that relying on reading the URL is not regarded as a viable method for distinguishing links.

HelenBurge (approved):

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -16,13 +16,18 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:1.1.1: # Non-text Content (A)
+    secondary: This success criterion is mapped as a Secondary requirement because the SC applies only to non-text content. When links have visual information as context, a failed outcome for this rule may result in SC 1.1.1 Non-text Content being not satisfied.
+  wcag20:1.3.1: # Info and Relationships (A)
+    secondary: This success criterion is mapped as a Secondary requirement because the SC applies to information and relationships conveyed through presentation. When links rely on visual cues for conveying information and/or relationships, and these cues are not programmatically determined or available in text, a failed outcome for this rule may result in SC 1.3.1 Info and Relationships being not satisfied.
```
WilcoFiers: This should follow the format for secondary requirements. Please have a look at the [design doc](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/design/rule-design.md#accessibility-requirements).

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
WilcoFiers: I don't think this needs to be an HTML page that changes based on the query string. Actually having two HTML files would be better. There is nothing in this rule that requires a tester to compare the query string of two resources. So having an example that checks that you do that feels to me like a disconnect between what we say needs to be tested, and how we in practice want people to test.

I don't think these examples are wrong, but they're not quite right either. They have a complexity to them which they either don't need, or which hint that we don't think this rule is tested in the way we say it is. Neither of which is quite right I think.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
WilcoFiers: I don't think this needs to have 1.3.1 / 1.1.1 in here any more than the [link has accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/c487ae/) rule does. Sure if 1.1.1 fails you might also fail 2.4.4 but it's entirely possible to for an image with a link in it to pass 1.1.1 and fail 2.4.4 and vice versa. That's not the case for something like an image button which always fails both 1.1.1 and 4.1.2, or something like a 2:1 text contrast which always fails both 1.4.3 and 1.4.5.

Rule of thumb here is if you can separate them, it's not a secondary requirement. There has to be some element for which its not possible to fail one without failing the other SC, or to pass one without passing the other.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
WilcoFiers: I don't mind having one or two examples of images in here. I don't think the examples should be exclusively about images. The far more common scenario is with just link text. I think the examples we had before did a better job of demonstrating that problem than the new examples do.

WilcoFiers (changes requested):

giacomo-petri:
I will promptly go through all the feedback. Regarding your specific point:

> I don't mind having one or two examples of images in here. I don't think the examples should be exclusively about images. The far more common scenario is with just link text. I think the examples we had before did a better job of demonstrating that problem than the new examples do.

It's worth noting that the earlier examples posed a challenge for users in general, which is an exception for the 2.4.4 criterion evaluation.

Honestly, I never seen an example of a couple of links with identical accessible names and context serving different purposes without being ambiguous to users in general, except for the ones where something visual (background, icon, image) is making clear the difference.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
WilcoFiers: Talked to @giacomo-petri. @Jym77 can we put this topic on the agenda of an upcoming meeting?

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
WilcoFiers: Discussed. These should be removed.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
WilcoFiers: Suggested failure examples without images:

```html
<p>W3C pages for ACT:</p>
<p><a href="//">ACT Rules</a></p>

<p>Community group for ACT:</p>
<p><a href="//Different">ACT Rules</a></p>
```

```html
<p>
  To reach us you can <a href="">Contact</a> us via e-mail, or <a href="">contact</a> us on Facebook.
</p>
```

giacomo-petri:
@WilcoFiers,

- updated "content of the page" with ACT "web page" definition;
- removed 1.1.1 and 1.3.1 from secondary requirements as they are incidental;
- added 2 failing examples for textual elements without including images.

Only query parameters topic is left out (pending agenda), then we should be ready to go for the call for review :D

colabottles (approved):

WilcoFiers (approved):

giacomo-petri:
Call for Review ends on February 23rd.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -62,10 +67,17 @@ There are no accessibility support issues known.
 
 There is a difference between two contexts being the _same_ and being _identical_. This rule specifically targets links within the _same_ context. The same context means exactly the same set of DOM nodes. Identical (but not the same) contexts might have a different set of DOM nodes, but those DOM nodes have equivalent content - such as text content, attribute values, and so on. This difference is similar to the difference in some programming languages between pointer equivalence and deep object equivalence. Links with identical name that are in identical (but not the same) contexts also fail [2.4.4 Link Purpose (In Context)][sc244]. However, defining "identical context" unambiguously has been deemed infeasible at this time, and so has been left out of this rule.
 
+Links that are [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general) are covered by the exception mentioned in Success Criterion 2.4.4 Link Purpose (In Context). If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links, which makes it more of a general user experience concern than an accessibility issue.
```
Jym77: **Chore (blocking):** The links need to be updated to WCAG 2.2 URLs.
```suggestion
Links that are [ambiguous to users in general](https://www.w3.org/TR/WCAG22/#dfn-ambiguous-to-users-in-general) are covered by the exception mentioned in Success Criterion 2.4.4 Link Purpose (In Context). If the links are ambiguous to users in general, users of assistive technologies are not at a disadvantage when viewing the links, which makes it more of a general user experience concern than an accessibility issue.
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -62,10 +67,17 @@ There are no accessibility support issues known.
...
+
+Pages with links that are not [ambiguous to users in general][], but are ambiguous to some users are likely to fail [success criterion 1.3.1 Info and Relationships][sc131] if the disambiguation information is conveyed through presentation, but not semantically. Moreover, when this information is non-text content, such a page is likely to fail [success criterion 1.1.1 Non-text Content][sc111]. Therefore, both these success criteria are secondary requirement for this rule.
+
 ### Bibliography
 
 - [Understanding Success Criterion 2.4.4: Link Purpose (In Context)](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html)
 - [HTML Specification - URL parsing](https://html.spec.whatwg.org/#resolving-urls)
+- [Ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general)
```
Jym77:
```suggestion
- [Ambiguous to users in general](https://www.w3.org/TR/WCAG22/#dfn-ambiguous-to-users-in-general)
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -62,10 +67,17 @@ There are no accessibility support issues known.
...
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html) (secondary requirement)
```
Jym77:
```suggestion
- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) (secondary requirement)
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -62,10 +67,17 @@ There are no accessibility support issues known.
...
+- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html) (secondary requirement)
```
Jym77:
```suggestion
- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) (secondary requirement)
```

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
```diff
@@ -207,11 +219,9 @@ These two SVG `a` and HTML `a` elements have the same [accessible name][], same
 </html>
 ```
 
-### Failed
-
-#### Failed Example 1
+#### Passed Example 9
 
-These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context] but go to different resources.
+These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], but resolve to different resources. However, the purpose of these links is [ambiguous to users in general](https://www.w3.org/TR/WCAG21/#dfn-ambiguous-to-users-in-general). Thus all readers are unsure about the destination and the person with a disability is not at any disadvantage.
```
Jym77:
```suggestion
These two HTML `a` elements have the same [accessible name][] and [context][programmatically determined link context], but resolve to different resources. However, the purpose of these links is [ambiguous to users in general](https://www.w3.org/TR/WCAG22/#dfn-ambiguous-to-users-in-general). Thus all readers are unsure about the destination and the person with a disability is not at any disadvantage.
```

Jym77 (changes requested):
Links to WCAG 2.1 need to be updated. This doesn't block the Call for Review but does block merging.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Per conversation with @WilcoFiers, 1.1.1 and 1.3.1 removed from secondary requirements.
Bibliography list items related to 1.1.1 and 1.3.1 removed completely.

_rules/links-with-identical-names-and-context-serve-equivalent-purpose-fd3a94.md
giacomo-petri: Per conversation with @WilcoFiers, 1.1.1 and 1.3.1 removed from secondary requirements.
Bibliography list items related to 1.1.1 and 1.3.1 removed completely.

Jym77 (approved):

giacomo-petri:
@Jym77 

this seems to be ready to merge.

Jym77:

PR #2010, opened 2023-01-18
Add low-level fails to AAA rules

WilcoFiers:
This is a partial revert of https://github.com/act-rules/act-rules.github.io/pull/1958/, now that secondary requirements have been added to AAA rules (#1986) we need to make sure that they always include an example of failing any lower-level success criterion. I went through all AAA rules, some of them have these examples. These two did not.

I'm listing this as a priority issue since for axe-core this impacts the consistency on Color Contrast (enhanced).

---

## How to Review And Approve

### DIFF
--- a/_rules/meta-refresh-no-delay-no-exception-bisz58.md
+++ b/_rules/meta-refresh-no-delay-no-exception-bisz58.md
@@ -96,7 +96,7 @@ The first valid `meta` element redirects immediately.
 ```html
 <head>
 	<meta http-equiv="refresh" content="0; https://w3.org" />
-	<meta http-equiv="refresh" content="72001; https://w3.org" />
+	<meta http-equiv="refresh" content="30; https://w3.org" />
 </head>
 ```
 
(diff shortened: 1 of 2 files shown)


carlosapaduarte (approved):

Jym77 (approved):
I'm not fully sure I understand how this impact the consistency, but there is certainly no problems with getting these example back now.

HelenBurge (approved):

WilcoFiers:
@Jym77 What this does is it lets implementors split the implementation of this rule between the level AA part, and the level AAA part. If you have two contrast procedures in your tool; The first fails things below 4.5:1, the second fails things between 4.5:1 and 7:1, then the combination of those two procedures should be consistent with the AAA rule. In practice that means that we need to trigger at least one AA procedure with a failed example, otherwise it isn't part of the mapping.

It's sort of like if the Button has name rule only had example of native buttons. If your implementation split tested native buttons separately from ARIA buttons, then the ARIA button rule wouldn't be mapped, even though it should be.

In our case we needed the AA color-contrast procedure to get triggered so we can include the cantTells. Those are intentionally not included on the AAA procedure to avoid duplicates.

catdro (approved):

kengdoj (approved):

Jym77:
> In our case we needed the AA color-contrast procedure to get triggered so we can include the cantTells. Those are intentionally not included on the AAA procedure to avoid duplicates.

Thanks, that's the bit I missed. Makes perfect sense 😄

tombrunet (approved):

colabottles (approved):

PR #2012, opened 2023-01-18
ARIA state or property: various editorial improvements

### LINKED ISSUE #1849: Should ARIA IDREFs existing be required? [6a7281]
For the [ARIA state or property has valid value](https://www.w3.org/WAI/standards-guidelines/act/rules/6a7281/proposed/) rule, the expectation says: 

> For value types ID Reference and ID Reference List for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.1/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or in the same [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.

This is then tested with failed example 4. This is actually a pretty good example of why we shouldn't strictly require these elements to exist. Take for example the following component (in Vue):

```html
<p v-if="error" id="error">{{ error }}<p>
<label>
  name <input aria-errormessage="error" aria-invalid="{{ error ? true : false }}">
</label>
```

In this scenario, while there is an error, aria-errormessage works. When there is no error, the `p` element is not in the DOM, and so `aria-errormessage` does nothing. I think this is a legitimate way of building such a component, one that I've seen more than a few times. I think we should get rid of this requirement. Either that or be more selective about which attributes do, and which co not require IDREF to exist. The most important one is use of aria-labelledby, but if those reference don't work we have rules that catch problems with empty accessible names.

WilcoFiers:
This is addressing feedback from the recent ACT TF survey. I've also updated several examples, and filled in a few gaps.

Closes #1849

Need for Call for Review: 1 week, since I'm updating a number of examples

This PR is closely related to #2041, which introduces a new rule that ensures required ID references exist in the page. We're basically splitting that part of this rule off into its own.

---

## How to Review And Approve


_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -28,31 +28,27 @@ acknowledgments:
...
 
 ## Expectation
 
-Each test target has a valid value according to its [WAI-ARIA 1.2 value type](https://www.w3.org/TR/wai-aria-1.2/#propcharacteristic_value).
+Each test target has an [attribute value][] that is valid according to its [WAI-ARIA value type][value type].
 
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.2/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or in the same [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
-
-For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+**Exception**: For `ID Reference` and `ID Reference List` value types, if the test target is not a [WAI-ARIA required states and properties][] for the [semantic role][] of its element, no ID referenced elements are required. Otherwise at least one of the elements with the given IDs exists in the same [document tree][] or in the same [shadow tree][] as the test target's element.
```
Jym77: To fully close #1849, I think we need to consider `aria-controls` as not required on closed `combobox` (and thus include it in this exception).
This matches the `combobox` authoring practice, and follows suit with changes made on ARIA 1.3 on `combobox`. And this is what we discussed on [June 23rd](https://www.w3.org/2022/06/23-act-r-minutes.html#item06).

From the [authoring practices](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#wai-aria-roles-states-and-properties-6) (second bullet):
> Note that `aria-controls` only needs to be set when the popup is visible.

ARIA Issue and PR: https://github.com/w3c/aria/issues/1334 / https://github.com/w3c/aria/pull/1335

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -28,31 +28,27 @@ acknowledgments:
...
-
-Only for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.2/#requiredState) with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement. For example, the value of the `aria-errormessage` attribute on an `input` does not need to reference an `id` that exists within the same document, because an [HTML element](https://html.spec.whatwg.org/#htmlelement) with such and `id` may be created in response to an [event](https://dom.spec.whatwg.org/#event) that may or may not happen.
+Using invalid ARIA attribute values is often the result of a typo or other developer error. These attributes are then either ignored, or a default value is assumed by browsers and assistive technologies. This often means that a state or property which should exist is missing or has an unexpected value. This can cause issues under [success criterion 1.3.1 Info and Relationships][sc131] or [4.1.2 Name, Rule Value][sc412]. If the default value for invalid attribute values happens to match the author's intention for the value, there will not be an accessibility issue.
```
Jym77, line 49: That almost sounds like 1.3.1 and 4.1.2 should be secondary requirements 😄

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -113,43 +117,41 @@ For value type `URI`, this rule does not require that the destination URI exists
 <div id="item2">Bananas</div>
 ```
 
-#### Passed Example 6
+#### Passed Example 7
 
-`aria-rowindex` property with valid integer value
+The `aria-rowindex` [attribute value][] is a valid `integer` [value type][].
```
Jym77:
```suggestion
The `aria-rowindex` [attribute value][] of 2 is a valid `integer` [value type][].
```
matching the descriptions on previous examples.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -167,78 +169,62 @@ For value type `URI`, this rule does not require that the destination URI exists
...
 
 ```html
-<div role="gridcell" aria-rowindex="2.5">Fred</div>
+<div role="spinbutton" aria-valuemin="one" aria-valuemax="three" aria-valuenow="two" aria-label="Choose a value"></div>
 ```
 
 #### Failed Example 6
 
-`aria-valuemin`, `aria-valuemax` and `aria-valuenow` property with invalid number values
+The `aria-live` [attribute value][] of `page` is not a valid `token`, because `page` is a token for `aria-live`.
```
Jym77:
```suggestion
The `aria-live` [attribute value][] of `page` is not a valid `token`, because `page` is not a token for `aria-live`.
```
(maybe also add in the list of valid tokens for `aria-page` 🤔 )

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -167,78 +169,62 @@ For value type `URI`, this rule does not require that the destination URI exists
...
 
 ```html
-<div role="spinbutton" aria-valuemin="one" aria-valuemax="three" aria-valuenow="two" aria-label="Choose a value"></div>
+<div role="main" aria-live="page"></div>
 ```
 
 #### Failed Example 7
 
-`aria-live` property with invalid token value
+The `aria-relevant` [attribute value][] has the `text` and `always` tokens. The `always` token is not valid for the `aria-relevant` `token list`.
```
Jym77:
```suggestion
The `aria-relevant` [attribute value][] has the two tokens `text` and `always`. The `always` token is not valid for the `aria-relevant` `token list`. In order to be a valid value, all tokens must be valid.
```

Jym77 (dismissed):
The whole `aria-controls` on `combobox` may be better taken in a separate PR, but I think this PR won't close #1849 until it's done.
And it may also require us to have the same exception on "Element with role attribute has required states and properties".

I did tried to give it a shot before seeing this PR 😅 : #2014, maybe things are salvageable from it.

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: That's a fair point!

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: That's what I was going for! They are secondary requirements after all. 😉

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: aria-pabe? This is a token of aria-current, But I think this is enough.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -72,39 +70,47 @@ For value type `URI`, this rule does not require that the destination URI exists
...
 ```
 
 #### Passed Example 4
 
-`aria-errormessage` property with valid ID reference value
+The `aria-pressed` [attribute value][] of `mixed` is allowed for the `tristate` [value type][].
 
 ```html
-<div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div>
+<div role="checkbox" aria-pressed="mixed">Partially checked checkbox</div>
```
tombrunet: Did you mean to use `aria-checked` here or `role="button"`?  `aria-pressed` isn't valid for `role="checkbox"`

tombrunet (dismissed):

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: Yes! That should have been a button. Copy/paste error I think.

```suggestion
<div role="button" aria-pressed="mixed">Partially pressed button</div>
```

colabottles (approved):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -272,7 +282,7 @@ Element has ARIA role, but no ARIA states or properties
 
 #### Inapplicable Example 3
 
-`aria-checked` state with empty value
+The `aria-checked` attribute does not have a value.
```
tombrunet: Not so much a question about this specific change, but I received a question from my team about the example as a whole. ARIA spec notes for aria-checked:
> undefined (default): The element does not support being checked.

The example to "Accept terms" that doesn't support being checked is strange. Does it not support being checked because it's currently disabled? Adding `aria-disabled="true"` would cause this rule to pass.

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -28,31 +28,29 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [WAI-ARIA 1.2 state or property](https://www.w3.org/TR/wai-aria-1.2/#state_prop_def) that is not empty (""), and that is specified on an [HTML or SVG element][].
+This rule applies to any [WAI-ARIA state or property][] that has a non-empty ("") [attribute value][], and that is specified on an [HTML or SVG element][].
 
 ## Expectation
 
-Each test target has a valid value according to its [WAI-ARIA 1.2 value type](https://www.w3.org/TR/wai-aria-1.2/#propcharacteristic_value).
+Each test target has an [attribute value][] that is valid according to its [WAI-ARIA value type][value type], except if one of the following is true:
```
carlosapaduarte: The following are descriptions, not conditions that can be evaluated to true or false.

```suggestion
Each test target has an [attribute value][] that is valid according to its [WAI-ARIA value type][value type], except in the following instances:
```

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -113,43 +119,41 @@ For value type `URI`, this rule does not require that the destination URI exists
...
 
 ```html
 <div role="gridcell" aria-rowindex="2">Fred</div>
 ```
 
-#### Passed Example 7
+#### Passed Example 8
 
-`aria-valuemin`, `aria-valuemax` and `aria-valuenow` properties with valid number values
+The `aria-valuemin`, `aria-valuemax` and `aria-valuenow` [attribute values][attribute value] are a valid for a `number` [value type][].
```
carlosapaduarte:
```suggestion
The `aria-valuemin`, `aria-valuemax` and `aria-valuenow` [attribute values][attribute value] are valid for a `number` [value type][].
```

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -113,43 +119,41 @@ For value type `URI`, this rule does not require that the destination URI exists
...
-	MM-DD-YYYY
-</div>
+<a href="/" aria-current="page">Home</a>
 ```
 
-#### Passed Example 9
+#### Passed Example 10
 
-`aria-relevant` property with valid token list values
+The `aria-relevant` [attribute value][] has a `text` and `removals`. Both are valid tokens for a `aria-relevant` `token list` [value type][].
```
carlosapaduarte:
```suggestion
The `aria-relevant` [attribute value][] has a `text` and `removals` tokens. Both are valid tokens for a `aria-relevant` `token list` [value type][].
```

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -163,82 +167,77 @@ For value type `URI`, this rule does not require that the destination URI exists
...
 
 ```html
-<div role="button" aria-expanded="mixed">A button</div>
+<div role="button" aria-expanded="collapsed">A button</div>
 ```
 
 #### Failed Example 3
 
-`aria-pressed` state with invalid tristate value
+The `aria-expanded` [attribute value][] of `horizontal` is not valid for a `tristate` [value type][].
```
carlosapaduarte: The example has aria-pressed, instead of aria-expanded

carlosapaduarte (changes requested):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -28,31 +28,29 @@ acknowledgments:
...
 
-For value types `ID Reference` and `ID Reference List` for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.2/#requiredState) at least one of the elements with the given ids exists in the same [document tree](https://dom.spec.whatwg.org/#document-trees) or in the same [shadow tree](https://dom.spec.whatwg.org/#shadow-trees) as the element that specifies the target attribute.
+- <dfn id="off6ek:id-reference">ID Reference</dfn>: For `ID Reference` value types an ID referenced elements is only required for `aria-controls` with [semantic][semantic role] `scrollbar` elements, and with [semantic][semantic role] `combobox` elements that have an `aria-expanded` [attribute value][] of `true`. The ID referenced element must exist in the same [document tree][] or [shadow tree][] as the test target's element.
 
-For value type `URI` the value matches the [generic URI syntax](https://www.ietf.org/rfc/rfc3986.txt).
+- <dfn id="off6ek:id-reference-list">ID Reference List</dfn>: For `ID Reference List` value types, no ID referenced elements are required.
```
Jym77: I feel the exceptions are getting somewhat out of hand and hard to read. Trying something in a different direction to see where it goes…

> For each test target, depending on its type, either:
> - the type is neither ID reference, nor ID reference list, and the value is valid; or
> - the type is ID reference, and the attribute is required, and this is not `aria-controls` on a `combobox` with `aria-expanded != true`, and the value is an existing `id`; or 
> - all other cases pass.

(plus adding Background note that currently no ID reference list is required, and that apart from combobox, the only required ID reference is `aria-controls` on a `scrollbar`)

I prefer listing the combobox as an exception as it makes the rule more future proof (if a new required ID ref comes up, we don't need to change the rule)

---
Maybe it can be make clearer by splitting the requirements:

> Expectation 1
> For each test target, either the type is ID reference/ID reference list; or the value is valid.
>
> Expectation 2
> For each test target, either the the type is not ID reference, or it is not required, or it is `aria-controls` on an closed `combobox`, or the value is an existing `id`.

(again with Background note to explain the reasoning and effect)
This one might makes it harder to see when Expectation 2 triggers since it gets all the negation 🤔

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -28,31 +28,29 @@ acknowledgments:
...
 
-Only for [WAI-ARIA required properties](https://www.w3.org/TR/wai-aria-1.2/#requiredState) with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given ids actually exists. For non-required properties, this is not a requirement. For example, the value of the `aria-errormessage` attribute on an `input` does not need to reference an `id` that exists within the same document, because an [HTML element](https://html.spec.whatwg.org/#htmlelement) with such and `id` may be created in response to an [event](https://dom.spec.whatwg.org/#event) that may or may not happen.
-
-For value type `URI`, this rule does not require that the destination URI exists.
+Only for [WAI-ARIA required properties][] with value types `ID Reference` and `ID Reference List` is there a requirement that the elements with the given IDs actually exists. For non-required properties, having the referenced element is optional. For example, `aria-errormessage` attribute on an `input` element may have a fixed value, but the element with the error message is only added to the page when an error actually occurred.
```
Jym77: We should expand on that to explain the `combobox` exception.

Jym77 (changes requested):
I think that works.
Exploring a bit different ways of phrasing it, but the current version is good.

colabottles (approved):

carlosapaduarte:
From the [CG call](https://www.w3.org/2023/02/09-act-r-minutes.html#t05): remove the exceptions from this rule and create a second rule to address the exceptions (in the first rule include a background note stating the second rule is under development)

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: I'm just going to use a different attribute here I think. You're right, an `undefined` aria-checked prop on a checkbox is odd.

colabottles (approved):

carlosapaduarte (approved):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -72,184 +68,143 @@ For value type `URI`, this rule does not require that the destination URI exists
...
 
 ```html
-<div role="textbox" aria-errormessage="my-error" aria-label="A textbox"></div>
+<div role="button" aria-pressed="mixed">Partially pressed button</div>
 ```
 
 #### Passed Example 5
 
-`aria-owns` property with valid ID reference list value
+The `aria-errormessage` [attribute value][] is an `ID Reference` [value type][]. Because `aria-errormessage` is not a required property for `textbox`, an element with a matching ID is not required.
```
Jym77:
```suggestion
The `aria-errormessage` [attribute value][] is an `ID Reference` [value type][]. The presence of an element with a matching ID is not required by this rule.
```
The current version of the rule doesn't mention the fact that the property is required or not, so it shouldn't really be part of the description.

Jym77 (dismissed):
(requesting change for one description update)

---
I'm not a huge fan of splitting the rule. I feel the other rule could more or less just be an Expectation 2 in this one. I won't block anything on that account, but it feels we are diluting things a bit too much.

_rules/aria-state-or-property-valid-value-6a7281.md
colabottles: I agree with @Jym77 in that this feels like we're diluting things but I wouldn't block this either.

WilcoFiers:
@Jym77 The reason to split this is because that second expectation got unnecessarily complicated. A new rule lets us keep things simple. It also fits better with the requirement to write atomic rules. There are other rules for which we may do things like this too. For example an allowed ARIA attribute is different from a prohibited attribute, from a deprecated attribute.

Jym77 (approved):

PR #2022, opened 2023-01-31
Create focus-returns-trigger-9au0ou.md

HelenBurge:
First draft of the 2.4.3 focus returns to trigger rule.

Adding a rule for testing the focus returning to the trigger tests.

Closes issue(s):

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,144 @@
...
+  previous_authors:
+---
+
+## Applicability
+
+This rule applies to any [modal][] that when closed the [focus][] returns to the [trigger][].
+
+## Expectation
+
+On closing the [modal][], the assistive technology [focus][] will return to the item in the owning document that triggered the [modal][].
```
Jym77: I see two difficulties here:
- we need to tie the "open" button/link/… with the modal, which we can't really do if we target the modal (there is little way to know how it was opened);
- this sounds like a process rather than a static check, that's OK but that is also something we do not have in rules so far (they check a static snapshot of the page).

---
My suggestion would be to have an Applicability on the trigger, which with our definition could be "an [instrument](https://act-rules.github.io/glossary/#instrument-to-achieve-an-objective) to open a modal", and an Expectation in the line of "after opening + closing, the focus is back on the trigger that was used for opening".
(that would also catch cases where several buttons can open the same modal but focus always go back to the same, rather than the one that was used)

We still need to be careful about how we phrase the process, since it is something we do not have currently in our rules…

I think we can also define "modal" as "role of `dialog` + `aria-modal="true"`" to get a less ambiguous definition 🤔 And "open" would be something about making it visible.

HelenBurge: All good points, and yes - I need to work on it more 👍 

I like your suggestions. I will work on this in the next few days.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+The button that is activated to open the [modal][] gets the [focus][] returned to it when the [modal][] is dismissed using the close button in the modal.
+
+```html
```
dan-tripp-siteimprove, line 58: I've made some progress on some code for these examples.  I think I can't fit it into this file.  There's too much code.  So I'll put it into separate files.  Other ACT rules do this, so it will work in the long term.  I think that I'll need to do some github trickery and push to your topic branch, Helen.  Wish me luck.

dan-tripp-siteimprove:
Okay, I pushed a "Passed Example 1".  I'd like to know what you think of it, before I create other examples.  The problem is that it will be hard for you to test it.   You can't just copy and paste that HTML.  I think that your options are: 

1. You can copy the multiple .css and .js files to your local machine and then edit some paths.  (If you want to pursue this option, let me know.)
3. Wait until it's built and released on the ACT rules site (i.e. https://www.w3.org/WAI/standards-guidelines/act/rules/)   (this option seems like the worst one to me.)
4. Discover some other method of viewing the built HTML before it gets published on the ACT rules site.    (I don't know if this option exists.  Maybe Jean-Yves or Wilco would.)

dan-tripp-siteimprove:
I put ["Passed Example 1" on a live test page](https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-1.html).   What do you think?

HelenBurge:
Love it Dan! Thanks for all of your hard work :)

dan-tripp-siteimprove:
I've been thinking about the examples that use a link (instead of a button) to open the modal.  (i.e. passed examples 4, 5 and 6, and failed examples 3 and 4.)   Is it best to include these?  I gather that using a link in this way is a bad practice - discussed here, for example: [https://stackoverflow.com/questions/38594369/is-it-more-accessible-to-use-a-button-or-a-to-open-close-a-modal ](url).  Even if people do it, we wouldn't want to encourage it, right?

dan-tripp-siteimprove:
At any rate, I've updated and added some more examples:
[https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-1.html](https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-1.html)
[https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-2.html](https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-2.html)
[https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-3.html](https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/passed-example-3.html)
[https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/failed-example-1.html](https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/failed-example-1.html)
[https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/failed-example-2.html](https://dan-tripp-siteimprove.github.io/act-cg-focus-returns-trigger-9au0ou-examples/failed-example-2.html)

HelenBurge:
> I've been thinking about the examples that use a link (instead of a button) to open the modal. (i.e. passed examples 4, 5 and 6, and failed examples 3 and 4.) Is it best to include these? I gather that using a link in this way is a bad practice - discussed here, for example: [https://stackoverflow.com/questions/38594369/is-it-more-accessible-to-use-a-button-or-a-to-open-close-a-modal ](url). Even if people do it, we wouldn't want to encourage it, right?

I agree, but best to allow it as we do not say how to code, but how to test and it is possible to have links that activate modals :(

dan-tripp-siteimprove (commented):
I like these changes so far, but I think that before they're merged into the main ACT repo we need to resolve all of the 'code needed' and other 'to do' items.  Because they're not fit for public consumption.  At least, I think that's the protocol for these PRs to ACT - no PR until it's fit for public consumption.  I'll let more senior members than myself comment on that.

dan-tripp-siteimprove:
> > I've been thinking about the examples that use a link (instead of a button) to open the modal. (i.e. passed examples 4, 5 and 6, and failed examples 3 and 4.) Is it best to include these? I gather that using a link in this way is a bad practice - discussed here, for example: [https://stackoverflow.com/questions/38594369/is-it-more-accessible-to-use-a-button-or-a-to-open-close-a-modal ](url). Even if people do it, we wouldn't want to encourage it, right?
> 
> I agree, but best to allow it as we do not say how to code, but how to test and it is possible to have links that activate modals :(

Fair.  I'll try to do that later in May.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+  - DOM Tree
+acknowledgments:
+  authors:
+    -  Helen Burge
+  previous_authors:
+---
+
+## Applicability
+
+This rule applies to any [modal][] or item with the role of `dialog` + `aria-modal="true"` that when no longer visible, the [focus][] returns to the [instrument][] or [trigger][] that made the [modal][] visible.
```
Jym77: Some issues I see with this:
* This seems to mix the Applicability and intention of the rule. Especially, the "when no longer visible, the focus returns to…" part is what the rule actually checks, not what it applies to.
* I'm not sure we can really use "modal" from the ARIA authoring practices as a definition here. OTOH, we should be good by looking at `dialog` elements + `role="modal"` ones.
* By targeting an already opened modal, we do not really know what was the instrument used to open it. I.e. if there are two buttons to open the modal, but the test starts when the modal is already opened, how can we check that we return to the correct one. I think the Applicability should either target the instrument to open a modal, or the combination of the open modal plus the instrument that opened it. Targeting the instrument seems easier (i.e. it means that opening the modal is part of the test).
* I'm not sure anymore where we are with the full "test instructions" bit (i.e. this rule requires some precise manipulations). I think this is part of the "stateful rules" discussion. Let's just assume the problem will be solved elsewhere and not care about it here 😄 We may have to adapt the rule according to the decision reached, however.
* We can probably use the "open/close" modal vocabulary only (i.e.  not mention visibility, or enter/exit). This is linked to either `aria-modal` or the `showModal` status in HTML sepcs, so we have both the technical solid ground fnor it, and easy enough vocabulary to understand.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Expectation
+
+On exiting the [modal][] making it no longer visible, the assistive technology [focus][] will return to the item in the owning document that was used to make the [modal][] visible.
```
Jym77: I think we do not need to refer to AT, maybe somethnig like:
```suggestion
After the [modal][] is closed, the [instrument][] that was used to open it becomes [focused][].
```

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+Code needed!
+```
+
+#### Passed Example 3
+
+The button that is activated to display the [modal][] gets the [focus][] returned to it when the [modal][] is dismissed using the cancel button.
+
+```html
+Code needed!
+```
```
Jym77: I think that closing the modal with a "close" or "cancel" button is essentially the same thing and doesn't require separate examples.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Passed Example 6
+
+The link that is activated to display the [modal][] gets the [focus][] returned to it when the [modal][] is dismissed using the cancel button.
+
+```html
+Code needed!
+```
```
Jym77: Same remark about close/cancel buttons.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Passed Example 7
+The page has a list of buttons that is activated to display a [modal][], the [focus][] is returned to the correct button on dismissing the modal.
+
+```html
+Code needed!
+```
+
```
Jym77, line 127: I think we should also have an example where the modal has several ways to be closed (buttons, link, keyboard) and all work.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+#### Passed Example 6
+
+The link that is activated to display the [modal][] gets the [focus][] returned to it when the [modal][] is dismissed using the cancel button.
+
+```html
+Code needed!
+```
+
+#### Passed Example 7
+The page has a list of buttons that is activated to display a [modal][], the [focus][] is returned to the correct button on dismissing the modal.
```
Jym77: This could also be a mix of buttons and links to open the modal.

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+#### Passed Example 3
+
+The button that is activated to display the [modal][] gets the [focus][] returned to it when the [modal][] is dismissed using the cancel button.
+
+```html
+Code needed!
+```
+
```
Jym77: Should we have an example where the modal is closed by a link rather than a button (also a not good practice example, but we want to illustrate that any way to close it is taken in consideration).

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,188 @@
...
+#### Failed Example 5
+The page has a list of buttons that is activated to display a [modal][], the [focus][] is returned to the incorrect button on dismissing the modal.
+
+```html
+Code needed!
+```
+
```
Jym77, line 234: I think we need here an example with several ways to close the modal, and one of them returns the focus but not the other (this illustrate that **all** ways to close the modal are tested).

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,245 @@
...
+  - DOM Tree
+acknowledgments:
+  authors:
+    -  Dan Tripp
+    -  Helen Burge
+---
+
+## Applicability
+
+This rule applies to any element with a semantic role of dialog or alertdialog that is not visible.
```
kengdoj: is "modal" needed?

```suggestion
This rule applies to any element with a semantic role of modal dialog or alertdialog that is not visible.
```

_rules/focus-returns-trigger-9au0ou.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Expectation
+
+The element that has the focus when the target element becomes visible is the same element that receives focus when the target element stops being visible.
```
kengdoj:
```suggestion
The element that triggers the target element to become visible is the same element that receives focus when the target element stops being visible.
```

giacomo-petri:
Hi @HelenBurge,

I've reviewed this rule and believe it's a bit too strict. We may need to tweak it slightly:
- Applicability:
    - the phrase "role of modal dialog" isn’t quite accurate. We typically assign the `role="dialog"` with a modal experience (`aria-modal="true"`), but perhaps a better definition would be when focus is trapped within a region activated by a trigger control that is dismissible
    - the rule seems to apply more to the element that triggers a temporarily trapped keyboard focus in a modal experience, rather than the element that "is not visible"
- Expectation: the current expectation may be too strict. Some components, once activated, modify background content in ways that prevent focus from returning to the original trigger element. Examples include:
    - an alert dialog that reloads the page
    - a confirmation popup that dynamically updates content via ajax
    - ...
    
    We might need to define two separate expectations or clarify certain assumptions to account for these scenarios.
- Accessibility Support: while this behavior has recently improved, some assistive technologies on known browsers still exhibit inconsistencies. Specifically, if an element is removed from the accessibility tree after being activated (e.g., a close button), and focus is moved via JavaScript to an element that isn’t inherently operable (e.g., a `button` works, but a `div` with `tabindex="-1"` may not), the AT might automatically shift focus to the closest visible element, regardless of the intended JavaScript behavior.

HelenBurge:
> Hi @HelenBurge,
> 
> I've reviewed this rule and believe it's a bit too strict. We may need to tweak it slightly:
> 
> * Applicability:
>   
>   * the phrase "role of modal dialog" isn’t quite accurate. We typically assign the `role="dialog"` with a modal experience (`aria-modal="true"`), but perhaps a better definition would be when focus is trapped within a region activated by a trigger control that is dismissible
>   * the rule seems to apply more to the element that triggers a temporarily trapped keyboard focus in a modal experience, rather than the element that "is not visible"
> * Expectation: the current expectation may be too strict. Some components, once activated, modify background content in ways that prevent focus from returning to the original trigger element. Examples include:
>   
>   * an alert dialog that reloads the page
>   * a confirmation popup that dynamically updates content via ajax
>   * ...
>   
>   We might need to define two separate expectations or clarify certain assumptions to account for these scenarios.
> * Accessibility Support: while this behavior has recently improved, some assistive technologies on known browsers still exhibit inconsistencies. Specifically, if an element is removed from the accessibility tree after being activated (e.g., a close button), and focus is moved via JavaScript to an element that isn’t inherently operable (e.g., a `button` works, but a `div` with `tabindex="-1"` may not), the AT might automatically shift focus to the closest visible element, regardless of the intended JavaScript behavior.

Hi @giacomo-petri I have agreed and updated the main file with your feedback.

HelenBurge:
@dan-tripp-siteimprove - please add some more examples for the rule.

dan-tripp-siteimprove:
My status:

Where I'm currently stuck (and have been for the last six months) is getting the netlify preview to output examples in a "useful" way  "Useful" here means:
a) the rule page links to each example's "live" version (AKA it's "in a new tab" version) _also on the netlify preview_ (not w3.org), and 
b) the "live" version works in terms of importing CSS, JS, etc. correctly - i.e. from the netlify preview (not w3.org).

I consider the above work a prerequisite to adding more examples, because the alternative - i.e. writing examples without being able to run them - would be quite messy in my opinion.

I have some progress on the above work, uncommited, on my hard drive.  Specifically I have solved (a) and some of (b).

Jym77:
> My status:
> 
> Where I'm currently stuck (and have been for the last six months) is getting the netlify preview to output examples in a "useful" way "Useful" here means: a) the rule page links to each example's "live" version (AKA it's "in a new tab" version) _also on the netlify preview_ (not w3.org), and b) the "live" version works in terms of importing CSS, JS, etc. correctly - i.e. from the netlify preview (not w3.org).
> 
> I consider the above work a prerequisite to adding more examples, because the alternative - i.e. writing examples without being able to run them - would be quite messy in my opinion.
> 
> I have some progress on the above work, uncommited, on my hard drive. Specifically I have solved (a) and some of (b).

Ah, yes, it seems that the examples pages link to `https://www.w3.org/WAI/content-assets/wcag-act-rules` instead of the netlify previews 🤔 (and also that the preview for this PR is down, probably too old?)
(e.g. preview https://deploy-preview-2387--act-rules.netlify.app/standards-guidelines/act/rules/73f2c2/ has its "open in new tab" links next to each example lead to https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/73f2c2/eabc191efa65e6613739042a0ae21937cda02428.html (for Passed Example 1))

@remibetin is this something you can look into and see if the examples can also be hosted on the previews?

dan-tripp-siteimprove:
To flesh out the website problem a bit:
1. Ideally we would like this to work on a local website build too.  Not just netlify. 
2. Naturally we want it to work for all rules - "proposed", "not proposed", "other" (?).   Rule oj04fd is a bit of a curve ball, on my hard drive anyway - it seems to exist in both a "proposed" and "not proposed" version.  That's over my head.

PR #2034, opened 2023-02-16
[cae760] Frame has non-empty accessible name

### LINKED ISSUE #1170: What does it actually mean to be part of sequential focus navigation? (cae760)

### LINKED ISSUE #1367: iframe has non-empty name - problems with examples and applicability [cae760]

tombrunet:
Adjustments to the rule based on conversation of task force conversations (Feb 16, 2023).

Fixes #1367
Fixes #1170

Note regarding the two above issues. There is no perfect answer here since the browser behaviors are all different. However, the approach in this PR was to ignore cases where it's clear that the author attempts to remove the iframe from the sequential focus navigation.

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


carlosapaduarte (approved):

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,11 +41,13 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
 
 ## Accessibility Support
 
-- Some browsers include `iframe` elements in the [sequential focus navigation][]. This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard. When an `iframe` is removed from the accessibility tree, this rule is still applicable for those browsers, unless the `iframe` is explicitly removed from [sequential focus navigation][] (by having the `tabindex` attribute set to a negative value).
-
 - Browser and assistive technology support for `iframe` elements is currently **inconsistent**. Some examples of inconsistencies include (but are not limited to):
   - There is a known combination of a popular browser and assistive technology that ignores `aria-label` and only announces `title` attribute as an [accessible name][]
   - Some assistive technologies ignore empty `iframe` elements, regardless of if they are focusable or if they have an accessible name.
+  - Some browsers do not include `iframe` elements in the [sequential focus navigation][] if content of the `iframe` is in the [sequential focus navigation][], even if the `iframe` has a non-negative `tabindex` [attribute value][].
```
WilcoFiers:
```suggestion
  - Some browsers instantly redirect focus from `iframe` elements  to the first focusable element inside that iframe. This redirect makes it appear as though the `iframe` never receives focus. This occurs even if the `iframe` has a non-negative `tabindex` [attribute value][].
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,11 +41,13 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
+  - Some browsers include `iframe` elements in the [sequential focus navigation][] unless the `iframe` is explicitly removed from [sequential focus navigation][] (by having the `tabindex` attribute set to a negative value). This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard.
```
WilcoFiers: I don't think we should mention this

```suggestion
  - Not all browsers redirect focus on `iframe` elements. This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard. This must not be circumvented by using a negative tabindex, as this will make the `iframe` completely inaccessible for keyboard navigation.
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -17,13 +17,17 @@ acknowledgments:
...
     - WAI-Tools
 ---
 
 ## Applicability
 
-This rule applies to `iframe` elements that are [included in the accessibility tree][] and do not have a negative `tabindex` [attribute value][].
+This rule applies to `iframe` elements that are [included in the accessibility tree][] and for which all of the following are true:
+- does not have a negative `tabindex` [attribute value][]; and
+- does not have a `role` [attribute value][] of `presentation`; and
+- does not have a `role` [attribute value][] of `none`.
```
WilcoFiers: I would suggest this:

```suggestion
- the `iframe` does not have a negative `tabindex` [attribute value][]; and
- the `iframe` does not have an [explicit semantic role][] of `presentation` or `none`.
```

Restated to follow the conditional list writing style, and then using explicit semantic role, so we don't require an exact attribute value match, to account for whitespace and fallback roles.

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -17,13 +17,17 @@ acknowledgments:
...
 
 **Note:** `frame` element is deprecated, this rule does not consider `frame` or `frameset` elements.
```
WilcoFiers: Would you mind moving this note to the background? That feels like a better place for it.

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,11 +41,13 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
   - There is a known combination of a popular browser and assistive technology that ignores `aria-label` and only announces `title` attribute as an [accessible name][]
   - Some assistive technologies ignore empty `iframe` elements, regardless of if they are focusable or if they have an accessible name.
+  - Some browsers do not include `iframe` elements in the [sequential focus navigation][] if content of the `iframe` is in the [sequential focus navigation][], even if the `iframe` has a non-negative `tabindex` [attribute value][].
+  - Some browsers include `iframe` elements in the [sequential focus navigation][] unless the `iframe` is explicitly removed from [sequential focus navigation][] (by having the `tabindex` attribute set to a negative value). This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard.
+
+Due to these inconsistencies, the applicability of this rule is scoped for `iframe` elements for which there is no indication of an attempt to remove them from accessibility consideration.
```
WilcoFiers:
```suggestion
Due to these inconsistencies, the applicability of this rule is scoped for `iframe` elements for which there is no indication of an attempt to remove them from accessibility consideration.  Whether `iframe` elements that are inapplicable to this rule still require an accessible name varies between browsers.
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -142,6 +146,14 @@ This `iframe` element has a negative `tabindex` [attribute value][].
 <iframe tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
 ```
 
+#### Inapplicable Example 4
+
+This `iframe` element has a `role` [attribute value][] of `none`.
+
+```html
+<iframe title=" " src="/test-assets/SC4-1-2-frame-doc.html" role="none"> </iframe>
```
WilcoFiers: This feels like a slightly better test of this. This could be passed because the title attribute is used.

```suggestion
<iframe src="/test-assets/SC4-1-2-frame-doc.html" role="none"> </iframe>
```

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -142,6 +146,14 @@ This `iframe` element has a negative `tabindex` [attribute value][].
 <iframe tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
```
WilcoFiers: Hmm, this one needs a height so that it isn't scrollable. 

https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/cae760/803b52f32a5d2193683f152aa5e00174d5f8416a.html

WilcoFiers (dismissed):

Jym77 (changes requested):
Agreeing with Wilco's comment.
It also seems that you need to add yourself as an contributor, iirc this is done by modifying [package.json](https://github.com/act-rules/act-rules.github.io/blob/develop/package.json). I suggest you do that bit in a separate PR (can be merged with one Approval), since it shouldn't be blocked by this PR (in case it takes longer than expected).

_rules/iframe-non-empty-accessible-name-cae760.md
tombrunet: Done

_rules/iframe-non-empty-accessible-name-cae760.md
tombrunet: Added

tombrunet:
@Jym77 https://github.com/act-rules/act-rules.github.io/pull/2048 for adding contributor

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,14 +38,18 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
   - Some assistive technologies ignore empty `iframe` elements, regardless of if they are focusable or if they have an accessible name.
+  - Some browsers instantly redirect focus from `iframe` elements  to the first focusable element inside that iframe. This redirect makes it appear as though the `iframe` never receives focus. This occurs even if the `iframe` has a non-negative `tabindex` [attribute value][].
+  - Not all browsers redirect focus on `iframe` elements. This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard. This must not be circumvented by using a negative tabindex, as this will make the `iframe` completely inaccessible for keyboard navigation.
+
+Due to these inconsistencies, the applicability of this rule is scoped for `iframe` elements for which there is no indication of an attempt to remove them from accessibility consideration.  Whether `iframe` elements that are inapplicable to this rule still require an accessible name varies between browsers.
 
 ## Background
 
+**Note:** `frame` element is deprecated, this rule does not consider `frame` or `frameset` elements.
```
Jym77:
```suggestion
The `frame` element is deprecated, this rule does not consider `frame` or `frameset` elements.
```
No need for the "note" marker in the Background section.

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -37,14 +38,18 @@ If an `iframe` is not perceived by the user as a single control, it does not qua
...
   - There is a known combination of a popular browser and assistive technology that ignores `aria-label` and only announces `title` attribute as an [accessible name][]
   - Some assistive technologies ignore empty `iframe` elements, regardless of if they are focusable or if they have an accessible name.
+  - Some browsers instantly redirect focus from `iframe` elements  to the first focusable element inside that iframe. This redirect makes it appear as though the `iframe` never receives focus. This occurs even if the `iframe` has a non-negative `tabindex` [attribute value][].
+  - Not all browsers redirect focus on `iframe` elements. This ensures that the contents of `iframe` elements can be scrolled and accessed by using the keyboard. This must not be circumvented by using a negative tabindex, as this will make the `iframe` completely inaccessible for keyboard navigation.
+
+Due to these inconsistencies, the applicability of this rule is scoped for `iframe` elements for which there is no indication of an attempt to remove them from accessibility consideration.  Whether `iframe` elements that are inapplicable to this rule still require an accessible name varies between browsers.
```
Jym77: **Suggestion:** I feel this is more of a Background note than an Accessibility Support one 🤔 
Maybe phrasing it as:
```suggestion
Due to inconsistencies in handling focus on `iframe`, this rule ignores `iframe` elements for which there is an attempt to hide them from assistive technologies.  Whether `iframe` elements that are inapplicable to this rule still require an accessible name varies between browsers.
```
and move it to the Background section.
(also trying to simplify the phrasing and avoid negations)

_rules/iframe-non-empty-accessible-name-cae760.md
```diff
@@ -139,7 +142,15 @@ This `iframe` is not [included in the accessibility tree][] because of setting a
 This `iframe` element has a negative `tabindex` [attribute value][].
 
 ```html
-<iframe tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html"> </iframe>
+<iframe tabindex="-1" src="/test-assets/SC4-1-2-frame-doc.html" style="height: 250px"> </iframe>
+```
+
+#### Inapplicable Example 4
+
+This `iframe` element has a `role` [attribute value][] of `none`.
```
Jym77:
```suggestion
This `iframe` element has an [explicit semantic role][] of `none`.
```
matching the newest Applicability.

Jym77 (dismissed):
Looks pretty good. A few polishing comments.

_rules/iframe-non-empty-accessible-name-cae760.md
tombrunet: @WilcoFiers I like the rewording, but wanted to get your thoughts on Accessibility Support section vs Background before accepting this.

WilcoFiers: LGTM 👍

tombrunet:
@Jym77 I think I have everything you suggested resolved, and I re-requested review, but it still shows a change requested, not sure why. Can you re-review? Thanks.

Jym77:
> @Jym77 I think I have everything you suggested resolved, and I re-requested review, but it still shows a change requested, not sure why. Can you re-review? Thanks.

Yes, the review needs to be dimissed…  (Bottom box, "change requested" dropdown, dots menu on the review to dismiss). I've done it.
dismiss review

WilcoFiers (approved):

HelenBurge (approved):

kengdoj (approved):

Jym77 (approved):

Jym77:
@tombrunet This looks like this is good to go in Call for Review 🎉 
Do you want me to do it?

tombrunet:
@Jym77 I was out on vacation, but it's sent now.

bruce-usab (approved):

carlosapaduarte (approved):

Issue #1507, opened 2020-11-23
Merge "Element with aria-hidden has no focusable content" (6cfa84) and "Element with presentational children has no focusable content" (307n5z)


Jym77:
[Element with aria-hidden has no focusable content](https://act-rules.github.io/rules/6cfa84) and [Element with presentational children has no focusable content](https://github.com/act-rules/act-rules.github.io/pull/1493) are checking very similar things: that an element should not be both excluded from the accessibility tree and focusable.

They differ in which part of "excluded from the accessibility tree" they check. We should merge them into a more generic "element not in the accessibility tree is not focusable" (or maybe avoid the double negation with "focusable element is included in the accessibility tree".

Unified rule would be more generic (if we found a new way to exclude an element from the accessibility tree, we're already covered), more future-proof (if UA/AT decide to ignore `aria-hidden` on focusable element, we're covered), and more technology-agnostic (the accessibility tree is an input aspect, not matter how it was built, so the rules don't need to make any supposition on how it was built) (that should also be safer with respect to Accessibility Support).

Postponing that for now since we don't have the resources to work on it. We'll likely need to deprecate the old rules when creating the new one.

ajanec01:
Sounds good to me.

WilcoFiers:
I think this issue can be closed. I would not personally be in favour of merging these two rules, they seem distinct enough to me.

Jym77:
They are both instances of "Presentational conflict is not triggered".
With the presentational conflict moving toward being an author error (in the ARIA 1.3 draft, [Presentational Roles Conflict Resolution](https://w3c.github.io/aria/#conflict_resolution_presentation_none) is currently part of [Handling Author Errors](https://w3c.github.io/aria/#document-handling_author-errors)), I feel it makes even more sense to group them.

(vendor hat on) I also see a lot of confusion from our customers on these rules, essentially because whenever the rules fail, the presentational conflict does trigger and the page "works" (i.e. the focusable content is exposed to AT). I feel that getting a generic wording ("presentational conflict") of "you did it wrong, the UA is fixing it for you but you still shouldn't do it" would make for an easier explanation 🤔

dan-tripp-siteimprove:
Also with my vendor hat on: it's contributing to the confusion when we use the words "hidden" and "presentational" interchangeably.  They mean different things, right?  For example: 

```
	<h2 aria-hidden="true">this is hidden.
		<span>... so none of this text...
			<span>will appear in the a11y tree.
[...]

PR #2036, opened 2023-02-23
Split aria-state-or-property-permitted-5c01ea.md

tbostic32:
Splitting the aria-state-or-property permitted into multiple rules. Essentially changing from having the 2 expectation where Expectation 1 is the state or property is permitted and Expectation 2 is the state or property is not prohibited, to 2 separate rules for each of the expectations. We also expect to have a rule for aria state or property is not deprecated in the future.

Need for Call for Review:
This will require a 2 week Call for Review 

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


package.json
```diff
@@ -116,6 +116,9 @@
 			"name": "Stein Erik Skotkjerra",
 			"url": "https://github.com/skotkjerra"
 		},
+		{
+			"name": "Trevor Bostic",
+			"url": "https://github.com/tbostic32"
```
tbostic32, line 125:
```suggestion
			"url": "https://github.com/tbostic32"
		},
```

carlosapaduarte (dismissed):
Please correct "Expectation 1" to "Expectation"

Jym77 (approved):
I'm not a fan of splitting the rule.
It seems the new rule (#2037) will have exactly the same Applicability, and the former Expectation 2 as its only Expectation. I do not think this add value compared to having two expectation in a rule (this is pretty much the reason why rules can have more than one expectation).

It feels we are going to duplicate bits, need to maintain in two places (e.g. if we figure out the Applicability needs fixing), … So I get the feeling we are making things more complex to maintain without bringing any real value (we are even diluting the rule and loosing a bit of cohesion).

Thus said, I won't block it for that reason.

tbostic32:
> It feels we are going to duplicate bits, need to maintain in two places (e.g. if we figure out the Applicability needs fixing), … So I get the feeling we are making things more complex to maintain without bringing any real value (we are even diluting the rule and loosing a bit of cohesion).

@Jym77 I do agree, and I left a comment on that new rule proposal (https://github.com/act-rules/act-rules.github.io/issues/2037) that we were reusing the applicability and the the passed examples from this rule. I think the larger conversation we had is that there is also likely to be a third rule around `deprecated` properties which will follow a very similar format to each of these rules as well. In the end, I think the decision was to just split them fully apart (each to it own atomic rule) for simplicity in covering each of those cases specifically with the knowledge that it does introduce a bit more overhead.

_rules/aria-state-or-property-not-prohibited.md
```diff
@@ -0,0 +1,150 @@
...
+    secondary: true
+input_aspects:
+  - Accessibility Tree
+  - CSS styling
+  - DOM Tree
+acknowledgments:
+  authors:
+    - Anne Thyme Nørregaard
+    - Jean-Yves Moyen
+    - Trevor Bostic
```
tbostic32: Should I keep all of the authors? Much of the content was created by Anne and Jean-Yves so I want to make sure proper credit is given, but not have people feel like I am putting words in their mouth.

WilcoFiers: It's a significant change. Probably better to put them under previous_authors

Jym77 (approved):

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,150 @@
...
+    - JFK's "We Choose the Moon" speech excerpt is courtesy of NASA.
+---
+
+## Applicability
+
+This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is [included in the accessibility tree][].
+
+## Expectation
+
+No test target (i.e., state or property) is prohibited on the semantic role of the element on which it is specified.
```
WilcoFiers: I think it would be good for the word "prohibited" to link to the corresponding section in WAI-ARIA. There are lots of ways in which an attribute may not be allowed, and without a link this could be misunderstood to be broader than it actually is.

WilcoFiers: I'm not a fan of the "(i.e. state or property)" bit. If we were to do this in all rules I'm concerned this would get messy. If we feel that "test target" as a term is confusing we should have a separate conversation about that, and then adjust all rules to fit the outcome. I don't think we want to introduce changes like this without thinking through the broader implications.

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,150 @@
...
+- [Understanding Success Criterion 4.1.1: Parsing](https://www.w3.org/WAI/WCAG21/Understanding/parsing.html)
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+- [WAI-ARIA 1.2, Supported States and Properties](https://www.w3.org/TR/wai-aria-1.2/#states_and_properties)
+- [WAI-ARIA 1.2, Global States and Properties](https://www.w3.org/TR/wai-aria-1.2/#global_states)
+- [ARIA5: Using WAI-ARIA state and property attributes to expose the state of a user interface component](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA5)
+- [Document conformance requirements for use of ARIA attributes in HTML](https://www.w3.org/TR/html-aria/#docconformance)
+
+## Test Cases
+
+### Passed
```
WilcoFiers, line 78: `aria-pressed` is never prohibited. I think it would be good to (also?) include examples of attributes that can be prohibited.

WilcoFiers (dismissed):

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,160 @@
...
+    - JFK's "We Choose the Moon" speech excerpt is courtesy of NASA.
+---
+
+## Applicability
+
+This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is [included in the accessibility tree][].
+
+## Expectation
+
+No test target is [prohibited](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes) on the semantic role of the element on which it is specified.
```
WilcoFiers, line 45: I wonder if semantic role is right here. It creates this strange interaction with presentational conflict resolution. If you agree with my suggestion on the applicability, the perhaps what needs to go here is something like this:

```suggestion
No test target is [prohibited](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes) on the element's role. The role here is the [explicit role][] if one exists, or the [implicit role][] otherwise.

**note** The role here is different from the [semantic role][], as it does not [consider presentational role conflict resolution][].
```

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,160 @@
...
+    - Jean-Yves Moyen
+  funding:
+    - WAI-Tools
+  assets:
+    - JFK's "We Choose the Moon" speech excerpt is courtesy of NASA.
+---
+
+## Applicability
+
+This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is [included in the accessibility tree][].
```
WilcoFiers, line 41: I don't know if you meant for this rule to skip prohibited attributes on role=none/presentation/generic, but by using Included in the accessibility tree that's what happens. Those aren't in the accessibility tree. That stuff is covered by Decorative element not exposed though, so maybe that's okay? I don't think ARIA WG intended for those to be ignored. In ARIA 1.2 that includes almost all prohibited attributes.

One thing you could do here is to use programmatically hidden instead:

```suggestion
This rule applies to any [WAI-ARIA state or property][] that is specified on an [HTML or SVG element][namespaced element] that is not [programmatically hidden][].
```

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,160 @@
...
+Implementation of [Presentational Roles Conflict Resolution][] varies from one browser or assistive technology to another. Depending on this, some elements can have a [semantic role][] of `none` and their attributes fail this rule with some technologies but users of other technology would not experience any accessibility issue.
+
+## Background
+
+The presence of prohibited ARIA attributes is often the result of a developer using an incorrect role, or a misunderstanding of the attribute. These attributes are ignored by browsers and other assistive technologies. This often means that a state or property which should exist is missing. This can cause issues under success criterion [1.3.1 Info and Relationships][sc131] or [4.1.2 Name, Rule Value][sc412].
+
+Assessing the value of the attribute is out of scope for this rule.
+
+### Related rules
+
```
WilcoFiers, line 62: This rule has an interesting interaction with Decorative element not exposed. It almost seems to me that this rule should cover everything that rule does, except that you've sidestepped all the overlapping cases by looking at elements included in the accessibility tree, which avoid anything role=none/presentation.

```suggestion

- [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/)
```

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+### Failed
+
+#### Failed Example 1
+
+The `aria-label` property is [prohibited](https://www.w3.org/TR/wai-aria-1.2/#prohibitedattributes) for an element with a generic role.
+
+```html
+<div aria-label="Bananas"></div>
```
WilcoFiers, line 112: I'm not sure this element is actually in the accessibility tree. ARIA 1.3 draft contradicts itself on that:

> https://w3c.github.io/aria/#scope
> user agents MUST ignore any elements with the role generic or none.

> https://w3c.github.io/aria/#generic
> However, unlike elements with role presentation, generic elements are exposed in accessibility APIs

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
```diff
@@ -0,0 +1,160 @@
+---
+id: yyw7k3
+name: ARIA state or property is not prohibited
```
WilcoFiers, line 3: How would you feel about having this rule as a replacement of "Element marked as decorative is not exposed"? I think right now this rule avoids most overlap between the two rules, but it seems to me that the way the ARIA WG intends prohibited attributes to work, this rule isn't broad enough, and most of what it's missing is covered by the other rule. I left a couple suggestions, if you want to go that direction, I think we should consider deprecating the other rule.

@Jym77 I would appreciate it you'd weigh in on this too.

WilcoFiers (changes requested):

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
tbostic32: Failed example 1 shows a failure on a generic role, so we would either need to change the applicability here, or we need to remove that example.

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
tbostic32: With your suggested change to the applicability, we would have some overlapping cases, which I think might just complicate things. As I've been digging more into it, I'm becoming more inclined to leave the applicability and expectation as they are. 

With that said, I think its still work adding this as a related rule so I'm going to add this suggestion in.

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
tbostic32: I could leave or delete this example, I think we could point to this sentence and say that it should be in the tree. 

> However, unlike elements with role presentation, user agents expose generic elements in [accessibility APIs](https://w3c.github.io/aria/#dfn-accessibility-api) when permitted accessibility attributes have been specified.

_rules/aria-state-or-property-not-prohibited-yyw7k3.md
tbostic32: Sorry, I apparently missed this comment on my first look through. I agree with the sentiment here though, if we want to expand this rule with the applicability you suggested, then I think we should deprecate "Element marked as decorative is not exposed". Additionally, since "Element marked as decorative" doesn't have any accessibility requirements, including those failed cases under this rule would give more direction on what we think reporting requirements are.

TL;DR - In favor of expanding this rule and deprecating "Element marked as decorative"

Jym77: Seemed to have missed the initial comment also… No fully sure I remember all the details, but…

* I believe that "element marked as decorative" (together with "`aria-hidden` no focusable content" and "presentational children no focusable content") can be consolidated into "presentational conflict is not triggered" (#1507). It seems all 3 check a different aspect of the conflict. I've also gathered from a couple of discussions that ARIA group more or less considers triggering the conflict as an author error (even though it didn't make to the spec).
* This rule is conceptually a bit different since prohibited attributes could be something else. Notably, some roles have prohibited naming (and thus attributes `aria-label(led-by)`) without triggering the conflict.
* I somewhat remember a discussion (maybe here 😅) about the fact that `role="none" aria-label="foo"` does trigger the conflict, thus the role is not `none` and the element is included in the accessibility tree (plus, the `aria-label` may be not prohibited on the effective role). I think we concluded that ARIA intention was to consider this as a prohibited attribute nonetheless.

So, I'm rather in favour of deprecating "element marked as decorative" which is a bit half-baked and too narrow in its focus, and that we can do much better in terms of intention now. But I'm not sure that this rule is the best fit for the deprecation. I'm also not against a bit of overlap between "attribute not prohibited" and "conflict not triggered" rules, given that they do test with very different focus in mind, and also report a lot of different things on their own.

Issue #1965, opened 2022-10-27
Should inert iframes with focusable content fail? [akn7bn]


WilcoFiers:
This related to the Iframe Iframe with interactive elements is not excluded from tab-order rule. It indicates that iframes with negative tabindex are a problem, but there's now another way to disable focus on iframe, by using `inert`.

I don't think iframes with `inert` should fail this rule, but I'd like other's opinion about that.

giacomo-petri:
> I don't think iframes with inert should fail this rule, but I'd like other's opinion about that.

I agree; if the author consciously set the inert attribute to an iframe we should (?) assume that the intent is to prevent users navigating the iframe.

Side note: it's interesting how, setting the inert property to the iframe, it:
- allows mouse users to scroll and perceive visually the content of the iframe
- prevents keyboard and AT (such as screen readers) users to perceive the content of the iframe

setting instead the inert property to the body inside the iframe, it:
- allows mouse users to scroll and perceive visually the content of the iframe
- allows keyboard users to scroll using up/down arrows and perceive visually the content of the iframe
- prevents AT (such as screen readers) users to perceive the content of the iframe

carlosapaduarte:
Initial discussion [in the CG call](https://www.w3.org/2023/02/09-act-r-minutes.html#t07). To be continued in the next meeting

carlosapaduarte:
Update the applicability to exclude inert iframes and create our definition of inert, that includes the inter attribute and content outside a modal

Issue #1953, opened 2022-10-13
Stateful Rules


tbostic32:
Problem: Defining states and transitions between states can be very difficult due to the number of ways in which states can be implemented within web pages. 

Aria spec definition of state:
> State: A state is a dynamic [property](https://www.w3.org/TR/wai-aria-1.2/#dfn-property) expressing characteristics of an [object](https://www.w3.org/TR/wai-aria-1.2/#dfn-object) that may change in response to user action or automated processes. States do not affect the essential nature of the object, but represent data associated with the object or user interaction possibilities. See [clarification of states versus properties](https://www.w3.org/TR/wai-aria-1.2/#statevsprop).

In the past, in 2021, we decided that the states we cared about should be tied to CSS psuedo classes since they are an inherently objective measurement of state. https://www.w3.org/TR/selectors-4/#useraction-pseudos. This would include things like:
	1. Focus
	2. Hover
	3. Visited (for links)
	4. Active
	5. Valid/Invalid (Only with html input specification)
	6. Enabled/Disabled
	7. Required
	8. not Checked/Checked
	9. And possibly more

As each of these only denotes a single state, we could also potentially expand the definition to include transitions between states (e.g., between not checked and checked), allowing us to test some user interactions with the page.  However, the downside of this approach is that CSS-psuedo classes cannot represent all possible states or state changes of interest that might occur. A quote from our findings in early 2021:

> CSS pseudo-classes are well-defined and easy to refer to. Javascript manipulations of the DOM that are not reflected in the CSS pseudo-class selectors are ill defined and extremely hard to specify in the objective manner that is requested for ACT rules. 

So, to give examples, with just CSS Psuedo selectors we would struggle to handle the following scenarios:
	1. An error message pops up as a result of bad form input (would only trigger :invalid if HTML5 input specs used)
	2. A dropdown menu is expanded (initial element would be hovered and eventually active, but its difficult to tie that to the "expansion")
	3. An image carousel rotates either due to human action or on a timer

So what can we do:
	1. We can stay with CSS psuedo selectors as our measurement of state, understanding its gaps and working around them. Perhaps try to estimate the number of rules that would be difficult to write with this definition.
	2. Expand our state definition to include states given by other W3C standards such as aria. For example, aria would (at least) include the states busy (loading) and expanded.
	3. Allow subjective applicability that allows us to define new states as needed, likely requiring new glossary entry's. 

Pros and Cons of 1&2
[...]

PR #2038, opened 2023-02-23
Update "iframe with interactive elements is not excluded from tab-order" rule to manage inert iframe elements

giacomo-petri:
This pull request manages inert iframe elements in [iframe with interactive elements is not excluded from tab-order](https://www.w3.org/WAI/standards-guidelines/act/rules/akn7bn/proposed/) rule, excluding them from the applicability of the rule.

Updates:
- edited applicability excluding inert iframe elements;
- added definition of _inert_ in applicability;
- added definition of _modal_ (modal dialog);
- added 2 inapplicable examples:
  - iframe with inert attribute
  - modal which overlaps the iframe, making the iframe inoperable

Closes issue(s): https://github.com/act-rules/act-rules.github.io/issues/1965

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,17 @@ acknowledgments:
...
 
-This rule applies to any `iframe` element that [contains](#akn7bn:contain) at least one element for which all the following are true:
+This rule applies to any `iframe` element that is not [inert](#akn7bn:inert) and that [contains](#akn7bn:contain) at least one element for which all the following are true:
 
 - the element is [visible][]; and
 - the element is part of the [sequential focus navigation order][] of the `iframe`'s [document][].
 
 An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
 
+An `iframe` element is <dfn id="akn7bn:inert">inert</dfn> if:
```
carlosapaduarte: I think we can have this as a definition also, to make it reusable in other (future?) rules

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,17 @@ acknowledgments:
...
+- it has an `inert` [attribute value][] of true; or
+- an element which is not the iframe itself and that is not [contained](#akn7bn:contain) in it behaves as a [modal][], making the iframe inoperable.
```
carlosapaduarte: What does "behaves as a modal" mean?
And do we want to tie this to the behaviour or the state? If there is a modal open, the iframe is inert and not applicable, but if the modal is closed, then the iframe is applicable. In both states, the component behaves as a modal, right?

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
carlosapaduarte: And probably it would be better to have a definite of inert element, not just inert iframe

pages/glossary/modal-dialog.md
```diff
@@ -0,0 +1,16 @@
...
+key: modal-dialog
+unambiguous: true
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+A modal dialog is a type of user interface element that appears on top of the main content of a [html web page][] or application and requires the user to take some action or make a decision before they can continue interacting with the program.
```
carlosapaduarte: I don't think "type of user interface" is objective.
Same for "main content". 
What is the "program" you mention?

I think this needs to focus on the element preventing interaction with all other elements that are not their children. 

This definition really needs to be objective, because it is used in the applicability of the rule.

carlosapaduarte (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
dan-tripp-siteimprove:
```suggestion
- an element which is not the iframe itself and that is not [contained](#akn7bn:contain) in it makes the iframe and all of its contents [inoperable][].  (For example: a modal.)
```
Does that improve the situation at all?

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +126,32 @@ This `iframe` element contains a link that is not part of its [sequential focus
...
+  <h2 id="modal-heading">Hello</h2>
+  <p>Hello World!</p>
+</div>
+<iframe tabindex="-1" srcdoc="<a href='/'>Home</a>"></iframe>
+```
+
 [attribute value]: #attribute-value 'Definition of Attribute Value'
 [container document]: https://html.spec.whatwg.org/#bc-container-document 'HTML browsing context container document, 2020/12/18'
 [document]: https://html.spec.whatwg.org/multipage/dom.html#document 'HTML definition of document'
 [flattened tabindex-ordered focus navigation scope]: https://html.spec.whatwg.org/multipage/interaction.html#flattened-tabindex-ordered-focus-navigation-scope 'HTML - Living Standard, 2022/07/08'
```
dan-tripp-siteimprove, line 161:
```suggestion
[flattened tabindex-ordered focus navigation scope]: https://html.spec.whatwg.org/multipage/interaction.html#flattened-tabindex-ordered-focus-navigation-scope 'HTML - Living Standard, 2022/07/08'
[inoperable]: https://www.w3.org/TR/wai-aria/#dfn-operable
```

pages/glossary/modal-dialog.md
dan-tripp-siteimprove:
```suggestion
A modal dialog is an element which, until it is closed, prevents user interaction with any content on the web page that is not a descendant element of the modal.  Visually, a modal often (but not necessarily) appears on top of all other content of the [html web page][] and visually darkens or otherwise obscures that other content.
```
Trying to put Carlos' advice into action.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -28,13 +28,17 @@ acknowledgments:
...
-This rule applies to any `iframe` element that [contains](#akn7bn:contain) at least one element for which all the following are true:
+This rule applies to any `iframe` element that is not [inert](#akn7bn:inert) and that [contains](#akn7bn:contain) at least one element for which all the following are true:
 
 - the element is [visible][]; and
 - the element is part of the [sequential focus navigation order][] of the `iframe`'s [document][].
 
 An element is <dfn id="akn7bn:contain">contained</dfn> in a [nested browsing context][] if its [owner document][] is the [container document][] of the [nested browsing context][].
 
+An `iframe` element is <dfn id="akn7bn:inert">inert</dfn> if:
+- it has an `inert` [attribute value][] of true; or
```
WilcoFiers: The attribute value doesn't matter. It just has to be present. Plus this should apply to ancestors too. Inert disables the entire subtree.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
WilcoFiers: What I meant by modal dialogs was the actual `<dialog>` element with `showModal()` called on it. That programmatically makes all other content on the page inert. I think you could define it like this:

> The element is in a page with a `dialog` element with the [`is-modal`](https://html.spec.whatwg.org/multipage/interactive-elements.html#is-modal) flag set to `true`, and is not a descendant of that `dialog` element.


I'm not sure we should try for a generic modal definition. I don't know how you'd do that in a way that's objective.

pages/glossary/modal-dialog.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Modal dialog
```
WilcoFiers: I'm concerned this could be confused by the `<dialog>`'s `showModal()` method.

WilcoFiers (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: > The attribute value doesn't matter. It just has to be present. 

As per [our definition of attribute value](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/glossary/attribute-value.md), boolean attribute (like `inert`) have an "attribute value" of `true` if the attribute is present.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: See https://html.spec.whatwg.org/multipage/interaction.html#modal-dialogs-and-inert-subtrees

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
dan-tripp-siteimprove:
```suggestion
- it has an `inert` [attribute value][] of true; or
- one of its ancestor elements has an `inert` [attribute value][] of true; or
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
dan-tripp-siteimprove:
```suggestion
- it is on the same page as a `dialog` element which has its [`is-modal`](https://html.spec.whatwg.org/multipage/interactive-elements.html#is-modal) flag set to `true`, and is not a descendant of that `dialog` element.
```

dan-tripp-siteimprove (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: I agree, we should have a separate definition of "inert element".

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: I think we can probably refer directly to the [HTML definition of inert](https://html.spec.whatwg.org/multipage/interaction.html#inert).
It doesn't refer explicitly to modals, but [being blocked by a modal](https://html.spec.whatwg.org/multipage/interaction.html#blocked-by-a-modal-dialog) causes most of the document to be inert, so we can mention that.

If we want to have our own definition of inert, we should probably refer to "being blocked by a modal", since it is objective and unambiguous, and a common definition from higher spec.
This may leave room for custom modals where somebody uses a `div` and a ton of Javascript to turn it into a modal manually. We might be OK enough by having an assumption about people not being that crazy… (i.e. I do not know how frequent it is for devs to re-code modal behaviour on their own rather than relying on native semantics…)

Jym77: Then, we are indeed going for [blocked by a modal](https://html.spec.whatwg.org/multipage/interaction.html#blocked-by-a-modal-dialog):
> A [Document](https://html.spec.whatwg.org/multipage/dom.html#document) document is blocked by a modal dialog subject if subject is the topmost [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element in document's [top layer](https://fullscreen.spec.whatwg.org/#top-layer). While document is so blocked, every node that is [connected](https://dom.spec.whatwg.org/#connected) to document, with the exception of the subject element and its [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) descendants, must become [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert).

> subject can additionally become [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert) via the [inert](https://html.spec.whatwg.org/multipage/interaction.html#the-inert-attribute) attribute, but only if specified on subject itself (i.e., subject escapes inertness of ancestors); subject's [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) descendants can become [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert) in a similar fashion.

> Note: The [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element's [showModal()](https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-showmodal) method causes this mechanism to trigger, by [adding](https://fullscreen.spec.whatwg.org/#top-layer-add) the [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element to its [node document](https://dom.spec.whatwg.org/#concept-node-document)'s [top layer](https://fullscreen.spec.whatwg.org/#top-layer).

Jym77 (dismissed):
To sum up, I think we can just use HTML definition of inert (instead of trying to rebuild it) + add some background description about `showModal` triggering "blocked by a modal" triggering inert + make the examples descriptions clear (this `iframe` is inert because it is blocked by the modal that has been brought to the top layer with `showModal`).

giacomo-petri:
Updates:
- added "inert" definition to be re-used;
- updated the definition of inert by using the "blocked by a modal" html definition;
- removed the modal definition completely (no longer needed);
- added example (inapplicable 6) of showModal() method.

I've kept the inapplicable example n. 7 with the custom modal. Let me know if it's better to remove it completely.

bruce-usab (changes requested):
The name of the referenced rule `Iframe with interactive elements is not excluded from tab-order` is fine -- as is the content of the rule.

But the name of this PR seems to be a double negative?  It is very confusing!  And maybe it does not matter?

Please change the PR name.  If it were just "update iframe not focusable" I would have twice about the rule!

giacomo-petri:
> The name of the referenced rule `Iframe with interactive elements is not excluded from tab-order` is fine -- as is the content of the rule.
> 
> But the name of this PR seems to be a double negative? It is very confusing! And maybe it does not matter?
> 
> Please change the PR name. If it were just "update iframe not focusable" I would have twice about the rule!

Title edited.
The initial intent was to provide information about the specific update, avoiding a generic label such as "Update iframe rule", which doesn't help in filtering PR.
Please, let me know if the new title makes sense.

Jym77:
> Title edited. The initial intent was to provide information about the specific update, avoiding a generic label such as "Update iframe rule", which doesn't help in filtering PR. Please, let me know if the new title makes sense.

As a rule of thumb, I feel it is good to include the full name of the updated rule in the PR title. Including the rule ID can also help tracking them across change of name (the id is the stable name). I usually like to frontload these, but we have no strict PR naming convention.
> "rule name" [rule ID]: verb something something.
e.g.
> "iframe with interactive elements is not excluded from tab-order" [akn7bn]: Manage inert iframe element

(I feel that "update rule" is a bit redundant in a PR title, since PR are mostly made to update rules)
(the current title is fine, no need to change it again)

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +122,61 @@ This `iframe` element contains a link that is not part of its [sequential focus
...
+
+This `iframe` element is [inert][] because of its own `inert` [attribute value][].
+
+```html
+<iframe inert srcdoc="<a href='/'>Home</a>"></iframe>
+```
+
+#### Inapplicable Example 6
+
+Once "Privacy policy details" button is activated, the `iframe` element become [inert][] because of showModal() method that causes the `iframe` to be [blocked by a modal].
```
dan-tripp-siteimprove:
```suggestion
Once the "Privacy policy details" button is activated, the `iframe` element becomes [inert][] because of the showModal() method, which causes the `iframe` to be [blocked by a modal].
```

dan-tripp-siteimprove (dismissed):

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: I think we should make the example so that the modal is shown at load time.
Currently, examples do not require manipulation before testing. There are some steps in that direction as part of the "states" discussion (#1953, #2045).

In this case, we can simply call the `showModal` in an `onload` call to ensure this is testable out of the box. I think the example will still look "real" enough that people will understand why the modal is opened by default. (otherwise a more realstic case could be to have a cookies banner or region selection modal that is opened on page load and blocking interaction with the rest of the page).

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +122,61 @@ This `iframe` element contains a link that is not part of its [sequential focus
...
+
+    cancelButton.addEventListener("click", () => {
+        dialog.close();
+    });
+</script>
+```
+
+#### Inapplicable Example 7
+
+This `iframe` element is [inert][] because it is [blocked by a modal].
```
Jym77:
```suggestion
This `iframe` element is [inert][] because it is [blocked by a modal][].
```
This will also need an entry in the references list.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +122,61 @@ This `iframe` element contains a link that is not part of its [sequential focus
...
+```html
+<div style="width:100vw; height:100vh; position:absolute; left:0; top:0; background:#000; opacity: 0.8;"></div>
+<div role="dialog" aria-labelledby="modal-heading" aria-modal="true" tabindex="-1" style="width:calc(100% - 20vw); height:calc(100% - 20vh); margin-left:10vw; margin-top:10vh; position:absolute; left:0; top:0; background:#fff; z-index:1; padding:40px; box-sizing:border-box;">
+  <h2 id="modal-heading">Hello</h2>
+  <p>Hello World!</p>
+</div>
+<iframe tabindex="-1" srcdoc="<a href='/'>Home</a>"></iframe>
+```
+
 [attribute value]: #attribute-value 'Definition of Attribute Value'
```
Jym77, line 157:
```suggestion

[attribute value]: #attribute-value 'Definition of Attribute Value'
[blocked by a modal]: https://html.spec.whatwg.org/multipage/interaction.html#blocked-by-a-modal-dialog
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +122,61 @@ This `iframe` element contains a link that is not part of its [sequential focus
 <iframe tabindex="-1" srcdoc="<a href='/' tabindex='-1'>Home</a>"></iframe>
 ```
 
+#### Inapplicable Example 5
+
+This `iframe` element is [inert][] because of its own `inert` [attribute value][].
+
+```html
+<iframe inert srcdoc="<a href='/'>Home</a>"></iframe>
```
Jym77: I think the `iframe` shuld have a negative `tabindex` under the rule of thumb "the Inapplicable example should look bad if it wasn't inapplicable". The reason is that in implemetation reports "Passed" and "Inapplicable" results are essentially equivalent, so to detect wrong Applicability in a tool, we need the tool to (wrongly) report "Failed" rather than to wrongly report "Passed".
```suggestion
<iframe tabindex="-1" inert srcdoc="<a href='/'>Home</a>"></iframe>
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +122,61 @@ This `iframe` element contains a link that is not part of its [sequential focus
...
+```
+
+#### Inapplicable Example 6
+
+Once "Privacy policy details" button is activated, the `iframe` element become [inert][] because of showModal() method that causes the `iframe` to be [blocked by a modal].
+
+```html
+<iframe srcdoc="<a href='/'>Home</a>"></iframe>
```
Jym77:
```suggestion
<iframe tabindex="-1" srcdoc="<a href='/'>Home</a>"></iframe>
```

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
Jym77: It's technically not true. The HTML definition of "blocked by a modal" requires a `dialog` element to do the blocking, while this is done by a `div`.

Also, this actually doesn't works. Due to non-interference with the host language, `aria-modal` does not make the rest of the page inert. [ARIA for `aria-modal`](https://www.w3.org/TR/wai-aria-1.2/#aria-modal) states:
> When a modal element is displayed, authors SHOULD mark all other contents as inert (such as "inert subtrees" in HTML)

When testing (without the negative `tabindex`), I have no problem tabbing onto the link inside the `iframe`. I haven't tested whether screen reader consider it as inert, but Chrome doesn't.

pages/glossary/inert.md
```diff
@@ -0,0 +1,15 @@
...
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An [HTML or SVG element][] is inert if:
+- it has an `inert` [attribute value][] of true; or
+- one of its ancestor elements has an `inert` [attribute value][] of true; or
+- it is [blocked by a modal](https://html.spec.whatwg.org/multipage/interaction.html#blocked-by-a-modal-dialog).
```
Jym77: All the references need to be defined in a references list at the bottom of the file.

Jym77 (dismissed):

giacomo-petri (commented):
suggestions added

giacomo-petri:
- Accepted all the @Jym77's suggestions about adding the tabindex="-1" attribute to the iframe element;
- Updated the Failed example n. 6 with dialog opened `onload` instead of after user activation;
- Removed failed example n.7 with "custom modal" as not aligned with the rule;
- Updated references in "inert" definition.

Jym77 (approved):

dan-tripp-siteimprove (approved):

colabottles (approved):

carlosapaduarte (approved):

carlosapaduarte:
Call for review ends June 21

bruce-usab (approved):
Thanks for this!

From the list email, I had expected I would have concern about defining `inert`.  Instead, I was delighted to learn that it is a reasonably well defined attribute.

I don't think my own knowledge is too below average, so I recommend adding links for background reading.  Here are two I found useful:

- [Introducing inert](https://developer.chrome.com/articles/inert/)
- [MDN reference inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert)

Jym77:
> I don't think my own knowledge is too below average, so I recommend adding links for background reading. Here are two I found useful:
> 
> * [Introducing inert](https://developer.chrome.com/articles/inert/)
> * [MDN reference inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert)

@bruce-usab Since ACT rules are intended to be published as official W3C documentation, we tend to avoid direct links (or citation) of commercial third parties, especially when they are themselves major players (we do not want to look like we favour one above the others). As much as we can, we'd rather stick with W3C documentation (HTML living standard, …)

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
```diff
@@ -122,10 +122,42 @@ This `iframe` element contains a link that is not part of its [sequential focus
...
+    <h2 id="dialogLabel">Privacy Policy</h2>
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+    <button id="cancel">Cancel</button>
+</dialog>
+<script>    
+			const dialog = document.getElementById("ppDialog");
+			window.addEventListener('load', function(event) {
+    			dialog.showModal();
+			});
+</script>
```
WilcoFiers: I'm not entirely happy with this example. You can easily get this in an inaccessible state. Before onload the page is inaccessible, and after the modal is closed too. I also think that using "DOMContentLoaded" would be better, since most tools start testing when onload is triggered. This ensures the modal is already open when automated testing starts. I also added a title attribute to the iframe, and removed the duplicate ID. Lastly, I adjusted whitespace so it's what our formatter expects.

```suggestion
<iframe id="myFrame" title="Links" srcdoc="<a href='/'>Home</a>"></iframe>
<div>
	<button id="ppButton" onclick="openDialog()">Privacy policy details</button>
</div>
<dialog id="ppDialog" aria-labelledby="dialogLabel">
	<h2 id="dialogLabel">Privacy Policy</h2>
	<p>We store no data.</p>
	<button id="cancel" onclick="ppDialog.close()">Cancel</button>
</dialog>
<script>
	const openDialog = () => {
		ppDialog.showModal();
		myFrame.tabIndex = '-1'
	}
	ppDialog.addEventListener('close', () => myFrame.tabIndex = 0)
	window.addEventListener('DOMContentLoaded', openDialog);
</script>
```

pages/glossary/inert.md
```diff
@@ -0,0 +1,19 @@
...
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An [HTML or SVG element][] is inert if:
+- it has an `inert` [attribute value][] of true; or
+- one of its ancestor elements has an `inert` [attribute value][] of true; or
```
WilcoFiers: We should indicate what tree to use when talking about ancestors.

```suggestion
- one of its ancestor elements in the [flat tree][] has an `inert` [attribute value][] of true; or
```

pages/glossary/inert.md
```diff
@@ -0,0 +1,19 @@
...
+- it is [blocked by a modal][].
+
+[HTML or SVG element]: #namespaced-element 'Definition of HTML or SVG element'
+[attribute value]: #attribute-value 'Definition of Attribute value'
+[blocked by a modal]: https://html.spec.whatwg.org/multipage/interaction.html#blocked-by-a-modal-dialog
```
WilcoFiers, line 19:
```suggestion
[blocked by a modal]: https://html.spec.whatwg.org/multipage/interaction.html#blocked-by-a-modal-dialog
[flat tree]: https://drafts.csswg.org/css-scoping/#flat-tree 'Definition of flat tree'
```

WilcoFiers (changes requested):
Couple minor points. Don't expect any discussion necessary on these.

giacomo-petri:
This PR is in call for review for 2 weeks. Review ends Sept 29th.

_rules/iframe-not-focusable-has-no-interactive-content-akn7bn.md
WilcoFiers: This looks outdated?

WilcoFiers (approved):
Good stuff!

PR #2043, opened 2023-04-04
invalid-form-field-value-36b590: updating failed example 2.

### LINKED ISSUE #2024: "Error message describes invalid form field value" [36b590]: Failed Example 2 should identify the errored field
[Error message describes invalid form field value](https://www.w3.org/WAI/standards-guidelines/act/rules/36b590/proposed/), Failed Example 2 has an error message that doesn't describe how to resolve the error.
However, it is also not very good at identifying the errored field, and should rather say something like "enter correct value for age" (like Passed Example 1). This would ensure we only test one condition at a time.

dan-tripp-siteimprove:
Closes issue(s):   #2024 

Need for Call for Review:
This will not require a Call for Review

### DIFF
--- a/_rules/invalid-form-field-value-36b590.md
+++ b/_rules/invalid-form-field-value-36b590.md
@@ -171,7 +171,7 @@ This `input` element has a [form field error indicator][] but its message does n
 <form>
 	<label for="age">Age (years)</label>
 	<input type="number" id="age" />
-	<span id="error">Please enter the correct text.</span><br />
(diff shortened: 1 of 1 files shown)


tbostic32 (approved):

WilcoFiers (approved):

Jym77 (approved):

Jym77:
Merging, as this does not require a Call for Review.

PR #2057, opened 2023-04-25
Fix typos in two examples

### LINKED ISSUE #197: Typos in 2 tests
There appears to be two typos in the tests:

The [Text has minimum contrast rule](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/proposed/) has a typo in the [Inapplicable Example 5](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/afw4f7/8df570a7bb69487e5f3b5d07a7af45cc881dc449.html) test: The **src** attribute is mispelt as **scr**

The [ARIA required context role rule](https://www.w3.org/WAI/standards-guidelines/act/rules/ff89c9/proposed/) refers to the **header** role in the [Inapplicable Example 4](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/ff89c9/84dd1831c0682a800e6af2fe992fb3d69d4222eb.html) test. There is no **header** role, so shouldn't this be role="**heading**" ?

WilcoFiers:
Closes https://github.com/w3c/wcag-act-rules/issues/197

Need for Call for Review: none - these should not impact implementation

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -225,7 +225,7 @@ This `section` element with an [explicit role][] of `doc-abstract` has a role fr
 There is no element whose role has [required context role][] because the `header` role does not have one.
 
 ```html
-<div role="header" aria-level="1">Hello!</div>
+<div role="heading" aria-level="1">Hello!</div>
 <p>Welcome to my homepage!</p>
 ```
 
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -222,7 +222,7 @@ This `section` element with an [explicit role][] of `doc-abstract` has a role fr
 
 #### Inapplicable Example 4
 
-There is no element whose role has [required context role][] because the `header` role does not have one.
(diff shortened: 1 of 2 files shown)


daniel-montalvo (approved):

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -225,7 +225,7 @@ This `section` element with an [explicit role][] of `doc-abstract` has a role fr
 There is no element whose role has [required context role][] because the `header` role does not have one.
```
HelenBurge: > There is no element whose role has [required context role][] because the `header` role does not have one.

"Header" should be amended here too?

HelenBurge (dismissed):

Jym77 (approved):

colabottles (approved):

_rules/aria-required-context-role-ff89c9.md
WilcoFiers:
```suggestion
There is no element whose role has [required context role][] because the `heading` role does not have one.
```

PR #2060, opened 2023-05-16
Move secondary requirement texts out of the background

WilcoFiers:
This PR moves the secondary requirements text out of the background and in with the requirement itself. This work is related to the following PRs:
- https://github.com/act-rules/act-tools/pull/29/files#diff-8e5975aabfe5082c95d89bb19ec33636a21993efc35ed73d032ebd3e6925b5cbR114
- https://github.com/w3c/wcag-act-rules/pull/213

I will do the rest of the texts once I've got agreement from the group that this is a good way to write these.

---

## How to Review And Approve


_rules/meta-refresh-no-delay-no-exception-bisz58.md
```diff
@@ -26,7 +26,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:2.2.1: # Timing Adjustable (A)
-    secondary: True
+    secondary: Because it allows no exceptions, this rule is stricter than this success criterion. Failing this rule often results in this success criterion being not satisfied.
```
WilcoFiers:
```suggestion
    secondary: Because this rule allows no exceptions, it is stricter than this success criterion. Failing this rule often results in this success criterion being not satisfied.
```

_rules/text-contrast-afw4f7.md
```diff
@@ -11,11 +11,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.4.6: # Contrast (Enhanced) (AAA)
-    forConformance: true
-    secondary: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
+    secondary: Because it has a higher minimum contrast, this success criterion is stricter than the rule. This is also why some passed examples do not satisfy this success criterion.
```
WilcoFiers:
```suggestion
    secondary: Because this success criterion has a higher minimum contrast, it is stricter than the rule. This is also why some passed examples do not satisfy this success criterion.
```

_rules/text-contrast-afw4f7.md
```diff
@@ -11,11 +11,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.4.6: # Contrast (Enhanced) (AAA)
-    forConformance: true
-    secondary: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
+    secondary: Because this success criterion has a higher minimum contrast, it is stricter than the rule. This is also why some passed examples do not satisfy this success criterion.
```
tbostic32:
```suggestion
    secondary: Because this success criterion has a higher minimum contrast, it is stricter than the rule. This is why some passed examples do not satisfy this success criterion.
```

Jym77 (approved):
Yes, I think that's the good place to put these explanations 👍

_rules/image-no-text-0va7u6.md
```diff
@@ -210,10 +209,13 @@ This image resource referenced by the `img` element contains text that provides
...
 ```html
-<img alt="WCAG Rocks" src="data:image/svg+xml;utf8,
+<img
+	alt="WCAG Rocks"
+	src="data:image/svg+xml;utf8,
 <svg xmlns='http://www.w3.org/2000/svg' height='20px' width='80px'>
   <text x='0' y='15'>WCAG Rocks</text>
-</svg>" />
+</svg>"
+/>
```
WilcoFiers, line 218: This was Prettier's doing.

_rules/link-non-empty-accessible-name-c487ae.md
```diff
@@ -27,7 +26,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.1.1: # Non-text content (A)
-    secondary: true
+    secondary: This success criterion **partially overlaps** this rule. This is because the HTML`area` element count as both a link, and as non-text content. This is why some of the failed examples may not satisfy this success criterion.
```
tbostic32:
```suggestion
    secondary: This success criterion **partially overlaps** this rule. This is because the HTML `area` element counts as both a link and as non-text content. This is why some of the failed examples may not satisfy this success criterion.
```

_rules/links-identical-name-equivalent-purpose-b20e66.md
```diff
@@ -11,7 +11,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
 wcag20:2.4.4: # Link Purpose (In Context) (A)
-secondary: true
+secondary: This success criterion is **less strict** than this rule. This is because the rule does not consider the context of the link. This is why some of the failed examples may not satisfy this success criterion.
```
tbostic32: If the success criterion is less strict, shouldn't the failed examples potentially satisfy the criterion (e.g., where the link text does not provide the purpose but the context does).

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -12,9 +12,9 @@ accessibility_requirements:
     passed: satisfied
     inapplicable: satisfied
   wcag20:1.3.1: # Info and Relationships (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because the rule does not ignore irrelevant ARIA properties. This is why some of the failed examples may not satisfy this success criterion.
```
tbostic32: Similar to m other comment, but shouldn't **less strict** criterion mean some of the failed examples may satisfy the criterion?

tbostic32 (commented):
My only final comment would be that we use the word "this" a lot in these definitions. The normal template seems to be "This success criterion .... this rule. This is because.... This rule". I'm not sure if we can or should cut down on the repetition, but it does feel a bit overly wordy to me, but maybe thats just personal preference.

_rules/aria-attr-defined-5f99a7.md
```diff
@@ -6,9 +6,9 @@ description: |
   This rule checks that each `aria-` attribute specified is defined in ARIA 1.2.
 accessibility_requirements:
   wcag20:1.3.1: # Info and Relationships (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because the rule does not ignore irrelevant ARIA properties. This is why some of the failed examples may not satisfy this success criterion.
```
WilcoFiers:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because the rule does not ignore irrelevant ARIA properties. Some of the failed examples satisfy this success criterion.
```

ChrisLoiselle: @WilcoFiers and others ... I would want us to define the "some" where as you'd want to map to what are the "some" examples that people can reference in context of what fails / passes as secondary reqs. Thanks!

colabottles:
```suggestion
secondary: This success criterion is less strict than this rule. This is because the rule does not ignore irrelevant ARIA properties. Additionally, some of the failed examples meet this success criterion.
```

_rules/link-non-empty-accessible-name-c487ae.md
```diff
@@ -27,7 +26,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.1.1: # Non-text content (A)
-    secondary: true
+    secondary: This success criterion **partially overlaps** this rule. This is because the HTML `area` element counts as both a link and as non-text content. This is why some of the failed examples may not satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **conditionally applicable** in this rule. It is applicable because the HTML `area` element counts as both a link and as non-text content. Most failed examples satisfy this success criterion.
```

kengdoj: I didn't see how an explanation would be written for [Keyboard Trap Example 7](https://w3c.github.io/wcag-act/act-rules-format.html#example-7be89ba3) in this PR, but I thought "partially" would work better there than here. 

For Ex.7: This SC is **partially covered** in this rule because the rule does not test for other possible solutions. As a result, this SC may be satisfied by some failed examples.

_rules/links-identical-name-equivalent-purpose-b20e66.md
mbgower: In regard to @tbostic32's concerns with the over-use of the word "this"... 
Without commenting on the accuracy of the statements, a rewrite of the prior secondary could be:

>This success criterion is **less strict** than this rule because the rule does not consider the context of the link. Some of the failed examples may not satisfy this success criterion.

@tbostic32's second comment seems correct to me just reading the sentence statements. If the following is not accurate, then I think there's a larger problem here...

>This success criterion is **less strict** than this rule because the rule does not consider the context of the link. Some of the failed examples may ~not~ satisfy this success criterion.

_rules/aria-attr-defined-5f99a7.md
maryjom: This rule is **more strict** than the success criterion because it does not ignore irrelevant ARIA properties. Therefore, some of the reported failures can be false positives rather than a success criterion failure.

My comment carries to the rest of the examples where this text appears. It was a little difficult to read and understand, as written.

__tests__/frontmatter.js
```diff
@@ -95,8 +95,11 @@ function validateRuleFrontmatter({ frontmatter }, metaData) {
 		 */
 		const accRequirementValues = Object.values(accessibility_requirements)
 		test.each(accRequirementValues)('has expected keys for accessibility requirement: `%p`', accReq => {
-			if (accReq.secondary) {
-				expect(accReq.secondary).toBeTrue()
+			if (!accReq) {
+				return
+			} else if (accReq.secondary) {
```
Jym77, line 100: **Note:** depending on our ES target, we may be able to use optional chaining:
```suggestion
			if (accReq?.secondary) {
```

_rules/aria-attr-defined-5f99a7.md
Jym77: * I agree that the second sentence should be "warning, some failed examples pass the SC" rather than "warning, some failed examples fail the SC", because the later is the normal behaviour of rules.
* I'd rather keep some emphasis on "less strict" (like the bolding here), to show this is some sort of keyword. I'd even push toward having one such emphasised term in each secondary explanation since it can help us (and readers) categorise them ("less strict", "more strict", "partial", …) Once we get a bit more usage on these secondary requirements, we may even include these terms in the ACt rules format (or a similar document), with detailed explanation about what they mean.
*  agree that the "this is why" is a bit of a mouthful after a sentence full of "thus rule" and "this SC". Another possibility could be:
    > Therefore, some FE do satisfy the SC.
* I'm also not a huge fan of "may (not)" in the "some (passed/failed) example may (not) satisfy the SC". I feel it collides a bit with the RFC 2119 usage of these terms which is somewhat different here. This sounds like an instruction to rule writers (or part of the format): "when a SC is **less strict**, you **MAY** write FE that satisfy this SC"; while here we describe a more static state: "there exist FE that satisfy the SC, deal with it".
  (plus "may not satisfy" is always slightly ambiguous between "(may not) satisfy" (= are forbidden to satisfy) and "may (not satisfy)" (=are not forced to satisfy); "do not satisfy" doesn't have this ambiguity)

Jym77: I'm not a huge fan of using the term "false positives" here, since it has a connotation of "the rule is bad and generate false positives" while we want to carry the information "this rule is designed for a slightly different use case and tangentially affects this use case".

Typically, in a color contrast case, saying that the 1.4.6 rule "reports false positives (for 1.4.3)" is technically true but sounds a bit dramatic.

_rules/image-no-text-0va7u6.md
```diff
@@ -12,7 +12,6 @@ accessibility_requirements:
     inapplicable: further testing needed
   wcag20:1.4.9: # Images of Text (No Exception) (AAA)
     forConformance: true
-    secondary: true
```
Jym77, line 15: While I understand why this has been removed (together with the Background note), I feel this requires a specific PR just to make 1.4.9 a primary requirement instead.

(implementer hat on), making the requirement primary may make my implementation "partially consistent" if it doesn't already report it. That requires me to update my report. An easy fix, but I'd rather track that change to an ACT PR that focuses on it.
(Alfa is actually not having an implementation of this rule, but you get the idea).

Jym77, line 15: Also, we made 1.4.6 a secondary requirement for the 1.4.3 rule, while it always fail.
Iirc, that was due to the fact that 1.4.6 is AAA and some tools don't report AAA failures, but that shouldn't prevent them to claim consistency on the 1.4.3 rule.

How is that case different?

Jym77 (dismissed):

_rules/image-no-text-0va7u6.md
WilcoFiers: The real reason I did this is because I couldn't come up with an explanation of why this was secondary. If you have one I don't mind putting it in, but I don't think there's anything I can put in here. This rule assumes that the difference between 1.4.5 and 1.4.9 doesn't occur, and so the rule maps to those criteria the same way.

I can put this into a separate PR, but I can't merge this one until that separate PR would then be merged. I did it this way to avoid introducing a blocking PR. Are you okay with that @Jym77?

WilcoFiers: See above. This rule assume there is the difference between 1.4.5 and 1.4.9 never occurs. That's why we don't have two rules for image-no-text. There is no difference in how we test 1.4.5 and 1.4.9. For color contrast the numbers are different, and so we have two different rules.

Jym77: I'm confused 😕 

Back in #1920 when we added secondary requirements, part of the reasoning in the PR description was:
> Their implementations are reported as partially consistent for not failing a AAA success criterion, even though the implementor does not support AAA to begin with. [in the 1.4.3/1.4.6 case]

This, in turn, linked to https://github.com/w3c/wcag-act-rules/issues/134 which has pretty much the same description:
> reporting against AAA criteria should not be required by tools that don't support AAA criteria.

The later is marked as fixed by https://github.com/w3c/wcag-act/pull/531 which updates the format (to allow secondary requirements) but doesn't change the tool consistency checking by discarding AAA mapping for AA tools or something like that.

---
How will that not recreate the same situation?
Tools/methods that only check AA level won't report AAA SC, hence they won't report 1.4.9 on these test cases. Which part of our machinery (something in `act-tools`?) will let them claim a consistent implementation of the rule?

pages/design/rule-design.md
```diff
@@ -63,6 +63,21 @@ The description is a short synopsis of what the rule does. This should be declar
...
+
+1. The requirement is more strict than the rule. If this occurs, and some but not all passed examples of the rule fail the accessibility requirement, the following test should be used:
+
+   > secondary: This success criterion is **more strict** than this rule. This is because {reason}. Some of the passed examples do not satisfy this success criterion.
+
+1. The requirement is related to, but different from the rule. This often happens when there is overlap between two accessibility requirements. If this occurs, and some but not all the failed examples of the rule fail the accessibility requirement, the following text should be used:
+   > This success criterion is **related** to this rule. This {is because / happens for} {reason}. Most failed examples satisfy this success criterion.
+
+Secondary requirements are optional, and are only necessary when fails some of the passed examples, or passes some of the failed examples. If there are no such examples, it is generally not necessary to include secondary requirements. Atomic rules used in composite rules should list the requirements from the composite rule as "more strict" secondary requirements.
```
kengdoj:
```suggestion

1. Atomic rules used in composite rules should list the requirements from the composite rule as "more strict" secondary requirements.

Secondary requirements are optional and are only necessary when a requirement fails some of the passed examples or passes some of the failed examples. If there are no such examples, it is generally not necessary to include secondary requirements. 
```

I like the last sentence and think it needs more attention, so suggest making it one of the list items. If this change is made, then line 66 should be updated to "four" instead of "three".

_rules/meta-viewport-b4f0c3.md
```diff
@@ -11,7 +11,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag21:1.4.10: # Reflow (AA)
-    secondary: true
+    secondary: This success criterion is **related** to this rule. This is because a page that cannot be zoomed up to 200% often does not reflow sufficiently either. Most failed examples satisfy this success criterion.
```
kengdoj, line 14: "Most failed examples satisfy this success criterion" is true of 1.4.4 also.

Should this rule be deprecated if it's only a problem in older mobile browsers?

_rules/text-contrast-enhanced-09o5cg.md
```diff
@@ -21,7 +21,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.4.3: # Contrast (Minimum) (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because this criterion has a lower minimum contrast. This is why some of the failed examples may satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because this criterion has a lower minimum contrast. Some of the failed examples may satisfy this success criterion.
```

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -16,9 +16,9 @@ accessibility_requirements:
     passed: satisfied
     inapplicable: satisfied
   wcag20:1.3.1: # Info and Relationships (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because browsers and assistive technologies will often fall back on a non-standard default value, which may be sufficient. This is why some of the failed examples may satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because browsers and assistive technologies will often fall back on a non-standard default value, which may be sufficient. Some of the failed examples may satisfy this success criterion.
```

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -16,9 +16,9 @@ accessibility_requirements:
...
   wcag20:4.1.2: # Name, Role, Value (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because browsers and assistive technologies will often fall back on a non-standard default value, which may be sufficient. This is why some of the failed examples may satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because browsers and assistive technologies will often fall back on a non-standard default value, which may be sufficient. Some of the failed examples may satisfy this success criterion.
```

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -16,9 +16,9 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.3.1: # Info and Relationship (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because the success criterion can be satisfied by an element's implicit role when the explicit role is incorrect. This is why some of the failed examples may satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because the success criterion can be satisfied by an element's implicit role when the explicit role is incorrect. Some of the failed examples may satisfy this success criterion.
```

_rules/role-attribute-valid-value-674b10.md
```diff
@@ -16,9 +16,9 @@ accessibility_requirements:
...
   wcag20:4.1.2: # Name, Role, Value (A)
-    secondary: true
+    secondary: This success criterion is **less strict** than this rule. This is because the success criterion can be satisfied by an element's implicit role when the explicit role is incorrect. This is why some of the failed examples may satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because the success criterion can be satisfied by an element's implicit role when the explicit role is incorrect. Some of the failed examples may satisfy this success criterion.
```

_rules/link-non-empty-accessible-name-c487ae.md
```diff
@@ -27,7 +26,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:1.1.1: # Non-text content (A)
-    secondary: true
+    secondary: This success criterion is **related** to this rule. This happens for HTML `area` elements which are both links and as non-text content. Most failed examples satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **related** to this rule. This is because HTML `area` elements are both links and non-text content. Most failed examples satisfy this success criterion.
```

_rules/meta-refresh-no-delay-no-exception-bisz58.md
```diff
@@ -26,7 +26,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   wcag20:2.2.1: # Timing Adjustable (A)
-    secondary: True
+    secondary: This success criterion is **less strict** than this rule. This is because this criterion allows redirects longer than 20 hours. This is why some of the failed examples satisfy this success criterion.
```
kengdoj:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because this criterion allows redirects longer than 20 hours. Some of the failed examples satisfy this success criterion.
```

pages/design/rule-design.md
```diff
@@ -63,6 +63,21 @@ The description is a short synopsis of what the rule does. This should be declar
 
 List all accessibility requirements that are not satisfied if one or more outcomes of a rule fails. For example, this could be one or more WCAG success criteria, or a WAI-ARIA rule. Each requirement should indicate the associated conformance level, and map the implications of each outcome, such as "satisfied", "not satisfied", or "further testing needed".
 
+Accessibility requirements should be listed as secondary if the requirement is related to the rule. There are three scenarios:
+
+1. The requirement is less strict than the rule. If this occurs, and some but not all failed examples of the rule pass the accessibility requirement, the following test should be used:
+
+   > This success criterion is **less strict** than this rule. This is because {reason}. Some of the failed examples satisfy this success criterion.
```
kengdoj, line 70:
```suggestion
   > This success criterion is **less strict** than this rule. This is because {reason}. Some of the failed examples may satisfy this success criterion.
```

kengdoj (changes requested):
mostly editorial, and one suggestion to deprecate the meta viewport rule.

_rules/meta-viewport-b4f0c3.md
WilcoFiers: Discussed, leaving this for a different PR.

pages/design/rule-design.md
WilcoFiers: Discussed, not using "may".

pages/design/rule-design.md
```diff
@@ -63,6 +63,21 @@ The description is a short synopsis of what the rule does. This should be declar
 
 List all accessibility requirements that are not satisfied if one or more outcomes of a rule fails. For example, this could be one or more WCAG success criteria, or a WAI-ARIA rule. Each requirement should indicate the associated conformance level, and map the implications of each outcome, such as "satisfied", "not satisfied", or "further testing needed".
 
+Accessibility requirements should be listed as secondary if the requirement is related to the rule. There are three scenarios:
```
WilcoFiers:
```suggestion
Accessibility requirements should be listed as secondary if the requirement is related to the rule. There are four scenarios:
```

__tests__/frontmatter.js
WilcoFiers: That's not quite the same. Doing it the way you suggests triggers the tests on line 107 / 108, which would then fail.

tbostic32 (approved):

HelenBurge (approved):

kengdoj (approved):

_rules/image-no-text-0va7u6.md
WilcoFiers: There are two reasons why an implementation may not report an accessibility requirement and still be consistent:
1. The requirement is secondary
2. The requirement is of a conformance level the implementation does not support

I think what you were expecting is for secondary requirements to cover both situations. That's not the intent. This is exactly an example of why that difference is necessary. If an implementation supports AAA, then it must report both 1.4.5 and 1.4.9 as not satisfied when this rule fails, Otherwise it is partially consistent. An implementation that only covers AA is not required to report 1.4.9.

Whether or not it can report 1.4.9 is an open question, and is possibly a topic for TPAC.

carlosapaduarte (approved):

colabottles (approved):

_rules/image-no-text-0va7u6.md
```diff
@@ -56,7 +55,7 @@ There are no accessibility support issues known.
 
 ## Background
 
-This rule is designed specifically for [SC 1.4.5 Images of Text][sc1.4.5] which includes exceptions to the images it applies to that are not part of [SC 1.4.9 Images of Text (No Exception)][sc1.4.9]. Therefore, some images that are inapplicable for this rule can be applicable to [SC 1.4.9 Images of Text (No Exception)][sc1.4.9].
+This rule is designed specifically for [SC 1.4.5 Images of Text][sc1.4.5]. There are however only minimal differences between this criterion and [SC 1.4.9 Images of Text (No Exception)][sc1.4.9]. The two differences are that customizable images of text are allowed, and that images of text are allowed when the presentation cannot otherwise be achieved. These scenarios are so rare the rule ignores them as part of the assumptions, and so the [accessibility requirements mapping](#accessibility-requirements-mapping) of these two criteria is the same.
```
bruce-usab, line 58: While I agree that SC 1.4.5 and 1.4.9 are very similar, I do not agree with characterization of "only minimal differences" because using the Essential exception for logotypes is _very_ common.

But maybe there is a precondition for the applicability of this test?

bruce-usab (approved):
Approve, but please see my inline comment/question regarding logotypes and the Essential Exception for 1.4.5.

_rules/aria-attr-defined-5f99a7.md
bruce-usab: I like the change from saying "the rule more strict" (than the SC)  to "the SC is less strict" (than the rule).

pages/design/rule-design.md
```diff
@@ -63,6 +63,24 @@ The description is a short synopsis of what the rule does. This should be declar
 
 List all accessibility requirements that are not satisfied if one or more outcomes of a rule fails. For example, this could be one or more WCAG success criteria, or a WAI-ARIA rule. Each requirement should indicate the associated conformance level, and map the implications of each outcome, such as "satisfied", "not satisfied", or "further testing needed".
 
+Accessibility requirements should be listed as secondary if the requirement is related to the rule. There are four scenarios:
+
+1. The requirement is less strict than the rule. If this occurs, and some but not all failed examples of the rule pass the accessibility requirement, the following test should be used:
```
Jym77:
```suggestion
1. The requirement is less strict than the rule. If this occurs, and some but not all failed examples of the rule pass the accessibility requirement, the following text should be used:
```

pages/design/rule-design.md
```diff
@@ -63,6 +63,24 @@ The description is a short synopsis of what the rule does. This should be declar
...
+
+   > This success criterion is **less strict** than this rule. This is because {reason}. Some of the failed examples satisfy this success criterion.
+
+1. The requirement is more strict than the rule. If this occurs, and some but not all passed examples of the rule fail the accessibility requirement, the following test should be used:
```
Jym77:
```suggestion
1. The requirement is more strict than the rule. If this occurs, and some but not all passed examples of the rule fail the accessibility requirement, the following text should be used:
```

_rules/image-no-text-0va7u6.md
Jym77: I still do not understand why 1.4.9 shouldn't be a **more strict** secondary requirement 🤔 

How is the 1.4.5/1.4.9 case (and the 2.1.1/2.1.3 one) different from the 2.4.4/2.4.9, the 1.4.3/1.4.6, the 1.2.3/1.2.5-1.2.8, or the 2.2.1/2.2.4-3.2.5 ones?
This PR keeps a **more strict** secondary requirement for the 4 later cases, but changes the first two as primary requirements instead. What is the criterion to decide that a more strict SC is a secondary or primary requirement? (shouldn't we instead try and find Passed Examples that fail the more strict SC?)

I won't block the PR due to that, it seems I'm misunderstanding something. I guess we can talk through that at TPAC.

_rules/scrollable-element-keyboard-accessible-0ssw9k.md
```diff
@@ -12,7 +12,6 @@ accessibility_requirements:
     inapplicable: further testing needed
   wcag20:2.1.3: # Keyboard (No Exceptions) (AAA)
     forConformance: true
-    secondary: true
```
Jym77, line 15: Also not sure why this shouldn't be a **more strict** secondary requirement instead.

Jym77 (commented):
* There are a couple of typos in the design document. These should be fixed but aren't a huge deal either.
* I still do not understand why some more strict SC are listed as **more strict** secondary requirement, and some are listed as primary requirements (without the less strict SC being listed as **less strict** secondary requirement either). Won't block because of that, though.
* I love the "design" bit 😄 Good work in clarifying the situation.

_rules/image-no-text-0va7u6.md
WilcoFiers: I'll take another crack at explaining, but yes, lets take this to TPAC.

The way I think of it, and why I ended up doing it this way is because this rule effectively treats those 1.4.5 and 1.4.9 as identical criteria. The assumption effectively says there is no difference, and so they are tested in the same way, and only require one rule. Compare that to 2.2.1 / 2.2.4. There are two rules for it. One with the 20 hour exception and one without. Similarly 1.4.3 / 1.4.6. Two rules, one with a 4.5:1 contrast, and one with a 7:1 contrast.

I can explain why the rule for 1.4.3 is less strict than 1.4.6: The number is lower. I can explain why the rule for 2.2.1 is less strict than 2.2.4: The rule has a 20 hour exception. There is no difference between how 1.4.5 is measured and how 1.4.9 is measured. The rule assume it.

_rules/image-no-text-0va7u6.md
WilcoFiers: I think the confusion here comes from this "No exception" thing in the title of that SC. 1.4.9 does allow text where presentation is essential (like logos) and text that is purely decorative. The only difference between 1.4.5 and 1.4.9 is that 1.4.5 allows images of text if the text can be customised.

Issue #2030, opened 2023-02-08
autoplay rule [x0paj4] failed examples don't autoplay in major browsers


carlosapaduarte:
https://github.com/act-rules/act-rules.github.io/pull/1655#discussion_r1098674794 raises the issue about the failed example not working in Chromium browsers. A quick test reveals that the same happens in Firefox and Safari.

The way I see it, we have two choices:

- Deprecate the rule, since browsers have addressed the accessibility issue caught by this rule
- Change the applicability as suggested by @WilcoFiers in https://github.com/act-rules/act-rules.github.io/pull/1655#discussion_r1098674794, i.e., make it applicable to media elements that get turned on by scripts within 3 seconds of the page loading

Jym77:
While building the Alfa semi-automated implementation, I had mixed result with that…
Sometimes the media did autoplay (often when opening the example from the rule page), sometimes it didn't (often when entering the URL directly).
I was also going to open such an issue, but more attempts show that stuff did autoplay sometimes 😖 And I'm not sure when exactly (this may also depend on devtools being open or not, on headset being connected or not, … 🙃 ) Tried both with Chrome and Firefox (on Windows 10).

carlosapaduarte:
I'll check support for audio autoplay and for scripted auto-play and report back here

carlosapaduarte:
I've run the tests. Results are the same for `audio` and `video` elements with audio content.

### Autoplay
- Firefox and Safari do not autoplay media content unless the user has instructed the browser to do so
- Chrome tries to learn the user behaviour and acts accordingly (which might explain why it [sometimes autoplayed](https://github.com/act-rules/act-rules.github.io/issues/2030#issuecomment-1424205479) for @Jym77). Chrome computes a [Media Engagement Index](https://developer.chrome.com/blog/autoplay/#media-engagement-index) to measure a user's "propensity to consume media on a site", and decides based on that to autoplay. I searched for this when, all of sudden, the examples that did not autoplay, started to autoplay after I manually played them.

### Scripted autoplay
- Chrome, Firefox and Safari all block it unless the user has instructed the browser to authorise it 
- This happens if there is an attempt to play the media before the user has interacted with the webpage, irrespective of the script running on load or after a timeout

Jym77:
[MDN usage note on `play()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play#usage_notes):
[...]

PR #2064, opened 2023-05-25
Updating the Transcript Rules

HelenBurge:
Are being updated from Task force feedback

Amending the rule from TF feedback: https://docs.google.com/spreadsheets/d/1OSkPFocXk4K3zYLnwS78WLsWO4PvE5yRcsauyefuIUI/edit#gid=1605073031

Closes issue(s):

Need for Call for Review:

---

## How to Review And Approve


_rules/audio-transcript-2eb176.md
```diff
@@ -5,6 +5,12 @@ rule_type: atomic
 description: |
   This rule checks that `audio` elements have a transcript that includes all auditory information.
 accessibility_requirements:
+  wcag20:1.2.1: # Audio-only and Video-only (Prerecorded) (A)
+    forConformance: true
+    secondary: true
```
Jym77: Should we start putting a reason here, as done in #2060 ?

_rules/audio-transcript-2eb176.md
```diff
@@ -5,6 +5,12 @@ rule_type: atomic
...
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
```
Jym77: Iirc, we do not have mappings for secondary requirements 🤔

_rules/audio-transcript-2eb176.md
```diff
@@ -24,12 +31,12 @@ acknowledgments:
 
 This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element that is:
 
-- playing; or,
-- has a [play button][] that is [visible][] and [included in the accessibility tree][].
+- autoplaying; or,
```
Jym77: What does "autoplaying" means? [Audio or video element avoids automatically playing audio](https://www.w3.org/WAI/standards-guidelines/act/rules/80f0bf/proposed/) has a bunch of conditions for it.

Also, if the video plays but not automatically (e.g. after user interaction, but maybe that interaction hides the play button thus removing the second condition), would the rule become inapplicable?

_rules/audio-transcript-2eb176.md
```diff
@@ -24,12 +31,12 @@ acknowledgments:
...
+- has a [play button][] that is [visible][].
 
 ## Expectation
 
-The auditory information of each test target is available through a text transcript. That text transcript is [visible][] and [included in the accessibility tree][], either on the page or through a link.
+The auditory information of each test target is available through a text transcript. That text transcript is [visible][], either on the page or through a clearly labeled location.
```
Jym77:
```suggestion
The auditory information of each test target is available through a text transcript. That text transcript is [visible][], either on the page or through a [clearly labeled location][].
```
Plus add
```
[clearly labeled location]: #clearly-labeled-location 'Definition of Clearly Labeled Location'
```
in the references list at the end.

Jym77: I'm not sure why it is OK to have a transcript that is not included in the accessibility tree.
Wouldn't that be problematic, e.g., for Braille display users?

_rules/video-only-transcript-ee13b5.md
```diff
@@ -10,6 +10,12 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:1.2.1: # Audio-only and Video-only (Prerecorded) (A)
+    forConformance: true
+    secondary: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
```
Jym77: Same remark.

_rules/video-only-transcript-ee13b5.md
```diff
@@ -36,7 +43,7 @@ This rule applies to any [non-streaming](#non-streaming-media-element) `video` e
 
 ## Expectation
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
```
Jym77, line 44: Same remarks.

For a transcript describing the visual content, the exclusion from the accessibility tree would also be problematic for screen reader users that have then no way to know what is the visual information in the video.

_rules/video-transcript-1a02b0.md
```diff
@@ -41,7 +42,7 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
 
 ## Expectation
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
```
Jym77, line 45: Same remarks.

Here also, non-accessible transcript is problematic for screen reader users.

Jym77 (changes requested):

_rules/audio-transcript-2eb176.md
HelenBurge: I will assign these questions to @WilcoFiers as not sure on the secondary requirements needs.

_rules/audio-transcript-2eb176.md
HelenBurge: I will assign these questions to @WilcoFiers as not sure on the secondary requirements needs.

_rules/audio-transcript-2eb176.md
WilcoFiers: I'll handle it in my PR. Don't worry about it here. @HelenBurge maybe just remove this change from the PR and let me handle it on my PR.

_rules/audio-transcript-2eb176.md
WilcoFiers: As above

_rules/audio-transcript-2eb176.md
HelenBurge: I updated the PR description with how we decided on this in a meeting.

_rules/audio-transcript-2eb176.md
HelenBurge: I believe it was due to the fact the transcript might not be included in the current page and therefore not in the accessibility tree.

_rules/video-only-transcript-ee13b5.md
HelenBurge: I believe it was due to the fact the transcript might not be included in the current page and therefore not in the accessibility tree.

_rules/video-transcript-1a02b0.md
HelenBurge: I believe it was due to the fact the transcript might not be included in the current page and therefore not in the accessibility tree.

_rules/audio-transcript-2eb176.md
Jym77: I have no problem with using autoplay in the rule, but it needs to be somewhat defined or clarified. In #2030 and the following discussion, we've seen that "autoplay" is not as easy as we could think.

Does that refers to the `autoplay` attribute or something else? How do we handle the fact that browsers do not always respect it (due to engagement index, …)?

I think that "autoplaying" in somewhat ambiguous and needs to be clarified.

_rules/audio-transcript-2eb176.md
```diff
@@ -24,12 +31,12 @@ acknowledgments:
 
 This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element that is:
 
-- playing; or,
-- has a [play button][] that is [visible][] and [included in the accessibility tree][].
+- autoplaying; or,
+- has a [play button][] that is [visible][].
```
Jym77: I do not really agree with the removal of the inclusion in the accessibility tree.
Looking at the survey report, the argument was that this might be a 1.3.1 failure rather than a 1.2.1. However, looking at the [Understanding document](https://www.w3.org/WAI/WCAG21/Understanding/audio-only-and-video-only-prerecorded.html) (which is not normative, but still very helpful) (emphasizes mine):
* The first sentence of the Intent is:
  > The intent of this Success Criterion is to make information conveyed by prerecorded audio-only and prerecorded video-only content **available to all users**.

  "available to all users" does imply that it is in the accessibility tree, since it would otherwise not be available to AT users.
  Especially, if a silent video has a hidden transcript, this does not provide any kind of alternative for blind users and I think that would fail 1.2.1.
* The first Benefit is:
  > This Success Criterion helps people who have **difficulty perceiving visual content**. Assistive technology can read text alternatives aloud, present them visually, or **convert them to braille**.

  People who have difficulties with visual content would not benefit at all from an alternative that is not included in the accessibility tree; as far as I know, braille displays also work with the accessibility tree and won't be able to present the alternative.
* The fourth benefit is:
  > People who are deaf-blind can read the text in braille.

  Again, not sure that this is possible if the alternative is not included in the accessibility tree.

I'm rather of the opinion that having an inaccessible alternative is actually a failure of 1.2.1 and thus should be part of this rule. This is maybe something that we need to discuss further.

Jym77 (dismissed):

daniel-montalvo:
Discussed at the TF meeting 29 Feb
https://www.w3.org/2024/02/29-wcag-act-minutes.html#t03

_rules/audio-transcript-2eb176.md
Jym77:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because the rule does not consider that the audio may be a media alternative for text. Some of the failed examples satisfy this success criterion.
```

_rules/audio-transcript-2eb176.md
```diff
@@ -24,25 +31,27 @@ acknowledgments:
 
 This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element that is:
 
-- playing; or,
-- has a [play button][] that is [visible][] and [included in the accessibility tree][].
+- [autoplaying](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#autoplay); or,
```
Jym77: We normally avoid quoting from non-W3C sites, especially vendor ones. If this is just about the `autoplay` attribute like the "3 seconds" rule:
```suggestion
- the element has an `autoplay` [attribute value][] of `true`; or
```

_rules/audio-transcript-2eb176.md
```diff
@@ -24,25 +31,27 @@ acknowledgments:
 
 This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element that is:
```
Jym77:
```suggestion
This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element for which at least one of the following is true:
```
(taking the opportunity to update to how we write these now)

_rules/audio-transcript-2eb176.md
```diff
@@ -24,25 +31,27 @@ acknowledgments:
...
 
-- playing; or,
-- has a [play button][] that is [visible][] and [included in the accessibility tree][].
+- [autoplaying](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#autoplay); or,
+- has a [play button][] that is [visible][].
```
Jym77:
```suggestion
- the element has a [play button][] that is [visible][].
```

_rules/audio-transcript-2eb176.md
```diff
@@ -24,25 +31,27 @@ acknowledgments:
...
-There are no assumptions.
+This rule assumes that if a transcript is present it is [included in the accessibility tree][].
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+It is known that the user can set the browser to prevent autoplay. This rule assumes the browser has not been set to do this.
```
Jym77:
```suggestion
Some major browsers do not automatically play the 'video' unless the 'video' is muted. For these, the rule is never applicable. But for browsers that autoplay not muted video, the rule is still applicable.
```
(copying from #1655 for consistency)

_rules/audio-transcript-2eb176.md
```diff
@@ -156,21 +165,6 @@ This `audio` element has native player controls and a [non-visible][visible] int
 </html>
 ```
 
-#### Failed Example 6
```
Jym77, line 159: I'd like to have a Passed example with a transcript that is only visible but not included in the accessibility tree.
This serves as documentation that the CG and TF have reached the conclusion that it is not necessarily a failure of 1.2.1.

If inaccessible transcript is always a failure of 1.2.1, I don't think we should remove it for the rule. We can add 1.3.1 as a secondary requirement if needed (to allow implementations that report it to have a consistent implementation), or we need a separate rule checking that transcripts are accessible (but I don't see plan for that other rule yet 🤔).

_rules/video-only-transcript-ee13b5.md
```diff
@@ -10,6 +10,12 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:1.2.1: # Audio-only and Video-only (Prerecorded) (A)
+    forConformance: true
+    secondary: true
```
Jym77:
```suggestion
    secondary: This success criterion is **less strict** than this rule. This is because the rule does not consider that the video may be a media alternative for text. Some of the failed examples satisfy this success criterion.
```

_rules/video-only-transcript-ee13b5.md
```diff
@@ -36,19 +43,19 @@ This rule applies to any [non-streaming](#non-streaming-media-element) `video` e
 
 ## Expectation
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
 
 ## Assumptions
 
-A mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+A mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster). This rule also assumes that if a transcript is present it is [included in the accessibility tree][].
```
Jym77: Putting it as an assumption for something that is so easily testable sounds weird 🤔

_rules/video-transcript-1a02b0.md
```diff
@@ -41,20 +42,22 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
...
-This rule assumes that a mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+This rule assumes that a mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster). The rule also assumes that if a transcript is present it is [included in the accessibility tree][].
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+It is known that the user can set the browser to prevent autoplay. This rule assumes the browser has not been set to do this.
```
Jym77:
```suggestion
Some major browsers do not automatically play the 'video' unless the 'video' is muted. For these, the rule is never applicable. But for browsers that autoplay not muted video, the rule is still applicable.
```
(but this rule doesn't seem to be about autoplay, so this probably shouldn't even be here in the first place)

Jym77 (changes requested):
I'm still not super happy with the inclusion in the accessibility tree part 😅 

---
From [the TF discussion](https://www.w3.org/2024/02/29-wcag-act-minutes.html#t03), I think I understand that TF reached the consensus that an inaccessible transcript is **not always** a 1.2.1 failure.

If this is indeed the case, then yes, we need to remove it (and the corresponding Failed Example). But in this case I'd like us to have a **Passed** example with an accessible transcript that is not failing 1.2.1. This will serve as documentation of this decision and essentially say: "no, inclusion in the accessibility tree is not required, look at this". (and if people disagree with this decision in the future, we'll have the example to foster concrete discussions instead of doing archaeology in Github history.

If I'm reading that wrong and that the issue with inaccessible transcripts is that they fail **both 1.2.1 and 1.3.1**, then I think that a better solution would be to add 1.3.1 as a **related** secondary requirement; this is similar to the [secondary requirement to 1.1.1 for "link have accessible name" (due to `area` elements)](https://github.com/act-rules/act-rules.github.io/blob/develop/_rules/link-non-empty-accessible-name-c487ae.md?plain=1#L28-L29)

_rules/video-only-transcript-ee13b5.md
HelenBurge: This was what we decided was the best way to fix the concern. Basically, we are not testing the fact it is in the accessibility tree for 1.2.1 but assume it is there.

_rules/audio-transcript-2eb176.md
HelenBurge: The assumption states the transcript is part of the accessibility tree so is this still true?

_rules/audio-transcript-2eb176.md
```diff
@@ -5,6 +5,12 @@ rule_type: atomic
 description: |
   This rule checks that `audio` elements have a transcript that includes all auditory information.
 accessibility_requirements:
+  wcag20:1.2.1: # Audio-only and Video-only (Prerecorded) (A)
+    forConformance: true
+    secondary: This success criterion is **less strict** than this rule. This is because the rule does not consider that the audio may be a media alternative for text. Some of the failed examples satisfy this success criterion.
```
Jym77:
```suggestion
  wcag20:1.2.1: # Audio-only and Video-only (Prerecorded) (A)
    secondary: This success criterion is **less strict** than this rule. This is because the rule does not consider that the audio may be a media alternative for text. Some of the failed examples satisfy this success criterion.
```

Jym77 (dismissed):

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
...
-There are no assumptions.
+This rule assumes that if a transcript is present it is [included in the accessibility tree][].
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless the 'video' is muted. For these, the rule is never applicable. But for browsers that autoplay not muted video, the rule is still applicable.
```
daniel-montalvo:
```suggestion
Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule stil applies to browsers that autoplay not muted 'video'.
```

daniel-montalvo (dismissed):
Just minor editorials on my part

_rules/video-transcript-1a02b0.md
```diff
@@ -41,20 +42,22 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
...
-This rule assumes that a mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+This rule assumes that a mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster). The rule also assumes that if a transcript is present it is [included in the accessibility tree][].
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless the 'video' is muted. For these, the rule is never applicable. But for browsers that autoplay not muted video, the rule is still applicable.
```
daniel-montalvo:
```suggestion
Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay not muted 'video'.
```
Sorry that I skipped this file earlier.

daniel-montalvo (dismissed):
Just adding the suggestion to the other occurrence of this paragraph, that I had skipped earlier.

tombrunet (approved):

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
...
 
-The auditory information of each test target is available through a text transcript. That text transcript is [visible][] and [included in the accessibility tree][], either on the page or through a link.
+The auditory information of each test target is available through a text transcript. That text transcript is [visible][], either on the page or through a [clearly labeled location][].
 
 **Note:** A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
 
 ## Assumptions
 
-There are no assumptions.
+This rule assumes that if a transcript is present it is [included in the accessibility tree][].
```
Jym77: It feels more like a [Background note](https://www.w3.org/TR/act-rules-format/#background) ("information about the background for the development of the rule") than an [Assumption](https://www.w3.org/TR/act-rules-format/#assumptions) 🤔 (sorry 😅)

How about moving it to Background and phrasing it like:
```suggestion
Transcripts need to be available for all users, not just sighted users in order to satisfy WCAG. This rule does not test this, and it needs to be tested separately.
```

(and do the same with the other rules)

Jym77 (commented):
Not daring to reject it again… So just leaving it as a comment 😐

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
...
 
-The auditory information of each test target is available through a text transcript. That text transcript is [visible][] and [included in the accessibility tree][], either on the page or through a link.
+The auditory information of each test target is available through a text transcript. That text transcript is [visible][], either on the page or through a [clearly labeled location][].
 
 **Note:** A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
 
 ## Assumptions
 
-There are no assumptions.
+Transcripts need to be available for all users, not just sighted users in order to satisfy WCAG. This rule does not test this, and it needs to be tested separately.
```
dan-tripp-siteimprove:
```suggestion
In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```
Rearranging this sentence.  My reason is that with the current arrangement, someone might think that it means this: `In order to satisfy WCAG, transcripts only need to be available to sighted users.  But WCAG doesn't go far enough - really, transcripts need to be available to all users.`   That's not true.  My suggestion clears it up.

_rules/video-only-transcript-ee13b5.md
```diff
@@ -36,19 +39,19 @@ This rule applies to any [non-streaming](#non-streaming-media-element) `video` e
 
 ## Expectation
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
 
 ## Assumptions
 
-A mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+Transcripts need to be available for all users, not just sighted users in order to satisfy WCAG. This rule does not test this, and it needs to be tested separately.
```
dan-tripp-siteimprove:
```suggestion
In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```
Same as my similar comment from a minute ago.

_rules/video-transcript-1a02b0.md
```diff
@@ -41,20 +42,22 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
...
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
 
 **Note:** A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
 
 ## Assumptions
 
-This rule assumes that a mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+Transcripts need to be available for all users, not just sighted users in order to satisfy WCAG. This rule does not test this, and it needs to be tested separately.
```
dan-tripp-siteimprove:
```suggestion
In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```
Same as my similar comment from a minute ago.

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
...
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay not muted 'video'.
```
dan-tripp-siteimprove:
```suggestion
Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay unmuted 'video'.
```

_rules/video-only-transcript-ee13b5.md
```diff
@@ -36,19 +39,19 @@ This rule applies to any [non-streaming](#non-streaming-media-element) `video` e
...
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
-A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
+A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
```
dan-tripp-siteimprove:
```suggestion
A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time-based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media). 
```

_rules/video-transcript-1a02b0.md
```diff
@@ -41,20 +42,22 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
...
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay not muted 'video'.
```
dan-tripp-siteimprove:
```suggestion
Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay unmuted 'video'.
```

dan-tripp-siteimprove (dismissed):

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element that is:
+This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element for which at least one of the following is true:
 
-- playing; or,
-- has a [play button][] that is [visible][] and [included in the accessibility tree][].
+- the element has an `autoplay` [attribute value][] of `true`; or
+- the element has a [play button][] that is [visible][].
```
kengdoj: why was "and included in the accessibility tree" removed here?

HelenBurge: Because from the survey discussion we agreed that to test the rule 1.2.1 the AT part is for a different checkpoint so it was made as an assumption instead: https://docs.google.com/spreadsheets/d/1OSkPFocXk4K3zYLnwS78WLsWO4PvE5yRcsauyefuIUI/edit#gid=1605073031

_rules/video-transcript-1a02b0.md
HelenBurge: amending this as got the spell checker unhappy!

_rules/audio-transcript-2eb176.md
kengdoj:
```suggestion
- the element has a [play button][] that is [visible][] and [included in the accessibility tree][].
```

_rules/audio-transcript-2eb176.md
kengdoj: Agree to keep this example but as passed

HelenBurge: Moved up :)

_rules/audio-transcript-2eb176.md
```diff
@@ -5,6 +5,8 @@ rule_type: atomic
 description: |
   This rule checks that `audio` elements have a transcript that includes all auditory information.
 accessibility_requirements:
```
kengdoj, line 7: Add SC 1.3.1 as a secondary requirement because it only applies when the transcript is visible.

Reference [Rules Format 1.1 Example 10](https://w3c.github.io/wcag-act/act-rules-format.html#example-cfb70282)

_rules/video-transcript-1a02b0.md
```diff
@@ -145,5 +148,4 @@ A video element that is not [visible][].
 </html>
 ```
 
-[included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of included in the accessibility tree'
```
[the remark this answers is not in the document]
kengdoj: I'm not seeing this added back in yet so have unresolved this.

dan-tripp-siteimprove (approved):

_rules/audio-transcript-2eb176.md
HelenBurge: We did not include 1.3.1 as made the transcript as an assumption - I am not sure this should be done.

kengdoj: Since we added Passed Ex 4, we would want to add 1.3.1 as secondary so an implementation that fails 1.3.1 is still consistent with this rule.

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
...
-There are no assumptions.
+In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay an audible 'video'.
```
kengdoj, line 49: I don't think this is needed. The applicability includes a play button, so the rule could still apply if the audio does not autoplay.

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +25,29 @@ acknowledgments:
...
 
-The auditory information of each test target is available through a text transcript. That text transcript is [visible][] and [included in the accessibility tree][], either on the page or through a link.
+The auditory information of each test target is available through a text transcript. That text transcript is [visible][], either on the page or through a [clearly labeled location][].
 
 **Note:** A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
 
 ## Assumptions
 
-There are no assumptions.
+In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```
kengdoj:
```suggestion
If a transcript is present, it is assumed it is in the accessibility tree. In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```

_rules/video-only-transcript-ee13b5.md
```diff
@@ -36,19 +39,19 @@ This rule applies to any [non-streaming](#non-streaming-media-element) `video` e
 
 ## Expectation
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
 
 ## Assumptions
 
-A mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```
kengdoj:
```suggestion
If a transcript is present, it is assumed it is in the accessibility tree. In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```

kengdoj:
I couldn't add this in my review comments - Like the change you made in 2eb176, please change this failed example 4 to passed example 3 in ee13b5. 

#### Failed Example 4
This `video` element, which has no audio, has a text transcript available on the same page, but the transcript is not [included in the accessibility tree][].
```html
<html lang="en">
<video controls>
  <source src="/test-assets/rabbit-video/silent.mp4" type="video/mp4"></source>
  <source src="/test-assets/rabbit-video/silent.webm" type="video/webm"></source>
</video>
<p aria-hidden="true">The above video shows a giant fat rabbit climbing out of a hole in the ground.
He stretches, yawns, and then starts walking.
Then he stops to scratch his bottom.</p>
</html>
```

_rules/video-transcript-1a02b0.md
```diff
@@ -41,20 +42,22 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
...
 
-The visual information of each test target is available through a text transcript that is [visible][], [included in the accessibility tree][], and is either on the page or linked.
+The visual information of each test target is available through a text transcript that is [visible][], and is either on the page or in a clearly labeled location.
 
 **Note:** A "text transcript" in the context of this rule is defined in WCAG 2 as an [alternative for time based media](https://www.w3.org/TR/WCAG22/#dfn-alternative-for-time-based-media).
 
 ## Assumptions
 
-This rule assumes that a mechanism is available to start the video and that the video element is not simply used to display the [poster](https://www.w3.org/TR/html5/semantics-embedded-content.html#element-attrdef-video-poster).
+In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```
kengdoj:
```suggestion
If a transcript is present, it is assumed it is in the accessibility tree. In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
```

_rules/video-transcript-1a02b0.md
```diff
@@ -41,20 +42,22 @@ This rule applies to every [non-streaming](#non-streaming-media-element) `video`
...
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that autoplay an audible 'video'.
```
kengdoj: Autoplay isn't in the applicability, so I don't think this is needed.

kengdoj (dismissed):

_rules/video-transcript-1a02b0.md
HelenBurge: Amended it to add the definition of the text transcript - as I think this is applicable here.

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +27,29 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element that is:
+This rule applies to every [non-streaming](#non-streaming-media-element) `audio` element for which at least one of the following is true:
 
-- playing; or,
-- has a [play button][] that is [visible][] and [included in the accessibility tree][].
+- the element has an [autoplay][] [attribute value][] of `true`; or
```
carlosapaduarte, line 32: Since autoplay here is referring to the HTML attribute I don't think it needs to link to the definition.

_rules/audio-transcript-2eb176.md
```diff
@@ -81,7 +88,7 @@ This `audio` element has native player controls and an external transcript.
 
 #### Passed Example 3
 
-This `audio` element has an `autoplay` attribute and an external transcript.
+This `audio` element has an [autoplay][] attribute and an external transcript.
```
carlosapaduarte, line 91: Since autoplay here is referring to the HTML attribute I don't think it needs to link to the definition.

_rules/audio-transcript-2eb176.md
```diff
@@ -131,7 +154,7 @@ This `audio` element has native player controls and an incorrect external transc
 
 #### Failed Example 4
 
-This `audio` element has an `autoplay` attribute and an incorrect external transcript.
+This `audio` element has an [autoplay][] attribute and an incorrect external transcript.
```
carlosapaduarte, line 157: Since autoplay here is referring to the HTML attribute I don't think it needs to link to the definition.

_rules/audio-transcript-2eb176.md
```diff
@@ -22,27 +27,29 @@ acknowledgments:
...
-There are no assumptions.
+If a transcript is present, it is assumed it is in the accessibility tree. In order to satisfy WCAG, transcripts need to be available for all users, not just sighted users. This rule does not test this, and it needs to be tested separately.
 
 ## Accessibility Support
 
 There are no accessibility support issues known.
 
 ## Background
 
+Some major browsers do not automatically play the 'video' unless it is muted, which makes the rule inapplicable in these browsers. The rule still applies to browsers that [autoplay][] an audible 'video'.
```
carlosapaduarte, line 51: This one should link to the autoplay definition, but since it is being used in the Background section, the definition doesn't need to be objective.

carlosapaduarte (changes requested):
Comments about the autoplay attribute

HelenBurge:
After discussion, this pull request is no longer required.

PR #2069, opened 2023-05-31
Updating Set of Instruments Definition to have Objective=False

tbostic32:
The definition for set of instruments has objective set to true, but relies on `instrument` and `clearly labelled location` which both have objective set to false, which I think should logically mean this definition isn't objective. 

I do not believe that this definition is used in any place where it would need to be objective (i.e., the applicability) so I think it is fine to change. 

## How to Review And Approve

### DIFF
--- a/pages/glossary/set-of-clearly-labeled-instruments.md
+++ b/pages/glossary/set-of-clearly-labeled-instruments.md
@@ -2,7 +2,7 @@
 title: Set of clearly labeled instruments
 key: set-of-clearly-labeled-instruments
 unambiguous: true
-objective: true
+objective: false
 ---
 
 A _set of clearly labeled instruments_ is a set of [instruments][instrument], where each [instrument][] is in the same [web page][] as the test target or can be found in a [clearly labeled location][] from that [web page][].


Jym77 (approved):
It is indeed only used in Expectations (and example description):
```shell
$ grep -l set-of-clearly-labeled-instruments pages/glossary/*.md _rules/*.md
pages/glossary/set-of-clearly-labeled-instruments.md
_rules/device-motion-disabled-c249d5.md
_rules/printable-characters-shortcut-ffbc54.md
$ grep -B5 "\[set of clearly labeled instruments]\[]" _rules/*.md
_rules/device-motion-disabled-c249d5.md-## Expectation
_rules/device-motion-disabled-c249d5.md-
_rules/device-motion-disabled-c249d5.md-For each registered [device orientation event][device orientation] or [device motion event][device motion] in the test target at least one of the following is true:
_rules/device-motion-disabled-c249d5.md-
_rules/device-motion-disabled-c249d5.md-- **no changes:** The registered event does not cause [changes to the content][changes in content] of the [web page][] within a 1 minute time span of the [event firing][event firing]; or
_rules/device-motion-disabled-c249d5.md:- **disabled:** There is at least one [set of clearly labeled instruments][] to [block the event][blocked event][] for at least 1 minute.
--
_rules/printable-characters-shortcut-ffbc54.md-
_rules/printable-characters-shortcut-ffbc54.md-## Expectation
_rules/printable-characters-shortcut-ffbc54.md-
_rules/printable-characters-shortcut-ffbc54.md-For each test target at least one of the following is true:
_rules/printable-characters-shortcut-ffbc54.md-
_rules/printable-characters-shortcut-ffbc54.md:- **disable/remap**: there is at least one [set of clearly labeled instruments][] to [block events][blocked event] that use the [same key][same key events] as the test target and whose `getModifierState` method returns `false` for each of the [valid modifier keys][]; or
--
_rules/printable-characters-shortcut-ffbc54.md-</html>
_rules/printable-characters-shortcut-ffbc54.md-```
_rules/printable-characters-shortcut-ffbc54.md-
_rules/printable-characters-shortcut-ffbc54.md-#### Passed Example 6
_rules/printable-characters-shortcut-ffbc54.md-
_rules/printable-characters-shortcut-ffbc54.md:This [HTML document][] is listening to [keyboard events][keyboard event] for which the attribute `key` is a [printable character][] and the method `getModifierState` returns `false`, and which cause [changes in content][]. There exists a [set of clearly labeled instruments][] to **disable** or to **remap** the [keyboard event][] so that [same key][same key events] events are [blocked][blocked event] unless `getModifierState("Control")` returns `true`.
```

WilcoFiers (approved):

daniel-montalvo (approved):

Jym77:
Since this is just fixing an editorial mismatch, I don't think it needs a Call for Review. Merging now.

PR #2072, opened 2023-06-06
Rule-issue-templates

daniel-montalvo:
Proposes to add issue templates for composite and atomic rules, and to change to new GH issue 
template config
Closes issue(s):

- none

Need for Call for Review:
other: Issue templates

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,8 +1,0 @@
-Please describe the issue with references and relevant examples where necessary.
-
---- or ---
-
-If authoring a proposal for a new rule, choose from one of the issue templates:
-
-- [Atomic Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/atomic-template-empty.md)
-- [Composite Rule Template](https://raw.githubusercontent.com/act-rules/act-rules.github.io/develop/pages/design/composite-template-empty.md)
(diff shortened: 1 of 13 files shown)


WilcoFiers:
@daniel-montalvo I guess I'm just not sure what you're hoping to get from this. We don't write rules in issues. We use pull requests for that. Why would rule templates need to be available as templates for issues?

daniel-montalvo:
I thought rough concept draft rules were first put in issues and, if consensus exists that these should be worked on, they’d be move to a PR for further refinement.

 

Then this does not seem to make much sense.

 

Best.

---

Daniel Montalvo

Accessibility Harmonization and Training Specialist

W3C/WAI

Jym77:
I agree that issues are not a good place to discuss fully written rules (i.e. using the full template) because they don't make it possible to comment on specific bits as PR review do, and as we do on every single rule proposed.

Thus said:
* The current template seems unclear about that and seems to push people toward **issue* rather than PR to author new rules. Maybe we should update it to say something like "if authoring a new rule, open a PR with the good template".
* For not fully fleshed out ideas, we may want a simplified template like "Applicability/Expectation/mapping/some explanation why this rule is good" which is enough to discuss the rule before writing it (and we often have these simplified Applicability/Expectation description in discussions). Maybe we could have that as a "rule idea template" or something 🤔 
* All up for switching to the new template config 😄

PR #2076, opened 2023-06-22
"Audio or video element avoids automatically playing audio" [80f0bf]: Add Accessibility Support note about autoplay not being respected

### LINKED ISSUE #2030: autoplay rule [x0paj4] failed examples don't autoplay in major browsers

Jym77:
~Make 1.4.2 as secondary requirement since `autoplay` is not always respected depending on user engagement with the page.~
See https://www.w3.org/2023/04/13-act-r-minutes.html#item09

Closes issue(s):

- closes #2030 

Need for Call for Review:
This will require a 1 weeks Call for Review (adding a non-normative note)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


Jym77:
Note: will need to be updated once #2060 is merged.

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -79,6 +76,8 @@ The native `video` and `audio` controls in several browser and assistive technol
 
 The [instruments][instrument] used to pass this rule (if any), must meet all level A Success Criteria in order to fully satisfy [Success Criterion 1.4.2 Audio Control][sc142]. These extra requirements are left out of this rule, and should be tested separately.
 
+User agents do not always automatically play medias, even when an `autoplay` attribute is present. This is done precisely to avoid autoplaying media interrupting the user when they do not want to, especially when the media is likely to contain sound. The decision to respect the `autoplay` attribute or not depends on user settings and previous behavior (interaction with the site, …) Depending on the state of the user agent, failing this rule may cause [Success Criterion 1.4.2 Audio Control][sc142] to be not satisfied.
```
tbostic32: This case strikes me as more of an accessibility support issue than an issue needing the requirement to be listed as secondary. The failure test cases below very clearly fail the success criterion, and the only reason that a user would be unaffected would be if the user agent is setup to prevent it.

Jym77: Yes, I agree it should be an AS note, not a Background one, will move it there.
In #2030, both @carlosapaduarte and me experienced that the failure test cases do not autoplay (or not always), and thus were not actually failing the SC (depending on the UA, … and global accessibility support).

In my case, I was building the semi-automated implementation report for Alfa, going to the page for the manual part of the review, seeing that the video did not autoplay and answered so… only to have the report saying that my implementation is incorrect. Only after looking at the code I could see the `autoplay` attribute and the fact that my answer should have been "yes, it autoplays" rather than "no"… But as a user, that video didn't autoplay and the page wasn't failing 1.4.2.

On [CG call of April 13th](https://www.w3.org/2023/04/13-act-r-minutes.html#item09), we discussed it. Especially, around the fact that we should not have failed cases that depend on Accessibility Support (because rules do not consider it directly, so tester A could fail the rule on that case, while tester B passes it, and both are correct because they have different accessibility support). And so we came up with the solution of using it as a secondary requirement only.

tbostic32 (dismissed):

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -75,6 +72,8 @@ For each test target, the [outcome](#outcome) of at least one of the following r
 
 The native `video` and `audio` controls in several browser and assistive technology combinations are not keyboard accessible and the `video` or `audio` element itself may not be announced. Authors are recommended to use custom controls for keyboard navigation and cross browser accessibility support in general. Some major browsers no longer automatically play the 'video' unless the 'video' is muted.
 
+User agents do not always automatically play medias, even when an `autoplay` attribute is present. This is done precisely to avoid autoplaying media interrupting the user when they do not want to, especially when the media is likely to contain sound. The decision to respect the `autoplay` attribute or not depends on user settings and previous behavior (interaction with the site, …) Depending on the state of the user agent, failing this rule may cause [Success Criterion 1.4.2 Audio Control][sc142] to be not satisfied.
```
tbostic32:
```suggestion
User agents do not always automatically play media, even when an `autoplay` attribute is present. This is done precisely to avoid autoplaying media interrupting the user when they do not want to, especially when the media is likely to contain sound. The decision to respect the `autoplay` attribute or not depends on user settings and previous behavior (interaction with the site). Depending on the state of the user agent, failing this rule may cause [Success Criterion 1.4.2 Audio Control][sc142] to be not satisfied.
```

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -6,10 +6,7 @@ description: |
   This rule checks that audio or video that plays automatically does not have audio that lasts for more than 3 seconds or has an audio control mechanism to stop or mute it.
 accessibility_requirements:
   wcag20:1.4.2: # Audio Control (A)
-    forConformance: true
-    failed: not satisfied
-    passed: further testing needed
-    inapplicable: further testing needed
+    secondary: true
```
kengdoj: Please see https://github.com/act-rules/act-rules.github.io/discussions/2095#discussioncomment-6824094

> the task force concluded that accessibility support is not a cause for having an accessibility requirement listed as "secondary".

kengdoj (dismissed):
Please see https://github.com/act-rules/act-rules.github.io/discussions/2095#discussioncomment-6824094

Jym77:
@kengdoj @tbostic32 I've reverted to primary requirement, following #2095, and kept the (new) AS note.

WilcoFiers:
@Jym77 I think we decided we were not going to do this. Can this PR be closed?

Jym77:
@WilcoFiers Ah yes, this was actually reduced to adding an Accessibility Support note 😅 
I've updated the title.

WilcoFiers (approved):

carlosapaduarte (approved):

giacomo-petri (approved):

Jym77:
Call for Review ends on February 8th.

kengdoj (approved):

Jym77:
Call for Review has ended. Merging.

PR #2077, opened 2023-06-22
"Heading has non-empty accessible name" [ffd0e9]: Update from TF survey

ChrisLoiselle:
Per review of survey heading non-empty task force conclusions, edited the following:

* Pass ex 3, have the acc name be consistent with the text						
* Set 1.3.1 as a secondary requirement						
* Removed sentence on accessible name spec not being clear on whitespace						
* Removed inapp example 1


daniel-montalvo (approved):

_rules/heading-non-empty-accessible-name-ffd0e9.md
```diff
@@ -40,7 +42,7 @@ There are no assumptions.
 
 ## Accessibility Support
 
-- Some assistive technologies may hide headings with empty [accessible name][] from the users. This depends on the user agent, on how the [accessible name][] was computed (the [accessible name and description computation][] is not clear concerning which characters should be trimmed), and on the assistive technology itself. Hence, there are cases where the outcome of this rule is _failed_, but users of certain assistive technology and browser combinations will not experience an issue.
+- Some assistive technologies may hide headings with empty [accessible name][] from the users. This depends on the user agent, on how the [accessible name][] was computed (the [accessible name and description computation][] and on the assistive technology itself. Hence, there are cases where the outcome of this rule is _failed_, but users of certain assistive technology and browser combinations will not experience an issue.
```
Jym77:
```suggestion
- Some assistive technologies may hide headings with empty [accessible name][] from the users. This depends on the user agent, on how the [accessible name][] was computed and on the assistive technology itself. Hence, there are cases where the outcome of this rule is _failed_, but users of certain assistive technology and browser combinations will not experience an issue.
```
Looks like part of the parenthesis wasn't deleted.

_rules/heading-non-empty-accessible-name-ffd0e9.md
```diff
@@ -173,14 +175,6 @@ This `h1` element has an [explicit role][] of `none`. However, the [global][] [p
 
 #### Inapplicable Example 1
 
-There is no [semantic][semantic role] `heading` element.
-
-```html
-<div></div>
-```
```
Jym77, line 180: Removing this example breaks our own "test case design" guidelines of [Test every condition](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/design/test-cases.md#test-every-condition) 🤔

Jym77 (changes requested):

ChrisLoiselle:
@WilcoFiers @kengdoj  please let me know how you'd like to proceed per @Jym77 's comments.

_rules/heading-non-empty-accessible-name-ffd0e9.md
WilcoFiers: @Jym77 We felt this was not a useful test case, so we're thinking of leaving things like this out. Is that OK with you? Happy to update the test ever condition bit then.

Jym77: @WilcoFiers No strong feeling about that particular example which is, indeed, quite useless 🙂

It's more on the boundaries of things on how we decide that tests are useful or not. 
From a TDD point of view, we should indeed have test cases for every condition, and since we do use the test cases to validate tools, this also sounds like a good idea from an engineering point of view.
From an "illustrate the rule to human" point of view, this is indeed a fairly useless case…

=> this tends to hint that we may want to have some test cases for the computers and only show a meaningful subset to the humans.

ChrisLoiselle:
@WilcoFiers I committed @Jym77 feedback, please let me know if there is anything else needed to push this to close / done. Thanks for all your help @Jym77 and @WilcoFiers !

_rules/heading-non-empty-accessible-name-ffd0e9.md
```diff
@@ -10,6 +10,8 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+secondary_requirements:
+wcag20:1.3.1: # Info and Relationships (A)
```
WilcoFiers, line 14: This is not the correct syntax. What you should do is replace:
```yml
  wcag20:1.3.1: # Info and Relationships (A)
    forConformance: true
    failed: not satisfied
```

with this:
```yml
  wcag20:1.3.1: # Info and Relationships (A)
    secondary: true
```

WilcoFiers (changes requested):

_rules/heading-non-empty-accessible-name-ffd0e9.md
ChrisLoiselle: @WilcoFiers am I to update within this PR or do I need to create a new one with the desired changes you recommend per secondary requirements? Not sure if you wanted another PR on https://github.com/act-rules/act-rules.github.io/pull/2077/commits/7fb7a37b54b81d40a7f5be3e0d87326c5e7b3ea1 for the highlight you are mentioning in this comment here.

WilcoFiers: No please update in this PR.

ChrisLoiselle (commented):
@WilcoFiers  I don't have access to approve the changes only to comment. I'll pursue updating the file and submitting the updated PR.

_rules/heading-non-empty-accessible-name-ffd0e9.md
ChrisLoiselle: @WilcoFiers when I view the file in "edit file" it shows your changes. I can't approve per admin settings, thus not sure what else to do to "close" this issue and "update the PR" from my side.

Jym77 (approved):

ChrisLoiselle:
@WilcoFiers , I see this is still open and aged quite a bit. Did you want me to take any action on this or would you like to close as no longer applicable?

_rules/heading-non-empty-accessible-name-ffd0e9.md
giacomo-petri:
```suggestion
```

1.3.1 is both in accessibility_requirements and secondary_requirements; suggesting to remove secondary

WilcoFiers:
I think we need to discuss this again. Added an agenda item label. Empty headings generally do get ignored by AT. It's not an accessibility problem for those. Failing that under WCAG is incorrect IMO, so I don't think we should have this as a secondary requirement either. Secondary requirements isn't intended to allow failing things under WCAG that aren't failures.

WilcoFiers:
This is outdated, rule has been changed in a different direction.

PR #2083, opened 2023-07-11
scrollable element: clarify the title

WilcoFiers:
We got feedback from AGWG that this rule wasn't testing everything. We (ACT TF) believe this is in part because the rule's title suggests it does more than it actually does. To address that we'd like to change the title of this rule. 

Here is the original issue: https://github.com/w3c/wcag-act-rules/issues/223

Need for Call for Review: None, this is an editorial change.

---

## How to Review And Approve


carlosapaduarte (approved):

HelenBurge (approved):

Jym77 (approved):

WilcoFiers:
Thanks for the quick reviews everyone!!

PR #2084, opened 2023-07-13
New rule: role is permitted

### LINKED ISSUE #2042: Rule idea: role is permitted for element
The gist of it is to test that the explicit role is allowed on the element. This would test requirements outlined in HTML in ARIA

https://www.w3.org/TR/html-aria/#docconformance

Example: `<button role="heading">` would not be allowed because the [`button` element](https://www.w3.org/TR/html-aria/#el-button) is only allowed to have certain roles.

Jym77:
New rule to check that a `role` is permitted on this element.

* It currently checks only the explicit role. Should it check every token in the `role` attribute instead? (same as "role has valid value")
* ~There are very few examples. I'm struggling a bit to find the good balance… The spec here is essentially a list of cases and we definitely cannot have all of them. If you have ideas of relevant (=used in practice) examples to add, pleas let me know.~
* I am actually not sure that the rule should require the element to be in the accessibility tree. This is because `presentation` is not allowed on every element (e.g. it is allowed on `h1`, not `button`, I assume this is due to the focusability of `button`). But we cannot really test that if we restrict the rule to content in the accessibility tree… (well, of course, the presentational conflict would trigger but I'd rather not have the rule rely on that…) OTOH, checking `display: none` content seems really unecessarily…
* ~I think the link to "ARIA in HTML" in the conformance mapping will need some website code update to be correctly picked up: https://github.com/act-rules/act-tools/pull/31.~

Closes issue(s):

- closes #2042 

Need for Call for Review:
This will require a 2 weeks Call for Review (new rule)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


colabottles (approved):

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,125 @@
...
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+This rule applies to any [HTML element][namespaced element] that is [included in the accessibility tree][] and has an [explicit semantic role][explicit role].
+
+## Expectation
+
+For each test target, its [explicit semantic role][explicit role] is allowed on this element, according to [ARIA in HTML specifications][aria in html document conformance].
```
daniel-montalvo:
```suggestion
For each test target, its [explicit semantic role][explicit role] is allowed on this element, according to the [ARIA in HTML specification][aria in html document conformance].
```

daniel-montalvo (commented):
So glad to see this new rule coming along. I have one minor editorial and one question.

What is this rule supposed to do with `role="generic"` or with ARIA deprecated roels such as `directory`? According to how this is written I'd say these pass, but different implementers may have different interpretations in the form of stronger warnings against of these that we may want to qualify.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,125 @@
+---
+id: j7zzqr
+name: ARIA role is permitted.
```
WilcoFiers:
```suggestion
name: ARIA role is permitted for the element
```

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,125 @@
...
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: satisfied
+  wcag20:1.3.1: # Info and Relationships (A)
+    secondary: true
+  wcag20:4.1.2: # Name, Role, Value (A)
+    secondary: true
+  wcag-technique:ARIA4: # Using a WAI-ARIA role to expose the role of a user interface component
+    secondary: true
```
WilcoFiers: I'm not sure these should be here. I think the correlation between this rule and WCAG is pretty weak. The one failed example this rule has, I think anyone who would fail that under 1.3.1 and 4.1.2 would be wrong to do so.

There might be cases I didn't think of, but while we didn't constrain secondary requirements to only be used when there are corresponding examples, that is a fairly strong indicator, and perhaps even something we'd want to do in our authoring policy; state that if there is a secondary requirement, there should be examples that demonstrate it.

@kengdoj what's your thoughts about using secondary here?

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,125 @@
...
+
+There are no assumptions.
+
+## Accessibility Support
+
+There are no accessibility support issues known.
+
+## Background
+
+The presence of an invalid role often implies that the programmatic role do not correspond to the one that is conveyed visually, or that the interactions provided by the elements do not match the ones expected for this role. Therefore, both [Success Criterion 1.3.1 Info and Relationships][sc131] and [Success Criterion 4.1.2 Name, Role, Value][sc412] are secondary requirements for this rule.
```
WilcoFiers: I'm not sure how true this really is. See my comment above. Also, typo:

```suggestion
The presence of an invalid role often implies that the programmatic role does not correspond to the one that is conveyed visually, or that the interactions provided by the elements do not match the ones expected for this role. Therefore, both [Success Criterion 1.3.1 Info and Relationships][sc131] and [Success Criterion 4.1.2 Name, Role, Value][sc412] are secondary requirements for this rule.
```

We probably also want to say something that this rule is stricter than 1.3.1 and 4.1.2, if we keep those as secondary.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Passed Example 2
+
+This `h1` element has an [explicit role][] of `tab`, which is allowed
+
+```html
+<h1 role="tab">ACT rules</h1>
+```
+
+### Failed
```
WilcoFiers, line 140: I think I'd like a few more examples here of different types of elements. Maybe have a landmark and a heading?

WilcoFiers (dismissed):

Jym77:
I've done most changes. I've also removed the inclusion in the accessibility tree from the Applicability in order to be able to also test `role="presentation"`.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,167 @@
...
+  <title>This is an SVG</title>
+</svg>
+```
+
+#### Inapplicable Example 2
+
+This `button` element is not [included in the accessibility tree][].
+
+```html
+<button role="list" style="display:none;">Click me</button>
```
tbostic32: From the background:

> This rules apply to every element, even if they are not [included in the accessibility tree][].

Wouldn't that make this a failed example instead of an inapplicable example?

carlosapaduarte: I agree with @tbostic32 

The applicability are elements with an explicit role defined which is the case.

carlosapaduarte (commented):
Suggestions for further examples:

Something with role="presentation" and role="generic", both for failed and pass examples

Perhaps we might want an example with a list of tokens, where the first token is a valid token corresponding to a non permitted role and the second is a valid permitted role

Examples with aria-hidden (in addition to the one with display none)

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,167 @@
...
+
+There are no assumptions.
+
+## Accessibility Support
+
+There are no accessibility support issues known.
+
+## Background
+
+[ARIA in HTML][aria in html document conformance] also defines the [implicit semantic role][implicit role] of each element. Setting the [explicit role][] as the same as the [implicit one][implicit role] is not recommended but nonetheless allowed. This rule doesn't use that in any of its test cases.
```
tombrunet:
```suggestion
[ARIA in HTML][aria in html document conformance] also defines the [implicit semantic role][implicit role] of each element. Setting the [explicit role][] as the same as the [implicit role][] is not recommended but nonetheless allowed. This rule does not apply in that scenario.
```

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,167 @@
...
+
+This rules apply to every element, even if they are not [included in the accessibility tree][]. This is because the roles of `none` or `presentation` are only allowed on certain elements. If the rule was only looking at content [included in the accessibility tree][], it wouldn't flag incorrect use of these roles.
```
tombrunet: If an element and its children are not included in the accessibility tree, it doesn't really matter what the roles are. Therefore, it shouldn't be an accessibility problem if these roles are incorrectly used. While I agree that it's technically incorrect, I think it should be considered more as a recommendation rather than a violation - similar to the use of an explicit role that is redundant with the implicit role.

@scottaohara Care to comment on the ARIA in HTML position on this one?

_rules/aria-role-permitted-j7zzqr.md
tombrunet: See my other comment. I do think this should be inapplicable to this rule, and the note and applicability rather than this testcase should be modified. However, one or the other needs to change.

tombrunet (dismissed):

tombrunet:
One additional comment - a colleague of mine suggested that we might want to more clearly indicate the applicability to **explicit** roles in the title.

_rules/aria-role-permitted-j7zzqr.md
scottaohara: i mean, yes - if the element is hidden form the accessibility tree then sure, no one can access it / there can't be an issue.

the counter point being that if it's rendered in the DOM but marked as hidden, is that because at some point it would be expected that a user could then unhide that content to interact with it?  Could excluding something just because it's temporarily hidden not cause a bunch of gotchas for developers who might run a scan, it comes up clean, and then they still get bugs because someone had the audacity to try and use their UI? :)

but seriously, i can see the rational to just expose this as a recommendation / needs review - in the hidden state, as long as if it becomes shown it then would be marked as a violation.  it just seems like a delay in flagging some issues because maybe someone did a hack-job at 'removing' an errant element, and _that_ shouldn't be called out?

_rules/aria-role-permitted-j7zzqr.md
scottaohara: aria in html does not define the implicit roles.  HTML AAM does.  ARIA in HTML informatively identifies what these are.

Jym77:
> What is this rule supposed to do with `role="generic"` or with ARIA deprecated roels such as `directory`? According to how this is written I'd say these pass, but different implementers may have different interpretations in the form of stronger warnings against of these that we may want to qualify.

@daniel-montalvo As far as I can read, both are "NOT RECOMMANDED" (in ARIA in HTML) or "SHOULD NOT" (for `generic` in ARIA 1.2) cases, but are not conformance requirements. I think that this rule should therefore not flag them. For deprecated roles, I think that "Role attribute has valid value" will catch it 🤔 For `generic`, we could have a separate rule checking for that, but that would not map to any requirement given that it is only a "SHOULD NOT" case.

I've added a Background note to explicit that reasoning.

_rules/aria-role-permitted-j7zzqr.md
Jym77: True, moved it to an Failed case.

I think that this is still failing ARIA in HTML, even if it doesn't cause an accessibility problem in the end, so I'd rather keep it that way. ARIA in HTML doesn't specify that the roles are only allowed on elements that are exposed. Moreover, if we restrict the rule to exposed elements, suddenly Failed Example 4 (`<li role="presentation">`) becomes Inapplicable when it's a clear failure since the only allowed role is `listitem`.

This is a bit similar to the fact that, in HTML¹, duplicate `id` are not allowed even though they might very well never cause a single problem. It is nonetheless invalid HTML. Here also, as far as I understand, using the wrong role on a hidden element is invalid ARIA, even if the user never encounter that.

¹: this is not about SC 4.1.1, but HTML/DOM specs

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Accessibility Support
+
+There are no accessibility support issues known.
+
+## Background
+
+Each element also has an [implicit semantic role][implicit role] defined in the [HTML Accessibility API Mappings][html aam:roles]. Setting the [explicit role][] as the same as the [implicit role][] is not recommended but nonetheless allowed. This rule does not check that scenario.
+
+This rule does not check specifically whether the [explicit role][] is deprecated, or whether it is the special `generic` role that should not be used by authors. These are not conformance requirement, or are checked by other rules.
```
scottaohara, line 42:
```suggestion
This rule does not check specifically whether the [explicit role][] is deprecated, or whether it is the special `generic` role that should not be used by authors. These are not conformance requirements, or are checked by other rules.
```

are deprecated features surfaced by other rule(s)?  If so, can there be a link to that?

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,221 @@
...
+
+This rules apply to every element, even if they are not [included in the accessibility tree][]. This is because the roles of `none` or `presentation` are only allowed on certain elements. If the rule was only looking at content [included in the accessibility tree][], it wouldn't flag incorrect use of these roles.
```
scottaohara:
```suggestion
This rule applies to every element, even if they are not [included in the accessibility tree][]. This is because the roles of `none` or `presentation` are only allowed on certain elements. If the rule was only looking at content [included in the accessibility tree][], it wouldn't flag incorrect use of these roles.
```

this section should probably also mention content hidden from the a11y tree with aria-hidden=true, or other host language features (per the example on line 182)

giacomo-petri: I'm uncertain about the accuracy of the following statement:

> This rule applies to every element, even if they are not [included in the accessibility tree][]. This is because the roles of `none` or `presentation` are only allowed on certain elements. If the rule was only looking at content [included in the accessibility tree][], it wouldn't flag incorrect use of these roles.

According to section [7.3 Relationships in the Accessibility Tree](https://w3c.github.io/aria/#tree_relationships),

> The accessibility children of a DOM element are all of the children of that element's corresponding [accessible object](https://w3c.github.io/aria/#dfn-accessible-object) in the [accessibility tree](https://w3c.github.io/aria/#dfn-accessibility-tree). In terms of the DOM, that includes the following (with exclusions listed blow):
> - The DOM children of the [element](https://dom.spec.whatwg.org/#concept-element).
> - All DOM descendants of the [element](https://dom.spec.whatwg.org/#concept-element) with only elements of role [generic](https://w3c.github.io/aria/#generic) or [none](https://w3c.github.io/aria/#none) intervening.

This principle is equally applicable to the accessibility parent definition (https://w3c.github.io/aria/#dfn-accessibility-parent)

A clear illustration of this can be found in Example 43: [Example 43](https://w3c.github.io/aria/#example-43).

Consequently, it is reasonable to conclude that having a role of "generic," "none," or "presentation" should not impact the outcome of a test.

_rules/aria-role-permitted-j7zzqr.md
giacomo-petri: In accordance with my comment https://github.com/act-rules/act-rules.github.io/pull/2084#discussion_r1340254070, to me, Failed Example 4 is not failing because of the presence of the "presentation" role in a DOM children. Rather, its failure is attributed to the absence of an accessibility child with the role "listitem" within the DOM child.

Hence, it is not the "role=presentation" itself that impacts the test result; rather, the test fails due to the absence of a child with the "listitem" role.

I agree with Tom that we should consider elements included into the accessibility tree.

Jym77: @giacomo-petri The reason why `role="presentation"` was taken into account by this rule is that certain elements (e.g., `article` or `aside`) explicitly allow it in the list of allowed roles; while some other (like `button`) also have an explicit list of allowed roles but **without** `presentation` being listed.

On looking a bit more closely, it might be that the ones where `presentation` isn't listed are the elements which are normally focusable, so the conflict would trigger 🤔 I'll look a bit deeper.

_rules/aria-role-permitted-j7zzqr.md
Jym77: @giacomo-petri Can you get a look at it now? I tried to improve the situation…

colabottles (approved):

scottaohara:
@Jym77 per your requesting my review - i personally dont' agree that it makes sense to designate `<button role=list>` as inapplicable because it's hidden from the a11y tree with display none.  

if `display none` was used to hide errant code for the purpose of solving a bug, then it should be of no surprise to the developer that this would be flagged to them.  hiding something under the rug of CSS isn't really solving the problem, after all.

if `display none` is there because the element is temporarily hidden until rendered by the user, then why make someone _have_ to render that UI element before calling out the error with it?  That to me seems inconsistent results, and would, I assume, mean that this test couldn't be performed at the source code level?  It could only find the error on a rendered web page?  please correct me if i'm wrong.

if classifying that as inapplicable is to try and avoid any false positives.  i mean... sure?  i can't really argue with that.  i don't agree with it, but i'm not sure i see the point in diving into that topic any further.  i'd at least submit that it should be in a 'needs review' category rather than ignoring it until someone happens to make the element visible.

Jym77:
@scottaohara Thanks a lot for the feedback. This is a rule designed for ARIA, not WCAG, so it is ultimately up to ARIA group to decide what fails and what is a false positive 😉 

@giacomo-petri Does that change your opinion on whether Applicability should be restricted to exposed content? Or should we take the discussion to a CG call?

giacomo-petri:
@Jym77 @scottaohara,

I understand and generally agree with the points Scott has raised. However, there are scenarios where these principles may not be easily applicable.

Generally speaking, I've come across several instances where functional code remains hidden from all users. For example, developers often use "display:none" containers to enclose functional inputs within a form. While some argue in favor of using "input type hidden" elements, the former approach doesn't necessarily lead to accessibility issues. 
There are also situations where code dynamically changes based on user interactions, such as a hidden list or a combobox with initial placeholders that become visible after their content is replaced.

Since we can't anticipate the outcome until we actually interact with the page, I believe it's not our responsibility to assume the author's intent. Instead, we should concentrate on evaluating the output produced by user interactions, as it effectively reveals the author's intent and the solutions they've implemented.

Jym77:
@giacomo-petri I think this is where I make the difference between ARIA (authoring) error and end user accessibility problem (more or less the same as WCAG violation).

It's a bit the same that happens when we check for validity or `aria-*` attributes or the `role` (or triggers of the presentation conflict):
```html
<button aria-name="hello">world</button>
<a role="btn" href="#">Hello</a>
<button role="presentation">Hello</button>
<h1 role="presentation" aria-label="Hello">world</h1>
```
None of these create an actual problem for the end user, but these are clearly bad ARIA and author errors, that some of the rules we have are flagging.

This proposed rule is also checking for author error, rather than end user problems. If `<button role="list" style="display: none">Hello</button>` is considered as an author error by the ARIA group, I think the rule should follow suit and flag it as well, even if end users are never going to experience that button.

_rules/aria-role-permitted-j7zzqr.md
tombrunet: @scottaohara With the way that UIs are built these days, what is in the DOM changes with almost every state change. There are often scaffolds that are hidden, and then made whole when they're actually used / made visible. We don't report on a lot of issues in hidden content because of this - teams were frustrated by all of the false positives. 

While I agree that it's unlikely that an incorrectly used role will be corrected on such a state change, I don't think we should call out a tool as incorrect for ignoring that content completely.

tombrunet:
For the purposes of ACT, it's not really a question of whether or not it's correct / incorrect for the invalid role to be used on hidden content, but rather should a tool be flagged as inconsistent for choosing not to report that. I believe that it's perfectly acceptable for a tool to choose to fail a hidden invalid role, and it's also perfectly acceptable to completely ignore that since the role will not be exposed outside of the DOM.

The two ways we tend to handle that are to either scope this rule so that the questionable scenario isn't included, or to intentionally not include related test cases.

scottaohara:
i had written a response, but with @tombrunet's comment, i see that would just be going off topic... so i'll refrain.

if i understand your comment, Tom, are you saying that by marking such instances as "inapplicable" it then allows tools to make their own choices about whether to surface such instances or not?  But otherwise, it would be expected for the checkers to either all pass or all fail the rule - and the choice to or mark something as a warning or needs review is off the table?

tombrunet:
For the purposes of this rule, if the test case says that this is a fail and a tool does not flag it, the tool would be considered as inconsistent. But, if the test case says inapplicable and the tool says it fails, it would also be considered inconsistent. So, if we don't want either of these tools to be considered as inconsistent, we can either:
* Be clear that this situation is out of scope and have the test case as inapplicable - this doesn't imply anything regarding the testcase, just that that scenario isn't covered by **this** rule (it could be covered by some other rule).
* Avoid this scenario completely and not have a testcase for it (ACT just doesn't take a position on the matter).

Jym77:
Given the amount of discussion we're having, I think it is safer to take it to the rest of the group on a call.

Feel free to continue discussing it here 😄 We'll need a common decision in the end, but I don't want to cut the current discussion short until then.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,252 @@
...
+  authors:
+    - Jean-Yves Moyen
+---
+
+## Applicability
+
+This rule applies to any [HTML element][namespaced element] which has an [explicit semantic role][explicit role] and for which at least one of the following is true:
+
+- the element is [included in the accessibility tree][]; or
+- the element has an [explicit semantic role][] of `presentation` or `none`.
```
WilcoFiers: It's a little odd to me that this rule would apply to an element with role=none that's inside an aria-hidden=true container. Perhaps the applicability should just have an exception for elements that are programmatically hidden?

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,252 @@
...
+```html
+<a href="https://www.w3.org/WAI/standards-guidelines/act/rules/" role="button">All ACT rules</a>
+```
+
+#### Passed Example 2
+
+This `h1` element has an [explicit role][] of `tab`, which is allowed
+
+```html
+<h1 role="tab">ACT rules</h1>
```
WilcoFiers: I'm not a huge fan of doing this, as there are more requirements for tab to be valid, like that it needs to be in a tablist. I think either a different role is needed here, or this needs to be made a fully valid tab.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,252 @@
+---
+id: j7zzqr
+name: ARIA role is permitted for the element.
+rule_type: atomic
+description: |
+  This rule checks that explicit WAI-ARIA roles are allowed for the element they are specified on.
+accessibility_requirements:
+  html-aria:docconformance:
+    title: ARIA in HTML, 4. Document conformance requirements for use of ARIA attributes in HTML
```
WilcoFiers, line 9: Sorry, this is going to be very nit picky.

> The third cell in each row defines the ARIA role values and aria-* attributes which authors MAY specify on the element

I don't think the way you're interpreting this requirement is right. Or quite possibly, I think the requirement is written incorrectly. MAY in RFC2119 speak does not create a prohibitive statement. You can use these roles on those elements, and browsers have to support it when you do. But that's not the same as saying authors are not allowed to use other roles. It would have to be something like "if an author specifies a role it MUST be one from the third column."

I think this is a solid rule, and we should definitely have it, but I'm a little reluctant to say this is a conformance requirement for ARIA in HTML. Even if that was the intended meaning, that's not what it says. I don't really want to hold this rule up over a technicality like this, but I do think something needs to be done about this.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,252 @@
...
+
+- [Document conformance requirements for use of ARIA attributes in HTML](https://www.w3.org/TR/html-aria/#docconformance)
+
+## Test Cases
+
+### Passed
+
+#### Passed Example 1
+
+This `a` element with an `href` attribute has an [explicit role][] of `button`, which is allowed.
```
WilcoFiers, line 60: I think these descriptions would read better if we flip the role and element around:

```suggestion
The `button` role is allowed on an `a` element with an `href` attribute.
```

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,252 @@
...
+```html
+<h1 role="btn tab">ACT rules</h1>
+```
+
+#### Passed Example 7
+
+This `dialog` element has an [explicit role][] of `alertdialog`, which is allowed. Even though the `alert` role is not allowed, the first token is a valid role and is therefore used.
+
+```html
+<dialog role="alertdialog alert">This is not right.</dialog>
```
WilcoFiers, line 137: I'm not 100% sure about this one. The requirement for it is this:

> The third cell in each row defines the ARIA role values and aria-* attributes which authors MAY specify on the element.

That to me reads that fallback roles apply here too. I don't know that I would want them too. That feels a little on the strict side, but it might be good to put something in the background section about fallback roles, and straight up invalid or abstract roles being out of scope for this rule, even if they do seem like they'd fail the requirement.

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,252 @@
...
+	<li>Orange</li>
+</ul>
+```
+
+#### Failed Example 6
+
+This `label` element has an [explicit role][] of `generic`, which is not allowed.
+
+```html
+<label role="generic"> Name: <input /> </label>
```
WilcoFiers, line 200: ARIA is a little vague on whether generic elements are supposed to be in the accessibility tree. I think the applicability as you have it now probably means this passes. My suggested update to it should solve that.

WilcoFiers (dismissed):

_rules/aria-role-permitted-j7zzqr.md
scottaohara: @WilcoFiers the sentence you quote needs to be put into the context of section 1 Author requirements fo ruse of ARIA in HTML

>Authors MUST NOT use the ARIA role and aria-* attributes in a manner that conflicts with the semantics described in the [4. Document conformance requirements for use of ARIA attributes in HTML](https://w3c.github.io/html-aria/#docconformance) and [4.2 Requirements for use of ARIA attributes in place of equivalent HTML attributes](https://w3c.github.io/html-aria/#docconformance-attr) tables. ...

per section 1, it's an Authors MUST NOT - where the Author MAYs specificy what it allowed.

if you think this is unclear because one has to piece together these different parts of the specification, then that's fair.  The spec can be updated to restate the author requirements more overtly in tandem with the text you quoted.

_rules/aria-role-permitted-j7zzqr.md
scottaohara: 'passes' though a should not from aria.

i think aria in html's rules for label are a bit too strict here.  i'm going to make an issue to address this.

tombrunet: I understood the point of this example to be that generic is a user-agent assigned role, but not one that you are allowed to explicitly specify. Now that I think about it though, that may be more appropriate for https://www.w3.org/WAI/standards-guidelines/act/rules/674b10/proposed/

scottaohara: https://github.com/w3c/html-aria/issues/492

_rules/aria-role-permitted-j7zzqr.md
scottaohara: just going to state one more time that i'd err on the side of flagging this - as otherwise it falls on making a developer/tester have to do more work to find errors, by ensuring they fully comb through a UI and reveal all hidden content.  I do acknoweldge there are probably instances of devs hiding problematic markup, rather than fixing it... but should a dev really be surprised to get an error flagged for something they knew was an error and 'hid' it as their 'fix'?

Jym77:
CG decision: restrict rule to elements in the applicability tree and explicit role of presentation/none (or not in hidden state), add background note that stuff out of the accessibility tree still need to respect this but won't cause user problem.

scottaohara:
>add background note that stuff out of the accessibility tree still need to respect this but won't cause user problem

will "wont' cause user problem" be qualified in this note to state "won't cause the user a problem **so long as elements with non-permitted roles don't become rendered/available to users.** ??

Jym77:
> > add background note that stuff out of the accessibility tree still need to respect this but won't cause user problem
> 
> will "wont' cause user problem" be qualified in this note to state "won't cause the user a problem **so long as elements with non-permitted roles don't become rendered/available to users.** ??

@scottaohara Should be good now. Can you check if the new note is alright for you?

colabottles (approved):

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Background
+
+Each element also has an [implicit semantic role][implicit role] defined in the [HTML Accessibility API Mappings][html aam:roles]. Setting the [explicit role][] as the same as the [implicit role][] is not recommended but nonetheless allowed. This rule does not check that scenario.
+
+This rule does not check specifically whether the [explicit role][] is deprecated, or whether it is the special `generic` role that should not be used by authors. These are not conformance requirement, or are checked by other rules.
+
+This rule only check for elements that are not [programmatically hidden][]. Elements that are [programmatically hidden][], e.g. due to a CSS `display:none`, must still have valid roles in order to conform to [ARIA in HTML][aria in html document conformance]. However, these won't cause any actual accessibility issues as long as they stay hidden. This rule only checks a snapshot of a page in a given state; if interacting with the page reveals such an element without fixing its role, that could create an accessibility issue for the user. Examples only show [programmatically hidden][] element with a correct role.
```
scottaohara: i appreciate the note being added.  I've made some wording updates that I hope can be merged.  Let me know if there are any questions.

```suggestion
This rule only check for elements that are not [programmatically hidden][]. Elements that are [programmatically hidden][], e.g. due to a CSS `display:none`, must still have valid roles in order to conform to [ARIA in HTML][aria in html document conformance]. However, as long as the elements with the errant roles remain programmatically hidden, they are not expected to cause accessibility issues. This rule only checks a snapshot of a page in a given state; if interacting with the page programmatically exposes an element with an errant role that this rule previously ignored, the new state of the page should be reevaluated to ensure it still passes this rule, and that an accessibility issue has not been introduced to users. Examples only show [programmatically hidden][] elements with a correct role.
```

_rules/aria-role-permitted-j7zzqr.md
```diff
@@ -0,0 +1,253 @@
...
+
+This rule does not check specifically whether the [explicit role][] is deprecated, or whether it is the special `generic` role that should not be used by authors. These are not conformance requirement, or are checked by other rules.
+
+This rule only check for elements that are not [programmatically hidden][]. Elements that are [programmatically hidden][], e.g. due to a CSS `display:none`, must still have valid roles in order to conform to [ARIA in HTML][aria in html document conformance]. However, as long as the elements with the errant roles remain programmatically hidden, they are not expected to cause accessibility issues. This rule only checks a snapshot of a page in a given state; if interacting with the page programmatically exposes an element with an errant role that this rule previously ignored, the new state of the page should be reevaluated to ensure it still passes this rule, and that an accessibility issue has not been introduced to users. Examples only show [programmatically hidden][] elements with a correct role.
```
scottaohara: just a missing 's' i had previously not noticed in the first sentence.  otherwise lgtm.

```suggestion
This rule only checks for elements that are not [programmatically hidden][]. Elements that are [programmatically hidden][], e.g. due to a CSS `display:none`, must still have valid roles in order to conform to [ARIA in HTML][aria in html document conformance]. However, as long as the elements with the errant roles remain programmatically hidden, they are not expected to cause accessibility issues. This rule only checks a snapshot of a page in a given state; if interacting with the page programmatically exposes an element with an errant role that this rule previously ignored, the new state of the page should be reevaluated to ensure it still passes this rule, and that an accessibility issue has not been introduced to users. Examples only show [programmatically hidden][] elements with a correct role.
```

PR #2086, opened 2023-07-13
Deprecate HTML page lang and xml:lang attributes have matching values

### LINKED ISSUE #1921: HTML page lang and xml:lang match - problems with assumptions and applicability [5b7ae0]
This is a follow on issue for https://github.com/act-rules/act-rules.github.io/issues/1172 

We've done a lot of testing round `lang` and `xml:lang` and there seem to be problems with both the rule assumptions and applicability. 

Historically there was a lot of spec churn over use of `lang` vs `xml:lang` which must have caused lots of implementation inconsistencies, but this was resolved in HTML 5 in 2014 and implementations are now very consistent. The inconsistencies between Chromium and other browsers for pages served as application/xhtml+xml was resolved in Chrome 88, and all browsers now behave identically. Notably Chrome 88 was released in January 2021 *after* this rule was created.

I think the key piece of text in the rule is:

> Since most assistive technologies will consistently use lang over xml:lang when both are used, violation of this rule may not necessarily be a violation of WCAG 2. Only when there are inconsistencies between assistive technologies as to which attribute is used to determine the language does this lead to a violation of SC 3.1.1.

## Rule problems

1) We weren't able to find any inconsistencies in AT voicing for pages served as text/html, but did find inconsistencies in pages served as application/xhtml+xml documents. Unfortunately the rule only applies to text/html so I think it's never detecting true positives, and only flagging false positives.

2) The rule assumes that fr, fr-FR and fr-CA are interchangeable and all voice as French. This isn't true for NVDA with the default OneCore synthesiser (see below).

## text/html documents

- all current screen readers uses the `lang` attribute for voicing when both `lang` and `xml:lang` are specified (matches the HTML spec)
- the `:lang()` selector in browsers matches the `lang` attribute when both `lang` and `xml:lang` are specified (matches the HTML spec)

There are no inconsistencies in assistive technologies for documents served as text/html. Test results here:
https://www.powermapper.com/tests/screen-readers/content/html-page-lang-with-xml-lang/

## application/xhtml+xml documents

- all current screen readers uses the `lang` attribute for voicing when both `lang` and `xml:lang` are specified (this doesn't match the HTML spec)
- the `:lang()` selector in browsers matches the `xml:lang` attribute when both `lang` and `xml:lang` are specified (matches the HTML spec)

Test results here:
https://www.powermapper.com/tests/screen-readers/content/xhtml-page-lang-with-xml-lang/

This means the language used for voicing may not match the CSS language leading to the wrong voice being used for content displayed using the :lang() selector.

This example displays 'C'est le français' and voices it correctly as French when served as text/html but displays 'Das ist deutsch' voiced as French when the page is served as application/xhtml+xml

```
		<html xmlns="http://www.w3.org/1999/xhtml" lang="fr" xml:lang="de">
		<head> 
			<title>Test for mismatching lang and xml:lang</title>
			<meta charset="utf-8"/>
			<link rel="stylesheet" href="SR-content-lang.css"/>
			<style>
				div:lang(fr)::before { content: "Un, deux, trois"; } 
				div:lang(de)::before { content: "Eins, zwei, drei"; } 
			</style>
		</head>
		<body>
			<h1 lang="en">Following elements inherit page language - hover to view CSS :lang()</h1>

			<p>garage</p>
			<p>double</p>
			<p>dame</p>
			<div></div>
		</body>
		</html>
```

## Subtag matching

Most screen readers will voice `lang=fr`, `lang=fr-FR` and `lang=fr-CA` as French if a French voice are installed. The exception is NVDA using the default OneCore speech synthesiser which voices `lang=fr`, `lang=fr-FR` if the French (France) language pack is installed, but voices `lang=fr-CA` as English unless the French (Canada) language pack is installed. The same thing happens with lang=de, lang=de-DE and lang=de-AT and the language packs for German (Germany) and German (Austria)

Test results here:
https://www.powermapper.com/tests/screen-readers/content/html-lang-subtags/

If you have a French (France) language pack language pack installed you can hear this happening in NVDA on https://www.canada.ca/fr.html if you use dev tools to change lang=fr to lang=fr-CA

The problem doesn't happen with the legacy NVDA eSpeak synthesiser which maps all fr- subtags to the same robotic French voice.

WilcoFiers:
Modern assistive technologies do not use xml:lang when the lang attribute is used, regardless of how a page is served. There is no situation in which a difference between xml:lang and lang can cause an accessibility issue as far as any of us can tell. The lang attribute still has to be correct though.

Closes #1921

Need for Call for Review: 2 weeks, this is a substantial change

---

## How to Review And Approve

### DIFF
--- a/_rules/html-page-lang-xml-lang-match-5b7ae0.md
+++ b/_rules/html-page-lang-xml-lang-match-5b7ae0.md
@@ -13,7 +13,7 @@ accessibility_requirements:
 input_aspects:
   - DOM Tree # The tree that HTML is parsed into.
 deprecated: |
-  This rule has been deprecated, as modern screen readers no longer use xml:lang when the lang attribute is given, regardless of which MIME type the pages is served with. This rule is not maintained anymore and should not be used.
+  This rule has been deprecated, as modern screen readers no longer use xml:lang when the lang attribute is given, regardless of which MIME type the page is served with. This rule is not maintained anymore and should not be used.
 acknowledgments:
   authors:
     - Jey Nandakumar
--- a/_rules/html-page-lang-xml-lang-match-5b7ae0.md
+++ b/_rules/html-page-lang-xml-lang-match-5b7ae0.md
@@ -1,6 +1,6 @@
 ---
 id: 5b7ae0
-name: HTML page lang and xml:lang attributes have matching values
+name: DEPRECATED — HTML page lang and xml:lang attributes have matching values
 rule_type: atomic
 description: |
   This rule checks that both `lang` and `xml:lang` attributes on the root element of a non-embedded HTML page, have the same primary language subtag.
@@ -12,6 +12,8 @@ accessibility_requirements:
     inapplicable: further testing needed
 input_aspects:
   - DOM Tree # The tree that HTML is parsed into.
+deprecated: |
+  This rule has been deprecated, as modern screen readers no longer use xml:lang when the lang attribute is given, regardless of which MIME type the pages is served with. This rule is not maintained anymore and should not be used.
 acknowledgments:
   authors:
     - Jey Nandakumar


_rules/html-page-lang-xml-lang-match-5b7ae0.md
```diff
@@ -12,6 +12,8 @@ accessibility_requirements:
     inapplicable: further testing needed
 input_aspects:
   - DOM Tree # The tree that HTML is parsed into.
+deprecated: |
+  This rule has been deprecated, as modern screen readers no longer use xml:lang when the lang attribute is given, regardless of which MIME type the pages is served with. This rule is not maintained anymore and should not be used.
```
carlosapaduarte:
```suggestion
  This rule has been deprecated, as modern screen readers no longer use xml:lang when the lang attribute is given, regardless of which MIME type the page is served with. This rule is not maintained anymore and should not be used.
```

carlosapaduarte (dismissed):
Just a single editorial

carlosapaduarte (approved):

Jym77 (approved):

tbostic32 (approved):

kengdoj (approved):

0066135 (approved):

colabottles (approved):

Issue #2081, opened 2023-07-10
How do we test definitions? (Add visibility tests with (horizontal) scroll?)


Jym77:
A problem I've been considering while trying to improve the visibility tests in Alfa…

As per our definition of visibility, content that can be scrolled into viewport is considered visible. I guess this was mostly added for vertical scrolling, which is the most frequent case, but it also covers horizontal scrolling:
```html
<style>
  .container {
      overflow: scroll;
  }
  .target {
      position: relative;
      left: 99999px
  }
</style>

<div class="container">
  Hello World <br />
  <div class="target">
    Lorem Ipsum
  </div>
</div>
```
Here, the `:target` element is pushed offscreen to the right, but a scrollbar is added to its parent `.container` making it visible as per our definition. A similar (and more natural) example can easily be made with vertical scrolling (which could be either the `.container` own scrolling, or the default scrolling of the browser window, depending on the case).

---
Now, in our examples for visibility (which are mostly spread around the various rules that use the definition), we tend to have cases with `left: -9999px` to verify that tools handle this case (invisible content). But iirc we do not have any test with content that appear to be invisible but can be scrolled in. That is, we never exerce the "can be scrolled in" part of the definition when validating tools implementation.
As always, testing definition is a bit clumsy in ACT rules since we can only test rules that rely on them and we tend to end up either copying the same-ish test in many rules, or having somewhat partial coverage for the definitions.

Anyway, should we add some test cases on rules concerned about visibility to verify that tools consider the "can be scrolled in" part of the definition?

Jym77:
This touches the more global question of "how do we test definitions?"

In that case, we have 50 rules who use "visible" and this would maybe need 3-4 cases to fully test the various corner cases (maybe more). It clearly doesn't make sense to add these to every rule.

Another way could be to spread examples across rules, adding 1 to some of the rules, another one to some other rules. However, this means that tools who do not correctly support the definition would randomly fail one rule but not another, which seems a bit arbitrarily. It also makes it harder for us to ensure that the definition is correctly covered by all the test cases, since the tests cases would be spread in the rules instead of being gathered in one place. This can create a big extra maintenance cost.

We did try to add some [examples of the definition](https://act-rules.github.io/pages/examples/visible/) for the big definitions. These are intended mostly for tool developers. However, these are not automatically testable, and as far as I know they are not shown on the WAI website either (and not really maintained…)

---
Some ideas and questions:
[...]

PR #2087, opened 2023-07-14
Turn "visible" examples into a rule

Jym77:
As per CG discussion, trying to turn definition's examples into a "rule" to see if we can benefit from the existing machinery to provide these in a testable way to tool implementers.
This is meant as a convenience for tool implementers, not to test any sort of requirements.

We will still need to add a bit of machinery to correctly export these test cases. We want them separated from the actual test cases of actual rules, so that testing them is purely opt-in from implementers.

Closes issue(s):

- Part of #2081 

Need for Call for Review:
This will require a 2 weeks Call for Review (new process)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


pages/examples/target-is-visible-exuqdg.md
```diff
@@ -0,0 +1,183 @@
+---
+id: exuqdg
+name: Target is visible
+rule_type: atomic
```
WilcoFiers: All of this makes this look like a regular rule. I wonder if we should do some things to this that move it away from that; in part not to confuse people, and in part just because it isn't that relevant. Like, I'm not sure this needs accessibility requirements, input aspect, applicability, expectation, etc. 

```suggestion
rule_type: definition-test
```

pages/examples/target-is-visible-exuqdg.md
```diff
@@ -0,0 +1,184 @@
...
+
+This `span` element is not visible because its text content has size 0.
+
+```html
+<span id="target" style="font-size: 0px">That you just can't place</span>
+```
+
+#### Failed Example 8
+
+This `span` element is not visible because it has the exact same color as its background.
```
carlosapaduarte: I will be able to see the `<span>`, but I won't be able to see its text content.

carlosapaduarte (commented):
Looking really good

carlosapaduarte:
And the visible definition will also have to be edited, because it links to the examples. It should link to the new "rule" instead.

pages/examples/target-is-visible-exuqdg.md
Jym77: 🤔 Ah, yes, indeed…

Jym77:
> And the visible definition will also have to be edited, because it links to the examples. It should link to the new "rule" instead.

Yes. Not sure what URL the new "rule" would have, though 😅

tbostic32 (approved):

carlosapaduarte (approved):

PR #2099, opened 2023-08-17
"Element with presentational children has no focusable content" (rule 307n5z) - adding examples 

dan-tripp-siteimprove:
Adding examples to rule presentational-children-no-focusable-content-307n5z.  Mostly uncontroversial, I think, except for Inapplicable Example 4 ("At the time of writing..."), which I expect will ruffle some feathers.

Closes issue(s): none 

Need for Call for Review:
This will not require a Call for Review : editorial changes (including to the applicability, expectation or examples section)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +136,93 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+```html
+<span role="img" aria-label="some ASCII art">****** This ASCII art ******* <a href="#">contains a link.</a></span>
+```
+
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-This element has a `link` role which does not have [presentational children][].
+None of the roles involved in this semantic table have [presentational children][].  (The roles are `table`, `row`, and `cell`, respectively, for the `<table>`, `<tr>`, and `<td>` elements.)  So this rule does not apply.
```
Jym77: There are actually also a `button` and a `link` role in the example, so the list of roles is not correct.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +136,93 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+<ul>
+<li>
+<q>Many individuals erroneously consider role="presentation" to be synonymous with aria-hidden="true"</q>
+</li>
+<li>
+<q>... the presentation role causes a given element to be treated as having no role or to be removed from the accessibility tree, but does not cause the content contained within the element to be removed from the accessibility tree.</q>
+</li>
+</ul>
+
+Adding to this confusion is a third term: "decorative".  The words "decorative" and "presentational" are often used interchangeably, but that usage is inaccurate.  The word "decorative" often appears in a sentence such as "marking an image as decorative" - that is, by adding `alt=""` to an `<img>` element.  "Decorative" in that context <i>does</i> mean "hidden" - and "hidden", again, is different from "presentational" - so using "decorative" and "presentational" interchangeably is inaccurate.  At the time of writing (August 2023), the ACT definition of "[marked as decorative][]" unfortunately encourages this inaccurate usage: it states that <q>An element is marked as decorative if ... it has an explicit role of none or presentation</q>.
```
Jym77: That's why we use "**marked** as decorative", and not "decorative" 😄 
Being "marked as decorative" is showing author **intention** to somehow remove the content. It doesn't necessarily means that the content will effectively be removed/hidden/… (because the presentational role conflict resolution can trigger).

I also consider "decorative" as being a semantic property of the page. "presentational" (or role of `presentation` (post-conflict resolution)) is more of a functional/effective property of the page. The same way that "being a label (WCAG sense of the term)" is a semantic property while using a `label` element is the functional property (and the way to convey  that semantics programmatically). That is, an image is "decorative" not because it has `alt=""` but because it is "serving only an aesthetic purpose, providing no information, and having no functionality" (WCAG's "pure decoration"). Adding `alt=""` is a way to convey programmatically the fact that the image is decorative, so that ATs can treat it differently from a non-decorative image.

In that spirit, "marked as decorative" is "the author tried to convey programmatically the semantic information that something is decorative". This may work as intended (and the stuff will be either presentational or hidden), but this may also fail (and the stuff will be exposed as normal).

Jym77 (commented):
The examples are good. The descriptions are very verbose and going into much more details compared to what we usually do. Some of these discussions might be moved to the background section instead (e.g. the discussion about presentational/hidden/decorative.

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove:
```suggestion
None of the roles involved in this semantic table have [presentational children][].  (The roles are `table`, `row`, `cell`, `button`, and `link`, for the `<table>`, `<tr>`, `<td>`, `<button>`, and `<a>` elements, respectively.)  So this rule does not apply.
```

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: Okay, so decorative != marked as decorative.  Even if that's okay (which is unclear to me), I still think we have a problem.  

I would draw a venn diagram here if I could, but I can't, so here is a tree which represents a venn diagram of the situation you described.  (In this venn diagram tree, all child sets totally consume the parent set.  eg. Set B minus Set C minus Set D = an empty set.)

- Set A: marked as decorative 
  - Set B: "marked as decorative" worked 
    - Set C: presentational 
    - Set D: hidden
  - Set E: "marked as decorative" did not work (==> still has semantics) 

That presents two problems:
1) Sets C and D are so different, I think it's confusing for us to include them both under "Set B: "marked as decorative" worked".
2) I think that "Set D: hidden" corresponds to what most people think when they hear "marked as decorative".  Also, I suspect that it overlaps precisely with their ideas of "decorative" and WCAG's "pure decoration" - and that's a good thing.  If we go with the the above venn diagram, someone might say that ACT is out of touch, and they would be right.

dan-tripp-siteimprove:
> Some of these discussions might be moved to the background section instead (e.g. the discussion about presentational/hidden/decorative.

I can do that.  How about I move Inapplicable Examples 3 and 4 to the Background section?

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +136,93 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+<ul role="tablist">
+	<li role="tab">
+		<a href="#">Tab 1</a>
+	</li>
+</ul>
+```
+
+#### Failed Example 5
+
+This element with the `img` role contains an `<a>` element.  The `img` role has [presentational children][].  The `<a>` element is included in [sequential focus navigation][].  So the element with the `img` role fails the rule.  (Interestingly, no similar failure is possible with a standard HTML `<img>` element, because according to the HTML specification, the `<img>` element has no closing tag.  So it's impossible to put anything inside of an `<img>` element.)
```
carlosapaduarte:
```suggestion
This element with the `img` role contains an `<a>` element.  The `img` role has [presentational children][].  The `<a>` element is included in [sequential focus navigation][].  So the element with the `img` role fails the rule.
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +136,93 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+						<td><button>button in semantic table - no problem</button></td>
+						<td><a href="#">anchor in semantic table - no problem</a></td>
+				</tr>
+		</tbody>
+</table>
 ```
 
+#### Inapplicable Example 2
+
+This `<a>` element has a `link` role, which does not have [presentational children][].  So this element - perhaps surprisingly - does not fail this rule, because it's inapplicable.  This is the case even though it's invalid HTML.  Any time there's a link as the <i>outer</i> element, that will be inapplicable under this rule.  (If the elements were switched - that is, if the `<button>` element contained the `<a>` element - then that would fail this rule.)
```
carlosapaduarte:
```suggestion
This `<a>` element has a `link` role, which does not have [presentational children][].  So this element does not fail this rule, because it's inapplicable. 
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +136,93 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+
+```html
+<a href="https://w3.org"><button>W3C Website</button></a>
+```
+
+#### Inapplicable Example 3
+
+The applicability of this rule does not include elements with an <i>explicit</i> role of `presentation`.  Only <i>implicit</i> (through roles with [presentational children][]).  This example shows that.
```
carlosapaduarte:
```suggestion
The applicability of this rule does not include elements with an <i>explicit</i> role of `presentation`.  Only <i>implicit</i>, through roles with [presentational children][]. 
```

_rules/presentational-children-no-focusable-content-307n5z.md
carlosapaduarte: While I think this feedback about how some definitions can be perceived is important, I don't think this is a good place to discuss it. I would suggest @dan-tripp-siteimprove opens an issue to discuss these points, and rewrite the example description is a less verbose manner.

carlosapaduarte (dismissed):
Tried to cut down on the verbosity of the example's descriptions.

dan-tripp-siteimprove:
> I would suggest @dan-tripp-siteimprove opens an issue to discuss these points, and rewrite the example description is a less verbose manner.
@carlosapaduarte  That's fair.  I haven't opened a new issue yet - I'll discuss it in-person first.  I did rewrite the example description and pushed another commit (d161d08ba6eba31686eee65f35f54bc425725e79).  Let me know what you think.

tombrunet:
@dan-tripp-siteimprove I think what's here is fine. Is it worth mentioning for Inapplicable 4 that that issue is covered by another rule (https://www.w3.org/WAI/standards-guidelines/act/rules/6cfa84/)?

tombrunet (approved):

dan-tripp-siteimprove:
> @dan-tripp-siteimprove I think what's here is fine. Is it worth mentioning for Inapplicable 4 that that issue is covered by another rule (https://www.w3.org/WAI/standards-guidelines/act/rules/6cfa84/)?

I think yes, and I just did it in commit d7ea3d32c889b8a2deb9f1273337c19599ceebae.  Thank you.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +135,79 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+```html
+<span role="img" aria-label="some ASCII art">****** This ASCII art ******* <a href="#">contains a link.</a></span>
+```
+
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-This element has a `link` role which does not have [presentational children][].
+None of the roles involved in this semantic table have [presentational children][].  (The roles are `table`, `row`, `cell`, `button`, and `link`, for the `<table>`, `<tr>`, `<td>`, `<button>`, and `<a>` elements, respectively.)  So this rule does not apply.
```
Jym77: Wouldn't the rule apply to the `button`? (and pass because it has no focusable child).

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +135,79 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+						<td><button>button in semantic table - no problem</button></td>
+						<td><a href="#">anchor in semantic table - no problem</a></td>
+				</tr>
+		</tbody>
+</table>
 ```
 
+#### Inapplicable Example 2
+
+This `<a>` element has a `link` role, which does not have [presentational children][].  So this element does not fail this rule, because it's inapplicable.
```
Jym77: Same here, the rule would be Applicable to the `button` (only), and pass.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +135,79 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+
+The applicability of this rule does not include elements with an <i>explicit</i> role of `presentation`.  Only <i>implicit</i>, through roles with [presentational children][]. 
+
+```html
+<ul>
+	<li role="presentation">
+		<a href="#">inapplicable</a>
+	</li>
+</ul>
+```
```
Jym77: As per our consistency rules, we need to make sure that the Inapplicable examples do not fail the accessibility requirement (in this case, 4.1.2). I'm not sure exactly what happens in that case (and whether the presentational conflict triggers or not).

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,16 +135,79 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+
+#### Inapplicable Example 4
+
+This example shows that the applicability of this rule does not include hidden elements.  The applicability only includes presentational elements (through roles with [presentational children][]).  (The difference between "hidden" and "presentational" is discussed in the [ARIA 1.2 specification](https://www.w3.org/TR/wai-aria-1.2/#presentation)).  This example is inapplicable, but if it didn't have aria-hidden="true", then it would fail the rule.  This example (as it is, with aria-hidden) fails the related rule [Element with aria-hidden has no content in sequential focus navigation](https://www.w3.org/WAI/standards-guidelines/act/rules/6cfa84/).
+
+```html
+<button aria-hidden="true">
+	<a href="#">inapplicable</a>
+</button>
+```
```
Jym77: Same here. Dose that pass 4.1.2?
Also, `aria-hidden` is actually ignored on focusable stuff as per [Including Elements in the Accessibility Tree](https://www.w3.org/TR/wai-aria-1.2/#tree_inclusion): 
> user agents MUST provide an [accessible object](https://www.w3.org/TR/wai-aria-1.2/#dfn-accessible-object) in the [accessibility tree](https://www.w3.org/TR/wai-aria-1.2/#dfn-accessibility-tree) for DOM [elements](https://www.w3.org/TR/wai-aria-1.2/#dfn-element) that meet any of the following criteria:
> 
> * Elements that are not [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden) and may fire an [accessibility API](https://www.w3.org/TR/wai-aria-1.2/#dfn-accessibility-api) [event](https://www.w3.org/TR/wai-aria-1.2/#dfn-event), including:
>   * Elements that are currently focused, even if the element or one of its ancestor elements has its [aria-hidden](https://www.w3.org/TR/wai-aria-1.2/#aria-hidden) attribute set to true.

So, I'm not sure how this ends up ultimately.

Jym77 (dismissed):

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: Yes.  I muddled this example. I was trying to get across the idea that the `<table>`, `<tr>`, and `<td>` elements (only) are inapplicable.  I just pushed a commit (9dd1ac652891bf99d4d7ea1f6c8598ccdb7c092a) that tries to make it more clear.

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: Fair point.  My intent was to comment on the inapplicability of the `<a>` element - the `<button>` element was just there as an example of a focusable child element.  I just pushed 527cfd96cec86da0fa5850834c83cc1691302482 which, I hope, clears it up.

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: Good point.  I read [test-cases.md](https://github.com/act-rules/act-rules.github.io/blob/d703022a0298f1996f980ee204b5304fa9aecaf4/pages/design/test-cases.md?plain=1).  Its guidance of "Consistency With Accessibility Requirements", combined with [Presentational Roles Conflict Resolution](https://www.w3.org/TR/wai-aria-1.2/#conflict_resolution_presentation_none) make it so that I don't know how to create an inapplicable example which explains this aspect of the rule.

I have seen customers wrongly report a false positive for this rule, and they write something like: "This is a false positive because this element does not have role=presentation.  It has role=tab."   Then I reply: "This rule isn't about an explicit role of presentation.  It's about an /implicit/ role of presentation."  So with this inapplicable example, I was trying to help explain that.  

So I tried something different just now in commit 9f81ae4a95da9ae5b412ccb31199917d9cbef491: I removed this inapplicable example and added some equivalent wording in the "background" section.  What do you think?

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: For this one too, I replaced this example with some wording in the "background" section.  I did this in commit c8d188d7589b27abbcc7793b356bc04e60c0d2ed.  Let me know what you think.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,14 +137,54 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+```html
+<span role="img" aria-label="some ASCII art">****** This ASCII art ******* <a href="#">contains a link.</a></span>
+```
+
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-This element has a `link` role which does not have [presentational children][].
+The roles that build this semantic table structure (`table` for `<table>`, `row` for `<tr>`, and `cell` for `<td>`) do not have [presentational children][].  So this rule does not apply to them.  (This rule does apply to the `<button>` element within the table.  The `<button>` element passes the rule.)
```
Jym77: I still believe the `button` should go away…
From an implementer point of view, tools will report "Passed" rather than "Inapplicable" on that example. This is not a huge deal as it they are equivalent for consistency, but it still creates some weirdness. 

We could have this example as a Passed Example. It would still fulfil the same goal (show that `table` role and its friends are not applicable) (because tools that apply the rule to them would report a Failed outcome). OTOH, I am not sure that the `button` is adding much value to the example (the link is already a focusable element inside the table, we can have two links and the example is still working as intended).

Jym77 (changes requested):

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: Fair points, I'll discuss one-on-one some time.

carlosapaduarte: I agree with @Jym77 

If we want this as an inapplicable example then we should not ask implementations to output "Passed". I agree with either solution proposed: make this a passed example or remove the `button` element.

carlosapaduarte (dismissed):

_rules/presentational-children-no-focusable-content-307n5z.md
dan-tripp-siteimprove: I fixed this a minute ago by removing the button element.

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -42,16 +42,17 @@ This rule applies to elements with a [semantic role][] that defines its [childre
 
 Elements with a [semantic role][] that has [presentational children][] will not have any descendants in the accessibility tree. If any of those descendants are included in [sequential focus navigation][], this causes the focus to land on an element that has no corresponding node in the [accessibility tree][]. The result is that there is no programmatic name or role available for assistive technologies. There are other problems that can come from [presentational children][] too. These must be tested separately.
 
+This rule is often misunderstood as applying to elements with an <i>explicit</i> role of `presentation`.  In fact, this rule only applies to elements which have been given an <i>implicit</i> role of `presentation` through the [presentational children][] mechanism.  Similarly, this rule does not apply to elements with `aria-hidden="true"`.
```
Jym77:
```suggestion
This rule is often misunderstood as applying to elements with an _explicit_ role of `presentation`.  In fact, this rule only applies to elements which have been given an _implicit_ role of `presentation` through the [presentational children][] mechanism.  Similarly, this rule does not apply to elements with `aria-hidden="true"`.
```
(which Markdown  conversion then translates to `em` elements)

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -93,6 +94,15 @@ This element with `menuitemcheckbox` role has an `input` element as a descendant
 </ul>
 ```
 
+#### Passed Example 4
+
+This `<button>` element has an `<a>` element as a [child][].  The `<a>` element has no `href` attribute, so it isn't included in [sequential focus navigation][].  So this `<button>` element passes the rule.
```
Jym77:
```suggestion
This `<button>` element has an `a` element as a [child][].  The `a` element has no `href` attribute, so it isn't included in [sequential focus navigation][].  So this `button` element passes the rule.
```
We don't use angled brackets for elements' names in ACT Rules…

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,14 +137,53 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
 </ul>
 ```
 
+#### Failed Example 4
+
+This element with the `tab` role contains an `<a>` element.  The `tab` role has [presentational children][].  The `a` element is included in [sequential focus navigation][].  So the element with the `tab` role fails the rule.  (This tablist implementation is non-functional for users.  It's not meant to function - it's only meant to show roles.)
```
Jym77:
```suggestion
This element with the `tab` role contains an `a` element.  The `tab` role has [presentational children][].  The `a` element is included in [sequential focus navigation][].  So the element with the `tab` role fails the rule.  (This tablist implementation is non-functional for users.  It's not meant to function - it's only meant to show roles.)
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,14 +137,53 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+<ul role="tablist">
+	<li role="tab">
+		<a href="#">Tab 1</a>
+	</li>
+</ul>
+```
+
+#### Failed Example 5
+
+This element with the `img` role contains an `<a>` element.  The `img` role has [presentational children][].  The `<a>` element is included in [sequential focus navigation][].  So the element with the `img` role fails the rule.
```
Jym77:
```suggestion
This element with the `img` role contains an `a` element.  The `img` role has [presentational children][].  The `a` element is included in [sequential focus navigation][].  So the element with the `img` role fails the rule.
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,14 +137,53 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+```html
+<span role="img" aria-label="some ASCII art">****** This ASCII art ******* <a href="#">contains a link.</a></span>
+```
+
 ### Inapplicable
 
 #### Inapplicable Example 1
 
-This element has a `link` role which does not have [presentational children][].
+None of the roles that build this semantic table structure (`table` for `<table>`, `row` for `<tr>`, `columnheader` for `<th>`, and `cell` for `<td>`) have [presentational children][].  So this rule does not apply to them.
```
Jym77:
```suggestion
None of the roles that build this semantic table structure (`table` for `table`, `row` for `tr`, `columnheader` for `th`, and `cell` for `td`) have [presentational children][].  So this rule does not apply to them.
```

_rules/presentational-children-no-focusable-content-307n5z.md
```diff
@@ -127,14 +137,53 @@ This element with `menuitemcheckbox` role has a checkbox as a child. Because the
...
+		<tr>
+			<td><a href="#">link in table data cell - no problem</a></td>
+		</tr>
+	</tbody>
+</table>
+```
+
+#### Inapplicable Example 2
+
+This `<a>` element has a `link` role, which does not have [presentational children][].  So this `<a>` element does not fail this rule, because it's inapplicable.  To have a "focusable element within a focusable element" like this is a bad practice, but this rule doesn't directly check for it.
```
Jym77:
```suggestion
This `a` element has a `link` role, which does not have [presentational children][].  So this `a` element does not fail this rule, because it's inapplicable.  To have a "focusable element within a focusable element" like this is a bad practice, but this rule doesn't directly check for it.
```

Jym77 (changes requested):
Just some cleanup.

Jym77 (approved):

shunguoy (approved):

PR #2100, opened 2023-08-21
Update element-lang-valid-de46e4.md

### LINKED ISSUE #2091: Improve examples in de46e4
Remove `<html>` and `<body>` from examples in element-lang-valid-de46e4 so these don't cause failures.

kengdoj:
Add valid lang attribute and value to `<html>` tags in all examples to remove unintended failures.

Closes https://github.com/act-rules/act-rules.github.io/issues/2091, which was identified during the annual re-review of this rule.

Call for review: 1 week

### DIFF
--- a/_rules/element-lang-valid-de46e4.md
+++ b/_rules/element-lang-valid-de46e4.md
@@ -71,7 +71,7 @@ There are differences in how assistive technologies handle unknown and invalid l
 This `article` element has a `lang` [attribute value][] which has a [known primary language tag][].
 
 ```html
-<html>
+<html lang="es">
 	<body>
 		<article lang="en">
 			They wandered into a strange Tiki bar on the edge of the small beach town.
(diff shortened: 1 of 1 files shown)


colabottles (approved):

sujisreerama (approved):

tbostic32 (approved):
Only question is how you picked what lang tag to put in. Was it just random? Some of the tags don't feel particularly relevant to the content. Did we just need the lang tag to be different than the content?

kengdoj:
> Only question is how you picked what lang tag to put in. Was it just random? Some of the tags don't feel particularly relevant to the content. Did we just need the lang tag to be different than the content?

Yes, to all. It was fairly random and also the language abbreviations I knew. These could have been any valid language really.

Jym77 (approved):

Issue #1458, opened 2020-09-23
Visible label is part of accessible name (2ee8b8): Expectation seems to have unintended consequences


kasperisager:
The expectation currently reads:

> For each target element, all text nodes in the visible text content either match or are contained within the accessible name of this target element, except for characters in the text nodes used to express non-text content. Leading and trailing whitespace and difference in case sensitivity should be ignored.

According to this expectation, something like the following passes the rule:

```html
<button aria-label="How are you">
  <span>you</span>
  <span>How</span>
  <span>are</span>
</button>
```

That seems a little odd 🤔 Shouldn't the rule instead be looking at the concatenation of the data of the relevant text nodes?

Summoning @WilcoFiers as you authored #1419.

kasperisager:
Both Alfa, aXe, and QualWeb concatenate the text nodes so I'm guessing we'll want to reflect that in the rule 🙈 That does bring up an interesting case though for code such as this:

```html
<div role="button" aria-label="Hello world">
  <p>Hello</p><p>world</p>
</div>
```

That button visually renders as two separate words, `Hello` and `world`, but the concatenated text node data is `Helloworld`. We're currently seeing a handful of cases like this across customer sites in Siteimprove and I'm leaning towards considering them false positives. A more realistic case, which causes issues when minified, is this:

```html
<a href="#" aria-label="Some article by John Doe">
  <h6>Some article</h6>
  <p>by John Doe</p>
</a>
```

When minified, the concatenated text will be `Some articleby John Doe`.

WilcoFiers:
I tried to dig up the use case for this, but I can't find it. I remember why we did it though. If we're concatenating, we need to make the assumption that the text will be part of the same piece of text, and that it isn't rearranged with CSS in some way to appear in a different order.

Jym77:
Example 2 of [Failure technique F96](https://w3c.github.io/wcag/techniques/failures/F96) seems to close the case of adding text in the middle of a label to create an accessible name (which, as far as I understand, is the shopping cart example @WilcoFiers mentioned during call):
> A download link reads "Download specification" but there is invisible link text so that the accessible name of that link is "Download gizmo specification". While the visible label text is contained in the accessible name, there is no string match which may prevent the link from being activated by speech input.
> `<a href="#">Download <span class="accessibly-hidden">gizmo</span> specification</a>`

dan-tripp-siteimprove:
[...]

PR #2075, opened 2023-06-22
Rule 2ee8b8 ("Visible label is part of accessible name"): introducing a new "label in name algorithm".  

dan-tripp-siteimprove:
<< Describe the changes >>

Closes issue(s):

- closes #2051, closes #1989, closes #1619, closes #1615, closes #1458 
- does not handle: #2040 #1500 #451 

Need for Call for Review:
This will require a 2 weeks Call for Review 

---

### **When creating PR:**

- [ x] Make sure you're requesting to **pull a branch** (right side) to the `develop` branch (left side).
- [x ] Make sure you do not remove the "How to Review and Approve" section in your pull request description

### **After creating PR:**

- [ x] Add yourself (and co-authors) as "Assignees" for PR.
- [ x] Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- [x ] [Link the PR to any issue it solves](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue). This will be done automatically by [referencing the issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) at the top of this comment in the indicated place.
- [ x] Optionally request feedback from anyone in particular by assigning them as "Reviewers".

### **When merging a PR:**

- [ ] Close any issue that the PR resolves. This will happen automatically upon merging if the PR was correctly linked to the issue, e.g. by referencing the issue at the top of this comment.

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Call for Review period. In case of disagreement, the longer period wins.

PR #2101, opened 2023-08-21
Update visible-label-in-accessible-name-2ee8b8.md

kengdoj:
Update examples to include https://github.com/act-rules/act-rules.github.io/issues/1458#issuecomment-1555077770
Closes issue(s):

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -38,7 +38,7 @@ This rule applies to any element for which all the following is true:
 
 ## Expectation
 
-For each target element, all [text nodes][] in the [visible text content][] either match or are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+For each target element, all the [text nodes][] in the [visible text content][] match in sequence and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
kengdoj:
```suggestion
For each target element, all [text nodes][] in the [visible text content][] match in sequence and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -38,7 +38,7 @@ This rule applies to any element for which all the following is true:
 
 ## Expectation
 
-For each target element, all [text nodes][] in the [visible text content][] either match or are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+For each target element, all [text nodes][] in the [visible text content][] match in sequence and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
tbostic32: I'm wondering if we should have a glossary definition for `match in sequence`. I know what you mean from being part of the conversations, but I don't think it reads very well. Some more explanation for how you are matching the sequence of characters exactly might be helpful.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -97,10 +97,10 @@ This button has [visible][] text that is contained within the [accessible name][
 
 #### Passed Example 5
 
-This button has [visible][] text that does not need to be contained within the [accessible name][], because the "x" text node is [non-text content][].
+This button has [visible][] text that does not need to be contained within the [accessible name][], because the "x" text node is [non-text content][]. Note: this would need to meet SC 1.1.1 Non text content.
```
tbostic32, line 100: Should we be listing SC 1.1.1 as a secondary requirement? Feels like we are leaning no, but figure we should consider it if its possibly related.

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -97,10 +97,10 @@ This button has [visible][] text that is contained within the [accessible name][
...
 
 ```html
-<button aria-label="close">X</button>
+<button aria-label="anything">X</button>
```
Jym77, line 103: Shouldn't we keep the "Close" label, which makes this pass 1.1.1?
(assuming this is indeed a close button)

Jym77 (dismissed):
How do the changes relate to https://github.com/act-rules/act-rules.github.io/pull/2075?

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj: I think SC 1.1.1 would be a secondary requirement, but I'd like to leave that for a separate PR.

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj: Because the Expectation says

> except for characters in the [text nodes][] used to express [non-text content][]

there is no expectation, and this example passes. I used `aria-label="anything"` to make it clear that really any description would pass this rule. But, you are right that SC 1.1.1 would then be not satisfied. But I'd like to consider SC 1.1.1 being secondary and handle that in a separate PR.

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj: There is an existing definition for "matching characters" from [WAI Glossary Harmonization](https://docs.google.com/spreadsheets/d/16C3vsiNmrRBwHcNx_WhQYoMaCkUIc6YjWTfqfEnXU9c/edit#gid=0):

> A sequence of characters is considered to match another if, after removing leading and trailing space characters and replacing remaining occurrences of one or more space characters with a single space, the two sequences of characters are equal character-by-character, ignoring any differences in letter casing.

Would this work?
```suggestion
For each target element, all [text nodes][] in the [visible text content][] match characters and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```

dan-tripp-siteimprove: I noticed that ACT already has a copy of that WAI Glossary Harmonzation definition of "matching characters".  [ACT's copy of the definition is here](https://github.com/act-rules/act-rules.github.io/blob/develop/pages/glossary/matching-characters.md).   So it would be easy to add a reference to it.

_rules/visible-label-in-accessible-name-2ee8b8.md
dan-tripp-siteimprove: This comes up often around here: when passed examples fail something else.  

I like it best when there is a note on the passed example which documents that "something else".  In this case that note could be: "This example passes the rule and will fail SC 1.1.1".

We could call these "curve ball passed examples". 

To have examples like these, and to document them as such, meets two goals, both important:
* It emphasizes _why_ the example passes the rule.  (In this case: "to make it clear that really any description would pass this rule", as Kathy put it).  I think that this aids understanding, a lot.
* It tells everyone that even though this is a "passed example" _for this rule_, it's not a good example overall, so no one should copy and paste the code.

Jean-Yves did this a lot in the [examples](https://github.com/act-rules/act-rules.github.io/blob/c302b7504ccc66acc2715fa5884d7d2411dbe45e/_rules/aria-role-permitted-j7zzqr.md) for his recent [PR 2084](https://github.com/act-rules/act-rules.github.io/pull/2084): 
* "Using an explicit role of generic for any element is globally not recommended, but not an error by itself."
* "Using an explicit role which is the same as the [implicit role](https://github.com/act-rules/act-rules.github.io/blob/c302b7504ccc66acc2715fa5884d7d2411dbe45e/_rules/aria-role-permitted-j7zzqr.md#implicit-role) is not recommended, but not an error by itself."
* "The btn token is not a valid role and is therefore ignored."
* "Even though the alert role is not allowed, the first token is a valid role and is therefore used."

I think it's a good thing

dan-tripp-siteimprove (approved):

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -38,7 +38,7 @@ This rule applies to any element for which all the following is true:
 
 ## Expectation
 
-For each target element, all [text nodes][] in the [visible text content][] either match or are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
+For each target element, all [text nodes][] in the [visible text content][] match characters and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```
kengdoj:
```suggestion
For each target element, all [text nodes][] in the [visible text content][] [match characters][] and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. Leading and trailing [whitespace][] and difference in case sensitivity should be ignored.
```

0066135 (approved):

dd8 (commented):
The updated rules references the #matching-characters glossary definition which says:

>A sequence of characters is considered to match another if, after removing leading and trailing [space characters](https://html.spec.whatwg.org/#white_space) and replacing remaining occurrences of one or more space characters with a single space, the two sequences of characters are equal character-by-character, ignoring any differences in letter casing

The link to https://html.spec.whatwg.org/#white_space in the #matching-characters definition is broken so it's not clear which definition of space characters is intended. The various WHATWG specs have moved to a common definition of ASCII whitespace defined here:
https://infra.spec.whatwg.org/#ascii-whitespace

This is different to the Unicode whitespace definition the rule currently uses:
https://www.w3.org/WAI/standards-guidelines/act/rules/2ee8b8/proposed/#whitespace

Notably ASCII whitespace doesn't include `&nbsp;` which would be problematic for this example from @dan-tripp-siteimprove 

```
<a aria-label="compose email" href="#">Compose &nbsp;&nbsp;<br> email</a>
```
Desired behaviour: pass this rule

kengdoj:
Thank you @dd8 for catching that. 

I've opened https://github.com/act-rules/act-rules.github.io/pull/2144 to update the matching characters definition.

Jym77:
Call for Review ends on December 21st.

Jym77:
Removing from Call for Review as two reviewers were from the same organisation…

_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -143,6 +143,22 @@ This link has [visible][] text with mathematical symbols, that does not match th
...
+```html
+<a href="#" aria-label="non-standard">nonstandard</a>
+```
+
+#### Failed Example 5
+
+This link has [visible][] text does not match the [accessible name][] because there are extra spaces in the accessible name. 
+
+```html
+<a aria-label="Call 1 2 3. 4 5 6. 7 8 9 0." href="tel:1234567890">123.456.7890</a>
```
giacomo-petri: Was the inclusion of a full stop at the end of the aria-label attribute value deliberate? If so, do we need to provide clarification regarding its potential impact on the outcome? In the issue description only extra spaces are mentioned.

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj:
```suggestion
For each target element, all [text nodes][] in the [visible text content][] [match characters][] and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. 
```

kengdoj:
```suggestion
For each target element, all [text nodes][] in the [visible text content][] [match characters][] and are contained within the [accessible name][] of this target element, except for characters in the [text nodes][] used to express [non-text content][]. 
```

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj:
```suggestion
<a aria-label="1 2 3. 4 5 6. 7 8 9 0" href="tel:1234567890">123.456.7890</a>
```

kengdoj: thanks! I edited the aria-label.

WilcoFiers (approved):

tbostic32 (approved):

carlosapaduarte (approved):

_rules/visible-label-in-accessible-name-2ee8b8.md
tombrunet: This works, but perhaps a cleaner way to handle this would be to modify the second condition of the applicability to filter out not just non-visible text content, but also non-text content. That would cause this example to move to the non-applicable section.

tombrunet:
Added a comment regarding Pass 5. It feels to me like this scenario should be considered inapplicable, but that would require a modification to the applicability. The rule is consistent as-is, but it seems more accurate to say that the non-text content scenario is not applicable to this rule rather than to say that it passes it...

colabottles (approved):

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj: Agree, Tom. 

But, Applicability has to be objective at this time, so the _except non-text content_ part is under the Expectation section for now.

tombrunet (approved):

PR #2107, opened 2023-08-31
Images of text, tweak the assumption

WilcoFiers:
From my annual review. 1.4.9 should not be in the assumption since the exception we're assuming doesn't occur is not part of SC 1.4.9.

Need for Call for Review: 1 week, not just editorial

---

## How to Review And Approve


carlosapaduarte (approved):

Jym77 (approved):

HelenBurge (approved):

tombrunet (approved):

Jym77:

PR #2110, opened 2023-08-31
"Headers attribute specified on a cell refers to cells in the same table element" [a25f45]: require headers to be `th` elements

### LINKED ISSUE #1910: Rule a25f45 table headers - mismatch between rule applicability, examples and HTML spec
There's a mismatch between the rule applicability, the rule example wording and the HTML spec:

- Rule applicability allows headers to reference `td` or `th` elements (cells, as defined by the HTML Standard)
- Rule wording in the examples suggest `role=rowheader` or `role=columnheader` is cause of pass/fail, but these aren't relevant to the rule applicability
- The HTML Standard only allows headers to reference `th` elements 

https://act-rules.github.io/rules/a25f45#expectation-1

## HTML Standard
The HTML living standard requires `headers` to reference `th` elements

> The [headers](https://html.spec.whatwg.org/multipage/tables.html#attr-tdth-headers) attribute, if specified, must contain a string consisting of an [unordered set of unique space-separated tokens](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#unordered-set-of-unique-space-separated-tokens), none of which are [identical to](https://infra.spec.whatwg.org/#string-is) another token and each of which must have the value of an [ID](https://dom.spec.whatwg.org/#concept-id) of a [th](https://html.spec.whatwg.org/multipage/tables.html#the-th-element) element taking part in the same [table](https://html.spec.whatwg.org/multipage/tables.html#concept-table) as the [td](https://html.spec.whatwg.org/multipage/tables.html#the-td-element) or [th](https://html.spec.whatwg.org/multipage/tables.html#the-th-element) element (as defined by the [table model](https://html.spec.whatwg.org/multipage/tables.html#table-model)).

## Rule Examples 
The rule examples suggest that `role=rowheader` or `role=columnheader` is relevant to passing or failing the rule:

https://act-rules.github.io/rules/a25f45#passed-example-5
https://act-rules.github.io/rules/a25f45#failed-example-4

tombrunet:
Updates based on #1910 
* Added to the expectation that the referenced element is expected to be a th
* Updated Pass 3 and Fail 4.

Closes issue(s):

- closes #1910 

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


tombrunet:
@WilcoFiers I put in a change to the expectation based on the issue description, but I'm a bit hesitant about that. I think ATs handle this regardless of whether or not it points to a td or a th. We certainly don't have any fail examples here to demonstrate that. I certainly agree that the examples should be in line with the spec. But, should we:
1) Keep this change to the expectation
2) Revert the expectation and add a failure case where a headers attribute points to a td element
3) Revert the change and do not add new test cases

_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
```diff
@@ -36,7 +37,7 @@ This rule applies to any `headers` attribute specified on a [`cell`][] within a
 
 ## Expectation 1
 
-Each target's [attribute value][] is a [set of space separated tokens][]. Each token is the value of the `id` attribute of an element, that is a [`cell`][] of the same [`table`][].
+Each target's [attribute value][] is a [set of space separated tokens][]. Each token is the value of the `id` attribute of an element, which is a [`cell`][] of the same [`table`][] and is a `th` element.
```
WilcoFiers: I would feel much better about this if we had some test data showing that referencing `td`s, with and without role=row/columnheader doesn't work. Would it be possible to get someone to do some testing before we say this definitely fails WCAG?

WilcoFiers: Also, I think we can simplify this a bit:

```suggestion
Each target's [attribute value][] is a [set of space separated tokens][]. Each token is the value of the `id` attribute of an element, which is a `th` element of the same [`table`][].
```

_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
```diff
@@ -110,14 +111,14 @@ The `headers` attribute on the cell refers to a `th` element within the same `ta
...
+The `headers` attribute on the data cells in the second row refers to a `th` element with a role of `columnheader` within the same `table`.
 
 ```html
 <table>
 	<thead>
 		<tr>
-			<td role="columnheader" id="header1">Projects</td>
-			<td role="columnheader" id="header2">Objective</td>
+			<th role="columnheader" id="header1">Projects</th>
+			<th role="columnheader" id="header2">Objective</th>
```
WilcoFiers: Probably don't need the role anymore if we do this. If we're going to fail `headers` referencing `td` elements we should have a new failed example for it. What we have now references an element inside the cell, which I do think is a problem.

WilcoFiers (changes requested):

_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
Jym77: It is at least invalid HTML: [headers attribute](https://html.spec.whatwg.org/multipage/tables.html#attr-tdth-headers), and shouldn't create a relationship: [Forming relationship](https://html.spec.whatwg.org/multipage/tables.html#header-and-data-cell-semantics), step 7 only keeps the "headers".

Due to non-interference with the host language, I'd be surprised if that worked. But I agree we should test it nonetheless.

tombrunet: I used this as a test with just regular HTML:
https://codesandbox.io/s/table-test-68d78y?file=/index.html

JAWS doesn't care if the referenced element is a `td` or a `th`, it reads the referenced headings in both cases. Surprisingly, NVDA and desktop Voiceover seem to ignore the `headers` attribute and use the automatic calculation - they won't read `td`'s, but they read the `th` at the top of the column rather than the `th` that I referenced via `headers`. 

So, 2 of the 3 desktop screen readers don't handle headers properly even when they're referencing a th. The 3rd doesn't care if it's a td or a th. Given that, I'm having a hard time seeing how this rule has any real world value - the results don't seem to be a factor of what the headers attribute is referencing.

tombrunet: To be clear, I was using Chrome on Mac, and Edge on Windows.

WilcoFiers (approved):

PR #2118, opened 2023-09-15
Remove "DEPRECATED" from rule names

WilcoFiers:
This "DEPRECATED" in the rule title is no longer necessary as these rules are now sorted in their own section on the ACT site, and the deprecated rule text is added to the rule page.

Need for Call for Review: none, editorial change

---

## How to Review And Approve


carlosapaduarte (approved):

Jym77 (approved):

HelenBurge (approved):
Looks good to me

Issue #2119, opened 2023-09-15
Rule idea: ARIA required accessible names


WilcoFiers:
Basically, a rule that checks that when ARIA says "Accessible Name Required: true" on a role, that whenever that role is present, it has an accessible name.

This rule has a **lot** of overlap with other rules, image names, button names, link names, etc. Implementation would likely then end up with many-to-one mappings. This will require various secondary requirements to be added from WCAG, as this overlaps with a number of criteria.

giacomo-petri:
I'll address that in combination with #2189.

PR #2120, opened 2023-09-15
Map Empty-heading rule to ARIA instead of WCAG

WilcoFiers:
PR related to #2119. As has been discussed a few times in TF and CG, empty headings is not a failure of 1.3.1, at least not anymore as far as we know. Assistive technologies consistently ignore empty headings now.

However there is a requirement in ARIA that says headings, among various other things require an accessible name. As part of #2119 we'd like to add such a rule. This would be a more generic rule covering all cases where accessible names are required. Instead of deprecating this rule and adding it back in later, at TPAC we talked about changing the Heading not empty rule to map exclusively to ARIA, and deprecating it once the broader rule is completed.

---

## How to Review And Approve


carlosapaduarte (approved):

daniel-montalvo (approved):

Jym77 (approved):

giacomo-petri (approved):

dd8 (approved):

colabottles (approved):

Jym77:

PR #2121, opened 2023-09-15
Clarified and included CJK requirements in large text definition and updated rules affected by this change

### LINKED ISSUE #2068: Contrast rules should account for CJK text size [afw4f7] [09o5cg]
This was brought up by Makoto in the most recent AGWG survey. A CJK text should not be considered large text until it has 22pt, or 18pt if bold.

This is a problem with the [large scale text](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/proposed/#large-scale-text) definition. It impacts both our contrast minimum and contrast enhanced rule.

giacomo-petri:
Since the WCAG is not unambiguously defining what  "equivalent size for Chinese, Japanese and Korean (CJK) fonts"  means, I've added the definition of large text for CJK languages, based on the [Japanese translation of WCAG 2.1](https://waic.jp/translations/WCAG21/#dfn-large-scale) with the assumption that it applies also to the other CJK languages.

Edits:
- updated the large text definition;
- updated ACT contrast minimum examples to reflect updates;
- updated ACT contrast enhanced examples to reflect updates.

Closes #2068

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


pages/glossary/large-scale-text.md
WilcoFiers, line 1: I rewrote a few things, including updating the links as I found this a little difficult to read with all the URLs in them:

```md
**Exception**: If the text nodes contain Unicode characters in the 4E00–9FFF range:

- at least 22 [points][], or
- at least 18 [points][] and a [computed][] [font weight][] CSS property of 700 or higher.

**Notes:**  
The WCAG definition of [large scale text](https://www.w3.org/TR/WCAG21/#dfn-large-scale) includes an exception for CJK (Chinese, Japanese, and Korean) text. The Unicode range 4E00–9FFF includes these characters. The "equivalent size" numbers are not included in WCAG, and come from the [Official Japanese Translation of WCAG 2.1](https://waic.jp/translations/WCAG21/#dfn-large-scale).  This translation offers the following insight (translated from Japanese to English to allow everyone understanding the specs):  

> TRANSLATION NOTE  
> In the case of Japanese full-width characters, 22 point or 18 point bold type is considered to be the "equivalent" size, based on the First Report of the Enlarged Textbook Promotion Council, "Chapter 2 Standards for Enlarged Textbooks." is reasonable."
Assuming all CJK characters pose similar contrast challanges, this definition extends the Japanese requirement to the whole CJK characters set included in the [Unicode standard](https://unicode.org/charts/PDF/U4E00.pdf) within the range of 4E00–9FFF.

[computed]: https://www.w3.org/TR/css-cascade-3/#computed-value
[font-size]: https://www.w3.org/TR/css-fonts-3/#propdef-font-size
[points]: https://www.w3.org/TR/css-values/#pt
[font weight]: https://www.w3.org/TR/css-fonts-3/#font-weight-prop
```

pages/glossary/large-scale-text.md
```diff
@@ -12,3 +12,18 @@ Text nodes with a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value
...
+>[large scale (text)](https://www.w3.org/TR/WCAG21/#dfn-large-scale): with at least 18 point or 14 point bold or font size that would yield equivalent size for Chinese, Japanese and Korean (CJK) fonts
+
+Since there is no official WCAG definition of **equivalent size** for CJK characters, we rely on the guidelines provided in the [Japanese translation of the WCAG 2.1](https://waic.jp/translations/WCAG21/#dfn-large-scale). This document offers the following insight (translated from Japanese to English to allow everyone understanding the specs):  
+
+>TRANSLATION NOTE  
+>In the case of Japanese full-width characters, 22 point or 18 point bold type is considered to be the "equivalent" size, based on the First Report of the Enlarged Textbook Promotion Council, "Chapter 2 Standards for Enlarged Textbooks." is reasonable."
+
+Assuming all CJK characters pose similar contrast challanges, this definition extends the Japanese requirement to the whole CJK characters set included in the [Unicode standard](https://unicode.org/charts/PDF/U4E00.pdf) within the range of 4E00–9FFF.
```
carlosapaduarte: We might want to indicate that that Unicode range describes the CJK Unified Ideographs

pages/glossary/large-scale-text.md
```diff
@@ -12,3 +12,18 @@ Text nodes with a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value
 
 - at least 18 [points](https://www.w3.org/TR/css-values/#pt), or
 - at least 14 [points](https://www.w3.org/TR/css-values/#pt) and a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font weight](https://www.w3.org/TR/css-fonts-3/#font-weight-prop) CSS property of 700 or higher.
+
+For CJK (Chinese, Japanese and Korean) characters with a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font-size](https://www.w3.org/TR/css-fonts-3/#propdef-font-size) of:
```
carlosapaduarte:
```suggestion
For CJK (Chinese, Japanese and Korean) characters within text nodes, a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font-size](https://www.w3.org/TR/css-fonts-3/#propdef-font-size) of: 
```

carlosapaduarte (commented):
Some editorial suggestions for the definition

pages/glossary/large-scale-text.md
giacomo-petri: Added at the beginning of the Exception for a better understanding

_rules/text-contrast-afw4f7.md
```diff
@@ -189,6 +189,36 @@ This text is using the default user agent text color and background color. By de
...
+
+```html
+<p style="color: #000; font-size:22pt; font-weight:700; background: #666;" lang="ja">
+	中国語、日本語、及び韓国語
+</p>
+```
+
+#### Passed Example 14
+
+This 18pt black Japanese text has a contrast ratio of 3.6:1 on the gray background. Since the characters composing the sentence are not included within the Unicode standard range of 4E00–9FFF, it's passing the rule.
```
Jym77:
```suggestion
This 18pt black Japanese text has a contrast ratio of 3.6:1 on the gray background. Since the characters composing the sentence are not included within the Unicode standard range of 4E00–9FFF, this is [large scale text][] and passes the rule.
```

_rules/text-contrast-enhanced-09o5cg.md
```diff
@@ -183,6 +183,36 @@ This text is using the default user agent text color and background color. By de
...
+
+```html
+<p style="color: #000; font-size:22pt; background: #777;" lang="ja">
+	中国語、日本語、及び韓国語
+</p>
+```
+
+#### Passed Example 13
+
+This 18pt black Japanese text has a contrast ratio of 4.6:1 on the gray background. Since the characters composing the sentence are not included within the Unicode standard range of 4E00–9FFF, it's passing the rule.
```
Jym77:
```suggestion
This 18pt black Japanese text has a contrast ratio of 4.6:1 on the gray background. Since the characters composing the sentence are not included within the Unicode standard range of 4E00–9FFF, this is [large scale text][] and passes the rule.
```

pages/glossary/large-scale-text.md
```diff
@@ -8,7 +8,27 @@ input_aspects:
...
+
+
+**Exception**:  
+If the text nodes contain Unicode characters in the 4E00–9FFF range ([CJK Unified Ideographs](https://unicode.org/charts/PDF/U4E00.pdf)), these characters are:
+
+- at least 22 [points][], or
+- at least 18 [points][] and a [computed][] [font weight][] CSS property of 700 or higher.
+
+**Notes:**  
+The WCAG definition of [large scale text](https://www.w3.org/TR/WCAG21/#dfn-large-scale) includes an exception for CJK (Chinese, Japanese, and Korean) fonts. The Unicode range 4E00–9FFF includes these characters. The "equivalent size" numbers are not included in WCAG, and come from the [Official Japanese Translation of WCAG 2.1](https://waic.jp/translations/WCAG21/#dfn-large-scale).  This translation offers the following insight (translated from Japanese to English to allow everyone understanding the specs):
```
Jym77: Isn't that an *un*official translation?

pages/glossary/large-scale-text.md
```diff
@@ -8,7 +8,27 @@ input_aspects:
...
 - at least 18 [points](https://www.w3.org/TR/css-values/#pt), or
-- at least 14 [points](https://www.w3.org/TR/css-values/#pt) and a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font weight](https://www.w3.org/TR/css-fonts-3/#font-weight-prop) CSS property of 700 or higher.
+- at least 14 [points](https://www.w3.org/TR/css-values/#pt) and a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font weight][] CSS property of 700 or higher.
+
+
+**Exception**:  
+If the text nodes contain Unicode characters in the 4E00–9FFF range ([CJK Unified Ideographs](https://unicode.org/charts/PDF/U4E00.pdf)), these characters are:
+
+- at least 22 [points][], or
+- at least 18 [points][] and a [computed][] [font weight][] CSS property of 700 or higher.
```
Jym77: **Suggestion:** I think we should rephrase the full definition (as it doesn't look like correct English now), in the line of:

> A text node is _large scale text_ if its [computed][] [font-size][] is:
> - at least 18 [points[]; or
> - at least 14 [points][] and its [computed][] [font+weight][] is at least 700.
>
> Exception:
> A text node containing CJK characters is _large scale text_ if …

Maybe even start the definition with "depending whether the text node contains CJK characters, …" 🤔 

Also, if the defined term instances (`_large scale text_`) are wrapped in a `dfn` element, we can reference them directly from the running text:
> `<dfn id="large-scale-text:cjk">large scale text</dfn>`
> `This is [large scale text][#large-scale-text:cjk]`

See, e.g. [attribute value in image button name rule](https://github.com/act-rules/act-rules.github.io/blob/develop/_rules/image-button-non-empty-accessible-name-59796f.md?plain=1#L182)

Additionally (but not directly related to this PR, so may be taken separately):
* These lists do not match how we write condition lists now; should be "if at least one of the following is true: the [computed][] [font-size][] is at least 18 points; or…"
* Technically, text nodes do not have CSS properties, only their parent element has, so the language used here is a bit incorrect (although not really confusing…)
* Also, not need to say "[font weight][] **CSS propert**" when the name already links to CSS specs.

Jym77 (changes requested):
Well done.

I think we should vary the languages in the new examples and have some in Korean or Chinese (OK, Chinese will be the same character set…)

pages/glossary/large-scale-text.md
giacomo-petri: Missed it while accepting Wilco suggestion. Good catch.
Updated

pages/glossary/large-scale-text.md
giacomo-petri: Suggestions applied rewriting the definition and starting with "A text node..."

---

> Additionally (but not directly related to this PR, so may be taken separately):
> 
> These lists do not match how we write condition lists now; should be "if at least one of the following is true: the [computed][] [font-size][] is at least 18 points; or…"

Happy to edit condition lists and include them in the current edits.

> Technically, text nodes do not have CSS properties, only their parent element has, so the language used here is a bit incorrect (although not really confusing…)

You're absolutely right. This was the case in the previous iteration as well. I'm struggling to find a straightforward method to enhance its clarity without adding more complexity to the definition.

> Also, not need to say "[font weight][] CSS propert" when the name already links to CSS specs.

updated.

_rules/text-contrast-afw4f7.md
```diff
@@ -189,6 +189,36 @@ This text is using the default user agent text color and background color. By de
 <div role="button">My button!</div>
 ```
 
+#### Passed Example 12
```
Jym77: I think some of the new examples should be in Korean or Chinese instead of Japanese. Not blocking, though.

Jym77 (approved):

_rules/text-contrast-afw4f7.md
giacomo-petri: @Jym77, 
I've made some revisions to the examples by substituting Japanese with Chinese, utilizing the Chinese rendition of the WCAG 2.1 title.

Regarding Korean, it appears that only "Hanja" (Chinese characters) are affected by this rule. However, given my unqualified Korean language knowledge, I'm unable to craft a meaningful sentence using these Hanja characters. Any assistance from the community in providing a valid Korean sentence would be greatly appreciated.

dd8: Korean Hangul might be another special case - lines are laid out in 2 dimensions unlike Chinese and Japanese (or European languages). It's unclear what point size and line height mean for Hangul layout.

> For example, the Korean word for "honeybee" (kkulbeol) is written as 꿀벌, not ㄲㅜㄹㅂㅓㄹ.
https://en.wikipedia.org/wiki/Hangul

The 꿀벌 above are are actually 2 groups of 3 characters stacked vertically (one group for each syllable).

dd8: It looks like Korean Hanja is very rare - its use and teaching was banned in 1968 - so we can probably ignore it. The ban has been repealed recently, but as a result most Korean people don't know Hanja:
https://en.wikipedia.org/wiki/Hanja#Decline_of_Hanja

The Hangul syllable blocks mentioned in the [comment above](https://github.com/act-rules/act-rules.github.io/pull/2121#discussion_r1346372231) are precomposed into 11,000+ Unicode code points so point size is the size of the precomposed glyph.
https://www.w3.org/TR/klreq/#hangulcodes

This is similar to the way accents like circumflex and vowels are precomposed Unicode code points in French, as in the letter ô in hôpital 

The Hangul Jamo code points can be combined into single glyphs representing a syllable. This is similar to how the letter ['o' U+006F](https://www.compart.com/en/unicode/U+006f) and combining circumflex ['◌̂' U+0302](https://codepoints.net/U+0302?lang=en) can be composed into the glyph ô

WilcoFiers (dismissed):

__tests__/spelling-ignore.yml
```diff
@@ -98,6 +98,7 @@
 - de-hello
 - 14pt
 - 18pt
```
WilcoFiers:
```suggestion
- 16pt
- 18pt
- 20pt
```

__tests__/spelling-ignore.yml
```diff
@@ -240,6 +244,9 @@
 - 000B
 - 000C
 - 000D
+- 4E00
```
WilcoFiers, line 248: For some reason this isn't working. Need to dig into this a bit.

dan-tripp-siteimprove: I found two different ways to fix this: 

1) In spelling-ignore.yml, replace this:
`- 4E00`
with this:
`- "4E00"`

or: 

2) In spelling.js, replace this: 
`const ignoreConfigured = yaml.safeLoad(fs.readFileSync('./__tests__/spelling-ignore.yml', 'utf8'))`
with this:
`const ignoreConfigured = yaml.safeLoad(fs.readFileSync('./__tests__/spelling-ignore.yml', 'utf8'), 
		{schema: yaml.FAILSAFE_SCHEMA})`

colabottles (approved):

__tests__/spelling-ignore.yml
Jym77: I'd much prefer the second which is more generic and will cover future cases, without the risk of somebody wondering why 4E00 is a special kid that needs double quotes. Can be improved with inlined comment in the code to explain that we need this option due to entries starting with a non-zero digit (?)

Jym77 (dismissed):

__tests__/spelling-ignore.yml
Jym77: @giacomo-petri It seems if you do the change suggested by Dan, this should pass the test and can go in Call for Review 🎉

giacomo-petri: spelling.js updated with the 2nd option and inline comment added to explain why

colabottles (dismissed):

giacomo-petri:
Call for review ends on January 25th.

carlosapaduarte (dismissed):

pages/glossary/large-scale-text.md
```diff
@@ -8,7 +8,27 @@ input_aspects:
...
+
+- the text node [computed][] [font-size][] is at least 22 [points][], or
+- the text node [computed][] [font-size][] is at least 18 [points][] and has a [computed][] [font weight][] of 700 or higher.
+
+**Notes:**  
+The WCAG definition of [large scale text](https://www.w3.org/TR/WCAG21/#dfn-large-scale) includes an exception for CJK (Chinese, Japanese, and Korean) fonts. The Unicode range 4E00–9FFF includes these characters. The "equivalent size" numbers are not included in WCAG, and come from the [Unofficial Japanese Translation of WCAG 2.1](https://waic.jp/translations/WCAG21/#dfn-large-scale).  This translation offers the following insight (translated from Japanese to English to allow everyone understanding the specs):  
+
+> TRANSLATION NOTE  
+> In the case of Japanese full-width characters, 22 point or 18 point bold type is considered to be the "equivalent" size, based on the First Report of the Enlarged Textbook Promotion Council, "Chapter 2 Standards for Enlarged Textbooks." is reasonable.
```
patrickhlauke: that final

`is reasonable.`

looks out of place. a remnant from a previous version of that sentence?

patrickhlauke:
```suggestion
> In the case of Japanese full-width characters, 22 point or 18 point bold type is considered to be the "equivalent" size, based on the First Report of the Enlarged Textbook Promotion Council, "Chapter 2 Standards for Enlarged Textbooks."
```

bobbytung:
Hello, 

I'm Bobby Tung, editor of [CLREQ](https://www.w3.org/TR/clreq/).

Some opinion here: 

1. Japanese 22pt equivalent issue.

The Japanese font-size 22pt rules is from [第2章　拡大教科書の標準的な規格について](https://www.mext.go.jp/a_menu/shotou/kyoukasho/1282361.htm), that is a formatting rule document for textbook designed for low vision student.

Usual Japanese textbook's base body text size is from 10.5pt, up to 18pt. In this document, its base body text size is from 22pt. However, it also asked for 0.8x (base body text is 18pt) and 1.2x (base body text is 30pt) layout as option for student to choose.

Since 0.8x version textbook's base body text size is 18pt, Should we keep the baseline on 18pt?

We do not apply Web a11y rule to print book, but if we do so, 0.8x version should be recheck for SC 1.4.3 and 1.4.6. 

2. Traditional Chinese textbook rule as reference.

In Taiwan, we have a rule for textbool's base text size: 14pt is minimal size for body text and other text must not be smaller than 11pt. Practically, large size textbook for low vision student usually directly scale up to 1.25% to 1.5%. That means 18pt~22pt. So I'd like to keep baseline on 18pt.

3. Character complexity for CJK.

Additional information for consideration. 

If we need to set minimal font size rule for CJK, I think glyph complexity and what type used inline (i.e. Katakana and Hiragana in Japanese) is important. The base font size:

Traditional Chinese ≥ Japanese ≥ Simplified Chinese

JediLin (changes requested):
Taiwan (Traditional Chinese, lang = zh-hant-tw or zh-tw) has different requirements than Japan (Japanese, ja). In the case of Traditional Chinese characters, for people with disabilities, 20 point or 16 point bold type is considered to be the "equivalent" size, based on the 臺灣易讀參考指南 (Taiwanese Readability Reference Guidelines), published by Social and Family Affairs Administration, Ministry of Health and Welfare, Taiwan.

Since the same characters may be used in different language, it is not sufficient simply using Unicode characters range to make exception rules. The value of lang attribution should be considered, too.

I propose 3 changes:
- Add 2 exception rules and additional notes text: https://github.com/act-rules/act-rules.github.io/pull/2121#discussion_r1456891552
- Add a passed example: https://github.com/act-rules/act-rules.github.io/pull/2121#discussion_r1456907290
- Add a passed example: https://github.com/act-rules/act-rules.github.io/pull/2121#discussion_r1456908149

pages/glossary/large-scale-text.md
JediLin: Taiwan has different requirements than Japan, I propose the following:

```
**Exception**:  
A text node containing Unicode characters in the 4E00–9FFF range ([CJK Unified Ideographs](https://unicode.org/charts/PDF/U4E00.pdf)) is <dfn id="large-scale-text:cjk">large scale text</dfn> if at least one of the following is true:

- the Japanese text node [computed][] [font-size][] is at least 22 [points][], or
- the Traditional Chinese text node [computed][] [font-size][] is at least 20 [points][], or
- the Japanese text node [computed][] [font-size][] is at least 18 [points][] and has a [computed][] [font weight][] of 700 or higher, or
- the Traditional Chinese text node [computed][] [font-size][] is at least 16 [points][] and has a [computed][] [font weight][] of 700 or higher.

**Notes:**  
The WCAG definition of [large scale text](https://www.w3.org/TR/WCAG21/#dfn-large-scale) includes an exception for CJK (Chinese, Japanese, and Korean) fonts. The Unicode range 4E00–9FFF includes these characters. The "equivalent size" numbers are not included in WCAG, and come from the [Taiwanese Readability Reference Guidelines](https://crpd.sfaa.gov.tw/BulletinCtrl?func=downloadFile&type=file&id=3749&code=F9FEF1F8F1FAF6FAFCFAFCFCFFFAF2F3) (PDF download) and the [Unofficial Japanese Translation of WCAG 2.1](https://waic.jp/translations/WCAG21/#dfn-large-scale). The Taiwanese Readability Reference Guidelines requires minimal font size 16 points for Traditional Chinese characters and 14 points for English characters to be fulfill the need of the people with disabilities, and the Unofficial Japanese translation offers the following insight (translated from Japanese to English to allow everyone understanding the specs):
```

_rules/text-contrast-afw4f7.md
```diff
@@ -183,6 +183,36 @@ This text is using the default user agent text color and background color. By de
...
+#### Passed Example 14
+
+This 18pt black Japanese text has a contrast ratio of 3.6:1 on the gray background. Since the characters composing the sentence are not included within the Unicode standard range of 4E00–9FFF, this is [large scale text][] and passes the rule.
+
+```html
+<p style="color: #000; font-size:18pt; background: #666;" lang="ja">
+	「ロック・アンド・ロール」
+</p>
+```
+
```
JediLin: #### Passed Example 15

This 16pt bold black Traditional Chinese text has a contrast ratio of 3.6:1 on the gray background.

```html
<p style="color: #000; font-size:16pt; font-weight:700; background: #666;" lang="zh-TW">
	內容可及性規範
</p>
```

_rules/text-contrast-enhanced-09o5cg.md
```diff
@@ -183,6 +183,36 @@ This text is using the default user agent text color and background color. By de
...
+#### Passed Example 13
+
+This 18pt black Japanese text has a contrast ratio of 4.6:1 on the gray background. Since the characters composing the sentence are not included within the Unicode standard range of 4E00–9FFF, this is [large scale text][] and passes the rule.
+
+```html
+<p style="color: #000; font-size:18pt; background: #777;" lang="ja">
+	「ロック・アンド・ロール」
+</p>
+```
+
```
JediLin: #### Passed Example 14

This 20pt black Traditional Chinese text has a contrast ratio of 4.6:1 on the gray background.

```html
<p style="color: #000; font-size:20pt; background: #777;" lang="zh-TW">
	內容可及性規範
</p>
```

giacomo-petri:
@Jym77 @carlosapaduarte,

What are your thoughts on adding this into the agenda items, if @bobbytung and @JediLin are able to join?

Jym77:
> @Jym77 @carlosapaduarte,
> 
> What are your thoughts on adding this into the agenda items, if @bobbytung and @JediLin are able to join?

Yes, definitely.

Jym77:
> Taiwan (Traditional Chinese, lang = zh-hant-tw or zh-tw) has different requirements than Japan (Japanese, ja). In the case of Traditional Chinese characters, for people with disabilities, 20 point or 16 point bold type is considered to be the "equivalent" size, based on the 臺灣易讀參考指南 (Taiwanese Readability Reference Guidelines), published by Social and Family Affairs Administration, Ministry of Health and Welfare, Taiwan.
> 
> Since the same characters may be used in different language, it is not sufficient simply using Unicode characters range to make exception rules. The value of lang attribution should be considered, too.

That's getting a bit tricky.
* We can of course link the requirement to the `lang` of the text node (in addition to the unicode range). This would also require an assumption that the `lang` attribute is correctly set (how valid is this assumption?)
* We can also go for a "maximum requirement" rule, with a background note stating that "different languages have different requirement even if they use the same script, thus less than X points is always 'small size', more than Y points is always 'big size', and between X and Y is unclear". Then, the rule consider "more than X points" as being 'big size' under the "no false positives" reasoning: anything less than X points with 3:1 contrast is sure to be bad, stuff more than X points with 3:1 contrast may be good.

JediLin:
> * We can also go for a "maximum requirement" rule, with a background note stating that "different languages have different requirement even if they use the same script, thus less than X points is always 'small size', more than Y points is always 'big size', and between X and Y is unclear". Then, the rule consider "more than X points" as being 'big size' under the "no false positives" reasoning: anything less than X points with 3:1 contrast is sure to be bad, stuff more than X points with 3:1 contrast may be good.

I like the "maximum requirement" rule option with "no false positives" reasoning :-)

carlosapaduarte:
> > @Jym77 @carlosapaduarte,
> > What are your thoughts on adding this into the agenda items, if @bobbytung and @JediLin are able to join?
> 
> Yes, definitely.

@bobbytung and @JediLin, would you be able to join one of our meetings? They take place from 10:00 EDT to 11:00 EDT every second and fourth Thursdays. The next one is on January 25.

bobbytung:
I'm not sure but try my best to attend.

However, I'd like to raise another issue.

As screenshot below, some system serif fonts for TC only have one font weight. (i.e. PMingLiu for MS Windows, Noto Han Serif for Android) That means browser may render as faux bold, it affects legibility a lot. 

截圖 2024-01-19 清晨6 11 04

JediLin:
> They take place from 10:00 EDT to 11:00 EDT every second and fourth Thursdays. The next one is on January 25.

I'll do my best to keep awake. (Local time 23:00 to 00:00)

Jym77:
@giacomo-petri CG decision was to got with this solution:

> * We can also go for a "maximum requirement" rule, with a background note stating that "different languages have different requirement even if they use the same script, thus less than X points is always 'small size', more than Y points is always 'big size', and between X and Y is unclear". Then, the rule consider "more than X points" as being 'big size' under the "no false positives" reasoning: anything less than X points with 3:1 contrast is sure to be bad, stuff more than X points with 3:1 contrast may be good.

the rule is pretty much already doing the same sort of approximation by only grabbing the highest possible contrast.

giacomo-petri:
Thanks @Jym77,

following the meeting's outcomes and the agreed-upon solution, we have designated the 18pt-22pt range as large text for CJK (and anything exceeding 22pt is considered large text as well). This aligns with the observed pattern in Latin languages. 

While we acknowledge some disparities in CJK requirements, we have universally recognized text => 18pt (and 14pt bold) as large text, below as standard text, without any exceptions.

In the next couple of days, I will revert the large text definition to the original one and include background notes for contrast minimum and enhanced rules.

giacomo-petri:
Files updated:
- Added a note within the "Large text" definition for CJK languages and how we decided to treat them.
- Eliminated passing CJK examples within the CJK range, since it was no longer pertinent based on the updated definition of large text.
- Similarly, removed irrelevant failing examples as there's no longer a distinction between CJK and Latin characters.

dan-tripp-siteimprove (approved):

colabottles (approved):

__tests__/spelling-ignore.yml
```diff
@@ -97,7 +97,10 @@
 - merchantability
 - de-hello
 - 14pt
+- 16pt
 - 18pt
+- 20pt
+- 22pt
```
Jym77: **Question:** Do we still need the 16, 20, and 22pt spelling exceptions?

pages/glossary/large-scale-text.md
```diff
@@ -8,7 +8,23 @@ input_aspects:
...
+
+
+**Notes:**  
+In the domain of languages such as Chinese, Japanese, and Korean (CJK languages), encompassing the Unicode character range from 4E00 to 9FFF ([CJK Unified Ideographs](https://unicode.org/charts/PDF/U4E00.pdf)), it becomes crucial to acknowledge the intricate nuances in their typographic requirements. Despite sharing a common Unicode spectrum, each CJK language and its corresponding country may exhibit distinct typographic preferences and standards.
+
+Particularly noteworthy is the lack of uniformity in defining text sizes as either large or small within the following ranges:
+- from 18 [points][] to 22 [points][]
+- from 14 [points][] to 18 [points][] with a [computed][] [font weight][] of 700 or higher.
+
+Given the diversity in these specifications, the ACT rules adhere to the large text definition outlined above. This definition is applicable across all languages under the rationale of "no false positives."
```
Jym77:
```suggestion
Whether CJK characters in these range should be considered as "large scale" depends on the language of the text, as well as the country for which the text is designed and whose requirements must be followed. Given that these pieces of information are not readily available, the ACT rules adhere to the large text definition outlined above. This definition is applicable across all languages under the rationale of "no false positives."
```

pages/glossary/large-scale-text.md
```diff
@@ -8,7 +8,23 @@ input_aspects:
...
-Text nodes with a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font-size](https://www.w3.org/TR/css-fonts-3/#propdef-font-size) of:
+A text node is large scale text if at least one of the following is true:
 
-- at least 18 [points](https://www.w3.org/TR/css-values/#pt), or
-- at least 14 [points](https://www.w3.org/TR/css-values/#pt) and a [computed](https://www.w3.org/TR/css-cascade-3/#computed-value) [font weight](https://www.w3.org/TR/css-fonts-3/#font-weight-prop) CSS property of 700 or higher.
+- the text node [computed][] [font-size][] is at least 18 [points][], or
+- the text node [computed][] [font-size][] is at least 14 [points][] and has a [computed][] [font weight][] of 700 or higher.
+
+
+**Notes:**
```
Jym77:
```suggestion
**Background:**  
```

Jym77 (dismissed):
* I believe we also need to have the note about CJK (or part of it) directly in the rules, in the lines of:
  > For CJK characters used in some languages and countries, a font size of 18-22pt should not be consider "large text". These may fail the SC while still passing the rule.
* We can keep some CJK examples, to show we've considered them.

pages/glossary/large-scale-text.md
```diff
@@ -14,14 +14,14 @@ A text node is large scale text if at least one of the following is true:
 - the text node [computed][] [font-size][] is at least 14 [points][] and has a [computed][] [font weight][] of 700 or higher.
 
 
-**Notes:**  
+**Background:**
```
Jym77:
```suggestion
#### Background
```
Let's make it into a section.

Jym77 (approved):

colabottles (approved):

carlosapaduarte (approved):

JediLin (approved):

giacomo-petri:
Call for review ends on ~April 16th~ April 23rd (2 weeks).

Jym77:

PR #2129, opened 2023-10-17
Fix missing definition aria-state-or-property-valid-value-6a7281

WilcoFiers:

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-state-or-property-valid-value-6a7281.md
+++ b/_rules/aria-state-or-property-valid-value-6a7281.md
@@ -53,6 +53,7 @@ This rule does not require the target of an `ID Reference` to exist. This is bec
 ### Related rules
 
 - [ARIA state or property is permitted](https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea/)
+- [ARIA required ID references exist](https://www.w3.org/WAI/standards-guidelines/act/rules/in6db8/)
 
 ### Bibliography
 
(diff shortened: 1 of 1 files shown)


kengdoj (approved):

tbostic32 (approved):

HelenBurge (approved):

_rules/aria-state-or-property-valid-value-6a7281.md
```diff
@@ -244,3 +245,4 @@ The `aria-hidden` attribute is not on an [HTML or SVG element][].
 [html or svg element]: #namespaced-element
 [value type]: https://www.w3.org/TR/wai-aria-1.2/#propcharacteristic_value
 [wai-aria state or property]: https://www.w3.org/TR/wai-aria-1.2/#state_prop_def
+[attribute value]: #attribute-value
```
Jym77, line 248: I'm curious how this passed tests if the `[attribute value]` ref was missing 🤔 Are we having flaky tests?

Jym77 (approved):

_rules/aria-state-or-property-valid-value-6a7281.md
WilcoFiers: The tests check that an ID exists. If there's no ID then it's ignored. We could definitely do with an extra test to prevent this if you're up for it.

PR #2141, opened 2023-11-23
[chore] Add test to detect duplicate rule id

### LINKED ISSUE #2135: [tooling]: Add test to check that each rule has a unique `id`
To avoid the mess of #2134 happening again, we should have test to detect duplicate rule `id` (or glossary item key).

Jym77:
Add a test to check that rule ids are unique, avoiding the mess of #2134.

Closes issue(s):

- closes #2135

Need for Call for Review:
This will not require a Call for Review (add test code)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/__tests__/unique-ids.js
+++ b/_rules/__tests__/unique-ids.js
@@ -1,0 +1,24 @@
+const getMarkdownData = require('../../utils/get-markdown-data')
+const rulesData = getMarkdownData(`./_rules`)
+
+describe('Rule ids verification', () => {
+	const duplicates = rulesData.filter(
+		ruleData => duplicate(rulesData, ruleData, (rule1, rule2) => rule1.frontmatter.id === rule2.frontmatter.id) !== 1
+	)
+
+	test('Rule ids are unique', () => {
+		expect(duplicates.length, `Duplicated rules: ${duplicates.map(ruleData => ruleData.filename).join(', ')}`).toBe(0)
+	})
+})
+
+function duplicate(array, element, test) {
+	let found = 0
+
+	for (let i = 0; i < array.length; i++) {
+		if (test(element, array[i])) {
+			found++
+		}
+	}
+
+	return found
+}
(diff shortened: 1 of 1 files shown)


dan-tripp-siteimprove (approved):
I found some cases where the YAML-style parsing of the frontmatter IDs can lead to unexpected results.  These cases might never happen in the real world.  I'll mention them anyway.

Case # 1: Octal: 077777 and 32767 would be flagged as duplicate IDs by unique-ids.js.  They probably shouldn't be.
Case # 2: Quotes: 123456 and "123456" would not be flagged as duplicate IDs by unique-ids.js.  They probably should be.

This is similar to the problem I found at https://github.com/act-rules/act-rules.github.io/pull/2121#discussion_r1374923716 .  But in this case I don't know how to change the parsing to make the rule frontmatter IDs get parsed as strings rather than numbers.  So if you want to prevent these cases (which might never happen in the real world), maybe the right place to do it is in the frontmatter.js test: assert that (typeof frontmatter.id) == 'string'.

_rules/__tests__/unique-ids.js
```diff
@@ -0,0 +1,24 @@
+const getMarkdownData = require('../../utils/get-markdown-data')
+const rulesData = getMarkdownData(`./_rules`)
+
+describe('Rule ids verification', () => {
+	const duplicates = rulesData.filter(
+		ruleData => duplicate(rulesData, ruleData, (rule1, rule2) => rule1.frontmatter.id === rule2.frontmatter.id) !== 1
```
tbostic32: This might be over simplification, but you could do
```js
ruleData => rulesData.findIndex(r => r.frontmatter.id == ruleData.frontmatter.id) >= 0
```
and then not need the duplicate function.

tbostic32 (approved):

carlosapaduarte (approved):

PR #2154, opened 2024-01-11
Add definition for implicit/explicit ARIA attributes

### LINKED ISSUE #2138: Required states can come from native elements [4e8ab6]
This came from a comment from @JAWS-test on a [review](https://github.com/w3c/aria/discussions/1850#discussioncomment-4309723)

The following shouldn't fail the [Element with role attribute has required states and properties](https://www.w3.org/WAI/standards-guidelines/act/rules/4e8ab6/proposed/) rule. The checked state is present even without the aria-checked attribute, because it is on a native checkbox.

```html
<input type="checkbox" role="menuitemcheckbox">
```

Jym77:
* Add a definition for "ARIA attribute is set", with variations whether it comes from an explicit `aria-*` attribute, an implicit HTML AAM, or a default value.
* Switch "role has all required ARIA attributes" to also apply to implicit roles. These normally have implicitly set attributes (which is why we skipped them initially), and can now be handled by this definition.
* Switch "role has all required ARIA attributes" to accept attributes set in any way.
* Switch "ARIA attribute permitted" to only look at explicitly set ones. It was more or less already doing that by looking at `aria-*` HTML attributes, this makes it a bit clearer.

Closes issue(s):

- closes #2138 

Need for Call for Review:
This will require a 2 weeks Call for Review (significant changes to Applicability)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -173,6 +175,14 @@ This `input` element does not have an [explicit role][] of `textbox`, but the `a
 <label>Password<input type="password" aria-required="true"/></label>
 ```
 
+#### Passed Example 12
```
giacomo-petri, line 178: @Jym77, correct me if I'm wrong, but per applicability this example should be inapplicable, as the aria-checked has not been explicitly set.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -205,23 +229,18 @@ This `div` does not have a [semantic role](#semantic-role).
```
giacomo-petri, line 226: I'd rephrase the description of "Inapplicable Example 1" to clarify that it is included here because, owing to its generic role, it does not feature in the accessibility tree.

Conversely, despite lacking a corresponding implicit role in ARIA, an `iframe` element, presumed to lack an implicit semantic role, is included in the accessibility tree, and therefore this rule is applicable.

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,30 @@
...
+key: aria-attribute-set
+unambiguous: true
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An ARIA [state][aria state] or [property][aria property] is <dfn>set</dfn> on an [HTML element][namespaced element] when it has a value. This may happen in three ways:
```
giacomo-petri, line 12: The term "value" brings to mind the explicit value of an attribute. What if we replace it with "when the accessibility tree exposes its value."

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,30 @@
...
+
+- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element, whose value is valid for this ARIA [state][aria state] or [property][aria property].
+
+  For example, `aria-label` is explicitly set on `<button aria-label="Next page">Next</button>`.
+
+- It is <dfn id="aria-attribute-set:implicit">implicitly set</dfn> if there is no corresponding `aria-*` HTML attribute on the element, but the element has an [implicit semantic role][implicit role] that has a value for this ARIA [state][aria state] or [property][aria property] defined in [HTML Attribute State and Property Mappings][aria attribute mapping].
+
+  For example, `aria-checked` is implicitly set on `<input type="checkbox" checked />`.
+
+- It is <dfn id="aria-attribute-set:default">set by default</dfn> if there is no corresponding `aria-*` HTML attribute on the element, and the element has an [implicit semantic role][implicit role] that has a default value for this ARIA [state][aria state] or [property][aria property].
```
giacomo-petri:
```suggestion
- It is <dfn id="aria-attribute-set:default">set by default</dfn> if there is no corresponding `aria-*` HTML attribute on the element, and the element, along with its attributes, has an [implicit semantic role][implicit role] that has a default value for this ARIA [state][aria state] or [property][aria property].
```

just to clarify that some attribute might affect the "default" value for the element.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -126,6 +128,28 @@ This `combobox` has the required properties `aria-controls` and `aria-expanded`.
 </ul>
 ```
 
+#### Passed Example 7
```
carlosapaduarte, line 133: I would update the description of, at least, all the passed examples to also include references to the necessary aria set definition.

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,30 @@
...
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An ARIA [state][aria state] or [property][aria property] is <dfn>set</dfn> on an [HTML element][namespaced element] when it has a value. This may happen in three ways:
+
+- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element, whose value is valid for this ARIA [state][aria state] or [property][aria property].
```
carlosapaduarte: "whose value is valid for this ARIA state or property"
Shouldn't this be checked by [ARIA state or property has valid value](https://www.w3.org/WAI/standards-guidelines/act/rules/6a7281/proposed/)?

pages/glossary/aria-attribute-set.md
carlosapaduarte: It is really hard to differentiate between implicitly set and set by default according to these definitions (the later is a subset of the former). Do we really need two definitions? Can we just have explicitly and implicitly set in the "Element with role attribute has required states and properties" rule?

carlosapaduarte (dismissed):
In addition to Giacomo's comments, I've added some more.

pages/glossary/aria-attribute-set.md
Jym77: This proved tricky, especially in combination of https://github.com/act-rules/act-rules.github.io/pull/2154#discussion_r1466221360 😅 

TLDR: DOM/HTML define the "value" of an attribute as the stuff written in the HTML code, which differs from our "attribute value", making everything hard to write. Here, we want to refer to the "value as written" since, presumably, the incorrect values could be dropped by parsing/validation and thus not make it to the "attribute value". Plus, as @carlosapaduarte stated, we do want rules to validate the "value as written" and thus not really bake it into this definition

Anyway, I tried to improve the situation…

pages/glossary/aria-attribute-set.md
Jym77: See https://github.com/act-rules/act-rules.github.io/pull/2154#discussion_r1474206470

pages/glossary/aria-attribute-set.md
Jym77: Indeed, definition was a bit weak. I do believe these are still different and it might be valuable to be able to speak about one or the other (maybe not immediately, though…)
I improved the definition to make the difference more obvious. It is still a bit of a weak difference (depends which specs is setting it, essentially…) 🤷

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,40 @@
...
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An ARIA [state][aria state] or [property][aria property] is <dfn>set</dfn> on an [HTML element][namespaced element] when it has a value. This may happen in three ways:
+
+- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element. The [value][html attribute value] as written in the HTML code may or may not be valid for this ARIA [state][aria state] or [property][aria property] and therefore the attribute may or may not have an [attribute value][]
```
tbostic32: The second sentence is phrased a bit confusingly to me. Could we change it to something like:
```suggestion
- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element. If the [value][html attribute value] as written in the HTML code is not valid for this ARIA [state][aria state] or [property][aria property], then the [value][html attribute value] will not have an [attribute value][]. 
```
We could also add a sentence about when the value is valid, but that operates as one would expect (the valid value is the value of the attribute), so I'm not sure the extra verbosity is worth it.

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,40 @@
...
+
+  For example, `aria-label` is explicitly set on `<button aria-label="Next page">Next</button>`; and `aria-checked` is explicitly set on `<input type="checkbox" aria-checked="yes" />`, even though it does not have an [attribute value][].
+
+- It is <dfn id="aria-attribute-set:implicit">implicitly set</dfn> if there is no corresponding `aria-*` HTML attribute on the element, but the element or one of its HTML attribute has an [ARIA attribute mapping][aria attribute mapping] setting this ARIA attribute.
+
+  For example, `aria-checked` is implicitly set both on `<input type="checkbox" checked />` (through the [presence of the HTML attribute `checked`][checked present]) and `<input type="checkbox" />` (through its [absence][checked absent]); however, it is not set on `<input type="text" />` given that the mapping for `checked` doesn't apply when the `input` element is not of type `checkbox` or `radio`.
```
tbostic32: Personal preference, but I would prefer to remove the `; however,` and just split it into two sentences. Its quite long as is.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -217,7 +219,18 @@ This `div` element is not [included in the accessibility tree][], hence its [WAI
 <div role="button" aria-sort="" style="display:none;"></div>
 ```
 
+#### Passed Example 12
```
tbostic32:
```suggestion
#### Inapplicable Example 3
```
I believe this is intended to be an inapplicable since it doesn't have an explicitly set ARIA attribute.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -217,7 +219,18 @@ This `div` element is not [included in the accessibility tree][], hence its [WAI
 <div role="button" aria-sort="" style="display:none;"></div>
 ```
 
+#### Inapplicable Example 3
+
+This `button` does not have any [explicitly set][aria set explicit] ARIA attribute. It has an [implicitly set][aria set implicit] `aria-checked` state, due to the [absence of a `checked` attribute][checked absent] on the `input` element with a `type` [attribute value][] of `checkbox`.
+
+`````html
+````html ```html <input type="checkbox" role="button" />
```
tbostic32: HTML markup is tripled up here

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -64,6 +64,8 @@ The presence of prohibited ARIA attributes is often the result of a developer us
 
 In HTML, there are language features that do not have corresponding implicit WAI-ARIA semantics. As per [ARIA in HTML](https://www.w3.org/TR/html-aria/), those elements can have [global states or properties][global]. Some of those elements can also have [inherited][], [supported][], or [required][] [states][state] or [properties][property] that correspond to a [WAI-ARIA role](https://www.w3.org/TR/wai-aria-1.2/#introroles). For example, the `audio` element has no corresponding ARIA semantics but it can have [inherited][], [supported][], or [required][] [states][state] or [properties][property] of the [`application` role](https://www.w3.org/TR/wai-aria-1.2/#application).
 
+This rule only applies to [explicitly set][aria set explicit] attributes since these have been provided by authors and using the wrong one is therefore an author issue. If an [implicitly set][aria set implicit] attribute is incorrect, this is not necessarily an author error, e.g., it can be the result of re-purposing an existing element.
```
tbostic32: Would this not still be a responsibility of the author to ensure the attributes are correct? It could still lead to violations, does the [ARIA state or property has valid value](https://www.w3.org/WAI/standards-guidelines/act/rules/6a7281/) check for this? If so, I think we should just call that out directly.

tbostic32 (dismissed):

_rules/aria-state-or-property-permitted-5c01ea.md
Jym77: If only it was so simple 😞 

`<input type="checkbox" role="button" aria-pressed="false" />` is [allowed by ARIA](https://www.w3.org/TR/html-aria/#el-input-checkbox) (as far as I can tell), but has an implicit `aria-checked` set by `checked` (automagically set on `<input type="checkbox" />`), which is not allowed on a role of `button`…

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,40 @@
...
+objective: true
+input_aspects:
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An ARIA [state][aria state] or [property][aria property] is <dfn>set</dfn> on an [HTML element][namespaced element] when it has a value. This may happen in three ways:
+
+- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element. If the [value][html attribute value] as written in the HTML code is not valid for this ARIA [state][aria state] or [property][aria property] and then the attribute is set but does not have an [attribute value][].
```
carlosapaduarte:
```suggestion
- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element. If the [value][html attribute value] written in the HTML code is not valid for this ARIA [state][aria state] or [property][aria property] then the attribute is set but does not have an [attribute value][].
```

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,40 @@
...
+
+  For example, `aria-label` is explicitly set on `<button aria-label="Next page">Next</button>`; and `aria-checked` is explicitly set on `<input type="checkbox" aria-checked="yes" />`, even though it does not have an [attribute value][].
+
+- It is <dfn id="aria-attribute-set:implicit">implicitly set</dfn> if there is no corresponding `aria-*` HTML attribute on the element, but the element or one of its HTML attribute has an [ARIA attribute mapping][aria attribute mapping] setting this ARIA attribute.
```
carlosapaduarte:
```suggestion
- It is <dfn id="aria-attribute-set:implicit">implicitly set</dfn> if there is no corresponding `aria-*` HTML attribute on the element, but the element or one of its HTML attributes has an [ARIA attribute mapping][aria attribute mapping] setting this ARIA attribute.
```

carlosapaduarte (approved):
Minor edits suggested

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -218,7 +220,18 @@ This `div` element is not [included in the accessibility tree][], hence its [WAI
 <div role="button" aria-sort="" style="display:none;"></div>
 ```
 
+#### Inapplicable Example 3
+
+This `button` does not have any [explicitly set][aria set explicit] ARIA attribute. It has an [implicitly set][aria set implicit] `aria-checked` state, due to the [absence of a `checked` attribute][checked absent] on the `input` element with a `type` [attribute value][] of `checkbox`.
+
+```html
+<input type="checkbox" role="button" />
```
zlayaAvocado, line 247: isn't this supposed to be used together with `aria-pressed` if `role="button" type="checkbox"`? https://www.w3.org/TR/html-aria/#el-input-checkbox

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -127,11 +129,41 @@ This `combobox` has the required properties `aria-controls` and `aria-expanded`.
...
+```html
+<label>
+	<input type="checkbox" />
+	Check me
+</label>
+```
+
+#### Passed Example 8
+
+This `menuitemcheckbox` has its required `aria-checked` property [implicitly set][aria set implicit].
```
zlayaAvocado, line 146:
```suggestion
This `menuitemcheckbox` has its required `aria-checked` property [implicitly set][aria set implicit] because type="checkbox" is set on the element.
```

If I am correctly understanding, `aria-checked` is only set because of `type="checkbox"`? Should this be clarified and then a failed example created (if we don't have one already)?

This `menuitemcheckbox` does not have its required `aria-checked` property [implicitly set][aria set implicit]. 
```html
<label>
	<input role="menuitemcheckbox" />
	Check me
</label>
```

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -127,11 +129,41 @@ This `combobox` has the required properties `aria-controls` and `aria-expanded`.
...
+```html
+<label>
+	<input type="checkbox" role="menuitemcheckbox" />
+	Check me
+</label>
+```
+
+#### Passed Example 9
+
+This `iframe` element is [included in the accessibility tree][], even without a role, and therefore has no [WAI-ARIA required states and properties][].
```
zlayaAvocado, line 157: trying to figure out this example, is it to clarify that `role` is not a requirement for this rule as long as other required properties are set (if the element is in the accessibility tree)?

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -198,31 +230,26 @@ This `combobox` uses `aria-owns` instead of using the required `aria-controls` p
...
-This `div` does not have a [semantic role](#semantic-role).
+This `div` is not [included in the accessibility tree][] due to its role of `generic`.
 
 ```html
 <div>Some Content</div>
 ```
 
 #### Inapplicable Example 2
 
-This `checkbox` has an [implicit semantic role](#implicit-role) that is identical to the [explicit semantic role](#explicit-role). This allows native HTML `checked` attribute to apply.
```
zlayaAvocado, line 197: can you explain why this was removed?

pages/glossary/aria-attribute-set.md
```diff
@@ -0,0 +1,40 @@
...
+  - Accessibility tree
+  - CSS styling
+  - DOM tree
+---
+
+An ARIA [state][aria state] or [property][aria property] is <dfn>set</dfn> on an [HTML element][namespaced element] when it has a value. This may happen in three ways:
+
+- It is <dfn id="aria-attribute-set:explicit">explicitly set</dfn> if there is a corresponding `aria-*` HTML attribute on the element. If the [value][html attribute value] written in the HTML code is not valid for this ARIA [state][aria state] or [property][aria property] then the attribute is set but does not have an [attribute value][].
+
+  For example, `aria-label` is explicitly set on `<button aria-label="Next page">Next</button>`; and `aria-checked` is explicitly set on `<input type="checkbox" aria-checked="yes" />`, even though it does not have an [attribute value][].
```
zlayaAvocado, line 16: just wondering if it would be better to use a different example for `<input type="checkbox" aria-checked="yes" />` since these attributes generally shouldn't be used together and the value of `checked` will be `false` on the accessibility tree

zlayaAvocado (changes requested):

_rules/aria-state-or-property-permitted-5c01ea.md
shunguoy: An implicit attribute doesn't show up in the code, rather, it is only added automatically by a user agent during accessibility API mapping.  A checker using the rule would never pick up any implicit attribute. Therefore, it's not necessary to call out the implicit attribute for the purpose of a rule.

_rules/aria-state-or-property-permitted-5c01ea.md
shunguoy: No need to mention the implicit attribute "aria-checked" because it doesn't exist in the code that a checker runs on.

_rules/role-required-states-and-properties-4e8ab6.md
```diff
@@ -32,17 +32,19 @@ acknowledgments:
...
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has an [explicit semantic role][], except if the element has an [implicit semantic role][] that is identical to the [explicit semantic role][].
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][].
 
 ## Expectation
 
-For each test target, the [WAI-ARIA required states and properties][] for the role are set and not empty (`""`), unless the state or property has a default value listed under [WAI-ARIA implicit value for role][].
+For each test target, the [WAI-ARIA required states and properties][] for the role are [set][aria set] and not empty (`""`).
+
+The attributes may be [explicitly set][aria set explicit], [implicitly set][aria set implicit], or [set by default][aria set default].
```
shunguoy, line 43: same with my early comment, an implicit attribute doesn't show up in the code, and thus never picks up by a checker, so it's not necessary to include any implicit attribute.

_rules/role-required-states-and-properties-4e8ab6.md
Jym77: Yes, that's a bit of a corner case difference with Inapplicable Example 1, where there is also no role which ends up with the `div` being out of the accessibility tree and the rule not triggering.

PR #2159, opened 2024-02-05
Add background info to meta refresh rules (bc659a, bisz58)

### LINKED ISSUE #2151: meta-refresh rules, add explanation of meta element [bisz58] [bc659a]
[Feedback from AGWG])(https://github.com/w3c/wcag/discussions/3549#discussioncomment-7832722)

Since meta refresh is a fairly dated technique, it would be useful to add a brief explanation of what it does and how it's used to the background of these two rules:

- [Meta element has no refresh delay](https://www.w3.org/WAI/standards-guidelines/act/rules/bc659a/)
- [Meta element has no refresh delay (no exception)](https://www.w3.org/WAI/standards-guidelines/act/rules/bisz58/proposed/)

carlosapaduarte:
Adds background info describing what a meta refresh is, to the meta refresh rules 

## Closes issue(s):

- closes #2151 

## Need for Call for Review:

- This will not require a Call for Review 

## How to Review And Approve


Jym77 (approved):
I wanted to say we should link to the specs… But we do so in the Expectation, and they are totally unreadable 😓 So, this is good.

tbostic32 (approved):

giacomo-petri (approved):

Jym77:
Call for review ends on April 25th.

Jym77:

PR #2160, opened 2024-02-08
Add accessibility support note about aria-busy to all rules with accessibility tree as input aspect

### LINKED ISSUE #2140: Should rules have an assumption about aria-busy?
This comes from an [ARIA review](https://github.com/w3c/aria/discussions/1850#discussioncomment-4309802)

We currently only have one rule that has an exception for aria-busy. I think for that rule it is a reasonable thing to have. I'm less sure about other rules, it seems unlikely that aria-busy is going to resolve issues such as invalid roles, or missing text alternatives. Even if not, should rules have an assumption that content is loaded?

carlosapaduarte:
Add accessibility support note about aria-busy to all rules with accessibility tree as input aspect

## Closes issue(s):

- closes #2140 

## Need for Call for Review:

This will not require a Call for Review: changes to accessibility support

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -42,6 +42,7 @@ The rule assumes that the [explicit role][] of the applicable elements is approp
 - User agents do not all have the same accessibility tree. This can lead to different results for this rule, depending on which accessibility tree is used as input.
 - `aria-owns` has limited support in some user agents.
 - There exist some combination of popular browsers and assistive technologies who do not announce correctly relationships based on a mix of [implicit][implicit role] and [explicit][explicit role] roles.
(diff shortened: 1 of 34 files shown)


giacomo-petri:
Should we contemplate including it in the rule template as well? Otherwise, there's a chance we might overlook this aspect when creating new rules. 

Alternatively, we could explore the option of implementing a script to automatically apply/remove this item (and potentially others in the future) to/from all rules.

Jym77 (approved):
I agree that it would be good to auto-inject this kind of things (we had case in the past with an assumption when Language is an input aspect) 🤔

tbostic32 (approved):

giacomo-petri (changes requested):
Sorry, I thought I had marked it as "request changes," but I never started the review. Please refer to my previous comment.

Issue #1967, opened 2022-11-09
[73f2c2] Autocomplete invalid - Pass example 8 is a fail, and justification is inconsistent


tombrunet:
Current text:
> This autocomplete [attribute value](https://act-rules.github.io/rules/73f2c2#attribute-value) only has the required token "bday-day". It remains programmatically identifiable even though it is inappropriate for the control's type [attribute value](https://act-rules.github.io/rules/73f2c2#attribute-value) "tel".
>
> ```<label>Birthday day<input name="bdayday" type="tel" autocomplete="bday-day"/></label>```

This example should be a failure since bday-day must be a numeric input type and is [not allowed](https://html.spec.whatwg.org/#inappropriate-for-the-control) for `type="tel".` It also fails in this basic checker:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.w3.org%2FWAI%2Fcontent-assets%2Fwcag-act-rules%2Ftestcases%2F73f2c2%2F35d0948a2cb309923e9a7cc5dd99a8ffc975088b.html

The description seems to indicate that it's a pass because the value can be programmatically accessed, however, that's true for all of the fail examples. The only distinction that I can make is that "bday-day" is valid for other input types, but not this type. 

As-is, this forces a tool that is correctly identifying appropriate values based on input type to pass something that is a failure.

WilcoFiers:
Right, so this is something that was changed a while back. Previously we would require type to be allowed. The issue is that autocomplete works perfectly fine, even if HTML spec the type isn't allowed. For the `type=tel` example, this is used quite a lot to trigger the number input on mobile devices. It's unfortunate that it has to work that way, but browsers implemented at that way, and as far as i can tell, even though it's not valid HTML, there's no negative impact to accessibility that I'm aware of.

tombrunet:
Okay, so 1.3.5:
> The purpose of each input field collecting information about the user can be [programmatically determined](https://www.w3.org/TR/WCAG21/#dfn-programmatically-determinable) when:
> * The input field serves a purpose identified in the [Input Purposes for User Interface Components section](https://www.w3.org/TR/WCAG21/#input-purposes); and
> * The content is implemented using technologies with support for identifying the expected meaning for form input data.

[...]

PR #2161, opened 2024-02-09
[73f2c2] Autocomplete invalid - Accessibility note clarifies HTML spec's disallowed autocomplete values for specific form controls

giacomo-petri:
- Included an accessibility support note to explain that discrepancies between form control types and autocomplete attribute values do not affect accessibility and meet the 1.3.5 SC, despite being disallowed by HTML specifications.
- Passing Example 8 was removed to avoid test cases that rely on the user agent or assistive technology used during testing.

Closes issue: #1967

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


tombrunet (approved):

colabottles (approved):

carlosapaduarte (approved):

giacomo-petri:
Call for review ends on February 21st.

dd8 (changes requested):
I agree that Passed Example 8 should not be a pass:

- the normative wording of the SC mentions 'programatically determinable' but it's only the understanding document that talks about autocomplete
- Passed Example 8 has two incompatible programatically determinable meanings: one from input=tel, the other from autocomplete=bday-day (both of these are exposed via MSAA / IAccessible2 / UIA )
 - The understanding document says the intent of the technique is for user agents to extract and present this purpose to users using different modalities
 
In this example some modalities convey the meaning as input=tel, and other modalities as autocomplete=bday-day. For example, VoiceOver with Safari announces this as "telephone number field", but somebody using AT to overlay icons on input fields might see a birthday cake icon next to the field (representing birthday day)

If this interpretation is correct, then I think this means the wording about "Thus, if the input possesses the suitable autocomplete ... regardless of type" probably needs to change.

**Edit:** ... and if this interpretation is correct, Passed Example 8 should probably be a fail rather removed completely

Jym77:
🤔 @dd8 @giacomo-petri 

Quickly browsing through the old issue this fixes, this looks like a case where UA + AT do not behave fully consistently. Moreover, in case of conflicting information (`type="tel"` vs `autocomplete="bday-day"`), the rule itself has no real way of deciding which of the two is the real "purpose" of the field and should be "programatically determined". So I think that always failing mismatching `autocomplete` would cause false positives (this was also the reason to remove that condition previously).

At the same time, I also agree that we shouldn't have a Passed Example like that. It is quite certainly failing the SC, as explained by Mark, so it shouldn't be a Passed Example (tools / methodologies that are smarter than the rule and correctly fail it should still have a consistent implementation).

It feels like we'd want to just discard the inappropriate `autocomplete` from the start, but we cannot exclude them from the Applicability given the nature of the rule which needs to check at invalid ones…

So, I think the only solution is to keep going as this PR does, maybe buffing up a bit the new note to explain that inappropriate `autocomplete` may or may not cause a fail of 1.3.5, and this rule doesn't care about them and simply checks validity.

---
We may want to add a rule checking that (valid) `autocomplete` are appropriate for their `type`. This would likely only have secondary mapping to 1.3.5.

carlosapaduarte:
CG: can you document any accessibility support issues related to passed example 8 before we merge this?

Jym77:
@giacomo-petri I'm removing Call for Review since it looks like https://github.com/act-rules/act-rules.github.io/pull/2161#issuecomment-1959659500 needs action.

giacomo-petri:
Replaced background note:

> Certain `autocomplete` attribute values are appropriate only for specific form controls. However, despite some combinations of `autocomplete` attribute values and form controls not being allowed by HTML specifications, the input purpose is programmatically available, and autocomplete functions correctly. Thus, if the input possesses the suitable autocomplete value, it fulfills the [Success Criterion 1.3.5 Identify Input Purpose](https://www.w3.org/TR/WCAG22/#identify-input-purpose), regardless of the input type.

with 

> Authors may assign inappropriate `autocomplete` attribute values. Moreover, HTML specifications restrict certain `autocomplete` attribute values to specific form controls. Mismatches between `autocomplete` attribute values and form control types may or may not lead to a failure of [Success Criterion 1.3.5 Identify Input Purpose](https://www.w3.org/TR/WCAG22/#identify-input-purpose). However, this rule focuses exclusively on validating valid `autocomplete` attribute values, disregarding their contextual appropriateness.

Jym77 (approved):

colabottles (approved):

giacomo-petri:
@dd8,

I believe we've covered all the points discussed. We've received 4 approvals, but I'd appreciate your final review, especially considering your previous concerns with the earlier version.

Thanks

dd8 (approved):
Perfect - that fixes it

giacomo-petri:
Call for review ends on May 8th

Jym77:

Issue #2179, opened 2024-04-10
HTML page has non-empty title [2779a5]: Shadow root example


giacomo-petri:
Test case:
A webpage lacks a `<title>` element but includes a shadow root that contains a `<title>`.

While the current phrasing of the expectations encompasses the scenario of a shadow root as a failure, I believe it would be beneficial to provide an example illustrating this case.

giacomo-petri:
PR merged, closing ticket

PR #2180, opened 2024-04-10
HTML page has non-empty title [2779a5]: Shadow root example - Included as failed example 6

giacomo-petri:
Closes issue(s): #2179

Description:
Adding a failed example of an HTML page lacking a title but featuring a shadow root with a non-empty `<title>` that does not affect the page title into the [HTML page has non-empty title [2779a5]](https://www.w3.org/WAI/standards-guidelines/act/rules/2779a5/) rule.

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/html-page-non-empty-title-2779a5.md
+++ b/_rules/html-page-non-empty-title-2779a5.md
@@ -210,7 +210,7 @@ This page does not have a title because the shadow root is not a [descendant](ht
 		</template>
 		<script>
       			const host = document.querySelector("body");
-      			const shadow = host.attachShadow({ mode: "closed" });
+      			const shadow = host.attachShadow({ mode: "open" });
       			const template = document.getElementById("shadow-element");
 
       			shadow.appendChild(template.content);
--- a/_rules/html-page-non-empty-title-2779a5.md
+++ b/_rules/html-page-non-empty-title-2779a5.md
@@ -198,6 +198,27 @@ This page has a `title` element that only contains a separator character.
 </html>
 ```
 
+#### Failed Example 6
+
+This page does not have a title because the shadow root is not a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of the [document element](https://dom.spec.whatwg.org/#document-element).
+
+```html
+<html>
+	<body>
+		<template id="shadow-element">
+			<title>This is the page title</title>
+		</template>
+		<script>
+      			const host = document.querySelector("body");
+      			const shadow = host.attachShadow({ mode: "closed" });
+      			const template = document.getElementById("shadow-element");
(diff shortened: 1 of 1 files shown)


Jym77 (approved):
I'm a bit wondering whether we should be clearer (maybe also in the Expectation) which tree is used for finding descendants (i.e., the shadow tree is a descendant **in the flat tree** of the root document, but it is not a descendant in the DOM tree, which the rule consider by default since no precision is made).

But that's a bit of a separate problem.

giacomo-petri:
Hi @Jym77,

to be honest, it took me a while to figure out if Expectation 2 was supposed to clarify how to deal with shadow roots, but after reviewing the definitions, I think it's accurate. However, I do agree that it could be clearer and less time-consuming for users to understand how to handle shadow roots.

Jym77:
> Hi @Jym77,
> 
> to be honest, it took me a while to figure out if Expectation 2 was supposed to clarify how to deal with shadow roots, but after reviewing the definitions, I think it's accurate. However, I do agree that it could be clearer and less time-consuming for users to understand how to handle shadow roots.

Yes, I think this is correct (this is also Alfa's interpretation). I think that this is a very old rule and we were not so conscious then about the differences between DOM and flat tree 😅 In any case, the example makes it clear that shadow DOM is ignored, so that's definitely going in the good direction.

_rules/html-page-non-empty-title-2779a5.md
```diff
@@ -198,6 +198,27 @@ This page has a `title` element that only contains a separator character.
...
+
+```html
+<html>
+	<body>
+		<template id="shadow-element">
+			<title>This is the page title</title>
+		</template>
+		<script>
+      			const host = document.querySelector("body");
+      			const shadow = host.attachShadow({ mode: "closed" });
```
WilcoFiers: Nitpick; I think this is a slightly stronger test case if it uses open shadow DOM, since closed shadow DOM can't be tested with regular DOM APIs.

```suggestion
      			const shadow = host.attachShadow({ mode: "open" });
```

WilcoFiers (changes requested):
Small suggestion

WilcoFiers (approved):

carlosapaduarte (approved):

giacomo-petri:
1 week call for review ends on September 11th.

colabottles (approved):

tombrunet (approved):

sage-putnam (approved):
Looks great to me!

Jym77:
Call for Review has ended, merging.

Issue #2164, opened 2024-02-27
Verbose examples (rules format)


dan-tripp-siteimprove:
I think that we would benefit from a new place in the rules format to put "verbose examples".   These would be examples that don't fit into our current categories (Passed / Failed / Inapplicable), and would not be counted in the tool implementation reports.  They would be more obscure, and would be interesting to a smaller group of people.  They would be informative nonetheless, for both ACT group members and some members of the general public.

For example, for the "Zoomed text node is not clipped with CSS overflow" rule, consider this case: zoomed text overlaps other text and does not clip.  Currently this passes the ACT rule and fails the SC.  I gather that overlapping text was left out of the ACT rule for now because it would be difficult to implement.  I think that overlapping text would be a useful as an example somewhere, to document the difference between the ACT rule and the SC.  We can't add it as a "Passed Example" because if we did, and then a keen vendor came up with an implementation that flags overlapping text and therefore matches the SC more closely, that vendor would be punished in their implementation report as having a false positive.  So a new "verbose examples" section would be the appropriate home for it.

Currently there is a lot of discussion of obscure cases which lives only in our github comments and meeting minutes.  This makes it difficult for us when we want to understand a decision a year or two after it was made.  A new "verbose examples" section could make that easier for us.

Other names for "verbose examples" could be: "illustrative examples" or "boundary cases".

dan-tripp-siteimprove:
Tagging @kengdoj because of our discussion of the examples for 'label in name'.

kengdoj:
The Editor's draft ACT Rules Format 1.1 has a [Background](https://w3c.github.io/wcag-act/act-rules-format.html#background) section that includes [Other Resources](https://w3c.github.io/wcag-act/act-rules-format.html#other-resources). Would "Other Resources" be a good place to include a link to the verbose examples?


On another thought:
> For example, for the "Zoomed text node is not clipped with CSS overflow" rule, consider this case: zoomed text overlaps other text and does not clip. Currently this passes the ACT rule and fails the SC. I gather that overlapping text was left out of the ACT rule for now because it would be difficult to implement. I think that overlapping text would be a useful as an example somewhere, to document the difference between the ACT rule and the SC.

Sould these be two atomic rules?

- Zoomed text node is not clipped with CSS
- Zoomed text does not overlap other text

(so the overlap example doesn't need to be included in the clipped rule)

dan-tripp-siteimprove:
[...]

Issue #2021, opened 2023-01-30
Audio element content has transcript[2eb176]: should the transcript files be HTML?


kengdoj:
These test cases link to a .txt file with HTML for a transcript:

- Pass ex 2
- Pass ex 3
- Fail ex 2
- Fail ex 3

Should the Pass examples link to html files instead? I don't think code is acceptable in a transcript for the pass examples?

carlosapaduarte:
We'll need to check all rules that use the same transcript files and update those also

PR #2185, opened 2024-04-25
Boundary / Verbose examples

Jym77:
Playing around the idea of boundary examples as discussed in #2164.

This adds one "boundary example" to some of the rules, showing how the rules are partial check and stuff that is inapplicable or even passing might still fail the SC (or other requirements).

These "boundary examples" must not be part of the normal examples that are required for tools to claim "consistent implementation" of the rule. Thus, tools can be more clever than the rule (i.e. fail these cases), but still have a consistent implementation.

This PR only shows a possible way to show these examples: in a collapsible section after the normal examples (so they are still together with the rule itself, but do not take space for people not interested in them). If we decide to go that way, we would need to build the website code (and tweak the validation) to automate part of this process.

I do believe that this won't require a format update since the format doesn't prohibit "extra" examples. Another possibility would be to place the boundary examples in the Background, but that makes them show a bit earlier in the text of the rule (even if the section is collapsed).

Closes issue(s):

- N/A

Need for Call for Review:
This will require a 2 weeks Call for Review (new concept)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-context-role-ff89c9.md
+++ b/_rules/aria-required-context-role-ff89c9.md
@@ -239,6 +239,23 @@ There is no element with an [explicit role][] different from its [implicit role]
 </ul>
 ```
 
+<details>
+<sumary>## Boundary Examples</sumary>
+
+These examples are not normative and not require for consistent implementations. They illustrate limitations and boundaries of the rule. Passed and Inapplicable examples in this category are likely to fail the rule's accessibility requirement, showing how the rule is only a partial test. Testing tools and methodologies are encouraged to be more clever than the rule, and to report the correct outcome for these examples. Therefore, they are not required for consistent implementation.
+
+### Inapplicable Examples
+
+#### Inapplicable Example 1
+
+This `li` element has an [explicit role][] identical to its [implicit role][] and therefore is not applicable. This nonetheless doesn't follow ARIA specifications since the resulting `listitem` does not have the required context role of `list`.
+
+```html
+<li role="listitem">List item 1</li>
+```
+
+</details>
+
 [accessibility tree]: https://www.w3.org/TR/act-rules-aspects/#input-aspects-accessibility 'Definition of accessibility tree'
 [aria 1.2]: https://www.w3.org/TR/wai-aria-1.2/ 'WAI ARIA 1.2 specifications'
 [content model]: https://html.spec.whatwg.org/multipage/dom.html#concept-element-content-model 'HTML definition of the Content Model'
--- a/_rules/image-non-empty-accessible-name-23a2a8.md
+++ b/_rules/image-non-empty-accessible-name-23a2a8.md
@@ -243,6 +243,23 @@ This `img` element is hidden with `visibility: hidden`.
 </div>
 ```
 
+<details>
+<sumary>## Boundary Examples</sumary>
+
+These examples are not normative and not require for consistent implementations. They illustrate limitations and boundaries of the rule. Passed and Inapplicable examples in this category are likely to fail the rule's accessibility requirement, showing how the rule is only a partial test. Testing tools and methodologies are encouraged to be more clever than the rule, and to report the correct outcome for these examples. Therefore, they are not required for consistent implementation.
+
+### Passed Examples
+
+#### Passed Example 1
+
+This `img` element has a non-empty [accessible name][]. The name is not descriptive and therefore the image fails [Success Criterion 1.1.1 Non-text content][sc111], but this rule only check for existence of a name. This `img` also fails rule [Image accessible name is descriptive](https://www.w3.org/WAI/standards-guidelines/act/rules/qt1vmo/).
+
+```html
+<img alt="Neil Armstrong walking on the Moon" src="/test-assets/shared/w3c-logo.png" />
+```
+
+</details>
+
 [accessible name]: #accessible-name 'Definition of accessible name'
 [explicit role]: #explicit-role 'Definition of explicit role'
 [focusable]: #focusable 'Definition of focusable'
--- a/_rules/html-page-lang-valid-bf051a.md
+++ b/_rules/html-page-lang-valid-bf051a.md
@@ -147,7 +147,7 @@ These examples are not normative and not require for consistent implementations.
 
 #### Inapplicable Example 1
 
-This `html` element has an empty `lang` attribute. It is not applicable for this rule, but nonetheless fails [Success Criterion 3.1.1 Language of Page][sc311]. It also fails rule [HTML page has non-empty title](https://www.w3.org/WAI/standards-guidelines/act/rules/b5c3f8/).
+This `html` element has an empty `lang` attribute. It is not applicable for this rule, but nonetheless fails [Success Criterion 3.1.1 Language of Page][sc311]. It also fails rule [HTML page has lang attribute](https://www.w3.org/WAI/standards-guidelines/act/rules/b5c3f8/).
 
 ```html
 <html lang="">
--- a/_rules/html-page-lang-valid-bf051a.md
+++ b/_rules/html-page-lang-valid-bf051a.md
@@ -138,6 +138,27 @@ This rule does not apply to `svg` elements.
 <svg xmlns="http://www.w3.org/2000/svg" lang="fr"></svg>
 ```
 
+<details>
+<sumary>## Boundary Examples</sumary>
+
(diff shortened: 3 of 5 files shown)


dan-tripp-siteimprove (commented):
I have some thoughts: 

1: 
The name "boundary examples" might not cover everything.  "Verbose examples" might.  It's true that "boundary examples" covers most of what we'll want to do with these.  But not everything.  All of your concrete examples in this PR fit into the category of "boundary examples".   However I can picture another kind of example which is not "boundary", and which might still be worth putting in this new section in the future.  Those would be examples in which there is no rule / SC discrepancy, and which might be already covered by previous examples in the main examples section, and which are still valuable to include because they are non-obvious.  "Verbose" would be a decent word for these.   Like these examples from [my long 'label in name' comment](https://github.com/act-rules/act-rules.github.io/issues/1458#issuecomment-1524088094): 
  - `<a href="#" aria-label="Some article by John Doe"><h6>Some article</h6><p>by John Doe</p></a> `
  -  `<div role="button" aria-label="Hello world"><p>Hello</p><p>world</p></div> `
  - `<button aria-label="how are you"><span>you</span><span>how</span><span>are</span></button> `
  
Someone might say that they're redundant, and they would have a point.  But they weren't always redundant - some (all?) of them used to fail.  So such examples will help future archeologists, as well as some implementors and users operating in the present.

Another set of "arguably redundant" examples which are not "boundary examples" and would have value as "verbose examples":
  -  `<a aria-label="WAVE" href="#">W A V E</a> `
  - `<a aria-label="Call 1 2 3. 4 5 6. 7 8 9 0." href="tel:1234567890">123.456.7890</a> `
  - `<a href="#2021" aria-label="20 21">2021</a> `


2: 
The new headings might cause confusion.  It would be unfortunate if someone confused the "Passed Example 1" under "boundary examples" with the "Passed Example 1" under the main "passed examples" section.  It seems to me that headings should be unique on a page when feasible.

-----

That said, in general I like this a lot.   In particular, the ideas of 1) putting it in a collapsible section, and 2) putting that section after the current examples.  

Let me know if I can help implement the website code for this.

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -239,6 +239,23 @@ There is no element with an [explicit role][] different from its [implicit role]
 </ul>
 ```
 
+<details>
+<sumary>## Boundary Examples</sumary>
+
+These examples are not normative and not require for consistent implementations. They illustrate limitations and boundaries of the rule. Passed and Inapplicable examples in this category are likely to fail the rule's accessibility requirement, showing how the rule is only a partial test. Testing tools and methodologies are encouraged to be more clever than the rule, and to report the correct outcome for these examples. Therefore, they are not required for consistent implementation.
```
giacomo-petri, line 246:
```suggestion
These examples are not normative and not required for consistent implementations. They illustrate limitations and boundaries of the rule. Passed and Inapplicable examples in this category are likely to fail the rule's accessibility requirement, showing how the rule is only a partial test. Testing tools and methodologies are encouraged to be more clever than the rule, and to report the correct outcome for these examples. Therefore, they are not required for consistent implementation.
```

typo, applies to all examples

giacomo-petri:
I believe there is no need to clarify the outcome of each specific example for the specific rule (passed, failed, inapplicable) within a heading. 

By definition, boundary examples "are not normative and are not required for consistent implementations." Therefore, in my opinion, and as recommended by @dan-tripp-siteimprove, even though I understand the intention to emphasize whether the current scenario is passing or inapplicable, we should avoid distinguishing this in the headings hierarchy to prevent confusion. 

The reason why an example is an edge case, in my opinion, should be provided in the description of the example, leaving just "Boundary Example 1," "Boundary Example 2," etc., as titles.

giacomo-petri:
And sorry, I forgot to mention the most important thing: I really love this idea!
We should definitely move forward with this idea!

_rules/aria-required-context-role-ff89c9.md
```diff
@@ -239,6 +239,23 @@ There is no element with an [explicit role][] different from its [implicit role]
 </ul>
 ```
 
+<details>
+<sumary>## Boundary Examples</sumary>
```
daniel-montalvo, line 244:
```suggestion
<summary>## Boundary Examples</summary>
```

_rules/image-non-empty-accessible-name-23a2a8.md
```diff
@@ -243,6 +243,23 @@ This `img` element is hidden with `visibility: hidden`.
 </div>
 ```
 
+<details>
+<sumary>## Boundary Examples</sumary>
```
daniel-montalvo, line 238:
```suggestion
<summary>## Boundary Examples</summary>
```

WilcoFiers (changes requested):
Just leaving my 2 cents.

1. I don't think we should put details / summary in here. I'd rather have that be added by our build tool. It's less error prone that way, and means we can more easily make changes to it in the future.
2. I think we should call these "Miscellaneous examples". As I understand the conversation, we'll want to use this for a variety of different kinds of examples. The only thing that they have in common is that they're not passed / failed / inapplicable examples.
3. I feel we should use different H3s / H4s here. Calling these a passed example, or an inapplicable example doesn't set it apart from the rest of the rule sufficiently in my opinion. People will get confused by what these passed examples mean compared to the other passed examples. We should call them something else. There are lots of things we can use here, like "Out of scope example 1", "Accessibility support example 2", "Explicit semantic role example 3", "Shadow DOM example 4". etc.

Jym77:
> Just leaving my 2 cents.
> 
> 1. I don't think we should put details / summary in here. I'd rather have that be added by our build tool. It's less error prone that way, and means we can more easily make changes to it in the future.
> 2. I think we should call these "Miscellaneous examples". As I understand the conversation, we'll want to use this for a variety of different kinds of examples. The only thing that they have in common is that they're not passed / failed / inapplicable examples.
> 3. I feel we should use different H3s / H4s here. Calling these a passed example, or an inapplicable example doesn't set it apart from the rest of the rule sufficiently in my opinion. People will get confused by what these passed examples mean compared to the other passed examples. We should call them something else. There are lots of things we can use here, like "Out of scope example 1", "Accessibility support example 2", "Explicit semantic role example 3", "Shadow DOM example 4". etc.

1. Definitely. This was just to show how it can look like.
2. 👍 
3. I think it can be good for tools vendors to have a way to retrieve these cases and use them for validation purpose. This will require us to provide the rule's result in a way that the build tool can understand and forward. But I agree that the names can be confusing.

dan-tripp-siteimprove:
I volunteered to be Project Leader for this on 2026-06-11.

dan-tripp-siteimprove:
I created a prototype page for this, on my github pages, [for the image alt text rule](https://dan-tripp-siteimprove.github.io/act-miscellaneous-examples/act-rule-23a2a8/deploy-preview-2185--act-rules.netlify.app/standards-guidelines/act/rules/23a2a8/proposed/index.html#misc).

I did it on my github pages because the details/summary didn't make it through to the netlify preview.  I'll look into doing this properly in the website build later. 

For this prototype, I made these tentative decisions:
- I called this thing "Miscellaneous examples" (not "boundary", nor "verbose"). 
- I put these "miscellaneous examples" in a single details/summary per rule page.   Not three of them (one for passed, one for failed, and one for inapplicable).
  - This is different than Jean-Yves' earlier prototype.  It's unclear to me how strong of an opinion he had on that part. 
- I put that details/summary after Inapplicable Examples and before the Glossary.
- Example heading: "Miscellaneous Example 2 (Passed)".
  - This satisfies Jean-Yves' idea that "I think it can be good for tools vendors to have a way to retrieve these cases and use them for validation purpose. This will require us to provide the rule's result in a way that the build tool can understand and forward." 
  - This contradicts Giacomo's suggestion that "I believe there is no need to clarify the outcome of each specific example for the specific rule (passed, failed, inapplicable) within a heading."   However, if I'm reading Giacomo's comment correctly, his goal is "to prevent confusion", and his suggestion is just one possible idea on how to reach that goal.  It's a similar situation with Wilco's goals to "set it apart from the rest of the rule sufficiently" and avoid "People will get confused".  I think that this prototype in it's current state reaches those goal in other ways: 1) it puts the word "Passed" in brackets, 2) it puts the word "Miscellaneous" also in the heading, and 3) it puts the whole thing in a details/summary that is after all of the real Passed/Failed/Inapplicable examples.

PR #2190, opened 2024-05-10
[NEW RULE] - ARIA required accessible names - aria-required-accessible-name-gp8n89 

giacomo-petri:
New rule: 
The aim of this new rule is to guarantee that ARIA roles requiring an accessible name meet the specified requirements.

Closes issue(s): #2119

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -108,7 +108,7 @@ The `div` element with role `dialog` has an [accessible name][] provided by the 
 The `div` element with role `heading` has an [accessible name][] provided by its content.
 
 ```html
-  <h1>Terms</h1>
+  <div role="heading" aria-level="1">Terms</div>
 ```
 
 #### Passed Example 6
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -93,8 +93,8 @@ The `div` element with role `img` has an [accessible name][] provided by the `ar
 The `div` element with role `checkbox` has an [accessible name][] provided thanks to the `aria-labelledby` attribute.
 
 ```html
-<div role="checkbox" aria-checked="false" tabindex="0" aria-labelledby="passed-example-3"></div>
-<div id="passed-example-3">I agree with terms and conditions</div>
+<div role="checkbox" aria-checked="false" tabindex="0" aria-labelledby="pass-agree-tc"></div>
+<div id="pass-agree-tc">I agree with terms and conditions</div>
 ```
 
 #### Passed Example 4
@@ -102,8 +102,8 @@ The `div` element with role `checkbox` has an [accessible name][] provided thank
 The `div` element with role `dialog` has an [accessible name][] provided by the `aria-labelledby` attribute.
 
 ```html
-<div role="dialog" aria-modal="true" tabindex="-1" aria-labelledby="passed-example-4">
-  <h1 id="passed-example-4">Terms</h1>
+<div role="dialog" aria-modal="true" tabindex="-1" aria-labelledby="pass-terms">
+  <h1 id="pass-terms">Terms</h1>
   <p>These are our terms.</p>
 </div>
 ```
@@ -185,8 +185,8 @@ The `div` element with role `heading` has an empty [accessible name][] due to th
 The `div` element with role `heading` has an empty [accessible name][] given by its `aria-labelledby` attribute.
 
 ```html
-<span id="failed-example-3"></span>
-<div role="heading" aria-level="1" aria-labelledby="failed-example-3">ACT Rules</div>
+<span id="fail-heading"></span>
+<div role="heading" aria-level="1" aria-labelledby="fail-heading">ACT Rules</div>
 ```
 
 #### Failed Example 7
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -49,7 +49,7 @@ The [accessible name and description computation][] suggests that if an `aria-la
 
 ## Background
 
-The applicability of this rule is limited to elements that have an [explicit semantic role][] that is not identical to their [implicit semantic role][]. Both [WAI-ARIA 1.2][wai-aria 1.2] and [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance) discourage authors from using ARIA when the host language provides a fetaure with equivalent role semantics and values.
+The applicability of this rule is limited to elements that have an [explicit semantic role][] that is not identical to their [implicit semantic role][]. Both [WAI-ARIA 1.2][wai-aria 1.2] and [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance) discourage authors from using ARIA when the host language provides a feature with equivalent role semantics and values.
 
 An example of an element that has an [implicit semantic role][] that is identical to its [explicit semantic role][] is a `<table role="table">` element. These elements are not applicable because they have extra requirements and should thus be checked separately.
 
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -60,7 +60,10 @@ Elements subjected to the [Presentational Roles Conflict Resolution][] are not c
 
 - [WAI-ARIA 1.2][wai-aria 1.2]
 - [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance)
+- [Understanding Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html)
 - [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value)
+- [Understanding Success Criterion 2.4.4 Link Purpose (In Context)]([https://www.w3.org/WAI/WCAG22/Understanding/name-role-value](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html))
+- [Understanding Success Criterion 2.4.9 Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-link-only.html)
 - [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value)
 
 ## Test Cases
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -257,14 +257,6 @@ This `button` element has an [explicit semantic role][] of `none`. leading to a 
 <button role="none">submit</button>
 ```
 
-<!-- Boundary example for future improvement -->
-<!--#### Inapplicable Example 5
-
-This `button` element has an [explicit semantic role][] of `none`. Although the [Presentational Roles Conflict Resolution][] requires the author to provide an [accessible name][], the [explicit semantic role][] does not require an [accessible name][] for this rule.
-
-```html
-<button role="none"></button>
-```-->
 
 [accessible name]: #accessible-name 'Definition of accessible name'
 [attribute value]: #attribute-value 'Definition of attribute value'
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -32,7 +32,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][wai-aria 1.2] [explicit semantic role][] that requires an [accessible name][] either from the [author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor) or the [content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent), unless the element [implicit semantic role][] matches its [explicit semantic role][].
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] that requires an [accessible name][] either from the [author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor) or the [content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent), unless the element has a presentational [explicit semantic role][] leading to a conflict resolved by [Presentational Roles Conflict Resolution][].
 
 ## Expectation
 
@@ -44,15 +44,15 @@ There are no assumptions.
 
 ## Accessibility Support
 
-Elements with [explicit semantic role][] of `heading` and with no [accessible name][] seem to be consistently ignored by assistive technologies. Nonetheless, in the context of this rule, they fail since [WAI-ARIA 1.2][wai-aria 1.2] mandates the presence of an [accessible name][].
+Elements with a [semantic role][] of `heading` and with no [accessible name][] seem to be consistently ignored by assistive technologies. Nonetheless, in the context of this rule, they fail since [WAI-ARIA 1.2][wai-aria 1.2] mandates the presence of an [accessible name][].
 
 The [accessible name and description computation][] suggests that if an `aria-labelledby` attribute refers to an existing but empty element, the computation should stop and return an empty name without defaulting to the next steps. Several user agents and assistive technologies chose to use the next step in the computation in this case (ultimately defaulting to the content).
 
 ## Background
 
-The applicability of this rule is limited to elements that have an [explicit semantic role][] that is not identical to their [implicit semantic role][]. Both [WAI-ARIA 1.2][wai-aria 1.2] and [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance) discourage authors from using ARIA when the host language provides a feature with equivalent role semantics and values.
+Although this ARIA rule mandates authors to ensure an [accessible name][] for each [HTML or SVG element][] [included in the accessibility tree][] and possessing a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] requiring an [accessible name][], it does not automatically constitute a WCAG accessibility violation.
 
-An example of an element that has an [implicit semantic role][] that is identical to its [explicit semantic role][] is a `<table role="table">` element. These elements are not applicable because they have extra requirements and should thus be checked separately.
+For instance, consider a `search` landmark comprising a `form` element with a search label, search input, and search submit button. According to ARIA specifications, both the `search` role and the `form` role necessitate an [accessible name][]. However, adhering strictly to this rule may lead to repetitive announcements by assistive technologies like screen readers, overwhelming users. Therefore, while this might technically breach ARIA specs, it does not necessarily translate to a violation of WCAG accessibility standards.
 
 Elements subjected to the [Presentational Roles Conflict Resolution][] are not part of this rule and must be tested separately.
 
@@ -72,13 +72,21 @@ Elements subjected to the [Presentational Roles Conflict Resolution][] are not p
 
 #### Passed Example 1
 
+This `button` element with an [implicit semantic role] of `button` has an [accessible name][] provided by its content.
+
+```html
+<button>Submit</button>
+```
+
+#### Passed Example 2
+
 The `div` element with role `button` has an [accessible name][] provided by its content.
 
 ```html
 <div role="button" tabindex="0">Submit</div>
 ```
 
-#### Passed Example 2
+#### Passed Example 3
 
 The `div` element with role `img` has an [accessible name][] provided by the `aria-label` [attribute value][].
 
@@ -88,7 +96,7 @@ The `div` element with role `img` has an [accessible name][] provided by the `ar
 </div>
 ```
 
-#### Passed Example 3
+#### Passed Example 4
 
 The `div` element with role `checkbox` has an [accessible name][] provided thanks to the `aria-labelledby` attribute.
 
@@ -97,7 +105,7 @@ The `div` element with role `checkbox` has an [accessible name][] provided thank
 <div id="pass-agree-tc">I agree with terms and conditions</div>
 ```
 
-#### Passed Example 4
+#### Passed Example 5
 
 The `div` element with role `dialog` has an [accessible name][] provided by the `aria-labelledby` attribute.
 
@@ -108,7 +116,7 @@ The `div` element with role `dialog` has an [accessible name][] provided by the 
 </div>
 ```
 
-#### Passed Example 5
+#### Passed Example 6
 
 The `div` element with role `heading` has an [accessible name][] provided by its content.
 
@@ -116,7 +124,7 @@ The `div` element with role `heading` has an [accessible name][] provided by its
   <div role="heading" aria-level="1">Terms</div>
 ```
 
-#### Passed Example 6
+#### Passed Example 7
 
 This `div` element with role `heading` is not [visible][], but is still [included in the accessibility tree][]. It has a non-empty [accessible name][] provided by its content.
 
@@ -124,7 +132,7 @@ This `div` element with role `heading` is not [visible][], but is still [include
 <div role="heading" aria-level="1" style="position: absolute; left: -9999px">ACT rules</div>
 ```
 
-#### Passed Example 7
+#### Passed Example 8
 
 The `div` element with role `link` has an [accessible name][] provided by its content.
 
@@ -144,6 +152,14 @@ The `div` element with role `button` doesn't have an [accessible name][].
 
 #### Failed Example 2
 
+This `button` element with an [implicit semantic role] of `button` doesn't have an [accessible name][].
+
+```html
+<button></button>
+```
+
+#### Failed Example 3
+
 The `div` element with role `img` doesn't have an [accessible name][].
 
 ```html
@@ -152,7 +168,7 @@ The `div` element with role `img` doesn't have an [accessible name][].
 </div>
 ```
 
-#### Failed Example 3
+#### Failed Example 4
 
 The `div` element with role `checkbox` doesn't have an [accessible name][].
 
@@ -161,7 +177,7 @@ The `div` element with role `checkbox` doesn't have an [accessible name][].
 <div>I agree with terms and conditions</div>
 ```
 
-#### Failed Example 4
+#### Failed Example 5
 
 The `dialog` role accepts an [accessible name][] only from the author and not from its content. Therefore, the `div` element with role `dialog` doesn't have an [accessible name][].
 
@@ -172,7 +188,7 @@ The `dialog` role accepts an [accessible name][] only from the author and not fr
 </div>
 ```
 
-#### Failed Example 5
+#### Failed Example 6
 
 The `div` element with role `heading` has an empty [accessible name][] due to the empty `aria-label` [attribute value][].
 
@@ -180,7 +196,7 @@ The `div` element with role `heading` has an empty [accessible name][] due to th
   <div role="heading" aria-level="1" aria-label="">Terms</div>
 ```
 
-#### Failed Example 6
+#### Failed Example 7
 
 The `div` element with role `heading` has an empty [accessible name][] given by its `aria-labelledby` attribute.
 
@@ -189,7 +205,7 @@ The `div` element with role `heading` has an empty [accessible name][] given by 
 <div role="heading" aria-level="1" aria-labelledby="fail-heading">ACT Rules</div>
 ```
 
-#### Failed Example 7
+#### Failed Example 8
 
 The `div` element with role `button` has an empty [accessible name][] because the `value` attribute does not count in the computation of the [accessible name][].
 
@@ -197,7 +213,7 @@ The `div` element with role `button` has an empty [accessible name][] because th
 <div role="button" value="test"></div>
 ```
 
-#### Failed Example 8
+#### Failed Example 9
 
 This `div` element with role `button` is not [visible][], but is still [included in the accessibility tree][]. It doesn't have an [accessible name][], therefore failing the rule.
 
@@ -205,7 +221,7 @@ This `div` element with role `button` is not [visible][], but is still [included
 <div role="button" style="position: absolute; left: -9999px" tabindex="0"></div>
 ```
 
-#### Failed Example 9
+#### Failed Example 10
 
 The `div` element with role `link` has an empty [accessible name][].
 
@@ -217,51 +233,26 @@ The `div` element with role `link` has an empty [accessible name][].
 
 #### Inapplicable Example 1
 
-This `button` element has no [explicit semantic role][].
-
-```html
-<button>Submit</button>
-```
-
-#### Inapplicable Example 2
-
-This `div` element is not [included in the accessibility tree][], hence its [explicit semantic role][] is not relevant.
+This `div` element is not [included in the accessibility tree][], hence its [semantic role][] is not relevant.
 
 ```html
 <div role="button" style="display:none;"></div>
 ```
 
-#### Inapplicable Example 3
+#### Inapplicable Example 2
 
-This `table` element has an [explicit semantic role][] of `table`, that is identical to its [implicit semantic role][].
-
-```html
-<table role="table">
-  <thead>
-    <tr>
-      <th>Table Header 1</th>
-      <th>Table Header 2</th>
-    </tr>
-  </thead>
-  <tbody>
-    <tr>
-      <td>Some content 1</td>
-      <td>Some content 2</td>
-    </tr>
-    <tr>
-      <td>Some content 3</td>
-      <td>Some content 4</td>
-    </tr>
-  </tbody>
-</table>
-```
-
-#### Inapplicable Example 4
-
-This `div` element has an [explicit semantic role][] of `group`, which does not require an [accessible name][].
+This `div` element has an [semantic role][] of `group`, which does not require an [accessible name][].
 
 ```html
 <div role="group">Some content</div>
+```
+
+#### Inapplicable Example 3
+
+This `button` element has an [explicit semantic role][] of `none`. leading to a conflict resolved by [Presentational Roles Conflict Resolution][], which is not covered by this rule.
+
+```html
+<button role="none">submit</button>
 ```
 
 <!-- Boundary example for future improvement -->
@@ -275,8 +266,9 @@ This `button` element has an [explicit semantic role][] of `none`. Although the 
 
 [accessible name]: #accessible-name 'Definition of accessible name'
 [attribute value]: #attribute-value 'Definition of attribute value'
-[explicit semantic role]: #explicit-role 'Definition of Explicit Role'
+[semantic role]: #semantic-role 'Definition of Semantic Role'
 [implicit semantic role]: #implicit-role 'Definition of Implicit Role'
+[explicit semantic role]: #explicit-role 'Definition of Explicit Role'
 [included in the accessibility tree]: #included-in-the-accessibility-tree 'Definition of Included in the Accessibility Tree'
 [presentational roles conflict resolution]: https://www.w3.org/TR/wai-aria-1.2/#conflict_resolution_presentation_none 'Presentational Roles Conflict Resolution'
 [wai-aria 1.2]: https://www.w3.org/TR/wai-aria-1.2/
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -32,7 +32,9 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] that requires an [accessible name][] either from the [author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor) or the [content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent), unless the element has a presentational [explicit semantic role][] leading to a conflict resolved by [Presentational Roles Conflict Resolution][].
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [semantic role][] that has the "Accessible Name Required: True" [characteristic][], unless the element has a [explicit][explicit semantic role] presentational role.
+
+[characteristic]: https://www.w3.org/TR/wai-aria-1.2/#Properties
 
 ## Expectation
 
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -32,7 +32,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][wai-aria 1.2] [explicit semantic role][] that requires an [accessible name][] either from the [author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor) or the [content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent), unless the element [implicit semantic role][] matches its [explicit semantic role][].
+Elements subjected to the [Presentational Roles Conflict Resolution][] are not part of this rule and must be tested separately. 
 
 ## Expectation
 
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -14,7 +14,7 @@ accessibility_requirements:
   wcag20:1.1.1: # Non-text Content (A)
     secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `img` or `image` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
   wcag20:1.3.1: # Info and Relationships (A)
-    secondary: This success criterion is **less strict** than this rule. Unlike WCAG, which might not mandate accessible names for certain elements like table elements, this rule requires them. Consequently, some of the failed examples satisfy this success criterion.
+    secondary: This success criterion is **related** to this rule. Unlike WCAG, which might not mandate accessible names for certain elements like table elements, this rule requires them. On the other hand, there are situations where information is not visually presented and is not required by WCAG, but is still required to have an accessible name according to ARIA requirements.
   wcag20:2.4.4: # Link Purpose (In Context) (A)
     secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
   wcag20:2.4.9: # Link Purpose (Link Only) (AAA)
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -257,7 +257,6 @@ This `button` element has an [explicit semantic role][] of `none`. leading to a 
 <button role="none">submit</button>
 ```
 
-
 [accessible name]: #accessible-name 'Definition of accessible name'
 [attribute value]: #attribute-value 'Definition of attribute value'
 [semantic role]: #semantic-role 'Definition of Semantic Role'
--- a/_rules/aria-required-accessible-name-gp8n89.md
+++ b/_rules/aria-required-accessible-name-gp8n89.md
@@ -11,6 +11,8 @@ accessibility_requirements:
     failed: not satisfied
     passed: satisfied
     inapplicable: satisfied
+  wcag20:1.1.1: # Non-text Content (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `img` or `image` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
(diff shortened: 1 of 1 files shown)


_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+  aria12:accessible_name_required:
+    title: ARIA 1.2, Accessible Name Required
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: satisfied
+  wcag20:1.3.1: # Info and Relationships (A)
+    secondary: This success criterion is **less strict** than this rule. Unlike WCAG, which might not mandate accessible names for certain elements like table elements, this rule requires them. Consequently, some of the failed examples satisfy this success criterion.
+  wcag20:2.4.4: # Link Purpose (In Context) (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that both passed and failed overlap with this success criterion.
```
Jym77:
```suggestion
    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+  wcag20:2.4.9: # Link Purpose (Link Only) (AAA)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that both passed and failed overlap with this success criterion.
```
Jym77:
```suggestion
    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+  wcag20:4.1.2: # Name, Role, Value (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned an ARIA role corresponding to a user interface component necessitate an accessible name according to WCAG standards. Some of the examples that both passed and failed overlap with this success criterion.
```
Jym77:
```suggestion
    secondary: This success criterion is **related** to this rule. This is because elements assigned an ARIA role corresponding to a user interface component necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
+---
+id: gp8n89
+name: ARIA required accessible name
+rule_type: atomic
+description: |
+  This rule checks that WAI-ARIA accessible name is available when required.
+accessibility_requirements:
+  aria12:accessible_name_required:
```
Jym77, line 8: IIrc, the second part is supposed to be the `id` of an element in the document so we can link to it (https://github.com/act-rules/act-tools/blob/cc2208ac0df7d53748d977cc7813887cbbf6a0fc/src/act/get-accessibility-requirement.ts and https://github.com/act-rules/act-tools/blob/cc2208ac0df7d53748d977cc7813887cbbf6a0fc/src/rule-transform/rule-content/get-requirements-map.ts), which this is not.
OTOH it doesn't look like there actually is a single place where this is defined 🤔

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+    title: ARIA 1.2, Accessible Name Required
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: satisfied
+  wcag20:1.3.1: # Info and Relationships (A)
```
Jym77, line 16: 1.1.1 should also be a secondary, since `img` role has name required.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that both passed and failed overlap with this success criterion.
+  wcag20:2.4.9: # Link Purpose (Link Only) (AAA)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that both passed and failed overlap with this success criterion.
+  wcag20:4.1.2: # Name, Role, Value (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned an ARIA role corresponding to a user interface component necessitate an accessible name according to WCAG standards. Some of the examples that both passed and failed overlap with this success criterion.
+input_aspects:
+  - Accessibility Tree
+  - CSS styling
+  - DOM Tree
+  - ARIA
```
Jym77: I don't think we have an ARIA aspect. This is more or less part of the Accessibility tree aspect.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+acknowledgments:
+  authors:
+    - Giacomo Petri
+---
+
+## Applicability
+
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][wai-aria 1.2] [explicit semantic role][] that requires an [accessible name][] either from the [author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor) or the [content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent), unless the element [implicit semantic role][] matches its [explicit semantic role][].
```
Jym77: I'm not sure about the "no implicit" exception in that case.
It makes some sense for the `aria-*` attribute because native elements do some magic to auto set them up (e.g. `aria-checked` is set by `checked`, and `aria-expanded` can be some deep magic).
However in that case, there is no magic to set up the name, so a native `` without a name is (also) an ARIA author eror and can be flagged by this rule.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Background
+
+The applicability of this rule is limited to elements that have an [explicit semantic role][] that is not identical to their [implicit semantic role][]. Both [WAI-ARIA 1.2][wai-aria 1.2] and [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance) discourage authors from using ARIA when the host language provides a feature with equivalent role semantics and values.
+
+An example of an element that has an [implicit semantic role][] that is identical to its [explicit semantic role][] is a `<table role="table">` element. These elements are not applicable because they have extra requirements and should thus be checked separately.
+
+Elements subjected to the [Presentational Roles Conflict Resolution][] are not covered by this rule, since their [explicit semantic role][] of `none` or `presentation` do not require an accessible name. However, if the [Presentational Roles Conflict Resolution][] is invoked, although the rule doesn't apply, an accessible name might still be necessary.
```
Jym77: I think this can be phrased a bit differently, hinting that triggering the conflict should be checked separately (and I do believe we should have a rule "conflict is not triggered", especially since it is [an explicit author error in ARIA 1.3](https://w3c.github.io/aria/#conflict_resolution_presentation_none)

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Bibliography
+
+- [WAI-ARIA 1.2][wai-aria 1.2]
+- [ARIA in HTML](https://www.w3.org/TR/html-aria/#docconformance)
+- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value)
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value)
```
Jym77: This should include the other secondary requirements?

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+</div>
+```
+
+#### Passed Example 3
+
+The `div` element with role `checkbox` has an [accessible name][] provided thanks to the `aria-labelledby` attribute.
+
+```html
+<div role="checkbox" aria-checked="false" tabindex="0" aria-labelledby="passed-example-3"></div>
+<div id="passed-example-3">I agree with terms and conditions</div>
```
Jym77: I'd recommend against using `passed-example-3` as an `id` to limit the changes needed if we add more examples.
(also applies to some other examples)

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+</div>
+```
+
+#### Passed Example 5
+
+The `div` element with role `heading` has an [accessible name][] provided by its content.
+
+```html
+  <h1>Terms</h1>
+```
```
Jym77: This is a native heading (which I think the rule can check nonetheless, but doesn't match the current wording).

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+The `div` element with role `checkbox` doesn't have an [accessible name][].
+
+```html
+<div role="checkbox" aria-checked="false" tabindex="0"></div>
+<div>I agree with terms and conditions</div>
+```
+
+#### Failed Example 4
+
+The `dialog` role accept an [accessible name][] only from the author and not from its content. Therefore, the `div` element with role `dialog` doesn't have an [accessible name][].
```
Jym77:
```suggestion
The `dialog` role accepts an [accessible name][] only from the author and not from its content. Therefore, the `div` element with role `dialog` doesn't have an [accessible name][].
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+```html
+<div role="dialog" aria-modal="true" tabindex="-1">
+  <h1>Terms</h1>
+  <p>These are our terms.</p>
+</div>
+```
+
+#### Failed Example 5
+
+The `div` element with role `heading` doesn't an [accessible name][] due to the empty `aria-label` [attribute value][].
```
Jym77:
```suggestion
The `div` element with role `heading` has an empty [accessible name][] due to the empty `aria-label` [attribute value][].
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+  <div role="heading" aria-level="1" aria-label="">Terms</div>
+```
+
+#### Failed Example 6
+
+The `div` element with role `heading` has an empty [accessible name][] given by its `aria-labelledby` attribute.
+
+```html
+<span id="failed-example-3"></span>
+<h1 aria-labelledby="failed-example-3">ACT Rules</h1>
```
Jym77: Also a native heading.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+<div role="group">Some content</div>
+```
+
+#### Inapplicable Example 5
+
+This `button` element has an [explicit semantic role][] of `none`. Although the [Presentational Roles Conflict Resolution][] requires the author to provide an [accessible name][], the [explicit semantic role][] does not require an [accessible name][] for this rule.
+
+```html
+<button role="none"></button>
+```
```
Jym77: Sadly, we can't have that as an Inapplicable example since these need to **not fail** the requirement, and this does.

That would be well suited for "boundary examples", though (#2185).

Jym77 (dismissed):

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,278 @@
...
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
+  wcag20:2.4.9: # Link Purpose (Link Only) (AAA)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `link` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
+  wcag20:4.1.2: # Name, Role, Value (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned an ARIA role corresponding to a user interface component necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
+input_aspects:
+  - Accessibility Tree
+  - CSS styling
+  - DOM Tree
+  - ARIA
```
giacomo-petri:
```suggestion
```

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: I missed it. The description is accurate "The `div` element with role `heading` has an [accessible name][] provided by its content." but the code doesn't reflect it. Updating the code

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,277 @@
...
+  <p>These are our terms.</p>
+</div>
+```
+
+#### Passed Example 5
+
+The `div` element with role `heading` has an [accessible name][] provided by its content.
+
+```html
+  <h1>Terms</h1>
```
giacomo-petri:
```suggestion
  <div role="heading" aria-level="1">Terms</div>
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,277 @@
...
+  <div role="heading" aria-level="1" aria-label="">Terms</div>
+```
+
+#### Failed Example 6
+
+The `div` element with role `heading` has an empty [accessible name][] given by its `aria-labelledby` attribute.
+
+```html
+<span id="failed-example-3"></span>
+<h1 aria-labelledby="failed-example-3">ACT Rules</h1>
```
giacomo-petri:
```suggestion
<div role="heading" aria-level="1" aria-labelledby="failed-example-3">ACT Rules</div>
```

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: Marked as comment so we don't forget about it if #2185 will be appoved.

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: @Jym77, exactly. Each role has its own unique requirements, which is why I've given it a general label... I'm uncertain about the best approach to address this.

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: @Jym77, the rationale behind this is that certain roles necessitate an accessible name per ARIA requirements, even if it's not mandated by WCAG requirements. For instance, consider a search form: there's a search landmark wrapping a form (which needs an accessible name per ARIA specs), followed by a Search label, a search input with its associated label, and finally, a search submit button.

When using a screen reader, we'd hear "search" multiple times, which seems more like an error than an accessibility feature. This holds true for many roles. That's why I've specifically excluded implicit roles from the applicability.

It's worth noting that during today's ARIA meeting, @scottaohara brought up this topic, and it appears there's consensus on reviewing each role that requires an accessible name to determine whether it's always necessary or not (minutes available: https://www.w3.org/2024/05/23-aria-minutes.html#t06).

I'd be more than willing to remove the exception with this change, but until then, I'm not comfortable doing so.

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: Replaced

> Elements subjected to the [Presentational Roles Conflict Resolution][] are not covered by this rule, since their [explicit semantic role][] of `none` or `presentation` do not require an accessible name. However, if the [Presentational Roles Conflict Resolution][] is invoked, although the rule doesn't apply, an accessible name might still be necessary.

with 

> Elements subjected to the [Presentational Roles Conflict Resolution][] are not part of this rule and must be tested separately. 

If we choose to proceed with eliminating the "implicit role" exception from the applicability and simply refer to it as "element role," perhaps it's not even necessary.

_rules/aria-required-accessible-name-gp8n89.md
Jym77: 🤔 It seems the WAI website currently doesn't show anything for non-WCAG requirements 🙈 ([example](https://www.w3.org/WAI/standards-guidelines/act/rules/6a7281/proposed/#accessibility-requirements-mapping)), where the old site did ([example](https://act-rules.github.io/rules/6a7281))…

Not sure why it went away, nor what we want to do. @carlosapaduarte do you have an input on that?

_rules/aria-required-accessible-name-gp8n89.md
Jym77: I'm not entirely sure as to why this relates to implicit roles 🤔

Jym77: I'm not entirely sure as to why this relates to implicit roles 🤔 
If the search landmark wrapping a search field, or any of the other "bad" cases discussed there, is built with `<div role="…">` all over the place, we end up in the exact same situation, don't we?

In any case, it seems to be a case where something is currently an ARIA error without being a WCAG one (nor an accessibility issue). Which does require deciding between flagging ARIA error, or only flagging accessibility issues…

Even if ARIA group is working on it, I guess this won't happen before ARIA 1.3, so we should find a solution without it.

Whatever we decide, I think this is worth a Background note to explain why the rule is shaped that way, or the limits of its use.

giacomo-petri: In my view, authors have the ability to utilize native semantic elements, such as a search element (`<search>`). If they do, even though it inherently carries the role of "search" as per ARIA, they may not be consciously utilizing ARIA. Conversely, if they consciously assign the role of search (`role="search"`), they are cognizant of ARIA requirements, which mandate the provision of an accessible name.

I'm not asserting that, in its current state, it does not constitute an ARIA violation (indeed it is), though it may not necessarily result in an accessibility issue. My concern lies in the potential confusion that may arise for those interpreting the rule if authors are required to adhere to this guideline for implicit roles as well.

However, again, I do acknowledge that, according to ARIA specs, it is indeed a requirement. Furthermore, once this concern is addressed by ARIA, there will no longer be a differentiation between implicit and explicit roles.

Therefore, I am inclined to remove the exception and update the background notes accordingly.

giacomo-petri: Side note: there were already a background note about that :D

giacomo-petri: @Jym77, I've made several updates:
- Expanded the applicability criteria to encompass any role, whether implicit or explicit.
- Adjusted the applicability to exclude conflicts with presentational roles. Previously, only explicit roles requiring acc name were included, so "none" and "presentation" were not part of the equation, excluding automatically presentational role conflicts. Now, considering the role of the element, also presentational roles causing conflicts apply, therefore I had to exclude them.
- Revised the background notes to address the discussed topic and emphasize that violating this ARIA rule does not necessarily breach WCAG standards.
- Enhanced the examples by incorporating implicit roles for both passed and failed examples, and removed implicit roles from inapplicable examples.
- Included a "non-failing" example into the inapplicable example of resolving presentational role conflicts, to clarify that this scenario is out of scope for this rule.

colabottles (approved):

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,275 @@
...
+  - CSS styling
+  - DOM Tree
+acknowledgments:
+  authors:
+    - Giacomo Petri
+---
+
+## Applicability
+
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] that requires an [accessible name][] either from the [author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor) or the [content](https://www.w3.org/TR/wai-aria-1.2/#namefromcontent), unless the element has a presentational [explicit semantic role][] leading to a conflict resolved by [Presentational Roles Conflict Resolution][].
```
Jym77: I'm not sure the "presentational explicit role" exception is even needed 🤔 
Because `` is pretty much bad anyway, no matter that the conflict was trigger to give the final role.
OTOH, might be worth waiting to see what others think.

Jym77 (dismissed):

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: Hi @Jym77,

I think I'm on the same page as you. Initially, the applicability was different, with only explicit roles, and there seemed to be some rationale behind that decision.

However, upon reviewing this matter now, I agree with what you said and I've also come to realize that our definition of semantic roles might not encompass all scenarios. Take, for instance:

```
<table role="none">
  <tbody>
    <tr>
      <th>Some content</th>
      <td>Some content</td>
    </tr>
  </tbody>
</table>
```

The `th` element inherits a presentational role, but as it stands, our current language implies (or at least seems to imply to me) that it holds a semantic role of `rowheader`, which isn't accurate (and shouldn't mandate an accessible name for this rule).

Lastly, I'm in the process of drafting three new atomic rules for the "Presentational Roles Conflict". They should be finalized soon; I'm just awaiting a response from the ARIA group regarding the last bullet point concerning the conflict of presentational roles, before presenting the PR to our group.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,275 @@
...
+  aria12:accessible_name_required:
+    title: ARIA 1.2, Accessible Name Required
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: satisfied
+  wcag20:1.1.1: # Non-text Content (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `img` or `image` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
+  wcag20:1.3.1: # Info and Relationships (A)
+    secondary: This success criterion is **less strict** than this rule. Unlike WCAG, which might not mandate accessible names for certain elements like table elements, this rule requires them. Consequently, some of the failed examples satisfy this success criterion.
```
WilcoFiers: This feels like a "related to" requirement as well. The way I read 1.3.1, it only applies if the accessible name should come from a visible label, who's relationship isn't available programmatically.

_rules/aria-required-accessible-name-gp8n89.md
WilcoFiers: This feels unnecessarily complex to me.
1. We don't need to mention an ARIA version here. The WAI-ARIA specifications definition has it, and that definition is used in explicit semantic role > semantic role. You've got this already
2. There are no other options besides name from author and name from content. You don't need to mention it.
3. I think we'll want to be explicit about what it means to have a required accessible name. 

```suggestion
This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [semantic role][] that has the "Accessible Name Required: True" [characteristic][], unless the element has a [explicit][explicit semantic role] presentational role.

[characteristic]: https://www.w3.org/TR/wai-aria-1.2/#Properties
```

The details of how to get an accessible name, of why an element can have a semantic role that's not a presentational role can go in the background.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Background
+
+Although this ARIA rule mandates authors to ensure an [accessible name][] for each [HTML or SVG element][] [included in the accessibility tree][] and possessing a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] requiring an [accessible name][], it does not automatically constitute a WCAG accessibility violation.
+
+For instance, consider a `search` landmark comprising a `form` element with a search label, search input, and search submit button. According to ARIA specifications, both the `search` role and the `form` role necessitate an [accessible name][]. However, adhering strictly to this rule may lead to repetitive announcements by assistive technologies like screen readers, overwhelming users. Therefore, while this might technically breach ARIA specs, it does not necessarily translate to a violation of WCAG accessibility standards.
```
WilcoFiers: There are some mistakes in this I think. Search doesn't require a name (searchbox does though). And while the `form` role requires an accessible name, the `form` element does not have `form` as its implicit role unless it has an accessible name, in the same way that the `img` element does not have the `img` role unless it has a name.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,275 @@
...
+  <p>These are our terms.</p>
+</div>
+```
+
+#### Failed Example 6
+
+The `div` element with role `heading` has an empty [accessible name][] due to the empty `aria-label` [attribute value][].
+
+```html
+  <div role="heading" aria-level="1" aria-label="">Terms</div>
```
WilcoFiers: I don't think this fails the rule. Browsers ignore empty aria-labels as far as I'm aware. The AccName spec say to do so too:

> https://w3c.github.io/accname/#comp_label
> D. AriaLabel: Otherwise, if the current node has an aria-label attribute whose value is not undefined, **not the empty string**, nor, when trimmed of whitespace, is not the empty string:

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+#### Failed Example 7
+
+The `div` element with role `heading` has an empty [accessible name][] given by its `aria-labelledby` attribute.
+
+```html
+<span id="fail-heading"></span>
+<div role="heading" aria-level="1" aria-labelledby="fail-heading">ACT Rules</div>
+```
```
WilcoFiers: I don't really like this example either. This rule is about checking that elements have accessible names. ACT rules aren't bug trackers to ensure tools have a flawless accessible name calculation.

WilcoFiers (dismissed):

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: I agree; we were discussing this matter here: https://github.com/act-rules/act-rules.github.io/pull/2190#discussion_r1621021885

Accepting your suggestion

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: Although I don't entirely agree with the interpretation, since there might be cases where something is visually clear due to the surrounding context or visual presentation of elements (e.g., a top promo banner carousel without an accessible name, but the section still requires one), I understand the rationale behind it being "related to."

Updating the secondary requirement.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,277 @@
...
+  aria12:accessible_name_required:
+    title: ARIA 1.2, Accessible Name Required
+    forConformance: true
+    failed: not satisfied
+    passed: satisfied
+    inapplicable: satisfied
+  wcag20:1.1.1: # Non-text Content (A)
+    secondary: This success criterion is **related** to this rule. This is because elements assigned the ARIA role of `img` or `image` necessitate an accessible name according to WCAG standards. Some of the examples that either pass or fail overlap with this success criterion.
+  wcag20:1.3.1: # Info and Relationships (A)
+    secondary: This success criterion is **less strict** than this rule. Unlike WCAG, which might not mandate accessible names for certain elements like table elements, this rule requires them. Consequently, some of the failed examples satisfy this success criterion.
```
giacomo-petri:
```suggestion
    secondary: This success criterion is **related** to this rule. Unlike WCAG, which might not mandate accessible names for certain elements like table elements, this rule requires them. On the other hand, there are situations where information is not visually presented and is not required by WCAG, but is still required to have an accessible name according to ARIA requirements.
```

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: I'm not sure I understand the point. 

I agree with the previous point (Failed Example 6), as the AccName specifications indicate that empty aria-labels should be ignored. 

However, I consider this a clear failure, since the element with the role of heading lacks an accessible name due to improper handling by the author.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,277 @@
...
+  <p>These are our terms.</p>
+</div>
+```
+
+#### Failed Example 6
+
+The `div` element with role `heading` has an empty [accessible name][] due to the empty `aria-label` [attribute value][].
+
+```html
+  <div role="heading" aria-level="1" aria-label="">Terms</div>
```
giacomo-petri:
```suggestion
The `div` element with role `heading` has an empty [accessible name][] due to the `display: none` CSS property set to its content.

```html
  <div role="heading" aria-level="1"><span style="display: none">Terms</span></div>
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Background
+
+Although this ARIA rule mandates authors to ensure an [accessible name][] for each [HTML or SVG element][] [included in the accessibility tree][] and possessing a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] requiring an [accessible name][], it does not automatically constitute a WCAG accessibility violation.
+
+For instance, consider a `search` landmark comprising a `form` element with a search label, search input, and search submit button. According to ARIA specifications, both the `search` role and the `form` role necessitate an [accessible name][]. However, adhering strictly to this rule may lead to repetitive announcements by assistive technologies like screen readers, overwhelming users. Therefore, while this might technically breach ARIA specs, it does not necessarily translate to a violation of WCAG accessibility standards.
```
giacomo-petri:
```suggestion
For instance, consider a `search` landmark comprising an element with the `form` role, a search label, a search input, and a search submit button. According to ARIA specifications, the `searchbox` role necessitate an [accessible name][]. However, adhering strictly to this rule may lead to repetitive announcements by assistive technologies like screen readers, overwhelming users. Therefore, while this might technically breach ARIA specs, it does not necessarily translate to a violation of WCAG accessibility standards.
```

colabottles (dismissed):

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+<!-- Boundary example for future improvement -->
+<!--#### Inapplicable Example 5
+
+This `button` element has an [explicit semantic role][] of `none`. Although the [Presentational Roles Conflict Resolution][] requires the author to provide an [accessible name][], the [explicit semantic role][] does not require an [accessible name][] for this rule.
+
+```html
+<button role="none"></button>
+```-->
```
giacomo-petri:
```suggestion
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Inapplicable Example 3
+
+This `button` element has an [explicit semantic role][] of `none`. leading to a conflict resolved by [Presentational Roles Conflict Resolution][], which is not covered by this rule.
+
+```html
+<button role="none">submit</button>
+```
+
+
```
WilcoFiers:
```suggestion

```

WilcoFiers (approved):

colabottles (approved):

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,268 @@
...
+This `div` element has an [semantic role][] of `group`, which does not require an [accessible name][].
+
+```html
+<div role="group">Some content</div>
+```
+
+#### Inapplicable Example 3
+
+This `button` element has an [explicit semantic role][] of `none`. leading to a conflict resolved by [Presentational Roles Conflict Resolution][], which is not covered by this rule.
```
carlosapaduarte:
```suggestion
This `button` element has an [explicit semantic role][] of `none`, leading to a conflict resolved by [Presentational Roles Conflict Resolution][], which is not covered by this rule.
```

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,268 @@
...
+
+The `div` element with role `heading` has an empty [accessible name][] due to the `display: none` CSS property set to its content.
+
+```html
+  <div role="heading" aria-level="1"><span style="display: none">Terms</span></div>
+```
+
+#### Failed Example 7
+
+The `button` element has an empty [accessible name][] given by its `aria-labelledby` attribute.
```
carlosapaduarte: I know it's mentioned in the background, but do we need to have a failed example where all major browsers ignore the accessible name computation rules and do provide the button with an accessible name?

carlosapaduarte (changes requested):
One edit requested and one question to discuss.

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: acc name computation is changes and now says:

> LabelledBy: Otherwise, if the current node has an aria-labelledby [attribute](https://dom.spec.whatwg.org/#concept-attribute) that contains at least one valid IDREF, and the current node is not already part of an ongoing aria-labelledby or aria-describedby traversal, process its IDREFs in the order they occur:

so, let's remove this example!

giacomo-petri:
the role="dialog" does no longer require the acc name, as per https://github.com/w3c/aria/pull/2297/files
So I've updated the examples.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Background
+
+Although this ARIA rule mandates authors to ensure an [accessible name][] for each [HTML or SVG element][] [included in the accessibility tree][] and possessing a [WAI-ARIA 1.2][wai-aria 1.2] [semantic role][] requiring an [accessible name][], it does not automatically constitute a WCAG accessibility violation.
+
+For instance, consider a `search` landmark comprising an element with the `form` role, a search label, a search input, and a search submit button. According to ARIA specifications, the `searchbox` role necessitate an [accessible name][]. However, adhering strictly to this rule may lead to repetitive announcements by assistive technologies like screen readers, overwhelming users. Therefore, while this might technically breach ARIA specs, it does not necessarily translate to a violation of WCAG accessibility standards.
+
+Elements subjected to the [Presentational Roles Conflict Resolution][] are not part of this rule and must be tested separately.
```
dan-tripp-siteimprove: I'm not sure what to make of this.  Does this affect the applicability?  If so, does it mean that the applicability should be changed to this: 

"... unless the element has a [explicit][explicit semantic role] presentational role which is not to be ignored according to [Presentational Roles Conflict Resolution][]".

?

edit: I just read Inapplicable Example 3, and it seems to contradict my interpretation in my preceding paragraph.  Regardless: I get the impression that confliction resolution is a factor in the applicability.  If that's true, then that fact should be documented in the applicabililty, not just in the background or examples.  

edit number 2: I find is strange that conflict resolution should be a factor in the applicability.  Wouldn't it be cleaner to make it not so?

edit number 3: I read some old resolved comments between you and Jean-Yves, and I think I get it now.  You really do want to make any element with role="none" and which is subject to conflict resolution inapplicable. That's fine, but I think we can help future confused readers like myself.  May I suggest changing this line in the background: 

"Elements subjected to the [Presentational Roles Conflict Resolution][] are not part of this rule and must be tested separately."

to this:

"Any element subjected to the [Presentational Roles Conflict Resolution][] is inapplicable under this rule, and must be tested separately.  Such an element is inapplicable because it has an [explicit][explicit semantic role] presentational role - even though that presentational role will be ignored by user agents, as a result of conflict resolution."

I don't think we use the term "computed role" around here, but maybe we should consider it at some point.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,245 @@
...
+```html
+<div role="group">Some content</div>
+```
+
+#### Inapplicable Example 3
+
+This `button` element has an [explicit semantic role][] of `none`, leading to a conflict resolved by [Presentational Roles Conflict Resolution][], which is not covered by this rule.
+
+```html
+<button role="none">submit</button>
```
dan-tripp-siteimprove, line 234: I gather there are really two reasons why this button does not have a "computed" role of none: 1) confliction resolution's "element is focusable" clause, and 2) the fact that https://www.w3.org/TR/html-aria/#el-button , third column, does not contain the words "Any role".  If I'm right about that, it means that this example is a little weirder than I think you meant it to be.  But maybe it doesn't matter - I'll wait to see the outcome of my other comment on conflict resolution.

dan-tripp-siteimprove (changes requested):

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: I'm not sure I'm recalling it exactly, but the idea, especially in light of my proposal for a [new rule checking for presentational roles conflict](https://github.com/act-rules/act-rules.github.io/pull/2195), was to encourage authors to address the presentational roles conflict issue first, rather than simply fixing the name. Once the original error is resolved, next steps can be considered.

I like your rewording suggestion and will apply it.

giacomo-petri:
```suggestion
Any element subjected to the [Presentational Roles Conflict Resolution][] is inapplicable under this rule, and must be tested separately. Such an element is inapplicable because it has an [explicit][explicit semantic role] presentational role - even though that presentational role will be ignored by user agents, as a result of conflict resolution.
```

_rules/aria-required-accessible-name-gp8n89.md
giacomo-petri: Answered your question above. This example is just to clarify is not in scope for this rule.

WilcoFiers (commented):
@giacomo-petri As discussed, I think we should compare this to existing rules to see which things this covers that we don't already have. We can then decide if we want to exclude those from this rule. Generally we've been trying to avoid overlap between rules - I think I'd prefer we do the same here.

Once we have the list we should put this on the agenda for discussion.

giacomo-petri:
I've reviewed the following specifications to help the group better understand the WAI-ARIA direction and intent regarding the reduction of roles that require an acc name:
- ARIA 1.2 W3C Recommendation: https://www.w3.org/TR/wai-aria-1.2/#namefromauthor
- ARIA 1.3 W3C First Public Working Draft: https://www.w3.org/TR/wai-aria-1.3/#namefromauthor 
- ARIA 1.3 W3C Editor's Draft: https://w3c.github.io/aria/#namefromauthor

I compiled a list of all and only the roles that require an accessible name in at least one of these versions. For each role, I noted any changes, such as if it was not originally required but is now, or if it has since been removed from the requirement.

In the final column, I included which of these roles are already covered by our rules, indicating whether we have specific test examples for them or if they are only applicable without associated examples.

| ARIA 1.2 Recommendation | ARIA 1.3 First Public Working Draft | ARIA 1.3 Editor's Draft | Already covered by ACT Rules
|----------|----------|----------|----------|
| alertdialog       | alertdialog       | does no longer require acc name | no |
| application       | application       | application                      | no |
| button            | button            | button                           | 97a4e1 - failed example 3 |
| checkbox          | checkbox          | checkbox                         | e086e5 - passed example 7 |
| columnheader      | columnheader      | columnheader                     | no |
| combobox          | combobox          | combobox                         | e086e5 - passed example 6 |
| dialog            | dialog            | does no longer require acc name | no |
| form            | form            | does no longer require acc name | no |
| grid              | grid              | does no longer require acc name | no |
| heading           | heading           | heading                          | ffd0e9 - all examples (proposed) |
| img               | img               | image | 23a2a8 - all examples |
| link              | link              | link                             | c487ae - all examples |
| listbox           | listbox           | listbox                          | e086e5 - no examples |
| marquee           | marquee           | does no longer require acc name | no |
| menuitem          | menuitem          | menuitem                         | m6b1q3 - all examples |
| menuitemcheckbox  | menuitemcheckbox  | menuitemcheckbox                 | e086e5 - passed example 8, failed example 8 |
| menuitemradio     | menuitemradio     | menuitemradio                    | e086e5 - no examples |
| meter           | meter           | meter                    | no |
| option            | option            | option                           | no |
| progressbar       | progressbar       | progressbar                      | no |
| radio             | radio             | radio                            | e086e5 - no examples |
| radiogroup        | radiogroup        | does no longer require acc name | no |
| region            | region            | region                           | no |
| rowheader         | rowheader         | rowheader                        | no |
| searchbox         | searchbox         | searchbox                        | e086e5 - no examples |
| slider            | slider            | slider                           | e086e5 - no examples |
| spinbutton        | spinbutton        | spinbutton                       | e086e5 - no examples |
| switch            | switch            | switch                           | e086e5 - no examples |
| tab            | tab            | tab                           | no |
| table             | table             | does no longer require acc name | no |
| tabpanel          | tabpanel          | tabpanel                         | no |
| textbox           | textbox           | textbox                          | e086e5 - failed example 5, failed example 6, failed example 7 |
| tooltip           | tooltip | does no longer require acc name | no |
| tree              | tree              | tree                             | no |
| treegrid          | treegrid          | treegrid                         | no |
| treeitem          | treeitem          | treeitem                         | no |

I understand your concern, @WilcoFiers, about avoiding duplicates and minimizing secondary requirements. However, I don't see a straightforward way to categorize the remaining missed roles in a clean manner. For example, while many fall under non-operable controls, we are still missing roles like "tab" in other rules, so a category like "ARIA non-UIC" doesn't fully hold up.

I'm inclined to suggest we keep it as is, accepting some overlap, given that ARIA roles inherently impact multiple rules. The alternative would be to explicitly list all the ARIA roles covered by this rule and justify why others were excluded in the background note. But IMO, that approach is more complex and harder for people to digest (and will probably require more maintenance over time).

Open to your thoughts on this.

carlosapaduarte:
Decision: Try to group roles in a meaningful way. Create rules for meaningful role groups. As a side effect, discuss the need for names of some of the roles.

_rules/aria-required-accessible-name-gp8n89.md
```diff
@@ -0,0 +1,245 @@
...
+  - CSS styling
+  - DOM Tree
+acknowledgments:
+  authors:
+    - Giacomo Petri
+---
+
+## Applicability
+
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [semantic role][] that has the "Accessible Name Required: True" [characteristic][], unless the element has a [explicit][explicit semantic role] presentational role.
```
Jym77, line 35:
```suggestion
This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [semantic role][] that has the "Accessible Name Required: True" [characteristic][], unless the element has an [explicit][explicit semantic role] presentational role.
```

Jym77 (approved):

giacomo-petri:
Before moving forward, I'd like to gather opinions and establish an initial consensus on the approach, to avoid creating rules that might not be adopted or fail to gain consensus with our CG.

Based on previous feedback, the idea of having a single rule for all ARIA roles requiring an accessible name was dismissed due to overlap with existing rules.

Missing rules for the following roles:
- `application`
- `columnheader`
- `option`
- `meter`
- `progressbar`
- `region`
- `rowheader`
- `tab`
- `tabpanel`
- `tree`
- `treegrid`
- `treeitem`

Proposed next steps:
- ARIA table header roles have a non-empty accessible name
  covers:
     - `columnheader`
     - `rowheader`
- ARIA tab interface roles have an accessible name
   covers:
     - `tab`
     - `tabpanel`
- ARIA tree structure roles have an accessible name
   covers:
     - `tree`
     - `treegrid`
     - `treeitem`
- ARIA range widget roles have an accessible name 
   covers:
     - `meter`
     - `progressbar`

The roles `application`, `option`, and `region` are not covered by these groupings. They appear unrelated to other rules, so we would need separate rules specifically for them. Should we go ahead and create those or just ignore them?

Additionally, roles like `listbox`, `menuitemradio`, `radio`, `searchbox`, `slider`, `spinbutton`, and `switch` are already covered by existing rules, but we currently lack examples. We could improve the guidance by adding examples for these roles, if interested.

dd8:
I think having separate rules for table / tab / tree / range widgets is the right approach. 

Error messages for failures are less generic than with a single rule, and you can provide more useful help (e.g. "Add a name describing the data in the row or column using aria-label" instead of "Provide an accessible name using aria-label")

tombrunet:
@dd8 
ACT rules define what conditions tools must detect, not how they present the detected conditions. While splitting rules could enable more specific messages, tools like ours already provide context-appropriate guidance through internal state management. More specific error messages are a tool implementation concern, not necessarily an ACT specification issue. 

On the other hand, splitting it in ACT would force tools to track that more specific condition, even though it would map back to the same conformance criteria.

Jym77:
> @dd8 ACT rules define what conditions tools must detect, not how they present the detected conditions. 

Yes.

> On the other hand, splitting it in ACT would force tools to track that more specific condition, even though it would map back to the same conformance criteria.

Not fully. A tool/method can have a single check mapping to several ACT rules (that is also why we insist that Passed (and Inapplicable) examples must not just pass the rule but also not fail the associated requirement. So, as long as the rule for tables does not include a Passed Example with a good table and a bad tree, a tool could have a single check testing both tables and trees in one go, and report it as implementing both ACT rules.

PR #2167, opened 2024-03-08
Target size minimum

Jym77:
Add rules for "Target Size (minimum)" and "Target Size (enhanced)" Success Criteria.

These are set up as composite rules in order to share the exceptions that are shareable.

"Target Size (minimum)" is messy due to the spacing condition that has to be failing in all Failed example, thus requiring quite a bit of scaffholding.

Closes issue(s):

- closes #1919 (supercedes it)

Need for Call for Review:
This will require a 2 weeks Call for Review (new rules 🎉)

---

### **When creating PR:**

- [X] Make sure you're requesting to **pull a branch** (right side) to the `develop` branch (left side).
- [X] Make sure you do not remove the "How to Review and Approve" section in your pull request description

### **After creating PR:**

- [X] Add yourself (and co-authors) as "Assignees" for PR.
- [X] Add label to indicate if it's a `Rule`, `Definition` or `Chore`.
- [X] [Link the PR to any issue it solves](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue). This will be done automatically by [referencing the issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) at the top of this comment in the indicated place.
- [X] Optionally request feedback from anyone in particular by assigning them as "Reviewers".

### **When merging a PR:**

- [ ] Close any issue that the PR resolves. This will happen automatically upon merging if the PR was correctly linked to the issue, e.g. by referencing the issue at the top of this comment.

## How to Review And Approve

- Go to the “Files changed” tab
- Here you will have the option to leave comments on different lines.
- Once the review is completed, find the “Review changes” button in the top right, select “Approve” (if you are really confident in the rule) or "Request changes" and click “Submit review”.
- Make sure to also review the proposed Call for Review period. In case of disagreement, the longer period wins.

PR #2195, opened 2024-06-07
[NEW RULES] - Define author's responsibility when presentational roles conflict arises

### LINKED ISSUE #2193: [New Rule] Presentational Roles Conflict is not triggered
New rule to check [Presentational Roles Conflict](https://w3c.github.io/aria/#conflict_resolution_presentation_none) is not triggered, since an author responsibility.

giacomo-petri:
New Rules:
- Check that elements with ARIA presentational role are not focusable
- Check that allowed child element of another element with presentational role does not cause presentational roles conflicts
- Check that elements with ARIA presentational role do not have global states or properties

Closes: #2193

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve


_rules/aria-presentational-role-not-focusable-18pg11.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `button` element with role attribute value `presentation` is [focusable][].
+
+```html
+<button role="presentation">Submit</button>
+```
+
+#### Failed Example 2
+
+The `button` element with role attribute value `none` is not tabbable but still [focusable][].
```
giacomo-petri:
```suggestion
The `button` element with role attribute value `none` is not reachable via keyboard but still [focusable][].
```

_rules/aria-allowed-child-element-with-presentational-parent-1g88p9.md
```diff
@@ -0,0 +1,171 @@
...
+  - CSS styling
+  - DOM Tree
+acknowledgments:
+  authors:
+    - Giacomo Petri
+---
+
+## Applicability
+
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][], is an [allowed child element](https://w3c.github.io/aria/#mustContain) of another element with an [explicit semantic role][] of `none` or `presentation`.
```
WilcoFiers: I don't think we should link to editors drafts. These are not stable documents.

_rules/aria-allowed-child-element-with-presentational-parent-1g88p9.md
```diff
@@ -0,0 +1,171 @@
+---
+id: 1g88p9
+name: ARIA allowed child element of another element with presentational role
+rule_type: atomic
+description: |
+  This rule checks that allowed child element of another element with presentational role does not cause presentational roles conflicts
```
WilcoFiers, line 6: This rule seems like a subset of our `ARIA required context role` rule. It requires that elements that can't exist without a parent have that parent. The rule you propose is more focused, but it doesn't seem to cover anything that isn't also covered by the `required context` rule. Or at least, I can't find anything that isn't.

Would a better solution be to add the one or both of the failed examples to `required context` instead?

_rules/aria-presentational-role-no-global-states-properties-p8g918.md
```diff
@@ -0,0 +1,175 @@
+---
```
WilcoFiers, line 1: I don't think you should have more than one new rule in a PR.

_rules/aria-presentational-role-no-global-states-properties-p8g918.md
```diff
@@ -0,0 +1,175 @@
...
+id: p8g918
+name: ARIA presentational role does not have global states or properties
```
WilcoFiers, line 3: I'm very sorry to do this a second time. I can tell you've put a lot of work into these, but this rule also seems like a duplicate to me. This time of [Element marked as decorative is not exposed ](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/). Your version makes slightly different trade-offs about explicitly stating which props cause this problem, whereas the existing rule is vaguer about that. But then in being vaguer and just saying "not in the accessibility tree" its means it can handle changes to the ARIA spec like how aria-label is not a global attr if on role=none, but still fails for creating presentation conflict.

_rules/aria-presentational-role-not-focusable-18pg11.md
```diff
@@ -0,0 +1,139 @@
+---
+id: 18pg11
+name: ARIA presentational role not focusable
```
WilcoFiers, line 3: See above, the [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/#marked-as-decorative) rule covers this.

I think we can debate splitting the rule up into the two rules you propose. I'd personally lean towards leaving this as is, since the "not exposed" rule is already approved by AG, but if others feel there are good reasons to deprecate it in favor of this I'm open to that conversation.

WilcoFiers (changes requested):

_rules/aria-allowed-child-element-with-presentational-parent-1g88p9.md
giacomo-petri:
```suggestion
This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][], is an [allowed child element](https://www.w3.org/TR/wai-aria-1.3/#mustContain) of another element with an [explicit semantic role][] of `none` or `presentation`.
```

_rules/aria-presentational-role-no-global-states-properties-p8g918.md
giacomo-petri: I apologize for this. I based my work on this pattern: https://github.com/act-rules/act-rules.github.io/pull/2167. These three rules are intended to fulfil the "Presentational Roles Conflict Resolution" topic from an author's perspective.

_rules/aria-allowed-child-element-with-presentational-parent-1g88p9.md
giacomo-petri: I see the point. 

The intent was to cover a specific ARIA requirement in detail and comprehensively (Presentational Roles Conflict Resolution) to guarantee authors are clear on what to do in these scenarios. 

I'm uncertain about the next steps. 

The other two existing ARIA requirements are rather more vague but do cover the necessary aspects. If we decide to revert these new rules, we'll still need to add examples to the existing rules.

_rules/aria-presentational-role-no-global-states-properties-p8g918.md
giacomo-petri: I see the point. 

The intent was to cover a specific ARIA requirement in detail and comprehensively (Presentational Roles Conflict Resolution) to guarantee authors are clear on what to do in these scenarios. 

I'm uncertain about the next steps. 

The other two existing ARIA requirements are rather more vague but do cover the necessary aspects. If we decide to revert these new rules, we'll still need to add examples to the existing rules.

_rules/aria-presentational-role-not-focusable-18pg11.md
giacomo-petri: I see the point. 

The intent was to cover a specific ARIA requirement in detail and comprehensively (Presentational Roles Conflict Resolution) to guarantee authors are clear on what to do in these scenarios. 

I'm uncertain about the next steps. 

The other two existing ARIA requirements are rather more vague but do cover the necessary aspects. If we decide to revert these new rules, we'll still need to add examples to the existing rules.

giacomo-petri:
After our conversation with @carlosapaduarte and @dan-tripp-siteimprove during the office hours meeting, I took another look at this item (since it was on me and I forgot about it 🫤).

Although there is certainly overlap with [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/) as mentioned by @WilcoFiers, I still see value in this PR. While it doesn't fully cover the same scope as [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/) - the applicability here is limited to elements that trigger the "presentational role conflict resolution", so examples like 
``
 are not within the scope of my PR - it does get rid of the ambiguous "marked as decorative" definition.

Additionally, it covers scenarios not currently within the scope of the test, such as:
```
<button>
  <a href="https://www.w3.org/WAI/standards-guidelines/act/rules/">All ACT Rules</a>
</button>
```
In this case, the link element is not marked as decorative by the author, even though the button has children presentational, so it falls outside the [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/) rule's applicability, but it is still an author error, and caught by this new rule.

and:

```
<ul role="none">
  <li role="listitem">WCAG</li>
  <li role="listitem">ARIA</li>
  <li role="listitem">ACT Rules</li>
</ul>
```
In [Element marked as decorative is not exposed](https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/) rule, the `<ul>` is technically meeting the current requirement because its role is ignored, but it still causes a "presentational role conflict," which leads to an author error.

Furthermore, while I agree there is repetition, the rule's focus is very clear and well-defined, since it specifically addresses all items that fall under the "presentational role conflict" definition.

Is it really a problem if something is duplicated, as long as the scope of the check is clear and unambiguous?

_rules/aria-allowed-child-element-with-presentational-parent-1g88p9.md
shunguoy: A "none" role only applies to three native elements that require hierarchies: 'table', 'ol', and 'ul'. 
   Not sure if the Failed Example 1 & 2 are already covered under
> ARIA required context role
Where the hierarchy for the child element is not met because of the parent's presentation role.

PR #2199, opened 2024-06-28
Test is clipped: "link to full version"

### LINKED ISSUE #2163: "Zoomed text node is not clipped with CSS overflow": "link to a full version": source unclear
[Passed Example 2](https://www.w3.org/WAI/standards-guidelines/act/rules/59br37/proposed/#passed-example-2) passes because it has a "link to a full version".

The problem is that I don't see where a "link to a full version" is mentioned in the applicability, expectations, or assumptions.

The closest thing I have found are these shaky references in [one of the corresponding SC's sufficient techniques](https://www.w3.org/WAI/WCAG22/Techniques/general/G178.html): 
* "The mechanism may include links or buttons that will switch the visual presentation..."
  *  Maybe a link to a full version on a separate page, like Passed Example 2 has, could be considered similar to this.
* "This technique can be used in combination with a style switching technique to present a page that is a [conforming alternate version](https://www.w3.org/TR/WCAG22/#dfn-conforming-alternate-versions) for non-conforming content"
  * This idea of a 'conforming alternate version' is something I don't know a lot about.  It seems to apply to all SCs implicitly.  I don't know what an ACT rule is supposed to do with this.

I'm looking forward to people's views on this.

dan-tripp-siteimprove:
<< Describe the changes >>

Closes issue(s):

- closes  #2163 

Need for Call for Review:

---

## How to Review And Approve


dan-tripp-siteimprove:
I made a mistake.  I'll redo this in a minute.

PR #2200, opened 2024-06-28
Text is clipped: "link to a full version"

### LINKED ISSUE #2163: "Zoomed text node is not clipped with CSS overflow": "link to a full version": source unclear

dan-tripp-siteimprove:
Closes issue(s):

- closes #2163

Need for Call for Review:

---

## How to Review And Approve

### DIFF
--- a/_rules/zoom-text-no-overflow-clipping-59br37.md
+++ b/_rules/zoom-text-no-overflow-clipping-59br37.md
@@ -85,7 +85,7 @@ This [text node][] is fully [visible][] at a [viewport size][] of 640 by 512.
 
 #### Passed Example 2
 
-This [text node][] is [horizontally clipped][] using `text-overflow: ellipsis` at a [viewport size][] of 640 by 512. A link to a full version of the poem is also provided.
+This [text node][] is [horizontally clipped][] using `text-overflow: ellipsis` at a [viewport size][] of 640 by 512. A link to a full version of the poem is also provided.  Strictly speaking, the link to a full version is not necessary to pass the rule.  This example passes the rule because its `white-space: nowrap` and `text-overflow: ellipsis` cause it to pass Expectation 1.  The link to a full version _is_ necessary to pass the success criterion.
 
 ```html
 <div style="white-space: nowrap; overflow: hidden; text-overflow: ellipsis; font-size: 16px;">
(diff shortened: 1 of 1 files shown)


_rules/zoom-text-no-overflow-clipping-59br37.md
```diff
@@ -85,7 +85,7 @@ This [text node][] is fully [visible][] at a [viewport size][] of 640 by 512.
 
 #### Passed Example 2
 
-This [text node][] is [horizontally clipped][] using `text-overflow: ellipsis` at a [viewport size][] of 640 by 512. A link to a full version of the poem is also provided.
+This [text node][] is [horizontally clipped][] using `text-overflow: ellipsis` at a [viewport size][] of 640 by 512. A link to a full version of the poem is also provided.  Strictly speaking, the link to a full version is not necessary to pass the rule.  This example passes the rule because its `white-space: nowrap` and `text-overflow: ellipsis` cause it to pass Expectation 1.  The link to a full version _is_ necessary to pass the success criterion.
```
WilcoFiers: This seems a little verbose. How about something like this:

```suggestion
This [text node][] is [horizontally clipped][] using `text-overflow: ellipsis` at a [viewport size][] of 640 by 512. This fulfills expectation 1. A link to a full version of the poem is also provided. The link to the full version is not required by this rule, but is necessary to satisfy [success criterion 1.4.4 Resize text][sc144].```

WilcoFiers (changes requested):
Left a suggestion

WilcoFiers (approved):

Jym77 (commented):
Looks good. Not formally approving to get the "3 organisations" in.

carlosapaduarte (approved):

daniel-montalvo (approved):

Jym77:
Call for review ends on February 27th.

colabottles (approved):

giacomo-petri (approved):

PR #2208, opened 2024-08-01
Add autoplay to glossary

HelenBurge:
A definition of autoplay
Relates to pull request #2064 

Need for Call for Review:

### DIFF
--- a/pages/glossary/autoplay.md
+++ b/pages/glossary/autoplay.md
@@ -5,4 +5,8 @@ unambiguous: true
 objective: false
 ---
 
-When media plays without the user starting it.
+An `audio` or `video` element _autoplays_ if the included media starts playing without the user interacting with the element.
+
+#### Accessibility Support
+
+User agents do not always automatically play media, even when an `autoplay` attribute is present. This is done to avoid autoplaying media interrupting the user when they do not want to, especially when the media is likely to contain sound. The decision to respect the `autoplay` attribute or not depends on user settings and previous behavior (interaction with the site).
--- a/pages/glossary/autoplay.md
+++ b/pages/glossary/autoplay.md
@@ -1,0 +1,8 @@
+---
+title: Autoplay
+key: Autoplay
+unambiguous: true
+objective: false
+---
+
+When media plays without the user starting it.
--- a/pages/glossary/autoplay.md
+++ b/pages/glossary/autoplay.md
@@ -6,7 +6,7 @@ objective: true
 
 ---
 
-An `audio` or `video` element _autoplays_ if the included media starts playing without the user interacting with the element.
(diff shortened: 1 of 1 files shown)


pages/glossary/autoplay.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Autoplay
+key: Autoplay
+unambiguous: true
+objective: false
```
kengdoj: should this be true?

```suggestion
objective: true

```

HelenBurge: I am not sure! @Jym77 - what should I do?

pages/glossary/autoplay.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Autoplay
+key: Autoplay
+unambiguous: true
+objective: false
+---
+
+When media plays without the user starting it.
```
Jym77:
```suggestion
An `audio` or `video` element _autoplays_ if the included media starts playing without the user interacting with the element.

#### Accessibility Support

User agents do not always automatically play media, even when an `autoplay` attribute is present. This is done to avoid autoplaying media interrupting the user when they do not want to, especially when the media is likely to contain sound. The decision to respect the `autoplay` attribute or not depends on user settings and previous behavior (interaction with the site).
```

Jym77 (dismissed):
Frontloading a bit the definition , and copying the accessibility support about browsers not always autoplaying stuff.

pages/glossary/autoplay.md
Jym77: I don't think it is really objective now.
"user starting it" or "interacting with it" is not really objective, I think. Even "play"/"playing" might be open to discussion.

HelenBurge: I copied it from the play button definition!

pages/glossary/autoplay.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Autoplay
+key: Autoplay
+unambiguous: true
+objective: false
+---
+
+An `audio` or `video` element _autoplays_ if the included media starts playing without the user interacting with the element.
```
WilcoFiers: 1. The controls to activate don't need to be part of the video / audio element. A separate play button should also pass
2. We should make sure this doesn't trigger on scroll or navigation, so just "without the user interacting" isn't enough here.


```suggestion
An `audio` or `video` element _autoplays_ if the included media starts playing without the user activating a control.
```

WilcoFiers (dismissed):

kengdoj:
There was discussion that https://html.spec.whatwg.org/#attr-media-autoplay exists so no need to add a glossary term

HelenBurge:
Closing this from Kathy's comment

PR #2218, opened 2024-10-17
Editorial tweaks to object name rule

WilcoFiers:
There is an inconsistency between the background which says that screen readers ignore object without an empty name, and the accessibility support section which says that some don't. 

Need for Call for Review: none, editorial upcate

---

## How to Review And Approve

### DIFF
--- a/_rules/object-has-accessible-name-8fc3b6.md
+++ b/_rules/object-has-accessible-name-8fc3b6.md
@@ -50,11 +50,7 @@ The [MIME type][] of the resource embedded in the `data` attribute impacts how t
 
 Testing that the [accessible name][] describes the purpose of the `object` element is not part of this rule and must be tested separately.
 
-Non-supported media formats make screen readers render the text content of the element instead of other attributes.
-
-`Object` elements without an accessible name are ignored by assistive technologies unless they have an [explicit role][].
-
-When the object resource is not loaded, the fallback content is rendered as shown in the Inapplicable Example: "This `object` element does not need an accessible name because it loads no image, audio, or video."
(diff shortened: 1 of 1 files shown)


dan-tripp-siteimprove (approved):

_rules/object-has-accessible-name-8fc3b6.md
```diff
@@ -50,11 +50,7 @@ The [MIME type][] of the resource embedded in the `data` attribute impacts how t
 
 Testing that the [accessible name][] describes the purpose of the `object` element is not part of this rule and must be tested separately.
 
-Non-supported media formats make screen readers render the text content of the element instead of other attributes.
-
-`Object` elements without an accessible name are ignored by assistive technologies unless they have an [explicit role][].
-
-When the object resource is not loaded, the fallback content is rendered as shown in the Inapplicable Example: "This `object` element does not need an accessible name because it loads no image, audio, or video."
+When the object resource is not loaded, the fallback content is rendered as shown in the Inapplicable Example: "This `object` element does not need an accessible name because it loads no image, audio, or video.". When screen readers encounter an unsupported media format they will also use the fallback content instead of other attributes.
```
carlosapaduarte:
```suggestion
When the object resource is not loaded, the fallback content, if present, is rendered as shown in the Inapplicable Example: "This `object` element does not need an accessible name because it loads no image, audio, or video.". When screen readers encounter an unsupported media format they will also use the fallback content instead of other attributes.
```

carlosapaduarte (approved):
Added an "if present" to clarify the usage of fallback content

carlosapaduarte (approved):
Added an "if present" to clarify the usage of fallback content

carlosapaduarte (approved):
Added an "if present" to clarify the usage of fallback content

Jym77 (approved):

Jym77 (approved):

giacomo-petri (approved):

PR #2225, opened 2024-11-21
Tweak image-description rule

WilcoFiers:
- Link to the definition doesn't work
- The name algorithm changed, update to account for that

---

## How to Review And Approve

### DIFF
--- a/_rules/image-button-non-empty-accessible-name-59796f.md
+++ b/_rules/image-button-non-empty-accessible-name-59796f.md
@@ -54,11 +54,11 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 ## Accessibility Support
 
-The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some user agents and assistive technologies combinations stop at the first existing one, even if empty.
+The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first defined attribute, even if the attribute has an empty value. Some user agents and assistive technologies also ignore empty attributes.
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


_rules/image-button-non-empty-accessible-name-59796f.md
```diff
@@ -54,11 +54,11 @@ Each target element has an [accessible name][] that is neither empty (`""`), nor
 
 ## Accessibility Support
 
-The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first non-empty name, but some user agents and assistive technologies combinations stop at the first existing one, even if empty.
+The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first defined attribute, even if the attribute has an empty value. Some user agents and assistive technologies also ignore empty attributes.
```
dan-tripp-siteimprove:
```suggestion
The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first defined attribute, even if the attribute has an empty value. On the other hand, some user agents and assistive technologies ignore empty attributes.
```

If I'm reading this correctly, the algorithm and the user agents act differently.  So the word "also" was misleading.  So I removed "also" and added "On the other hand".

dan-tripp-siteimprove (dismissed):

_rules/image-button-non-empty-accessible-name-59796f.md
sage-putnam:
```suggestion
The [input type="image" Accessible Name Computation algorithm](https://www.w3.org/TR/html-aam/#input-type-image-accessible-name-computation) uses the first defined attribute, even if the attribute has an empty value. In contrast, some user agents and assistive technologies ignore empty attributes.
```

I believe I'm reading this in the same way you are, @dan-tripp-siteimprove . 

If that is correct, I also believe "also" may not be the best choice, though I disagree with the use of "On the other hand". It may be helpful to avoid idioms, as they may not translate well to readers unfamiliar with English. My suggestion is to use "In contrast," instead!

sage-putnam (changes requested):

sage-putnam (approved):

kengdoj (approved):

dan-tripp-siteimprove (approved):

PR #2229, opened 2024-12-05
Add 2.1.3 to iframe focusable rule

### LINKED ISSUE #2026: "Iframe with interactive elements is not excluded from tab-order" [akn7bn] should have 2.1.3 as secondary requirement
[Iframe with interactive elements is not excluded from tab-order](https://www.w3.org/WAI/standards-guidelines/act/rules/akn7bn/proposed) is mapping to 2.1.1 Keyboard, which makes it an automatic secondary requirement on 2.1.3 Keyboard (No Exception).

WilcoFiers:
Closes https://github.com/act-rules/act-rules.github.io/issues/2026

- This added 2.1.3 Keyboard (No Exceptiosn) (AAA) to this rule.
- This seemed slightly better to me.

**IMPORTANT**: This is going to cause implementors to be inconsistent. That's why I'm using a 2 week review. It is incorrect for this rule not to fail 2.1.3 for the same reason that [Scrollable content can be reached with sequential focus navigation](https://www.w3.org/WAI/standards-guidelines/act/rules/0ssw9k/) fails 2.1.3. The only difference between 2.1.1 and 2.1.3 is an exception for path based input. That's not the case for this iframes or scrollable regions.

---

## How to Review And Approve

### DIFF
--- a/_rules/iframe-with-interactive-content-in-tab-order-akn7bn.md
+++ b/_rules/iframe-with-interactive-content-in-tab-order-akn7bn.md
@@ -4,7 +4,7 @@ name: Iframe with interactive elements is not excluded from tab-order
 rules_format: 1.1
 rule_type: atomic
 description: |
-  This rule checks that `iframe` elements which contain an interactive (tabbable) element are not excluded from sequential focus navigation.
+  This rule checks that `iframe` elements which contain keyboard focusable elements are not excluded from sequential focus navigation.
 accessibility_requirements:
   wcag20:2.1.1: # Keyboard (A)
     forConformance: true
--- a/_rules/iframe-with-interactive-content-in-tab-order-akn7bn.md
+++ b/_rules/iframe-with-interactive-content-in-tab-order-akn7bn.md
@@ -11,6 +11,11 @@ accessibility_requirements:
     failed: not satisfied
     passed: further testing needed
     inapplicable: further testing needed
+  wcag20:2.1.3: # Keyboard (No exceptions) (AAA)
+    forConformance: true
+    failed: not satisfied
+    passed: further testing needed
+    inapplicable: further testing needed
   wcag-technique:G202: # Ensuring keyboard control for all functionality
     forConformance: false
     failed: not satisfied


Jym77 (dismissed):
Isn't that a case where Scenario 1 (the rule is not as strict as a requirement) of [Secondary requirements](https://w3c.github.io/wcag-act/act-rules-format.html#secondary-requirements) applies?

That is, anything that fails 2.1.1 also automatically fails 2.1.3, similarly to the 1.4.3/1.4.6 situation. 

Additionally, 2.1.3 is a AAA criterion, so having this as a primary requirements means that tools that only check for level AA will never be able to have a consistent implementation of the rule (which was pretty much what led us to introduce secondary requirements for the 1.4.3/1.4.6 or 2.4.6/2.4.9 cases: [initial PR](https://github.com/act-rules/act-rules.github.io/pull/1920), linking to more issues in the wcag-act-rules repo)

WilcoFiers:
@Jym77 
1. While yes, criterion 2.1.3 is stricter then criterion 2.1.1, 2.1.3 isn't stricter then this rule. There are no passed examples that fail 2.1.3 but don't fail 2.1.1 like is the case for text contrast. That's essentially the difference between a conformance requirement and a secondary requirement. A secondary requirement may fail passed exampled (stricter) or it may pass failed examples (less strict), or in a few cases it can do both. When all failed examples fail a criterion, and all passed / inapplicable examples meet the requirement it's a conformance requirement, not a secondary requirement.

2. ACT allows implementations to be consistent at lower levels. I.e. you're only required to report AAA criteria if you test for AAA. The way we check that is by checking if any AAA criteria are reported. So you can't do AAA for some rules but not for others. There is an open bug in how this works though, so this isn't working correctly everywhere.

Jym77 (approved):

carlosapaduarte (approved):

giacomo-petri (approved):

shunguoy:
@WilcoFiers  another [issue related to this rule](https://github.com/act-rules/act-rules.github.io/issues/2320).

WilcoFiers:
I forgot to put the call for review label on this, but that e-mail went out on April 10th. That's over two weeks ago, so I'm merging this PR.

Issue #2256, opened 2025-02-01
[Editorial] - [5c01ea] ARIA state or property is permitted - Inapplicable Example 1


giacomo-petri:
In [ARIA state or property is permitted](https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea/proposed/), Inapplicable Example 1

`<div role="region">A region of content</div>`

does not fully adhere to the [ARIA specifications](https://www.w3.org/TR/wai-aria-1.3/#region), which require an accessible name for the region.

I’m opening a PR to add the necessary accessible name.

giacomo-petri:
PR already merged: #2257

Closing

PR #2257, opened 2025-02-01
[Editorial] - [5c01ea] ARIA state or property is permitted - Inapplicable Example 1 improvement (acc name)

giacomo-petri:
Closes issue(s): https://github.com/act-rules/act-rules.github.io/issues/2256

### Description:
This PR adds the required acc name to the element with explicit role="region"
from
`<div role="region">A region of content</div>`

to
`<div role="region" aria-label="Region name">A region of content</div>`

### Need for Call for Review:
This can be merged with 1 approval

### DIFF
--- a/_rules/aria-state-or-property-permitted-5c01ea.md
+++ b/_rules/aria-state-or-property-permitted-5c01ea.md
@@ -207,7 +207,7 @@ The `aria-label` property is [prohibited][] for an element with a `generic` role
 This `div` element has no [WAI-ARIA state or property][].
 
 ```html
-<div role="region" aria-label="Region name">A region of content</div>
+<div role="group">A group of content</div>
 ```
 
 #### Inapplicable Example 2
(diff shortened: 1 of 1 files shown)


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -207,7 +207,7 @@ The `aria-label` property is [prohibited][] for an element with a `generic` role
 This `div` element has no [WAI-ARIA state or property][].
 
 ```html
-<div role="region">A region of content</div>
+<div role="region" aria-label="Region name">A region of content</div>
```
Jym77: This makes the example Applicable (by having an `aria-*` attribute).

We should instead change the role into one that doesn't need any ARIA attribute.

Jym77 (changes requested):

_rules/aria-state-or-property-permitted-5c01ea.md
giacomo-petri: replaced with role="group", which doesn't require any aria state/property

carlosapaduarte (approved):

Jym77 (approved):

dan-tripp-siteimprove (approved):

WilcoFiers:
This should not be an editorial update. It changes an example, we'll always want a CfR for that.

WilcoFiers (approved):

Jym77:
Call for review end on February 28th.

colabottles (approved):

PR #2279, opened 2025-02-06
"Target Size Enhanced" - atomic rule

Jym77:
New take at "Target Size" rule, as an automated rule with definitions, following [CG decision](https://www.w3.org/2024/10/17-act-r-minutes.html#item04).

For first pass, please focus reviews on the overall structure rather than the details of each definition 🙂 

Closes issue(s):

- closes #2214 
- closes #1919 (supersedes it)
- Partly supersedes #2167 (missing the "minimum" rule) 

Need for Call for Review:
This will require a 2 weeks Call for Review (new rule)

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

- [X] Close any issue that the PR resolves. This will happen automatically upon merging if the PR was correctly linked to the issue, e.g. by referencing the issue at the top of this comment.

## How to Review And Approve


_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,819 @@
+---
+id: gi8qkf
```
Jym77, line 2: For first pass of review, please focus on global structure to se whether this is a good structure for the rule. We'll discuss details of the definitions later.

pages/glossary/can-be-targeted-by-pointer-event.md
```diff
@@ -0,0 +1,36 @@
...
+  - CSS styling
+  - DOM tree
+---
+
+An element _can be targeted by pointer events_ when all the following conditions are true:
+
+- the element is a [semantic `widget`][semantic role]; and
+- the element is [focusable][]; and
+- the element [computed value][] of the `pointer-events` CSS property is `auto`; and
+- through scrolling, it is possible to have parts of the element's [border box][] which intersect the viewport and are not entirely covered by the [border boxes][] of elements with greater [computed][computed value] `z-index` and a [computed][computed value] `pointer-events` of `auto`.
```
carlosapaduarte:
```suggestion
- through scrolling, it is possible to have parts of the element's [border box][] which intersect the viewport and do not overlap [border boxes][] of elements with greater [computed][computed value] `z-index` and a [computed][computed value] `pointer-events` of `auto`.
```

Perhaps it's more objective?

dan-tripp-siteimprove (dismissed):
The general structure looks good to me.  I am interested in the definition of "inline".  I look forward to that part.

pages/glossary/can-be-targeted-by-pointer-event.md
```diff
@@ -0,0 +1,36 @@
...
+
+#### Background
+
+This definition tries to capture which HTML elements can actually react to pointer events. It is not possible to have an exact definition of these for two main reasons:
+
+- Sometimes, the element that handles the event is not the element that appear to react to it, but an ancestor (or descendant) capturing the event during propagation or bubbling. In the most extreme case, the `body` element of a page could be the only one with an event handler, acting differently depending on where the event actually occurred. In such a case, a button would be perceived by users as something that can be targeted by a pointer event, while technically it is the `body` element which is targeted.
```
carlosapaduarte: Reading this (and the next) reason, I was wondering if we should name this definition as "is perceived as a pointer event target"?

pages/glossary/clickable-area.md
```diff
@@ -0,0 +1,51 @@
...
+title: Clickable area
+key: clickable-area
+unambiguous: true
+objective: false
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+The _directly clickable area_ of an element is the set of all viewport coordinates for which the element is the [topmost event target][]
```
carlosapaduarte:
```suggestion
The <dfn id="clickable-area:directly">_directly clickable area_</dfn> of an element is the set of all viewport coordinates for which the element is the [topmost event target][]
```

We might need this (it is used in the next sentence).

pages/glossary/clickable-area.md
```diff
@@ -0,0 +1,51 @@
...
+
+The _clickable area_ of an element is the union of its _directly clickable area_ and that of its [implicit][implicit label] or [explicit label][]. Clickable areas may contain several disconnected parts.
+
+An element has a <dfn id="clickable-area:empty">totally empty clickable area</dfn> if its clickable area is empty and cannot be made non-empty through scrolling.
+
+#### Examples of totally empty clickable areas
+
+This button has a totally empty clickable area because it is moved off-screen.
```
carlosapaduarte:
```suggestion
This button has a totally empty clickable area because it is moved off-screen and cannot be brought into the viewport through scrolling.
```

pages/glossary/clickable-area.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Clickable area
+key: clickable-area
+unambiguous: true
+objective: false
```
carlosapaduarte, line 5: Why is this definition not objective?

pages/glossary/rendered-on-a-line.md
```diff
@@ -0,0 +1,29 @@
...
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+> Comment: I do not believe this definition correctly captures the "inline" exception of 2.5.5/2.5.8 currently. Please disregard this for now and review the global structure.
+
+All nodes that generate an [inline box][] that are part of a [line box][] are considered _rendered on a line_. These are _line rendering nodes_.
+
+Any [flat tree][] [descendants][] of line rendering nodes is also considered part of a line if one of its [clientRects][] overlaps with the position of the [line box][], and if all the following is true for each [ancestor][] in the [flat tree][] up to the line rendering node:
```
carlosapaduarte:
```suggestion
Any [flat tree][] [descendant][] of line rendering nodes is also considered part of a line if one of its [clientRects][] overlaps with the position of the [line box][], and if all the following is true for each [ancestor][] in the [flat tree][] up to the line rendering node:
```

pages/glossary/rendered-on-a-line.md
```diff
@@ -0,0 +1,29 @@
...
+
+- the ancestor is an [inline box][]; and
+- the ancestor is not [floating][]; and
+- the ancestor does not have an [absolute][] or [fixed][] `position`.
+
+[absolute]: https://drafts.csswg.org/css-position-3/#valdef-position-absolute 'Definition of absolute'
+[ancestor]: https://dom.spec.whatwg.org/#concept-tree-ancestor 'Definition of ancestor'
+[clientrects]: https://drafts.csswg.org/cssom-view/#dom-element-getclientrects 'Definition of getClientRects'
+[descendants]: https://dom.spec.whatwg.org/#concept-tree-descendant 'Definition of descendant'
```
carlosapaduarte:
```suggestion
[descendant]: https://dom.spec.whatwg.org/#concept-tree-descendant 'Definition of descendant'
```

pages/glossary/user-agent-controlled-component.md
```diff
@@ -0,0 +1,31 @@
...
+- the [computed values][] of the element's `height` and `width` CSS properties do not depend on content provided by the author; and
+- the [computed values][] of the element's `height` and `width` CSS properties do not depend on any CSS property with a [cascaded value][] with "Author" [origin][].
+
+#### Examples of User-Agent Controlled Components
+
+Typically, radio buttons or checkboxes are User-Agent controlled, until an author does change their dimensions.
+
+Links and buttons usually aren't, because their (text) content is provided by the author and the width depends on it.
+
+Days in a calendar widget build with an `<input type="date">` element are also User-Agent controlled since their content isn't provided by the author and their dimension do not depend on values provided by the author.
```
carlosapaduarte:
```suggestion
Days in a calendar widget build with an `<input type="date">` element are also User-Agent controlled since their content isn't provided by the author and their dimension does not depend on values provided by the author.
```

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,819 @@
...
+
+## Background
+
+While the rule, and [Success Criterion 2.5.5 Target Size (enhanced)][sc255], apply targets of any shape, the test cases mostly focus on targets whose [clickable area][] is itself an [horizontal rectangle][]. This acknowledges the fact that the [border box][] of an element can easily be queried by automated tools (e.g., through the `getBoundingClientRect` function), and therefore it is expected that most automated tools will perform better on such elements. For elements with "weird" clickable shape, including `area` elements, nested targets, or elements that have been rotated or clipped, the actual [clickable area][] is much harder to determine and may be much smaller than the [border box][]. These elements could fail the rule while their [border box][] contain a large enough [horizontal rectangle][]. In order to allow automated tools to have a consistent implementation of this rule, it does not contain such test cases, notably all Failed test cases have a [border box][] which is too small.
```
carlosapaduarte:
```suggestion
While the rule, and [Success Criterion 2.5.5 Target Size (enhanced)][sc255], apply targets of any shape, the test cases mostly focus on targets whose [clickable area][] is itself an [horizontal rectangle][]. This acknowledges the fact that the [border box][] of an element can easily be queried by automated tools (e.g., through the `getBoundingClientRect` function), and therefore it is expected that most automated tools will perform better on such elements. For elements with irregular clickable shapes, including `area` elements, nested targets, or elements that have been rotated or clipped, the actual [clickable area][] is much harder to determine and may be much smaller than the [border box][]. These elements could fail the rule while their [border box][] contains a large enough [horizontal rectangle][]. In order to allow automated tools to have a consistent implementation of this rule, it does not contain such test cases, notably all Failed test cases have a [border box][] which is too small.
```

pages/glossary/horizontal-rectangle.md
```diff
@@ -0,0 +1,13 @@
...
+unambiguous: true
+objective: true
+input_aspects:
+---
+
+A set of coordinates R is an _horizontal rectangle_ if there exist a coordinate _(x, y)_ and numbers _w_ and _h_ such that all coordinates _(a, b)_ with _x ⩽ a ⩽ x+w_ and _y ⩽ b ⩽ y+h_ are part of R, and reciprocally all coordinates _(a, b)_ in R verify _x ⩽ a ⩽ x+w_ and _y ⩽ b ⩽ y+h_.
+
+R is then said to have corner _(x, y)_, width _w_, and height _h_.
+
+Note that the rectangle is "horizontal" in the sense that its sides align with the axis of the coordinate system.
```
carlosapaduarte: Do we need to have "horizontal" as part of the name of the definition? 
I understand that if people read the definition they will understand that it's not horizontal in the sense that w is bigger than h, but I expect that people will still think that is the case (and might not bother to read the definition).
What about "aligned rectangle"? Or "axis aligned rectangle"?

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,819 @@
...
+---
+
+## Applicability
+
+This rule applies to any [HTML element][namespaced element] which [can be targeted by a pointer event][], except when one of the following is true:
+
+- the element has a [totally empty clickable area][]; or
+- the element is [rendered on a line]; or
+- the element is [User Agent controlled][user agent controlled component]; or
+- the element has [essential size][].
```
carlosapaduarte: Just wondering if these should be in the expectation?

pages/glossary/programmatic-label.md
```diff
@@ -16,5 +16,9 @@ For more details, see [examples of programmatic label][].
 
 **Note**: a given element may have more than one programmatic label.
 
+The [labeled control][] L of an element T is its <dfn id="programmatic-label:explicit">explicit label</dfn> if L has a `for` attribute referencing T's `id` attribute.
```
carlosapaduarte: This is confusing. According to the definition (both in the spec and here), T is the labeled control, but in the explicit and implicit labels you are referring to L as the labeled control. 

```suggestion
The label L of a [labeled control][] T is its <dfn id="programmatic-label:explicit">explicit label</dfn> if L has a `for` attribute referencing T's `id` attribute.
```

pages/glossary/programmatic-label.md
```diff
@@ -16,5 +16,9 @@ For more details, see [examples of programmatic label][].
...
+
+The [labeled control][] L of an element T is its <dfn id="programmatic-label:implicit">implicit label</dfn> if L has a no `for` attribute and is an ancestor of T in the DOM tree. Note that explicit labels take precedence over implicit label, and that [labeled controls][labeled control] do not cross shadow boundaries nor content documents.
```
carlosapaduarte:
```suggestion
The label L of [labeled control][] T is its <dfn id="programmatic-label:implicit">implicit label</dfn> if L has no `for` attribute and is an ancestor of T in the DOM tree. Note that explicit labels take precedence over implicit label, and that [labeled controls][labeled control] do not cross shadow boundaries nor content documents.
```

_rules/target-size-enhanced-gi8qkf.md
```diff
@@ -0,0 +1,819 @@
...
+	</style>
+</head>
+<body>
+	<div id="target" role="button" onclick="alert('Hello')">
+		Hi
+	</div>
+</body>
+```
+
+### Inapplicable
```
carlosapaduarte, line 710: Shouldn't we have one with a clickable area that is not a semantic widget?

carlosapaduarte (dismissed):
Great work so far. Looking forward to moving this one forward.

pages/examples/programmatic-label.md
```diff
@@ -56,13 +56,13 @@ The `div` element is not [labelable][]. Therefore, it is not a [labeled control]
 <div id="bond">My name is Bond. James Bond.</div>
 ```
 
-## aria-labelledby on div
+## aria-labelledby on aside
 
-The `span` element is referenced by the `aria-labelledby` attribute on the `div` element. Therefore, the `span` element is a programmatic label of the `div` element. Note that the `aria-labelledby` attribute works on any element, not just on the [labelable][] elements.
+The `span` element is referenced by the `aria-labelledby` attribute on the `aside` element. Therefore, the `span` element is a programmatic label of the `div` element. Note that the `aria-labelledby` attribute works on any element, not just on the [labelable][] elements.
```
iadawn:
```suggestion
The `span` element is referenced by the `aria-labelledby` attribute on the `aside` element. Therefore, the `span` element is a programmatic label of the `aside` element. Note that the `aria-labelledby` attribute works on any element, not just on the [labelable][] elements.
```

pages/glossary/can-be-targeted-by-pointer-event.md
```diff
@@ -0,0 +1,36 @@
...
+An element _can be targeted by pointer events_ when all the following conditions are true:
+
+- the element is a [semantic `widget`][semantic role]; and
+- the element is [focusable][]; and
+- the element [computed value][] of the `pointer-events` CSS property is `auto`; and
+- through scrolling, it is possible to have parts of the element's [border box][] which intersect the viewport and are not entirely covered by the [border boxes][] of elements with greater [computed][computed value] `z-index` and a [computed][computed value] `pointer-events` of `auto`.
+
+#### Background
+
+This definition tries to capture which HTML elements can actually react to pointer events. It is not possible to have an exact definition of these for two main reasons:
```
iadawn:
```suggestion
This definition tries to capture which HTML elements can actually respond to pointer events. It is not possible to have an exact definition of these for two reasons:
```

pages/glossary/can-be-targeted-by-pointer-event.md
iadawn:
```suggestion
- Sometimes, the element that handles the event is not the element that appears to react to it, but an ancestor (or descendant) capturing the event during propagation or bubbling. In the most extreme case, the `body` element of a page could be the only one with an event handler, acting differently depending on where the event actually occurred. In such a case, a button would be perceived by users as something that can be targeted by a pointer event, while technically it is the `body` element which is targeted.
```

pages/glossary/can-be-targeted-by-pointer-event.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Sometimes, the element that handles the event is not the element that appear to react to it, but an ancestor (or descendant) capturing the event during propagation or bubbling. In the most extreme case, the `body` element of a page could be the only one with an event handler, acting differently depending on where the event actually occurred. In such a case, a button would be perceived by users as something that can be targeted by a pointer event, while technically it is the `body` element which is targeted.
+- It is not possible to query the list of event listeners on a given elements. Some User Agents offer way to monitor events fired at a given element, but none offer a way to query for event listeners. Additionally, an event listener might ultimately do nothing and thus, for users, the corresponding element wouldn't look like it can be targeted by pointer events (since it effectively wouldn't react to them).
```
iadawn:
```suggestion
- It is not possible to query the list of event listeners on a given element. Some User Agents offer ways to monitor events fired at a given element, but none offer a way to query for event listeners. Additionally, an event listener might ultimately do nothing and thus, for users, the corresponding element wouldn't look like it could be targeted by pointer events (since it effectively wouldn't react to them).
```

pages/glossary/clickable-area.md
```diff
@@ -0,0 +1,51 @@
...
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+The _directly clickable area_ of an element is the set of all viewport coordinates for which the element is the [topmost event target][]
+
+The _clickable area_ of an element is the union of its _directly clickable area_ and that of its [implicit][implicit label] or [explicit label][]. Clickable areas may contain several disconnected parts.
+
+An element has a <dfn id="clickable-area:empty">totally empty clickable area</dfn> if its clickable area is empty and cannot be made non-empty through scrolling.
```
iadawn: 'totally' may be unnecessary in this definition.... unless there is a definition for 'empty clickable area' that was slightly distinct from 'totally empty clickable area'

pages/glossary/clickable-area.md
iadawn:
```suggestion
This `button` has a totally empty clickable area because it is moved off-screen.
```

Making this consistent with the use of button in the next example.

pages/glossary/horizontal-rectangle.md
iadawn: Minor but it isn't specified in the definition that _w > h_.

+1 to a change of name to better clarify the key features of the rectangle.

pages/glossary/user-agent-controlled-component.md
```diff
@@ -0,0 +1,31 @@
...
+
+A _User-Agent Controlled Component_ is an [HTML element][namespace element] for which all the following are true:
+
+- the element has an [implicit role][] which is a [semantic `widget`][semantic role]; and
+- the [computed values][] of the element's `height` and `width` CSS properties do not depend on content provided by the author; and
+- the [computed values][] of the element's `height` and `width` CSS properties do not depend on any CSS property with a [cascaded value][] with "Author" [origin][].
+
+#### Examples of User-Agent Controlled Components
+
+Typically, radio buttons or checkboxes are User-Agent controlled, until an author does change their dimensions.
```
iadawn:
```suggestion
Typically, radio buttons or checkboxes are User-Agent controlled, until an author changes their dimensions.
```

pages/glossary/user-agent-controlled-component.md
```diff
@@ -0,0 +1,31 @@
...
+
+Links and buttons usually aren't, because their (text) content is provided by the author and the width depends on it.
```
iadawn:
```suggestion
Links and buttons usually aren't. Authors provide their (text) content which determines the width.
```

pages/glossary/user-agent-controlled-component.md
iadawn:
```suggestion
Days in a calendar widget build with an `<input type="date">` element are also User-Agent controlled. Their content is not provided by the author and dimensions do not depend on values provided by the author.
```

_rules/target-size-enhanced-gi8qkf.md
Jym77: The full point of #2214 and the CG discussion was to have them as Inapplicable 🤷

_rules/target-size-enhanced-gi8qkf.md
Jym77: 🤔 Wouldn't that be bad for 1.3.1?
a `<span onclick="alert('Hello')">Say hello</span>` would fail 1.3.1. We can have one (with associated caveats), but I'm afraid it would cause problems for tools that do not have the same separation as ACT rules and would (rightfully) fail it.

pages/glossary/can-be-targeted-by-pointer-event.md
Jym77: I see the problem. I'm not sure this formulation is much better since the point it also about `z-index` and the like (if the element is totally covered by others, it is not clickable).

pages/glossary/clickable-area.md
Jym77: The idea here is "not just empty right now, but stays empty no matter how you scroll". An element at the bottom of a page would have an empty clickable area until scrolled into view, so it would jot be "totally empty".

I agree the name is not really conveying that, though 🤔

pages/glossary/horizontal-rectangle.md
Jym77: Agreed with "aligned", this is the key difference in this definition (vs any rotated rectangle).

pages/glossary/can-be-targeted-by-pointer-event.md
Jym77: I've rephrased it in terms of intersections. Also referring directly to CSS "panting order" definition since just looking at the z-index is far from enough 🙈

pages/glossary/can-be-targeted-by-pointer-event.md
Jym77: I went for "Observed as a pointer events target". I felt "perceived" was having a too strong meaning, especially in our contexts.

pages/glossary/clickable-area.md
Jym77: Because I feel that "topmost event target" is not really objective.

pages/glossary/clickable-area.md
Jym77: Replaced with "always empty"

pages/glossary/observed-as-pointer-events-target.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Observed as a pointer events target
```
dan-tripp-siteimprove, line 2: Could the words "observed as" be cut out?

I know you added those words recently, so I assume you see considerable value in them.  Still I must ask this, because when I first read those words, it made me wonder: is there another term called "pointer events target" that is different from "_observed as_ a pointer events target"?  If so, what does the venn diagram of these two sets look like?  After I read the background and searched elsewhere for a minute, I arrived at the answers "no, or at least: not officially" and "n/a".  If I'm the only reader that will go through that thought process: then it doesn't matter.  If I'm not: then we have a cognitive load question here for future readers.  

The background is essential commentary, but I'm not sure that it's essential to have it in the wording of the term itself.  I figure most conversations about target size will just talk about "pointer event targets" and won't need to get into the details.

To summarize: the value of having "observed as" in the wording of the term itself is, IMO, outweighed by the cost of the added cognitive load.

pages/glossary/in-a-block-of-text.md
```diff
@@ -0,0 +1,19 @@
+---
+title: In a Block of Text
```
dan-tripp-siteimprove, line 2: I was going to look at the examples of this PR via [the new "deploy preview" functionality, which is here](https://deploy-preview-2279--act-rules.netlify.app/standards-guidelines/act/rules/gi8qkf/proposed/).  When I clicked on the "Open in a new tab" links for the examples, they didn't work, because they're pointing to w3.org rather than deploy-preview-2279--act-rules.netlify.app.  Is this a known limitation?  If so, can I help fix it?

dan-tripp-siteimprove (changes requested):

HelenBurge:
Please assign yourself if able to work on this PR as a complex one!

pages/glossary/observed-as-pointer-events-target.md
```diff
@@ -0,0 +1,37 @@
...
+unambiguous: true
+objective: true
+input_aspects:
+  - CSS styling
+  - DOM tree
+---
+
+An element is _observed as a pointer events target_ when all the following conditions are true:
+
+- the element is a [semantic `widget`][semantic role]; and
```
Jym77, line 13: **Question:** Some "composite widgets" are actually not the real targets and should probably be ignored by this definition: "menu", "menubar", "tablist", "radiogroup", "listbox", "tree", "grid", "treegrid". 

We need to look a bit more into that list to agree on it.

dan-tripp-siteimprove: I figure: let's start with the [ARIA spec for "widget roles"](https://www.w3.org/TR/wai-aria-1.3/#widget_roles) where it gives us two lists:
Let list1 = "The following roles act as standalone user interface widgets or as part of larger, composite widgets".
Let list2 = "The following roles act as composite user interface widgets".

AFAICS the "clickable roles" are: list1 minus progressbar minus tabpanel plus listbox plus combobox.

minus progressbar because it's not clickable.

minus tabpanel because it's not clickable.

plus listbox because I figure the listbox is the first thing that the user clicks, and it could be smaller than the options which it contains.  the "option" role (= child of listbox ~= HTML `<option>`) is already in list1.

plus combobox because in [this APG combobox example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/) the first / "outer" clickable thing is the element which has role="combobox".  The other four APG combobox examples: same.  I consulted the APG rather than the ARIA spec for this, because spec-wise, combobox is over my head.  But I understand the APG examples well enough.

And I figure this rule is obliged to follow the ARIA spec wording that "separator (when focusable)" (that is: _only_ when focusable) is included in list1.

I don't know what to do about `<div tabindex="0" onclick="alert('yes')">x</div>`.

Jym77: > I don't know what to do about `<div tabindex="0" onclick="alert('yes')">x</div>`.

I think this is handled by the Background of that definition:
> Not all HTML elements that can actually be targeted by a pointer event match this definition. For example, an author may build custom buttons without giving them an appropriate role or making them [focusable][]; or content overflowing the [border box][] of an element is clickable, but that element might not match the definition if the [border box][] is off-screen. Elements that can actually be targeted by pointer events but do not match this definition likely fail either [Success Criterion 4.1.2 Name, Role, Value][sc412] or [Success Criterion 2.1.1 Keyboard][sc211].

dan-tripp-siteimprove: This will make a great boundary/verbose example one day.

2biazdk, line 13: Small wording question: does “semantic `widget`” here mean a semantic role that inherits from ARIA’s abstract `widget` role? If so, could we say that explicitly?

Read alongside the Semantic Role glossary, it looks a little like the role has to be exactly `widget`. Since `widget` itself is abstract, I just want to make sure roles such as `button` and `link` are clearly covered.

2biazdk: I think the original list is a good fit:

- exclude `menu`, `menubar`, `tablist`, `radiogroup`, `listbox`, `tree`, `grid`, and `treegrid`. In the usual patterns, the pointer target is one of the children (`menuitem*`, `tab`, `radio`, `option`, `treeitem`, or `gridcell`), not the container itself.
- keep `combobox`. The combobox is normally the input or trigger that people click.

I would treat this as a sensible heuristic, rather than trying to detect every custom click handler. In that light, I am happy to err on the side of not testing a container unless people normally interact with it directly.

dan-tripp-siteimprove: Yes, and I agree.

dan-tripp-siteimprove: @2biazdk I can't tell whether you just agreed with me or disagreed.  

When I wrote this: 

> AFAICS the "clickable roles" are: list1 minus progressbar minus tabpanel plus listbox plus combobox.

... that evaluates to this list of roles:

button
checkbox
gridcell
link
menuitem
menuitemcheckbox
menuitemradio
option
radio
scrollbar
searchbox
separator (when focusable)
slider
spinbutton
switch
tab
textbox
treeitem
listbox
combobox


When you wrote "the original list is a good fit" I don't know what "the original list" is.

2biazdk: @dan-tripp-siteimprove Sorry – I think you're right about this list. :-)

PR #2286, opened 2025-02-07
button name move assumption to the correct section

WilcoFiers:
This is an assumption, not an accessibility support issue.

---

## How to Review And Approve

### DIFF
--- a/_rules/button-non-empty-accessible-name-97a4e1.md
+++ b/_rules/button-non-empty-accessible-name-97a4e1.md
@@ -39,11 +39,11 @@ This rule considers an exception for "image buttons" (i.e., `input` elements wit
 
 - The rule assumes that all buttons are [user interface components as defined by WCAG 2](https://www.w3.org/TR/WCAG22/#dfn-user-interface-components).
 
+- Some elements have a role of `button` and a default accessible name defined by the [HTML Accessibility API Mapping][html aam input button], for example `input` elements whose `type` [attribute value][] is either `submit` or `reset`. This rule considers that these default names can be descriptive and therefore does not fail them.
+
 ### Accessibility Support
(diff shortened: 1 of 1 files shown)


Jym77 (approved):

giacomo-petri (approved):

zlayaAvocado (approved):

_rules/button-non-empty-accessible-name-97a4e1.md
```diff
@@ -39,12 +39,12 @@ This rule considers an exception for "image buttons" (i.e., `input` elements wit
 
 - The rule assumes that all buttons are [user interface components as defined by WCAG 2](https://www.w3.org/TR/WCAG22/#dfn-user-interface-components).
 
+- Some elements have a role of `button` and a default accessible name defined by the [HTML Accessibility API Mapping][html aam input button], for example `input` elements whose `type` [attribute value][] is either `submit` or `reset`. This rule considers that these default names can be descriptive and therefore does not fail them.
```
zlayaAvocado, line 42: Should the HTML Accessibility API Mapping link be updated? It seems it's currently not referencing anything and goes to the whole file instead, new link: https://www.w3.org/TR/html-aam-1.0/#input-type-button-input-type-submit-and-input-type-reset-accessible-name-computation

The same issue is in the "Bibliography" section of the rule

zlayaAvocado (changes requested):

_rules/button-non-empty-accessible-name-97a4e1.md
WilcoFiers: Good catch

zlayaAvocado (approved):

PR #2304, opened 2025-02-11
summary name, fix markdown error

WilcoFiers:
There is no "explicit semantic role" link. Should be "explicit role" as seen in other places in the rule.

---

## How to Review And Approve


dan-tripp-siteimprove (approved):

Jym77 (commented):
Good. Not approving since Dan already did.

kengdoj (approved):

giacomo-petri (approved):

PR #2305, opened 2025-02-11
Updated required owned elms rule description

WilcoFiers:
When reviewing this rule I noticed the description is out of sync with what it actually requires. 

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-owned-element-bc4a75.md
+++ b/_rules/aria-required-owned-element-bc4a75.md
@@ -4,7 +4,7 @@ name: ARIA required owned elements
 rules_format: 1.1
 rule_type: atomic
 description: |
-  This rule checks that an element with an explicit role that restricts which elements it can own only owns such elements
+  This rule checks that an element with an explicit role that restricts which elements it can own only owns such elements.
 accessibility_requirements:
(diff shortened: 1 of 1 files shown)


_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -4,7 +4,7 @@ name: ARIA required owned elements
 rules_format: 1.1
 rule_type: atomic
 description: |
-  This rule checks that an element with an explicit semantic role has at least one of its required owned elements.
+  This rule checks that an element with an explicit role that restricts which elements it can own only owns such elements
```
dan-tripp-siteimprove:
```suggestion
  This rule checks that an element with an explicit role that restricts which elements it can own only owns such elements.
```

Nitpick.  Adding a period.  Most (approximately 82 out of 93) rules' descriptions have a period like this.

dan-tripp-siteimprove (approved):

Jym77 (commented):
Looks good. Not approving since Dan already did.

kengdoj (approved):

giacomo-petri (approved):

Issue #2268, opened 2025-02-03
[a25f45] Headers attribute specified on a cell refers to cells in the same table element - Inapplicable Example 6 ambiguous


giacomo-petri:
In [Headers attribute specified on a cell refers to cells in the same table element](https://www.w3.org/WAI/standards-guidelines/act/rules/a25f45/proposed/), Inapplicable Example 6 is ambiguous because role="region" requires an accessible name. Without one, browsers treat it as a generic element.

In Inapplicable Example 6

```
<table role="region">
	<td id="self" headers="self">World</td>
</table>
```

All browsers expose this element as a table:
- Chrome exposes it as a table.
- Safari exposes it as a table.
- Firefox exposes it as a table, but treats its children as generic elements!?

Since this example introduces ambiguity, IMO it should be adjusted. We can either add an aria-label to ensure role="region" is correctly applied or replace it with a role that does not require an accessible name.

giacomo-petri:
During the ACT meeting the group agreed to add an acc name.
Proceeding with creating a PR.

giacomo-petri:
Corresponding PR already merged #2312

Closing

PR #2312, opened 2025-03-06
[a25f45] Headers attribute specified on a cell refers to cells in the same table element - Updated Inapplicable Example 6

giacomo-petri:
Closes issue: #2268 

Role region without acc name is ignored.
Added an aria-label attribute to the `table` element with role `region` so that the table is exposed as a region rather than a table (that is currently invalidating the example, by making it applicable).

Need for Call for Review:

---

## How to Review And Approve

### DIFF
--- a/_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
+++ b/_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
@@ -405,7 +405,7 @@ The `table` is not [included in the accessibility tree][].
 This `table` doesn't have a role of `table`, `grid` or `treegrid`.
 
 ```html
-<table role="region">
+<table role="region" aria-label="Hello">
 	<td id="self" headers="self">World</td>
 </table>
 ```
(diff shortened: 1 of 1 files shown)


HelenBurge (dismissed):
Please add https://www.w3.org/TR/wai-aria-1.2/#region in the PR details

daniel-montalvo (dismissed):
Sounds reasonable

dan-tripp-siteimprove (dismissed):

_rules/table-headers-attribute-refer-to-data-cells-a25f45.md
```diff
@@ -405,7 +405,7 @@ The `table` is not [included in the accessibility tree][].
 This `table` doesn't have a role of `table`, `grid` or `treegrid`.
 
 ```html
-<table role="region">
+<table role="region" aria-label="Hello">
```
WilcoFiers: I wonder if this isn't a slightly clearer example:

```suggestion
<table role="heading" aria-level="1">
```

giacomo-petri: I'm okay with this change. We already have three approvals. If we apply this change, does it mean we need to request a review again? @WilcoFiers

daniel-montalvo (approved):

carlosapaduarte (approved):

WilcoFiers (approved):

giacomo-petri:
Call for review ends on April 18th.

giacomo-petri:
This is ready to be merged

Issue #2310, opened 2025-02-17
[4e8ab6] Element with role attribute has required states and properties - Combobox Failed Example 6


giacomo-petri:
In rule [Element with role attribute has required states and properties](https://www.w3.org/WAI/standards-guidelines/act/rules/4e8ab6/proposed/), combobox Failed Example 6 requires the aria-controls attribute. 

In [ARIA 1.2 specs](https://www.w3.org/TR/wai-aria-1.2/#combobox) the aria-controls attribute is required, but since aria-controls has no support at all, it has been removed in [ARIA 1.3](https://www.w3.org/TR/wai-aria-1.3/). 

Although this technically violates ARIA 1.2 (until version 1.3 is officially released), requiring support for a property that no UA/AT currently supports seems like overkill for the author to me.

I suggest switching aria-expanded and aria-controls so that the example remains meaningful, aligns with ARIA 1.3 when it's released, and requires no further changes.

giacomo-petri:
I just realised that Failed Example 5 is covering the missing aria-expanded scenario, so the suggestion doesn't work and probably we should simply remove the example.

carlosapaduarte:
CG decision: remove the example

WilcoFiers:
Here's the ARIA ticket that removed the aria-controls requirement from combobox:
https://github.com/w3c/aria/issues/2132

giacomo-petri:
Closed with #2313

PR #2313, opened 2025-03-06
[4e8ab6] Element with role attribute has required states and properties - Removed Combobox Failed Example 6

giacomo-petri:
Closes issue: #2310

ARIA has removed the aria-controls requirement from combobox, so failed example 6 has been removed. 

Need for Call for Review:

---

## How to Review And Approve

### DIFF
--- a/_rules/role-required-states-and-properties-4e8ab6.md
+++ b/_rules/role-required-states-and-properties-4e8ab6.md
@@ -181,19 +181,6 @@ This `combobox` does not have the required `aria-expanded` property. Prior to [W
 </ul>
 ```
 
-#### Failed Example 6
-
-This `combobox` uses `aria-owns` instead of using the required `aria-controls` property.
-
-```html
-<label for="tag_combo">Tag</label>
-<input type="text" id="tag_combo" role="combobox" aria-expanded="true" aria-owns="popup_listbox" />
-<ul role="listbox" id="popup_listbox">
-	<li role="option">Zebra</li>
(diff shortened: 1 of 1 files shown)


HelenBurge (approved):

daniel-montalvo (approved):

dan-tripp-siteimprove (approved):
I wonder if it would be beneficial for this PR to add a mention (say, in the rule's Background or Bibliography) that this rule now "partially targets" ARIA 1.3 (assuming I'm understanding this situation correctly).  We know that, and anyone could discover that by looking through this git history.  But that's not ideal.

giacomo-petri:
Thanks for the review @dan-tripp-siteimprove.

I'm not sure if we can reference a document that hasn't yet become a W3C Recommendation.
Perhaps we could add an Accessibility Support note stating that aria-controls is not supported and that ARIA plans to remove the requirement. Would that work for you?

dan-tripp-siteimprove:
> I'm not sure if we can reference a document that hasn't yet become a W3C Recommendation. Perhaps we could add an Accessibility Support note stating that aria-controls is not supported and that ARIA plans to remove the requirement. Would that work for you?

Yes.  Would you like to do that, or should I?

WilcoFiers (approved):
@giacomo-petri please send out the call for review.

giacomo-petri:
Call for Reviews ends on April 18th

carlosapaduarte (approved):

giacomo-petri:
This is ready to be merged

PR #2318, opened 2025-04-01
Update the aria-required-owned rule [bc4a75] to include the elements with implicit roles 

shunguoy:
The original description limits the rule to the elements with explicit semantic roles only.  The rule should apply to elements with either explicit or implicit roles. 
The changes include the following updates: (1) the description of implicit roles; (2) test cases accordingly; and (3) other minor code cleanups.  

Need for Call for Review:
This can be merged with 1 approval 

## How to Review And Approve

### DIFF
--- a/_rules/aria-required-owned-element-bc4a75.md
+++ b/_rules/aria-required-owned-element-bc4a75.md
@@ -27,7 +27,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [explicit or implicit semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
 
 ## Expectation
 
--- a/_rules/aria-required-owned-element-bc4a75.md
+++ b/_rules/aria-required-owned-element-bc4a75.md
@@ -27,7 +27,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [explicit semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [explicit or implicit semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
 
 ## Expectation
 
@@ -93,8 +93,8 @@ This element with the `menu` role only owns elements with the `menuitem`, `menui
 
 ```html
 <div role="menu">
-	<li role="none"></li>
-	<li role="menuitem">Item 1</li>
+	<div role="none"></li>
+	<div role="menuitem">Item 1</li>
 	<div role="menuitemradio" aria-checked="false">Item 2</div>
 	<div role="menuitemcheckbox" aria-checked="false">Item 3</div>
 </div>
@@ -137,6 +137,16 @@ This element with the `menu` role only owns an element with a `group` role. The 
 		</div>
 	</div>
 </div>
+```
+
+#### Passed Example 7
+
+The element `ul` with an implicit `list` role owns an element `li` with an implicit role `listitem` role.
+
+```html
+<ul>
+	<li>Item 1</li>
+</ul>
 ```
 
 ### Failed
@@ -237,11 +247,11 @@ This element with the `list` role is not included in the accessibility tree beca
 
 #### Inapplicable Example 2
 
-This `ul` element does not have an [explicit semantic role][].
+This `div` element with the `list` role is not included in the accessibility tree because it is hidden .
 
 ```html
-<ul>
-	<li>Item 1</li>
+<div role='list' hidden>
+	<div role='listitem'>Item 1</li>
 </ul>
 ```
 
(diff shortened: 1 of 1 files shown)


daniel-montalvo (approved):
Thanks @shunguoy 

Approving. Except that I think this does require a regular 2-week Call for review period.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -27,7 +27,7 @@ acknowledgments:
 
 ## Applicability
 
-This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [explicit semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
+This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [explicit or implicit semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
```
WilcoFiers: There is no `explicit or implicit semantic role` definition. 

```suggestion
This rule applies to any [HTML or SVG element][] that is [included in the accessibility tree][] and has a [WAI-ARIA 1.2][] [semantic role][] with [required owned elements][], except if the element has an [inclusive ancestor][] in the accessibility tree with an `aria-busy` [attribute value][] of `true`.
```

You'll also need to add this to the bottom of the file:

```md
[semantic role]: #semantic-role
```

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -93,8 +93,8 @@ This element with the `menu` role only owns elements with the `menuitem`, `menui
 
 ```html
 <div role="menu">
-	<li role="none"></li>
-	<li role="menuitem">Item 1</li>
+	<div role="none"></div>
+	<div role="menuitem">Item 1</div>
```
WilcoFiers, line 97: I see no reason to change this. This is an important example to demonstrate explicit roles override implicit ones.

_rules/aria-required-owned-element-bc4a75.md
```diff
@@ -139,6 +139,16 @@ This element with the `menu` role only owns an element with a `group` role. The
...
+
+The element `ul` with an implicit `list` role owns an element `li` with an implicit role `listitem` role.
+
+```html
+<ul>
+	<li>Item 1</li>
+</ul>
+```
+
 ### Failed
```
WilcoFiers, line 192: We need a couple more examples for this rule now.
1. Passed example showing how `<ul>` can pass with `<div role="listitem">`
3. A failed example of native HTML failing

WilcoFiers (changes requested):

_rules/aria-required-owned-element-bc4a75.md
shunguoy: This change was in the original suggestion, I didn't hear objection during the discussion (I could be wrong). Per [HTML 5](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element), the li can only be used as a child of ul, ol, or menu. Otherwise, no list relationship is defined:
> The [li](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) a list item. If its parent element is an [ol](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element), [ul](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element), or [menu](https://html.spec.whatwg.org/multipage/grouping-content.html#the-menu-element) element, then the element is an item of the parent element's list, as defined for those elements. Otherwise, the list item has no defined list-related relationship to any other [li](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element) element.
 
This change is to make sure no confusion on the relationship, and maintain the purpose of the example at the same time.

_rules/aria-required-owned-element-bc4a75.md
shunguoy: will make the changes. The term "implicit semantics" is defined in [ARIA](https://www.w3.org/TR/wai-aria-1.2/#implicit_semantics). The term "explicit semantics" is used throughout the [ARIA in HTML](https://w3c.github.io/html-aria/), but I don't find a definition.

_rules/aria-required-owned-element-bc4a75.md
shunguoy: Will add more examples.  But still think the mixing div and ul or ol is against the HTML 5 list relationship (see my earlier comment and quote).

shunguoy:
@WilcoFiers  added more Pass/Fail test cases for native elements. Please review again.

WilcoFiers (approved):

sage-putnam (approved):

giacomo-petri:
The applicability now includes implicit roles, but we currently lack any examples of them. IMO, if we intend this new applicability to focus on "semantic roles" rather than "explicit roles" only, we should include additional examples that illustrate implicit roles.

giacomo-petri (changes requested):
The applicability now includes implicit roles, but we currently lack any failing examples of them. IMO, if we intend this new applicability to focus on "semantic roles" rather than "explicit roles" only, we should include additional examples that illustrate implicit roles.

shunguoy:
@giacomo-petri  if you review the changes for the rule, I added a few examples with implicit roles: Passed Examples 7 - 9, Failed Examples  8 & 9, some with mixed explicit and implicit roles.

giacomo-petri:
@shunguoy, yes, I've noticed the improvements. I was specifically talking and curious if we could have something similar to the Passed Example 9:
```
<table>
	<tr>
		<td>Item 1</td>
		<td>Item 2</td>
		<td>Item 3</td>
	</tr>
</table>
```
but for failing examples.

For instance, this scenario is now matching the applicability (and previously it wasn't):
```
<ul>
    <div></div>
    <div></div>
</ul>
```
however, we don’t have any examples with this kind of issue.

shunguoy:
@giacomo-petri added this as Failed Example 10.  Also added a note `Note that this example is for demonstration purpose only because it's not a valid HTML 5 structure.`

giacomo-petri (approved):

WilcoFiers:
@shunguoy I don't think this was actually sent out as call for review. You will need to send an e-mail to the mailing list to notify people about this change coming up. Here is an example e-mail:
https://lists.w3.org/Archives/Public/public-act-r/2025Apr/0007.html

I'm removing the label, please add it back when the e-mail is sent.

shunguoy:
@WilcoFiers The review has been completed with four approvals, including yours. I thought the emails were sent out by GitHub.

carlosapaduarte (approved):

zlayaAvocado (approved):

PR #2321, opened 2025-04-14
[e88epe] Image not in the accessibility tree is decorative - Updated decorative SVG examples + Accessibility support notes

### LINKED ISSUE #2299: [e88epe] Image not in the accessibility tree is decorative - SVG elements treatment
[Image not in the accessibility tree is decorative](https://www.w3.org/WAI/standards-guidelines/act/rules/e88epe/proposed/) has the following applicability:

> ignored svg: The element is an svg with an empty ("") [accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/e88epe/proposed/#accessible-name) and a [semantic role](https://www.w3.org/WAI/standards-guidelines/act/rules/e88epe/proposed/#semantic-role) of graphics-document; or

and the following Passed Example 4:

```
<p>Happy new year!</p>
<svg height="200" xmlns="http://www.w3.org/2000/svg">
	<polygon points="100,10 40,180 190,60 10,60 160,180" fill="yellow" />
</svg>
```

Now, per [WAI-ARIA Graphics Module](https://www.w3.org/TR/graphics-aria-1.0/) specs, role="graphics-document" requires acc name. Moreover, [SVG AAM](https://www.w3.org/TR/svg-aam-1.0/) set the default platform WAI-ARIA role mappings to "graphics-document" and [ARIA in HTML](https://w3c.github.io/html-aria/#el-svg) maps the <svg> element to role="graphics-document".

Web authors expect svg not to be exposed by default; but if authors have to follow current specs, they are forced to either set an accessible name, or role=none to each svg element, even if the svg doesn't convey any information.

Moreover, different AT don't expose the svg element to the user unless it has an acc name, ignoring the browsers computed role; browsers are exposing svg elements differently ([see last WPT test](https://wpt.fyi/results/svg-aam/role/roles.tentative.html?label=pr_head&max-count=1&pr=50555)).

The Passed Example 4 currently creates accessibility failures (1.1.1) as browsers that expose the `svg` as an image cause the `svg` to be exposed as an image with no acc name.

Moreover, the applicability, IMO, is wrong anyway, since now, it's against specs, and if specs will be updated marking as presentational svg elements without acc name, is no longer needed.

As ACT group, we need to understand what to do.

In the meantime, I've opened an issue for ARIA specs: https://github.com/w3c/aria/issues/2427

giacomo-petri:
Closes: #2299

This update includes the following improvements:
- adds accessibility support notes for `svg` elements when used decoratively
- updates PE 4 to ensure the `svg` is consistently treated as decorative, including a note explaining the addition of role="none".
- updates IE 4 to similarly ensure the `svg` is treated as decorative, with a note clarifying the use of role="none".

Need for Call for Review:

### DIFF
--- a/_rules/image-not-in-acc-tree-is-decorative-e88epe.md
+++ b/_rules/image-not-in-acc-tree-is-decorative-e88epe.md
@@ -64,7 +64,7 @@ Each test target is [purely decorative][].
 
 According to the [WAI-ARIA Graphics Module](https://www.w3.org/TR/graphics-aria-1.0/) specification, the `role="graphics-document"` requires an accessible name. Additionally, the [SVG Accessibility API Mappings](https://www.w3.org/TR/svg-aam-1.0/) set the default ARIA role for SVG elements to `graphics-document`, and [ARIA in HTML](https://www.w3.org/TR/html-aria/) maps the `<svg>` element to the same role.
 
-However, browser implementations vary. Some browsers expose `<svg>` elements without accessible names as images without alternative text. To address this, in the passed and inapplicable examples where the `<svg>` element is intended to be purely decorative, the role="none" attribute has been added.
+However, browser implementations vary. Some browsers expose `<svg>` elements without accessible names as images without alternative text. To address this, in the passed and inapplicable examples where the `<svg>` element is intended to be purely decorative, the `role="none"` attribute has been added.
 
 ### Bibliography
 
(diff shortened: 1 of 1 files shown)


WilcoFiers (approved):

dan-tripp-siteimprove (approved):

shunguoy (approved):

giacomo-petri:
Call for reviews ends on May 16th 2025.

carlosapaduarte (approved):

PR #2322, opened 2025-04-16
Update permitted ARIA state or property for combobox role

### LINKED ISSUE #2311: [5c01ea] ARIA state or property is permitted - Passed Example 6-7-8-10-12 improvement
In [ARIA state or property is permitted](https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea/), Passed Examples 6, 7, 8 and 10
do not fully adhere to the [ARIA specifications](https://www.w3.org/TR/wai-aria-1.3) for required and supported states and properties.

1. The only required state and property for the combobox role is area-expanded, the aria-controls is no longer required.  This applies to the examples 6, 7, and 8. The descriptions and the code for those examples need to be updated.

2. The role='none' is invalid for a button element per [ARIA in HTML](https://w3c.github.io/html-aria/).  Thus, the description and code for the example need to be updated. 
     <button role="none" aria-pressed="false">ACT rules are cool!</button> 

3. remove aria-hidden="true" from example 12 because of the explicit style="display: none;".
     <span class="on" aria-hidden="true" style="display: none;">On</span>

shunguoy:
<< Describe the changes >>
Per the latest [ARIA spec](https://w3c.github.io/aria/#combobox), change the permitted ARIA state or property for "combobox" role from "aria-controls" to "aria-expanded" in the Pass example 6, 7 & 8.

Closes issue(s):  

- closes #2311

Need for Call for Review:

---

## How to Review And Approve


_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -126,26 +126,26 @@ The `aria-checked` [state][] is [required][] for the [semantic][semantic role] `
...
 ```
 
 #### Passed Example 7
 
-The `aria-controls` [property][] is [required][] for the [semantic][semantic role] `combobox`. [WAI-ARIA states and properties][wai-aria state or property] with empty value are still applicable to this rule.
+The `aria-expanded` [property][] is [required][] for the [semantic][semantic role] `combobox`.
 
 ```html
-<div role="combobox" aria-expanded="false" aria-controls>My combobox</div>
+<div role="combobox" aria-expanded="true" aria-controls="id1">My combobox</div>
```
WilcoFiers: This one feels off. If this is an expanded combobox it would fail for not having a popup.

```suggestion
<div role="combobox" aria-expanded="false" aria-controls="id1">My combobox</div>
```

WilcoFiers (dismissed):
Minor nitpick. I also think we should add something to the background noting the change in ARIA 1.3.

shunguoy:
@WilcoFiers   Added a reference to ARIA 1.3 in the background section, and other changes you mentioned.  ALso fixed some merge conflicts.

_rules/aria-state-or-property-permitted-5c01ea.md
```diff
@@ -57,6 +57,8 @@ The presence of prohibited ARIA attributes is often the result of a developer us
...
 
+Note that required WAI-ARIA states or properties for a role may change with the advance of the WAI-ARIA specifications. For example, both aria-expanded and aria-controls were required in [ARIA 1.2](<[https://](https://www.w3.org/TR/wai-aria-1.2/#combobox)https://>) for combobox role, but the aria-controls was removed from required states or properties in [ARIA 1.3](<[https://](https://www.w3.org/TR/wai-aria-1.3/#combobox)https://>). The examples in this rule follow [ARIA 1.3](<[https://](https://www.w3.org/TR/wai-aria-1.3/#combobox)https://>).
```
WilcoFiers:
```suggestion
Note that required WAI-ARIA states or properties for a role may change with the advance of the WAI-ARIA specifications. For example, both aria-expanded and aria-controls were required in [ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#combobox) for combobox role, but the aria-controls was removed from required states or properties in [ARIA 1.3](https://www.w3.org/TR/wai-aria-1.3/#combobox). The examples in this rule follow [ARIA 1.3](https://www.w3.org/TR/wai-aria-1.3/#combobox).
```

WilcoFiers (approved):

sage-putnam (approved):

adithyasai16 (approved):
The Code contains the changes of "aria-controls" to "aria-expanded" in the Pass example 6, 7 & 8. This is approved.

colabottles (approved):

carlosapaduarte (approved):

PR #2323, opened 2025-04-16
summary name, element position clarified

zlayaAvocado:
Adds clarification on how the `summary` element can be used since HTML specification requires the `summary` element to be the first child of `details`, while WCAG does not.

This will not require a Call for Review : editorial changes to background

## How to Review And Approve

### DIFF
--- a/_rules/summary-non-empty-accessible-name-2t702h.md
+++ b/_rules/summary-non-empty-accessible-name-2t702h.md
@@ -34,7 +34,7 @@ Each target element has an [accessible name][] that is not empty (`""`), nor jus
 
 ## Background
 
-This rule is only applicable to `summary` elements that the browser will use as controls for a `details` element. While this rule is not applicable to `summary` elements with an [explicit role][], most of the time these likely do still require an [accessible name][]. This is covered by other rules, such as the [Button has non-empty accessible name][97a4e1].
+This rule is only applicable to `summary` elements that the browser will use as controls for a `details` element. Even though the [HTML specification - The summary element][https://html.spec.whatwg.org/#the-summary-element] requires the `summary` element to be the first child of `details`, this is not required for WCAG. And while this rule is not applicable to `summary` elements with an [explicit role][], most of the time these likely do still require an [accessible name][]. This is covered by other rules, such as the [Button has non-empty accessible name][97a4e1].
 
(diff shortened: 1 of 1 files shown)


_rules/summary-non-empty-accessible-name-2t702h.md
```diff
@@ -34,7 +34,7 @@ Each target element has an [accessible name][] that is not empty (`""`), nor jus
 
 ## Background
 
-This rule is only applicable to `summary` elements that the browser will use as controls for a `details` element. While this rule is not applicable to `summary` elements with an [explicit role][], most of the time these likely do still require an [accessible name][]. This is covered by other rules, such as the [Button has non-empty accessible name][97a4e1].
+This rule is only applicable to `summary` elements that the browser will use as controls for a `details` element. Even though the [HTML specification - The summary element](https://html.spec.whatwg.org/#the-summary-element) requires the `summary` element to be the first child of `details`, this is not required for WCAG. And while this rule is not applicable to `summary` elements with an [explicit role][], most of the time these likely do still require an [accessible name][]. This is covered by other rules, such as the [Button has non-empty accessible name][97a4e1].
```
shunguoy, line 37: (1) WCAG can't override HTML 5; (2) per [ARIA in HTML](https://w3c.github.io/html-aria/#case-sensitivity), a 'summary' element neither has an implicit nor explicit role; and (3) a "summary" represent a [summary, caption, or legend for its parent "details" element](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element), rather than "use as control for a 'details' element".
 Suggest changing into:  "This rule is only applicable to a `summary` element that is a child of a `details` element.  An accessible name is required for a "summary" element because the [details element](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element) are often presented by browsers as a [disclosure widget](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element) from which the user can obtain additional information or controls.

WilcoFiers: I don't really understand your point here. We're not testing HTML, or HTML in ARIA in this rule. I like Sasha's suggestion better.

WilcoFiers (approved):

shunguoy (approved):

sage-putnam (approved):

PR #2329, opened 2025-05-12
Add test for validating aria properties for `accessibility_requirements`

zlayaAvocado:
Adds a test for validating aria12 properties under `accessibility_requirements`. Since we expect a particular `accessibility_requirements` structure for the rules, the `frontmatter` tests act similar to typescript but for the frontmatter structure.

References this issue: https://github.com/act-rules/act-rules.github.io/pull/2328
The test job was failing before the fix: https://github.com/act-rules/act-rules.github.io/actions/runs/14984632459/job/42096095113

Need for Call for Review:
This will not require a Call for Review: adds a test

---

## How to Review And Approve

### DIFF
--- a/__tests__/frontmatter.js
+++ b/__tests__/frontmatter.js
@@ -133,7 +133,7 @@ function validateAriaProperties(accessibilityRequirements) {
 		expect(aria12Key.split(':')[1]).toBeTruthy()
 
 		// Check that aria object has the required properties
-		const requiredAriaProps = ['title', 'forConformance', 'failed', 'passed']
+		const requiredAriaProps = ['title', 'forConformance', 'failed', 'passed', 'inapplicable']
 		test.each(requiredAriaProps)('has required property `%s`', requiredProp => {
 			expect(accessibilityRequirements[aria12Key]).toHaveProperty(requiredProp)
 		})
--- a/__tests__/frontmatter.js
+++ b/__tests__/frontmatter.js
@@ -1,7 +1,7 @@
 const describeRule = require('../test-utils/describe-rule')
 const describePage = require('../test-utils/describe-page')
 
-describe.only('frontmatter', () => {
+describe('frontmatter', () => {
 	/**
 	 * Rules
 	 */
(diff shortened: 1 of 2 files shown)


__tests__/frontmatter.js
```diff
@@ -123,3 +124,18 @@ function validateGlossaryFrontmatter({ frontmatter }) {
 		expect(frontmatter).toHaveProperty(requiredProp)
 	})
 }
+
+function validateAriaProperties(accessibilityRequirements) {
+	const aria12Key = Object.keys(accessibilityRequirements).find(key => key.startsWith('aria12:'))
```
zlayaAvocado: At first was thinking to search for `ariaxx` but if the number ever changes many more places will need to update this anyway

WilcoFiers: What would you say about flipping this around and requiring a title for everything except WCAG criteria and techniques? That would be anything that doesn't start with either `/wcag-technique:/` or `/wcag2\d:/`. We know we only have titles for those two, so everything else really should have a title.

__tests__/frontmatter.js
```diff
@@ -123,3 +124,18 @@ function validateGlossaryFrontmatter({ frontmatter }) {
...
+
+	if (aria12Key) {
+		// Check that aria key has an anchor name
+		expect(aria12Key.split(':')[1]).toBeTruthy()
+
+		// Check that aria object has the required properties
+		const requiredAriaProps = ['title', 'forConformance', 'failed', 'passed', 'inapplicable']
```
WilcoFiers: Do we not already have the forConformance / failed / passed / inapplicable tested? Those are required except if the requirement is secondary. That's also a scenario that shouldn't fail here.

It seems to me this only needs to require the title. The other props should be tested elsewhere.

WilcoFiers (dismissed):

__tests__/frontmatter.js
zlayaAvocado: you are right, didn't look into that part close enough

WilcoFiers (approved):

PR #2332, opened 2025-05-29
Update aria-hidden-no-focusable-content-6cfa84.md

WilcoFiers:
I noticed this title was slightly off. The format we use in other non-WCAG titles is {spec}, {heading text}

---

## How to Review And Approve

### DIFF
--- a/_rules/aria-hidden-no-focusable-content-6cfa84.md
+++ b/_rules/aria-hidden-no-focusable-content-6cfa84.md
@@ -12,7 +12,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   using-aria:fourth:
-    title: Using ARIA: Fourth Rule of ARIA Use
+    title: Using ARIA, 2.4 Fourth Rule of ARIA Use
     forConformance: false
     failed: not satisfied
     passed: further testing needed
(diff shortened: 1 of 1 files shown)


_rules/aria-hidden-no-focusable-content-6cfa84.md
```diff
@@ -12,7 +12,7 @@ accessibility_requirements:
     passed: further testing needed
     inapplicable: further testing needed
   using-aria:fourth:
-    title: Fourth rule of ARIA use
+    title: Using ARIA, 2.4 Fourth Rule of ARIA Use
```
WilcoFiers, line 15: Here's an example of how we did this for links to ARIA:

https://www.w3.org/WAI/standards-guidelines/act/rules/5c01ea/proposed/#accessibility-requirements-mapping

> ARIA 1.2, 8.6 State and Property Attribute Processing

zlayaAvocado (approved):

_rules/aria-hidden-no-focusable-content-6cfa84.md
adithyasai16: The change is as per the standard {Spec}, {Heading Text}. This is approved. 
Reference: https://www.w3.org/TR/using-aria/#fourth

adithyasai16 (approved):
The change is as per the standard {Spec}, {Heading Text}. This is approved.
Reference: https://www.w3.org/TR/using-aria/#fourth

carlosapaduarte (approved):

shunguoy (approved):

giacomo-petri (approved):

PR #2340, opened 2025-07-24
[80af7b] Focusable element has no keyboard trap - move Failed Example 2 to passed since it's currently passing the ACT rule

### LINKED ISSUE #2314: [80af7b] Focusable element has no keyboard trap - Failed Example 2 is not failing the ACT rule
In [Focusable element has no keyboard trap](https://www.w3.org/WAI/standards-guidelines/act/rules/80af7b/proposed/), [Failed Example 2](https://w3.org/WAI/content-assets/wcag-act-rules/testcases/80af7b/d2f5325f3fd5ddde38cd677a5ca36ba0d762fb84.html) is actually not failing the ACT rule, because navigating backward all buttons are focusable (as per background note).

giacomo-petri:
Closes: #2314

Description: 
Failed Example 2 is currently passing the ACT rule because it traps the user when navigating forward, but not backward. According to the Background note:

> This rule only requires navigation in one direction (either forward or backward), not both, and not a specific one.

Given this, I'm removing the specific test case.

Need for Call for Review:

### DIFF
--- a/_rules/focusable-no-keyboard-trap-80af7b.md
+++ b/_rules/focusable-no-keyboard-trap-80af7b.md
@@ -148,6 +148,22 @@ These focusable `button` elements have scripts that create a keyboard trap. The 
 <a id="link2" href="#">Link 2</a>
 ```
 
+#### Passed Example 7
+
+The [focusable][] `button` labeled "Button3" creates a keyboard trap when navigating forward, as it moves focus back to "Button1" preventing users to focus "Button3". However, when navigating backward, all focusable elements receive focus as expected. Therefore, according to the background notes, this behavior is considered compliant with the rule, despite being recognized as a poor practice in terms of keyboard accessibility.
+
+```html
+<button onblur="setTimeout(() => this.nextElementSibling.focus(), 10)">
+	Button1
+</button>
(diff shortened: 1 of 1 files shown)


_rules/focusable-no-keyboard-trap-80af7b.md
```diff
@@ -164,22 +164,6 @@ This [focusable][] element creates a keyboard trap bringing focus to the `button
 
 #### Failed Example 2
 
-These [focusable][] `button` elements create a keyboard trap preventing the last `button` to be reached using the keyboard.
```
WilcoFiers, line 167: I think we should move this to a passed example and explain why this is a narrow pass.

WilcoFiers (changes requested):

giacomo-petri:
Moved from failed to passed, as per CG meeting.

carlosapaduarte (approved):

daniel-montalvo (approved):

WilcoFiers (approved):

dan-tripp-siteimprove (approved):

giacomo-petri:
Call for review ends on August 18th

PR #2365, opened 2025-10-23
Resolve language-rule issues

### LINKED ISSUE #1614: Should ACT rules allow fictional and/or extinct languages as a valid language?
Currently ACT rules allow some fictional languages as valid languages:

https://act-rules.github.io/rules/de46e4#valid-language-tag

BCP-47 fictional languages:

tlh - Klingon (Star Trek)
sjn - Sindarin (Lord of the Rings)
qya - Quenya (Lord of the Rings)

Only a few fictional languages have BCP-47 codes, more recent ones like like Dothraki don't.

https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

A related issue is extinct languages with BCP-47 codes but no AT support:

nrn - Norn - was spoken on Scottish islands until 1850
xpi - Pictish - was spoken in mainland Scotland until 1150

Lots listed here with BCP-47 tags:
https://en.wikipedia.org/wiki/Lists_of_extinct_languages

WilcoFiers:
This PR resolves a number of smaller language rule issues.

- Closes https://github.com/act-rules/act-rules.github.io/issues/1614#issuecomment-957322693
- Clarify that body tag is optional, which is why some of the examples are applicable even without a body [off6ek]
- ...			

---

## How to Review And Approve

### DIFF
--- a/_rules/element-lang-matches-default-language-off6ek.md
+++ b/_rules/element-lang-matches-default-language-off6ek.md
@@ -141,6 +141,8 @@ This `div` element has a `lang` [attribute value][] of `en` (English), which mat
 
 This `span` element has a `lang` [attribute value][] of `fr` (French), which matches one of its [most common languages][most common language]. The most common languages are both English and French because all the words belong to both languages.
 
+**Note**: The `body` element in HTML are optional and will be added by the browser.
+
 ```html
 <html lang="en">
 	<p>
--- a/pages/glossary/most-common-language.md
+++ b/pages/glossary/most-common-language.md
@@ -10,7 +10,7 @@ input_aspects:
   - Language
 ---
 
-The _most common language of an element_ is determined by counting the number of _words_ in the [text inheriting its programmatic language][] from this element that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. In case of ties, the element has several most common languages. If there are no words in the [text inheriting its programmatic language][] from the element, then it has no most common language.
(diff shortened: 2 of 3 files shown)


_rules/element-lang-matches-default-language-off6ek.md
```diff
@@ -141,6 +141,8 @@ This `div` element has a `lang` [attribute value][] of `en` (English), which mat
 
 This `span` element has a `lang` [attribute value][] of `fr` (French), which matches one of its [most common languages][most common language]. The most common languages are both English and French because all the words belong to both languages.
 
+**Note**: The `body` element in HTML are optional and will be added by the browser.
```
Jym77, line 144: Given that it seems there are only 2 examples in that rule without a `body` element (Passed Examples 4 and 5), I'd say it is simpler to add the `body` element to both rather than adding this note.

pages/glossary/most-common-language.md
```diff
@@ -10,7 +10,7 @@ input_aspects:
   - Language
 ---
 
-The _most common language of an element_ is determined by counting the number of _words_ in the [text inheriting its programmatic language][] from this element that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. In case of ties, the element has several most common languages. If there are no words in the [text inheriting its programmatic language][] from the element, then it has no most common language.
+The _most common language of an element_ is determined by counting the number of _words_ in the [text inheriting its programmatic language][] from this element that are part of any of the languages in the [language subtag registry][]. The same word can be part of multiple languages. In case of a tie, the first of the tied languages used in the text is the most common language. If there are no words in the [text inheriting its programmatic language][] from the element or the tie cannot be resolved, then it has no most common language.
```
Jym77, line 13: This removes the possibility of having several most common languages (this is OK), but then the description of Passed Examples 4/5 of "HTML element language subtag matches language" (and maybe some other) need to be updated accordingly. They still read 
> This `span` element has a `lang` [attribute value][] of `fr` (French), which matches one of its [most common languages][most common language]. The most common languages are both English and French because all the words belong to both languages.

stating that there are 2 "most common languages".

Jym77 (changes requested):

Issue #2368, opened 2025-10-30
Element in sequential focus order has visible focus [oj04fd] expectation allows common scenarios to pass even when failing


giacomo-petri:
I understand that this rule is atomic, and passing it doesn't necessarily mean that rule 2.4.7 is also met. However, there are common scenarios where the focus indicator is invisible, yet the current rule still passes.

A typical example is a top navigation menu or toolbar, where each control uses `outline: none` when focused and triggers a fly-out menu to expand. This expansion changes the pixels on screen, so our rule considers it a pass. However, because `outline: none` removes the visible focus indicator, users can't tell which option is currently focused or which menu item has been expanded.

I also believe this rule currently lacks sufficient examples, so I've added several in https://github.com/act-rules/act-rules.github.io/pull/2369 PR.
Failed Example 3 illustrates the situation I mentioned earlier (even though the description is slightly different). I've marked it as a failure, even though (based on the current wording) it technically passes the rule's expectations.

The PR is currently in draft status, as I'd like to discuss these points with the group before making any changes.

dan-tripp-siteimprove:
You want to fail any case where the changed pixels might not be noticed by the user - am I right?

If I'm right: it seems to me that there are two categories of "changed pixels which might not be noticed by the user": 1) changed pixels which are outside the viewport, and 2) changed pixels that are inside the viewport but are "too far away to be noticed" (which is vague).

But I wonder if we could catch a lot of them this way: 

We change the expectation to "the changed pixels need to be either a) part of the target element (= within its bounding box including CSS margin), or b) on opposite sides of the target element".   And (b) means either: b1) both above and below the target element, or b2) both to the left and to the right of the target element.

Passed Examples 1 and 2 would pass because of (a).

[Passed Example 3](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/oj04fd/b44aa294df6f0de62f11f6ae732087d92c90b448.html) would pass because of (b1) and (b2).

[Passed Example 4](https://www.w3.org/WAI/content-assets/wcag-act-rules/testcases/oj04fd/dd9628d86628e285fe99ce98efdacbe441c20ca5.html) would pass because of (b2).

In [your PR 2369](https://github.com/act-rules/act-rules.github.io/pull/2369), Failed Example 3 would still fail and it would be obvious why.

Jym77:
[...]

PR #2369, opened 2025-10-30
[DRAFT] Element in sequential focus order has visible focus [oj04fd] - Added examples + review expectation

giacomo-petri:
Closes issue(s): #2368

For now, I've only added passing and failing examples.
To discuss:
- is the expectation strict enough?

Need for Call for Review:

---

### **When creating PR:**

### **After creating PR:**

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
+++ b/_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
@@ -135,7 +135,6 @@ Each of these three [focusable][] elements has a blue square in front of it when
 #### Passed Example 5
 
 Each of these three [focusable][] elements uses a `::before` pseudo-element with a black arrow as its `content`, which appears in front of the element when it is [focused][]. The arrow's `hsl` value (`hsl(0, 0%, 0%)`) is different to the `hsl` value of the background (`hsl(0, 0%, 100%)`) over which the arrow appears.
-<body>
 
 ```html
 <style>
--- a/_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
+++ b/_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
@@ -186,10 +186,10 @@ The [focusable][] element does not have any pixel changing color when it is [foc
 
 ```html
 <style>
-body {background:#FFF;}
-a:focus {outline-color:#FFF}
+	body {background: white;}
+	a:focus {outline-color: white;}
 </style>
-<a class="no-focus-default" href="https://act-rules.github.io/">ACT rules</a>
+<a href="https://act-rules.github.io/">ACT rules</a>
 ```
 
 #### Failed Example 3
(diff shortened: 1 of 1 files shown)


_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -143,6 +180,113 @@ The [focusable][] element does not have any pixel changing color when it is [foc
...
+    }
+    a {
+		display: inline-block;
+		color: black;
+		text-decoration: none;
+		padding: 20px 20px;
+		outline: none;
+		border-bottom: 2px solid #ccc;
+    }
+    a:focus {
```
dan-tripp-siteimprove: The focus indicator defined by this CSS block is never visible, correct?  Because it's hidden by the fly-out?  Assuming I'm correct: I think that there is no point in CSS block, and it should be removed.  The rule's expectation is about "device pixels" - not "occluded pixels".   If we want to make an example distinguishing those two things, that would be a good thing, and I think it should be done in a separate example.

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -143,6 +180,113 @@ The [focusable][] element does not have any pixel changing color when it is [foc
...
+<style>
+	body {background: white;}
+	a:focus {outline-color: white;}
+</style>
+<a href="https://act-rules.github.io/">ACT rules</a>
+```
+
+#### Failed Example 3
+
+The [focusable][] element does not have any pixel changing color when it is [focused][] because the focus indicator's is covered by another element.
```
WilcoFiers, line 201: We talked this through in the CG call today. We think this passes the rule, and probably the success criterion too. WCAG does not require different elements to have different focus indicators. There's a comment about this point in the background section:

> WCAG does not require that the focus indicator for each [focusable][] element is unique in appearance. Therefore, this rule can pass even if several focus indicators are identical. Such a situation may nonetheless cause confusion and all examples in this rule avoid it.

We also felt this is enough of an edge case that it isn't worth trying to address.

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
```diff
@@ -143,6 +180,113 @@ The [focusable][] element does not have any pixel changing color when it is [foc
...
 ```
 
+#### Failed Example 2
+
+The [focusable][] element does not have any pixel changing color when it is [focused][] because the focus indicator's color matches the background's color, making the outline imperceptible.
+
+```html
+<style>
+	body {background: white;}
+	a:focus {outline-color: white;}
```
WilcoFiers: This doesn't look like Prettier was run on it. Would you mind doing that?

WilcoFiers (changes requested):

_rules/sequentially-focusable-element-has-visible-focus-oj04fd.md
giacomo-petri: I'm not sure I'm following. Are we referring to point 3? When you run the example, nothing changes in the visual appearance of the page when the element receives focus. Can you please explain a bit more?

PR #2371, opened 2025-11-13
Add inapplicable examples for the audio/video autoplay rule

### LINKED ISSUE #2115: Audio or video element avoids automatically playing audio: passed example 2 should be inapplicable
Passed example 2 says:

> This video element does not play for longer than 3 seconds.

https://act-rules.github.io/rules/80f0bf#passed-example-2

but it looks like it should be inapplicable because the applicability says:

This rule applies to any audio or video element for which all the following are true:

- autoplay: the element has an autoplay [attribute value](https://act-rules.github.io/rules/80f0bf#attribute-value) of true; and
- not muted: the element has a muted [attribute value](https://act-rules.github.io/rules/80f0bf#attribute-value) of false; and
- not paused: the element has a paused [attribute value](https://act-rules.github.io/rules/80f0bf#attribute-value) of false; and
- duration: the element has a [media resource](https://html.spec.whatwg.org/multipage/media.html#media-resource) **lasting more than 3 seconds** and that contains audio.

zlayaAvocado:
Adds inapplicable examples for the audio/video autoplay rule.
https://deploy-preview-2371--act-rules.netlify.app/standards-guidelines/act/rules/80f0bf/proposed/

Closes issue(s):

- closes https://github.com/act-rules/act-rules.github.io/issues/2115

Need for Call for Review:

---

## How to Review And Approve

### DIFF
--- a/_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
+++ b/_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
@@ -57,7 +57,8 @@ This rule applies to any `audio` or `video` element for which all the following 
 - **autoplay**: the element has an `autoplay` [attribute value][] of `true`; and
 - **not muted**: the element has a `muted` [attribute value][] of `false`; and
 - **not paused**: the element has a `paused` [attribute value][] of `false`; and
-- **duration**: the element has a [media resource][] lasting more than 3 seconds and that contains audio.
+- **play time**: the element has a [media resource][] that contains audio and that stops no more than 3 seconds
+This can be changed by authors, for example using a [temporal media fragment URI](https://www.w3.org/TR/media-frags/#naming-time) (`#t`) in the source URI.
 
 ## Expectation
 
--- a/_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
+++ b/_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
@@ -192,7 +192,7 @@ This `video` element does not play for longer than 3 seconds.
 
 #### Inapplicable Example 4
 
-This `video` element autoplays but paused.
+This `video` element autoplays but is paused.
 
 ```html
 <video id="paused" autoplay>
--- a/_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
+++ b/_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
@@ -192,17 +192,6 @@ This `video` element does not play for longer than 3 seconds.
 
 #### Inapplicable Example 4
 
-This `video` element does not autoplay.
-
-```html
-<video>
-	<source src="/test-assets/rabbit-video/video.mp4" type="video/mp4" />
-	<source src="/test-assets/rabbit-video/video.webm" type="video/webm" />
-</video>
-```
-
-#### Inapplicable Example 5
-
 This `video` element autoplays but paused.
 
 ```html
@@ -214,6 +203,17 @@ This `video` element autoplays but paused.
 <script>
   document.getElementById('paused').pause();
 </script>
+```
+
+#### Inapplicable Example 5
+
+This `video` element does not play automatically.
+
+```html
+<video>
+	<source src="/test-assets/rabbit-video/video.mp4" type="video/mp4" />
+	<source src="/test-assets/rabbit-video/video.webm" type="video/webm" />
+</video>
 ```
 
 #### Inapplicable Example 6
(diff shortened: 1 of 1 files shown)


_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -192,6 +181,43 @@ This `video` element has no audio output.
...
+```html
+<video autoplay>
+	<source src="/test-assets/rabbit-video/video.mp4#t=8,10" type="video/mp4" />
+	<source src="/test-assets/rabbit-video/video.webm#t=8,10" type="video/webm" />
+</video>
+```
+
+#### Inapplicable Example 4
+
+This `video` element autoplays but paused.
```
Jym77:
```suggestion
This `video` element autoplays but is paused.
```

Jym77 (approved):
Typo. Approving already.

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -192,6 +181,43 @@ This `video` element has no audio output.
 
 #### Inapplicable Example 3
 
+This `video` element does not play for longer than 3 seconds.
+
+```html
+<video autoplay>
+	<source src="/test-assets/rabbit-video/video.mp4#t=8,10" type="video/mp4" />
+	<source src="/test-assets/rabbit-video/video.webm#t=8,10" type="video/webm" />
```
WilcoFiers, line 189: This one's actually kind of tricky. The media resource has a duration of 13 seconds, but the video has a play time of 2 seconds. I think we'll need to update the applicability.

Something like:

```
- **play time**: the element has a [media resource][] that contains audio and that stops no more than 3 

**Note**: For most videos the start time is 0, and the stop time is the end of the video. In HTML the `t` parameter allows content authors to set the start and stop time of a video.
```

@Jym77 WDYT?

WilcoFiers (dismissed):
Agree with JYM's comments. I also found a problem with the applicability because of this. Interesting edge case here.

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
Jym77: Yes, good point. Maybe phrasing the second sentence a bit more generically and with a link to specs:
```
This can be changed by authors, for example using a [temporal media fragment URI](https://www.w3.org/TR/media-frags/#naming-time) (`#t`) in the source URI.
```

Jym77 (approved):

WilcoFiers (approved):

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -57,7 +57,7 @@ This rule applies to any `audio` or `video` element for which all the following
 - **autoplay**: the element has an `autoplay` [attribute value][] of `true`; and
 - **not muted**: the element has a `muted` [attribute value][] of `false`; and
 - **not paused**: the element has a `paused` [attribute value][] of `false`; and
-- **duration**: the element has a [media resource][] lasting more than 3 seconds and that contains audio.
+- **play time**: the element has a [media resource][] that contains audio and that stops after no more than 3 seconds. This can be changed by authors, for example using a [temporal media fragment URI](https://www.w3.org/TR/media-frags/#naming-time) (`#t`) in the source URI.
```
giacomo-petri, line 60: I believe this is incorrect. In the context of:

> This rule applies to any `audio` or `video` element for which all the following are true:

If we only consider videos with a play time of < 3s, the SC would always pass by default.
Instead, I believe we should consider:
- all videos, with the two stated expectations applied OR
- only videos longer > 3s, with only the 1st expectation needed, since videos shorter than 3 seconds would be considered inapplicable.

_rules/audio-or-video-avoids-automatically-playing-audio-80f0bf.md
```diff
@@ -101,18 +101,7 @@ This `audio` element has an [instrument][] to pause, stop, or turn the audio vol
...
-<video autoplay>
-	<source src="/test-assets/rabbit-video/video.mp4#t=8,10" type="video/mp4" />
-	<source src="/test-assets/rabbit-video/video.webm#t=8,10" type="video/webm" />
-</video>
-```
-
-#### Passed Example 3
-
-This `video` element autoplays and has an [instrument][] to pause, stop, or turn the audio volume off.
+This `video` element is longer than 3 seconds, autoplays, and has an [instrument][] to pause, stop, or turn the audio volume off.
```
giacomo-petri, line 104:
```suggestion
This `video` element is longer than 3 seconds and autoplays; however, it passes the rule because it provides an [instrument][] that allows the user to pause, stop, or turn the audio volume off.
```

Suggestion to clearly separate the conditions that would cause the rule to fail from those that allow it to pass, explicitly highlighting the feature that makes the rule pass.

giacomo-petri (changes requested):
I’ve added a couple of comments. The one regarding applicability is mandatory in my view, the other is just editorial.

PR #2395, opened 2026-03-31
Added the glossary page

### LINKED ISSUE #2393: Add standalone ACT Rules glossary on WAI (`wcag-act-rules`)
### Summary

Add a **standalone glossary** to the WAI ACT Rules area, comparable to the glossary on the former ACT Rules community site: a single place with **all** definitions, **stable fragment IDs** for deep linking, **correct internal links**, and for each term a **list of rules that use** that definition.

This is an **additional** resource. It must **not** replace or remove the existing **Glossary** section on individual rule pages (those stay so each rule remains self-contained).

### Reference implementation (former public site)

The behaviour to match comes from the **Gatsby** site in **`act-rules-web`**. Useful pointers:

- **Glossary page:** renders full glossary markdown as HTML, one block per term with `id` = glossary `key`, plus a **“Used In Rules”** list per term (see `act-rules-web/src/pages/glossary.js`).
- **Reverse index:** `_data/glossary-usages.json`, generated by `build/create-glossary-usages.js` from rule + glossary markdown (rule id → list of `#key` references, including **transitive** references via links inside glossary entries).
- **Per-rule glossary in Gatsby:** `act-rules-web/src/components/glossary.js` shows only the terms relevant to that rule, but still uses **full** rendered HTML for each term.

### Full definition content vs what appears in WAI rule files today

**act-tools** (used when generating `content/rules/*/proposed.md` for WAI) **truncates** inlined glossary bodies: it keeps only content **before the first `##` heading** in each glossary source file (see `get-glossary.ts` / `getDefinitionBody` in act-tools). Anything after that—e.g. **Examples**, extra sections—is **omitted** from the glossary block **on the rule page**.

**Requirement for this issue:** the standalone WAI glossary must show the **complete** definition from source (including everything after the first `##`), **not** the act-tools truncation used for per-rule embeds.

### Scope and non-goals

- **In scope:** New glossary index (and/or per-term pages), navigation entry, working anchors and links on WAI, “used in rules” lists with correct WAI rule URLs, full definition bodies as above.
- **Explicit non-goal:** Do not remove or replace the existing per-rule **Glossary** section in published rules. Optional future work (separate decision): whether to add more cross-links from rules to the global glossary; that does not change keeping the inline glossary.

### Implementation notes (suggestions)

1. **Content source** today is still the CG repo: `act-rules.github.io` → `pages/glossary/*.md`. Publishing a full glossary on WAI may require extending **act-tools** and/or the **wai-build** workflow so glossary output is written into `wcag-act-rules` with appropriate frontmatter and permalinks—or an agreed alternative if authoring moves.

2. **“Used in rules”:** Reuse the same idea as `act-rules-web`’s `create-glossary-usages.js`: build-time map from rule → glossary `#key` (direct + transitive where applicable), then invert for display. Rule links must use **WAI** paths (e.g. `/standards-guidelines/act/rules/{id}/proposed/`), not Gatsby’s `/rules/{id}`.

3. **Jekyll:** `_config.yml` already declares a collection `wai-wcag-act-rules-patterns` titled “Glossary” but without content; that may be a hook, or use another pattern consistent with WAI theme requirements. Confirm any **main WAI site** config sync if your deployment requires it.

4. **Link hygiene:** Replace or avoid depending on `https://act-rules.github.io/...` for primary navigation where WAI URLs exist; fix remaining references in glossary-related content as needed.

5. **Edge cases:** Rules use non-glossary anchors (e.g. `#ruleid:hidden`); reverse-index logic should only count keys that correspond to real glossary entries. Glossary pages may define extra `<dfn id="...">` sub-ids; the primary `key` in frontmatter remains the main stable id unless you document otherwise.

### Acceptance criteria

- [ ] Public glossary on WAI under the ACT Rules area, listing all terms with stable `#key` (or equivalent) anchors.
- [ ] Each term shows the **full** source definition, including sections **not** present in act-tools-inlined rule glossaries.
- [ ] Internal and example links work in the WAI context.
- [ ] Each term lists **all rules** that use it (same notion of “use” as the reference: direct + transitive glossary references where applicable).
- [ ] ACT standalone nav includes a link to the glossary.
- [ ] Per-rule **Glossary** sections on rule pages remain unchanged in purpose and continue to be generated as today.

### Related repositories

- [w3c/wcag-act-rules](https://github.com/w3c/wcag-act-rules) — WAI Jekyll site  
- [act-rules/act-tools](https://github.com/act-rules/act-tools) — rule + glossary transform (`--glossaryDir`, `get-glossary` truncation)  
- [act-rules/act-rules.github.io](https://github.com/act-rules/act-rules.github.io) — glossary source markdown, `wai-build`, tests  
- **act-rules-web** (Gatsby) — reference UX for glossary page + `glossary-usages` generation

---

G0dwin:
Added a glossary top level page to the ACT rules W3C site.

Related PR: https://github.com/act-rules/act-tools/pull/53

Closes issue(s):

- closes [#2393](https://github.com/act-rules/act-rules.github.io/issues/2393)

Need for Call for Review:
* This can be merged with 1 approval: adding new top level page
* This will require a 2 weeks Call for Review

---

## How to Review And Approve

---

[Preview the changes here](https://act.g-o-d.win/wcag-act-rules/standards-guidelines/act/rules/terms/)

### DIFF
--- a/.github/scripts/wai-build.mjs
+++ b/.github/scripts/wai-build.mjs
@@ -2,24 +2,36 @@
 import 'zx/globals';
 import { config, cloneWcagActRules, commitAndPush } from './commons.mjs'
 
+const ACT_TOOLS_PATH = process.env.ACT_TOOLS_PATH || './node_modules/act-tools';
+
 await cloneWcagActRules(config);
 await generateProposedRulePages(config);
 await generateexamples(config);
+await generateGlossaryPage({ ...config, wcagActRulesDir: config.tmpDir });
 const commitMessage = (await $`git log -1 --pretty=%B`).stdout;
 await commitAndPush(config, commitMessage);
 
+async function generateGlossaryPage({ tmpDir, rulesDir, glossaryDir, testAssetsDir }) {
+  await $`node ${ACT_TOOLS_PATH}/dist/cli/generate-glossary.js \
+    --rulesDir "${rulesDir}" \
+    --glossaryDir "${glossaryDir}" \
+    --testAssetsDir "${testAssetsDir}" \
+    --outDir "${tmpDir}"
+  `;
+}
+
 async function generateProposedRulePages({ tmpDir, rulesDir, glossaryDir, testAssetsDir }) {
-  await $`node ./node_modules/act-tools/dist/cli/rule-transform.js \
-  --rulesDir "${rulesDir}" \
-  --glossaryDir "${glossaryDir}" \
-  --testAssetsDir "${testAssetsDir}" \
-  --outDir "${tmpDir}" \
-  --proposed
+  await $`node ${ACT_TOOLS_PATH}/dist/cli/rule-transform.js \
+    --rulesDir "${rulesDir}" \
+    --glossaryDir "${glossaryDir}" \
+    --testAssetsDir "${testAssetsDir}" \
+    --outDir "${tmpDir}" \
+    --proposed
   `;
 }
 
 async function generateexamples({ tmpDir, rulesDir, testAssetsDir }) {
-  await $`node ./node_modules/act-tools/dist/cli/build-examples.js \
+  await $`node ${ACT_TOOLS_PATH}/dist/cli/build-examples.js \
     --rulesDir "${rulesDir}" \
     --testAssetsDir "${testAssetsDir}" \
     --outDir "${tmpDir}" \
--- a/.github/scripts/wai-build.mjs
+++ b/.github/scripts/wai-build.mjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { config, cloneWcagActRules, commitAndPush } from './commons.mjs'
+import { config, cloneWcagActRules } from './commons.mjs'
 
 const ACT_TOOLS_PATH = process.env.ACT_TOOLS_PATH || './node_modules/act-tools';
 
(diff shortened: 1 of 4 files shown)


kengdoj (approved):
This looks great!

.github/scripts/wai-build.mjs
```diff
@@ -1,25 +1,35 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { config, cloneWcagActRules, commitAndPush } from './commons.mjs'
+import { config, cloneWcagActRules } from './commons.mjs'
+
+const ACT_TOOLS_PATH = process.env.ACT_TOOLS_PATH || './node_modules/act-tools';
```
Jym77: **Note** (non-blocking): I usually prefer nullish coalescence (`??`) in order to avoid the occasional blow-up with falsy-but-actually-correct values.

```suggestion
const ACT_TOOLS_PATH = process.env.ACT_TOOLS_PATH ?? './node_modules/act-tools';
```

Jym77: **Question:** If `act-tools` is expected to be in `node_modules`, wouldn't an `import { … } from "act-tools"` work fine? 

The `cli` code there seems to only call some other function with parameters read from CLI.
(a bit unrelated to this PR, since it was already how the other commands are run)

Jym77 (approved):

daniel-montalvo (approved):
Thanks @G0dwin for working on this. Approving, but I'd like for this to have @Jym77 suggestions included.

daniel-montalvo:
@remibetin We would benefit from your review of this functionality, also related to https://github.com/act-rules/act-tools/pull/53, because all of this is transforming content that's intended to land on the WAI website.

remibetin:
@daniel-montalvo Thanks for the mention, Daniel. We'll review this one too with @kfranqueiro.

.github/scripts/wai-build.mjs
G0dwin: This would be complicated for a couple of reasons:
1. `import` statements can't use dynamic paths, so we would need to use `await import(path)` which I don't think adds much of an improvement, but it's subjective
1. The files that we are referencing are not JS modules, they don't export functions. Instead, they are executables. So we would need to rewrite the target files to export functions instead of executing code when they are interpreted.

It might be worth doing at some point, but I think it's beyond the scope of this change.

Jym77: Fair point 👍

Issue #2406, opened 2026-04-16
Add content in testcases for language rules [bf051a, ucwvc8, b5c3f8]


giacomo-petri:
For rules that check for existing or appropriate language, it's important to include actual content if we want to map the rule to a SC. If a page is empty, the language used (whether valid or invalid) is not relevant.

giacomo-petri:
After a quick review, the issue appears to affect only the rule [HTML page lang attribute has valid language tag](https://www.w3.org/WAI/standards-guidelines/act/rules/bf051a/proposed/), since the other rules do not account for examples with empty content.

All the rules mapped to 3.1.2 Language of Parts include the "Not empty" applicability exception.

> Not empty: there is some non-empty [text inheriting its programmatic language](https://www.w3.org/WAI/standards-guidelines/act/rules/off6ek/proposed/#text-inheriting-language:text) from the element which is neither empty nor only [whitespace](https://www.w3.org/WAI/standards-guidelines/act/rules/off6ek/proposed/#whitespace).

I'm also adding an analogous exception for all the 3.1.1 Language of Page rules, together with content for the passing example and additional inapplicable examples.

PR #2407, opened 2026-04-16
Update rules mapping for SC 3.1.1 Language of Page to address empty content [bf051a, ucwvc8, b5c3f8]

giacomo-petri:
Closes issue(s): #2406

This PR introduces a new exception to the rules mapping for SC 3.1.1 Language of Page, ensuring that pages with no content are treated as an exception, since there is no lang to communicate to the user.

Need for Call for Review:

---

## How to Review And Approve

### DIFF
--- a/_rules/html-page-lang-b5c3f8.md
+++ b/_rules/html-page-lang-b5c3f8.md
@@ -21,6 +21,7 @@ input_aspects:
 acknowledgments:
   authors:
     - Jey Nandakumar
+    - Giacomo Petri
   previous_authors:
     - Annika Nietzio
   funding:
--- a/_rules/html-page-lang-b5c3f8.md
+++ b/_rules/html-page-lang-b5c3f8.md
@@ -32,7 +32,8 @@ acknowledgments:
 This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element for which all the following are true:
 
 - is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
-- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
+- has at least one [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) [text node][] which is neither empty nor only [whitespace][]..
 
 **Note:** `html` elements within `iframe` and `object` elements are not applicable as `iframe` and `object` elements create [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context). However, as these elements are meant to provide a layer of isolation, the declared language of their [parent browsing context](https://html.spec.whatwg.org/#parent-browsing-context) will likely not be inherited, making it possible for empty `lang` attributes in [nested browsing contexts](https://html.spec.whatwg.org/#nested-browsing-context) to also cause accessibility issues.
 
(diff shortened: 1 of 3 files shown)


_rules/html-page-lang-b5c3f8.md
```diff
@@ -32,7 +33,8 @@ acknowledgments:
 This rule applies to any [document element](https://dom.spec.whatwg.org/#document-element) if it is an `html` element for which all the following are true:
 
 - is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
-- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
+- has at least one [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) [text node][] which is neither empty nor only [whitespace][]..
```
Jym77: **Question:** Shouldn't the rule also apply if there is only replaced content that needs language (alt text of images, …)?
Maybe use "there is some [text inheriting its programmatic language](https://www.w3.org/WAI/standards-guidelines/act/rules/de46e4/proposed/#text-inheriting-language) from the document element, which is neither empty nor only [whitespace](https://www.w3.org/WAI/standards-guidelines/act/rules/de46e4/proposed/#whitespace)." like in the "element's `lang`" rule.

_rules/html-page-lang-matches-default-ucwvc8.md
```diff
@@ -39,7 +40,8 @@ This rule applies to any [document element][] if it is an `html` element for whi
 - The [document element][] has a `lang` attribute with a value that has a [known primary language tag][]; and
 - The [document element][] is in a [top-level browsing context][]; and
 - The [document element][] has a [content type][] of `text/html`; and
-- The [document element][] has a defined [default page language][].
+- The [document element][] has a defined [default page language][]; and
+- The [document element][] has at least one [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) [text node][] which is neither empty nor only [whitespace][].
```
Jym77: Same remark about having "text inheriting the language".

_rules/html-page-lang-valid-bf051a.md
```diff
@@ -33,7 +34,8 @@ This rule applies to any [document element](https://dom.spec.whatwg.org/#documen
 
 - has a `lang` attribute that is neither empty ("") nor only [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace); and
 - is in a [top-level browsing context](https://html.spec.whatwg.org/#top-level-browsing-context); and
-- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`.
+- has a [node document](https://dom.spec.whatwg.org/#concept-node-document) with a [content type](https://dom.spec.whatwg.org/#concept-document-content-type) of `text/html`; and
+- has at least one [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) [text node][] which is neither empty nor only [whitespace][].
```
Jym77: Same remark about "text inheriting the language".

Jym77 (dismissed):
Good point. We can tighten it a bit with existing definition.

_rules/html-page-lang-b5c3f8.md
giacomo-petri: Yeah, I had considered that but then forgot about it. And the reason why is that for 3.1.2 rules, the assumption is that the HTML has a lang defined. If no language is set, it becomes tricky. I'll think about how to handle that and how to address the problem of "equivalent" text.

WilcoFiers (changes requested):
I think this adds complexity for a problem that doesn't really need to be solved (pages without content or a title). I'd prefer we make sure all our examples have content so that there is no confusion around this.

Jym77:
> I think this adds complexity for a problem that doesn't really need to be solved (pages without content or a title). I'd prefer we make sure all our examples have content so that there is no confusion around this.

Should we then add an assumption to point out that we know about that?

PR #2419, opened 2026-06-08
Chore: Update package-lock.json

daniel-montalvo:
Update package-lock.json


WilcoFiers (approved):
That's fine, but you may want to put something about this bumping act-tools in the commit message so that that's clear. Since this is going to change rule pages.

Jym77 (approved):

PR #2427, opened 2026-07-13
Remove use of parentheses on its own sentence as per WAI Style Guide.

daniel-montalvo:
See https://wai-website-theme.netlify.app/style-guide/structure-presentation/#parentheses-optional

Need for Call for Review:
This will not require a Call for Review, editorial changes (including to the applicability, expectation or examples section).

---

### **When creating PR:**

### **After creating PR:**

- [x] If your PR is for a new rule with an Accessibility Support note, or for updating the Accessibility Support note of an existing section, make sure to open a corresponding Accessibility Support issue.

### **When merging a PR:**

## How to Review And Approve

### DIFF
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -142,7 +142,7 @@ This button has [visible inner text][] that, according to the [label in name alg
 
 #### Passed Example 8
 
-(Similar to previous example.) This link has [visible inner text][] that, according to the [label in name algorithm][], is contained within the [accessible name][]. This example shows why the [label in name algorithm][] uses the [visible inner text][] and not the [visible text content][]: the `<h6>` tags insert whitespace into the result in the former but not the latter.
+This link has [visible inner text][] that, according to the [label in name algorithm][], is contained within the [accessible name][]. This example shows why the [label in name algorithm][] uses the [visible inner text][] and not the [visible text content][]: the `<h6>` tags insert whitespace into the result in the former but not the latter.
 
 ```html
 <a href="#" aria-label="Some article by John Doe">
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -142,7 +142,7 @@ This button has [visible inner text][] that, according to the [label in name alg
 
 #### Passed Example 8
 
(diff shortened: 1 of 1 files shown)


_rules/visible-label-in-accessible-name-2ee8b8.md
```diff
@@ -142,7 +142,7 @@ This button has [visible inner text][] that, according to the [label in name alg
 
 #### Passed Example 8
 
-(Similar to previous example.) This link has [visible inner text][] that, according to the [label in name algorithm][], is contained within the [accessible name][]. This example shows why the [label in name algorithm][] uses the [visible inner text][] and not the [visible text content][]: the `<h6>` tags insert whitespace into the result in the former but not the latter.
+This link has [visible inner text][] that, according to the [label in name algorithm][], is contained within the [accessible name][]. This example shows why the [label in name algorithm][] uses the [visible inner text][] and not the [visible text content][]: the `<h6>` tags insert whitespace into the result in the former but not the latter.
```
dan-tripp-siteimprove:
```suggestion
Similar to previous example.  This link has [visible inner text][] that, according to the [label in name algorithm][], is contained within the [accessible name][]. This example shows why the [label in name algorithm][] uses the [visible inner text][] and not the [visible text content][]: the `<h6>` tags insert whitespace into the result in the former but not the latter.
```

I have no strong opinion on parenthesis.  I think there's value in the words "Similar to previous example" somehow.  My rationale is that, because this example's paragraph copied about 95% of its words from the previous example's paragraph, if we don't warn the reader about that somehow, then we will force the reader to read this example more carefully, and that's unnecessary mental effort.

dan-tripp-siteimprove (changes requested):

_rules/visible-label-in-accessible-name-2ee8b8.md
daniel-montalvo: Thanks for giving the context @dan-tripp-siteimprove 

It may be probably best to just focus these two descriptions on what makes the two examples different, not on what  makes them similar.

If we keep "Similar to previous example.", I would instead use a comma to avoid having an incomplete starting sentence.

> Similar to previous example, this link has [...]

dan-tripp-siteimprove: That sounds good to me.

WilcoFiers:
```suggestion
Similar to the previous example, this link has [visible inner text][] that, according to the [label in name algorithm][], is contained within the [accessible name][]. This example shows why the [label in name algorithm][] uses the [visible inner text][] and not the [visible text content][]: the `<h6>` tags insert whitespace into the result in the former but not the latter.
```

WilcoFiers (approved):

dan-tripp-siteimprove (approved):

PR #2436, opened 2026-07-31
Chore: prettier formatting in CI

daniel-montalvo:
This moves prettier formatting to CI and removes existing pre-commit hooks to ensure prettier always runs on PRs and not just when someone commits with deps installed on their local machine.


daniel-montalvo:
After discussing with @wilcofiers.


- Put a more explicit failing check for forks
- update the diffing to account for commits that are not in develop but are on the topic branch

daniel-montalvo:
Ready now @WilcoFiers ?

WilcoFiers (approved):

PR #2443, opened 2026-08-21
fix(label-in-name): remove SC-failing inapplicable examples

WilcoFiers:
Removed inapplicable examples regarding abbreviations and non-standard terms from the document. These two may be inapplicable to the rule, but IMO they fail the success criterion. There is are no exceptions for abbreviations in this WCAG criterion. By having these as inapplicable examples we prohibit implementors from failing these.

---

## How to Review And Approve

### DIFF
--- a/_rules/visible-label-in-accessible-name-2ee8b8.md
+++ b/_rules/visible-label-in-accessible-name-2ee8b8.md
@@ -401,22 +401,6 @@ This link has no [visible text content][].
 </a>
 ```
 
-#### Inapplicable Example 5
-
-This link's label contains an abbreviation.
-
-```html
-<a aria-label="University Avenue" href="#">University Ave.</a>
-```
-
-#### Inapplicable Example 6
-
-This word - non-standard / nonstandard - appears in both the element's accessible name and its visible label, using different hyphenation.
-
-```html
-<a href="#" aria-label="non-standard">nonstandard</a>
-```
-
 [accessible name]: #accessible-name 'Definition of accessible name'
 [label in name algorithm]: #label-in-name-algorithm 'Definition of Label in Name Algorithm'
 [non-text content]: https://www.w3.org/TR/WCAG22/#dfn-non-text-content 'WCAG Definition of Non-text content'


dan-tripp-siteimprove (approved):

HelenBurge (approved):
This one is interesting as I know the use of ABBR is not supported by mobile screen readers, but agree this is not inapplicable.

_rules/visible-label-in-accessible-name-2ee8b8.md
kengdoj, line 53:
```suggestion
If the target element contains an [image of text](https://www.w3.org/TR/WCAG22/#dfn-images-of-text), it may pass this rule but fail [2.5.3 Label in Name][understand253] (because the accessible name should then match the text inside the image, which is not taken into account by this rule). So further testing is needed. This case might be handled by a different rule in the future.

This rule does not include examples of abbreviations or differently hyphenated text because the label in name algorithm should not be used. While such examples are not covered by this rule, [Success Criterion 2.5.3 Label in Name][sc253] could apply.
```

kengdoj (dismissed):
Because these are inapplicable for the rule, I added an explanation as to why they are not included on line 53.

G0dwin (approved):

kengdoj (approved):

sorinfratila (approved):

mraccess77:
Should they be moved to failure examples then?

dan-tripp-siteimprove:
> Should they be moved to failure examples then?

I believe the answer is "no", because this rule (= the algorithm therein), as written, doesn't fail them.  Any example like this -  which fails the SC and has some other result X from the rule (where X = inapplicable, or pass) - doesn't currently have anywhere to live.  Hopefully they will one day.  That's the goal of "boundary / verbose examples": https://github.com/act-rules/act-rules.github.io/pull/2185

mraccess77:
> I believe the answer is "no", because this rule (= the algorithm therein), as written, doesn't fail them. Any example like this - which fails the SC and has some other result X from the rule (where X = inapplicable, or pass) - doesn't currently have anywhere to live. Hopefully they will one day. That's the goal of "boundary / verbose examples": #2185

Ok, so we have to remove for now - but it seems like we should update the algorithm then and later add back in.  If these are truly fails and they seem like common things then we'd want the algorithm to catch them - if the rules are too leaky then there is less value in using them.  If people don't consider these true failures then that's a different situation.

dan-tripp-siteimprove:
> but it seems like we should update the algorithm then and later add back in. If these are truly fails and they seem like common things then we'd want the algorithm to catch them - if the rules are too leaky then there is less value in using them. 

I fully agree.  This is a classic ACT rules conundrum.

dan-tripp-siteimprove:
This rule seems to have it worse than most, for some reason.

